PR #84, opened 2019-05-07
2019-05-07 Update to DNC 3.0.0 Preview 5

MarkStega:
Very simple Nuget update for DNC 3.0.0 P5


stsrki:
You should not add PackageReference in the csproj files, but instead through the .props files located in the Build folder.

PR #87, opened 2019-05-07
2019-05-07 DNC 3.0.0 P5 changes to Props files

MarkStega:
Here is the P5 change done in the props files (as well asd adding the props to the solution as solution items so that they are visible). Nuget got very confused on my system when I had the downlevel UnitTest and TestAppClient/Server included so I disabled them in the configuration manager. They need to be upgraded at some point but I was unclear as to their utility & what they didn't have the props file use in their csproj.


stsrki:
First thanks for PR.

1. There are some conflicts. Could you get the latest source from master and fix those? I have released 0.7.1 and there were some changes in those files. 
2. Can you remove props from the solution? I dont want them to be part of the solution.
3. Just update test projects directly in csproj files

MarkStega:
I resynchronized with master. And updated the one file that had a conflict.

The reason I added the props (and why you might consider doing the same) is that they aren't evident when they aren't included in the solution. This was the source of my doing a nuget update with the resulting changes to the csproj files which was not desirable.

I did update the two testapp projects inline, I did not update Blazorise.UnitTests since it does not seem to be an active part of the solution (?).

Issue #115, opened 2019-06-13
Wrong color filling of Charts in server-side demo


macropas:
Wrong color filling of Charts in server-side demo in comparison with the client-side demo
[Client-side](https://bootstrapdemo.blazorise.com/tests/charts):
![1](https://user-images.githubusercontent.com/7488502/59444182-1254b500-8e06-11e9-91f5-8b1fd1a4f009.png)
[Server-side](https://rcbootstrapdemo.blazorise.com/tests/charts):
![2](https://user-images.githubusercontent.com/7488502/59444243-33b5a100-8e06-11e9-9fa5-c159024eaf15.png)

stsrki:
Thanks for report, I will check the problem.

jdtcn:
@stsrki, I think I found the problem.

The problem is in this line:
https://github.com/stsrki/Blazorise/blob/38d04753cd458618905c2b41b91f9aba897980ec/Source/Extensions/Blazorise.Charts/ChartData.cs#L362

The "A" variable is a float, so it'll be converted to string using CurrentCulture. In "en-US" culture decimal delimiter is ".", but in some other cultures it is ",".

Thus when current culture decimal delimiter is "," we have
````JS
rgba(100,100,100,0,5)
````
instead of
````JS
rgba(100,100,100,0.5)
````

Possible fix:
````C#
$"{A.ToString(System.Globalization.CultureInfo.InvariantCulture)}"
````

stsrki:
@jdtcn You're right, that was the solution. Who knows when would I see that little dot missing :)

It will be in the next 0.8.1

Thanks a lot!

PR #155, opened 2019-07-22
Enable chart specific options to be serialized

johnbirdau:
It seems that one of the Blazor Json Serialization limitations is that it strictly limits properties to the declared `<T>` data types in `List<T>` and not looking at runtime subclasses of `T`.
This pull request works around that limitation.

For `ChartData<TItem>` by casting the `List<ChartDataset<TItem>> Datasets` to` List<object>` which happily serializes `TItem` properties and `TItem` subclass properties.
Unhappily, it does not honor `EmitDefaultValue = false` attribute instructions in this way. I investigated casting to specific ChartDataset subclasses however found that the codebase did not always honor the correct use of `ChartDataset` subclass for each graph type, so I was not sure if that was a design intention to accept it. (This can be revisited if you are OK for the PR to enforce that, for example, line charts must always have `List<LineChartDataset<TItem>> Datasets`)
As a result, I had to provide default options for all chart dataset properties as per your chart.md comment.
For the defaults I chose the default values from the chartjs website. These dont always have the best/consistent UX look to them, but were (for me) a better choice than arbitrarily choosing "looks good" values.

For `ChartOptions` a signature change to `Blazorise.Charts.JS.SetChartData` was sufficient to obtain correct serialization.

During testing I found that `BorderDashOffset` and `LineTension` were incorrectly specified as Int rather than the required Float, as their values can range between 0-1.

### DIFF
--- a/Source/Extensions/Blazorise.Charts/ChartData.cs
+++ b/Source/Extensions/Blazorise.Charts/ChartData.cs
@@ -1,9 +1,6 @@
 ﻿#region Using directives
-using System;
 using System.Collections.Generic;
-using System.Linq;
 using System.Runtime.Serialization;
-using System.Threading.Tasks;
 #endregion
 
 namespace Blazorise.Charts
@@ -39,6 +36,22 @@ namespace Blazorise.Charts
     [DataContract]
     public class ChartDataset<T>
     {
+
+        public ChartDataset() { }
+
+        protected ChartDataset(
+            string label,
+            List<string> backgroundColor,
+            List<string> borderColor,
+            int borderWidth
+        )
+        {
+            Label = label;
+            BackgroundColor = backgroundColor;
+            BorderColor = borderColor;
+            BorderWidth = borderWidth;
+        }
+
         /// <summary>
         /// Defines the dataset display name.
         /// </summary>
@@ -70,20 +83,31 @@ namespace Blazorise.Charts
         public int BorderWidth { get; set; } = 1;
     }
 
+    /// <remarks>
+    /// Defaults values as per https://www.chartjs.org/docs/latest/charts/line.html#dataset-properties
+    /// </remarks>
     [DataContract]
     public class LineChartDataset<T> : ChartDataset<T>
     {
+        public LineChartDataset() : base(
+            label: string.Empty,
+            backgroundColor: new List<string> { ChartColor.FromRgba(0, 0, 0, 0.1f) },
+            borderColor: new List<string> { ChartColor.FromRgba(0, 0, 0, 0.1f) },
+            borderWidth: 3
+        )
+        { }
+
         /// <summary>
         /// Length and spacing of dashes.
         /// </summary>
         [DataMember( EmitDefaultValue = false )]
-        public List<int> BorderDash { get; set; }
+        public List<int> BorderDash { get; set; } = new List<int>();
 
         /// <summary>
         /// Offset for line dashes.
         /// </summary>
         [DataMember]
-        public int BorderDashOffset { get; set; }
+        public float BorderDashOffset { get; set; }
 
         /// <summary>
         /// Fill the area under the line.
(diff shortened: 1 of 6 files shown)


stsrki:
Thanks for the PR. Excellent work.

I like the idea of setting all parameters to the default values. I think that will fix most of the issues with the serializer. At least for now.

There are some issue that I would like you to fix before I merge it. 

1. I think it would be best for now to leave the `DataJsonString` and `OptionsJsonString`, that is until the next version. In the meantime it should be noted in docs that it will be removed in the v0.8.x. Or of we're going to remove it that it **must** be noted in docs for the people to upgrade their code.
2. Fix minor formatting, missing spaces etc. I will leave comment so you can see what I mean..

Source/Extensions/Blazorise.Charts/JS.cs
```diff
@@ -8,9 +9,9 @@ namespace Blazorise.Charts
     static class JS
     {
         // TODO: clean this
-        public static Task<bool> SetChartData( IJSRuntime runtime, string id, ChartType type, object data, object options )
+        public static Task<bool> SetChartData<TItem, TOptions>( IJSRuntime runtime, string id, ChartType type, ChartData<TItem> data, TOptions options )
         {
-            return runtime.InvokeAsync<bool>( "blazoriseCharts.setChartData", id, ToChartTypeString( type ), data, options, data is string, options is string );
+            return runtime.InvokeAsync<bool>( "blazoriseCharts.setChartData", id, ToChartTypeString( type ), ToChartDataSet( data ), options);
```
stsrki: add space before closing bracket` );`

Source/Extensions/Blazorise.Charts/wwwroot/blazorise.charts.js
```diff
@@ -1,11 +1,5 @@
 window.blazoriseCharts = {
-    setChartData: (canvasId, type, data, options, dataIsString, optionsIsString) => {
-        if (dataIsString)
```
stsrki: maybe leave json strings for now?

Source/Extensions/Blazorise.Charts/ChartData.cs
```diff
@@ -70,20 +83,31 @@ public class ChartDataset<T>
...
 
+    /// <remarks>
+    /// Defaults values as per https://www.chartjs.org/docs/latest/charts/line.html#dataset-properties
+    /// </remarks>
     [DataContract]
     public class LineChartDataset<T> : ChartDataset<T>
     {
+        public LineChartDataset() : base(
+            label: string.Empty,
+            backgroundColor: new List<string> { ChartColor.FromRgba(0, 0, 0, 0.1f) },
```
stsrki: add missing spaces for ChartColor.FromRgba

stsrki:
PS. if you have time it would be great if you could also look at this issue with Charts https://github.com/stsrki/Blazorise/issues/115

I'm working on `DataGrid` now for v0.8 and that is the last issue left before release. Any help will be appreciated!

johnbirdau:
No problem with any of the feedback. Will address as requested.

Source/Extensions/Blazorise.Charts/JS.cs
johnbirdau: Done

Source/Extensions/Blazorise.Charts/ChartData.cs
johnbirdau: Done

Source/Extensions/Blazorise.Charts/wwwroot/blazorise.charts.js
johnbirdau: Done. Although had to pass string as separate parameter (to typed value on JS.cs caller)

Source/Extensions/Blazorise.Charts/JS.cs
johnbirdau: Done

johnbirdau:
Please review updated PR as per your comments

Re #115 I have not been able to reproduce locally on master branch nor on dev080 (Blazorise.Demo.Bootstrap.Server only as Blazorise.Demo.Bootstrap.RC project is now broken on dev080 - NRE in Container.razor).

Is https://rcbootstrapdemo.blazorise.com/tests/charts hosting the Blazorise.Demo.Bootstrap.Server or Blazorise.Demo.Bootstrap.RC project website?

stsrki:
I will see why it's broken on dev branch later after i get back home. Probably something with new features in the dev branch.
The problem with this bug is that it's visible only on production while on local it works just fine.

stsrki:
Merged. If you're going to look into #155 please open new PR. Once again great work, thanks!

stsrki:
Hi, I have fixed the error with _Container_ in server-side projects. I forgot to add theme options for that project. If you could check the #155 now that would be great. tnx

PR #310, opened 2019-10-12
Fix so that Autocomplete dropdown doesn't show empty dropdown

robalexclark:
Fixed up autocomplete show that the dropdown menu is only displayed when there is FilteredData available.


Source/Extensions/Blazorise.Components/Autocomplete.razor.cs
```diff
@@ -40,7 +40,7 @@ protected void HandleTextChanged( string text )
             CurrentSearch = text ?? string.Empty;
             dirtyFilter = true;
 
-            if ( text?.Length >= MinLength )
+            if ( text?.Length >= MinLength && FilteredData.Any())
```
stsrki: Can you just add space after `Any()`? Just to keep the formatting in check.

PR #311, opened 2019-10-12
Chart Options Not Updating

andrewwilkin:
If you made a change to any of the chart options after the initial render they were not getting applied. Data changes were.

Changing the lines in the update method to not include config fixed this.


stsrki:
Thanks for the bug fix!

PR #317, opened 2019-10-16
Fix nugets

stsrki:
The reason behind this PR is that `dotnet pack` is not working for referenced projects. And since every project is referencing Blazorise.csproj it means almost every nuget package will be invalid.

To overcome this limitation I have constructed .nuspec files to generate valid .nupkg files based on the dotnet pack.

This is only a workaround until hopefully dotnet pack is fixed. https://github.com/aspnet/AspNetCore/issues/14467


jzabroski:
Is there no simpler way to do this? I've been researching this for my own
project. FluentMigrator

stsrki:
I really don't know if there is a better way to handle this. I have tried to find it but eventually had to come up with this workaround or otherwise I would lose too much time until I release the new version.

stsrki:
PS. I like FluentMigrator a lot and being using it for all my personal and work projects 👍

jzabroski:
@stsrki Maybe we should submit an RFP to dotnetcli project and elaborate our needs?  It really sucks having to do verbose solutions like this.  I have run into this problem twice myself now.  The other situation was in a private work project dedicated to running PowerShell C# binary cmdlets.  Pack only packs the main file, but I need to pack everything in order to have my psd1 RequiredModules references all there.  My hack for now is to just download the dll with NuGet and save it to a lib folder in PowerShell.  That's "ok" for slowly changing dependencies like third party libs, but for a "global build at head" like a Solution file with all related projects moving a GlobalAssemblyInfo.cs AssemblyVersionAttribute in sync across all projects in the Solution, it SUCKS!

stsrki:
Yeah I think RFP on dotnetcli project would be the better route instead of in AspNetCore.

For me it's really strange that `dotnet pack` doesn't support referenced projects as I think it should be a fairly common scenario.

jzabroski:
This seems like one simplification: https://natemcmaster.com/blog/2018/02/02/dotnet-global-tool/#publish-output-into-pack

Issue #343, opened 2019-11-14
Add support for Regular and Brands FontAwesome Icons styles


WillianGruber:
The FA Icon Provider forces Solid Style (fas) even for string icon names.
Could you add support for Regular (far) and Brands (fab), and possibly Light (fal) and Duotone (fad) if someone buys the FontAwesome Pro? (I may buy it in the future).

Testing this code:
**fas-->** `<Button Color="Color.Primary" Clicked="@Save"><Icon Name="@("fas fa-square")" Margin="Margin.Is3.FromRight" />Save</Button>`
**far-->** `<Button Color="Color.Primary" Clicked="@Save"><Icon Name="@("far fa-square")" Margin="Margin.Is3.FromRight" />Save</Button>`
Renders to both solid icons:
![image](https://user-images.githubusercontent.com/56033692/68818584-2d04b180-0664-11ea-9f5c-48b0ff196d99.png)

I think if this line https://github.com/stsrki/Blazorise/blob/4755349c7080856b6bebcaed2bea768b4c72380d/Source/Blazorise/Icon.razor.cs#L23 was not executed for FA and when Name is string https://github.com/stsrki/Blazorise/blob/4755349c7080856b6bebcaed2bea768b4c72380d/Source/Blazorise/Icon.razor.cs#L28 then the others would automatically work. Probably checking for FA in the component would be a bad practice, and the FAIconProvider should handle this somehow.

I would greatly appreciate if you can implement this when reviewing and adding the new planned icons.

Thank you,
Willian Gruber

aboimpinto:
To help to the investigation, 
I try to use the "fab fa-docker" icon and it's not on the IconName enumeration and this is the result:

![image](https://user-images.githubusercontent.com/1231687/69766738-a4b10100-1179-11ea-901d-0ac8989a95e8.png)

after checking the generated HTML, I saw this:
![image](https://user-images.githubusercontent.com/1231687/69766772-c7dbb080-1179-11ea-9482-fbfb25cc75b2.png)
there are an extra "fas" in the class...
if I remove it, the docker symbol appears ...
![image](https://user-images.githubusercontent.com/1231687/69766817-04a7a780-117a-11ea-99de-bb48c0244b09.png)

![image](https://user-images.githubusercontent.com/1231687/69766828-12f5c380-117a-11ea-8ed6-075e9ff756d5.png)

[...]

PR #450, opened 2019-12-19
feat: added IconStyle to icon component

stsrki:
- Added new attribute IconStyle to Icon component
- Refactored icon provider so it can detect and remove duplicate icon style names from the icon name

#343

### DIFF
--- a/Source/Blazorise/Enums.cs
+++ b/Source/Blazorise/Enums.cs
@@ -1137,6 +1137,17 @@ namespace Blazorise
         SortDown,
     }
 
+    public enum IconStyle
+    {
+        Solid = 0,
+
+        Regular = 1,
+
+        Light = 2,
+
+        DuoTone = 4,
+    }
+
     public enum FormatType
     {
         None = 0,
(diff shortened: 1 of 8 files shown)


WillianGruber:
@stsrki , what is this IconStyle?
I have been working on something like this and was about to propose it, as I have already done it.

stsrki:
It's an enum and it's used as a "suggestion" for icon styles. For example, for FontAwesome there are Solid, Regular, DuoTone. And for Material there are similar styles, Filled, Outline, Duo-Tone.

The IconStyle is just an abstraction to combine style for different icon providers.

WillianGruber:
Ohh I see, and this won't affect my side because I use the "constants" from my custom class. Thanks.
Anyway, in a few days (maybe today) I'll present a new suggestion that you can adapt/include in Blazorise if you wish so.

stsrki:
@WillianGruber Now that I think off. There could be a problem if icon name is "fab fa-something". It will not work as it is expecting solid "fas" style. I think I will need to fix this.

PR #460, opened 2019-12-29
Dev088 datagrid improvements

Herdo:
Several DataGrid improvements for UX.

- Changed `TableHeaderCell` cursor to only use `pointer`, when the table is sortable.
- Added missing documentation for existing `DataGrid` members.
- Changed `_DataGridRow` cursor to use `pointer`, when the row is selectable.
- Unset `background-color` for `_DataGridDetailRow` - in case of hovering, the detail row doesn't need to be highlighted.


Source/Extensions/Blazorise.DataGrid/DataGrid.razor
```diff
@@ -79,7 +79,7 @@
                 }
                 else
                 {
-                    <_DataGridRow @key="@item" TItem="TItem" Item="@item" Columns="@Columns" Edit="@OnEditCommand" Delete="@OnDeleteCommand" Save="@OnSaveCommand" Cancel="@OnCancelCommand" Selected="@OnSelectedCommand" />
+                    <_DataGridRow @key="@item" TItem="TItem" Item="@item" Columns="@Columns" Style="@(RowSelectable?.Invoke(item) ?? false ? "cursor: pointer" : "")" Edit="@OnEditCommand" Delete="@OnDeleteCommand" Save="@OnSaveCommand" Cancel="@OnCancelCommand" Selected="@OnSelectedCommand" />
```
stsrki: So, I tried the PR and this line is not working as expected. First, the `Style `is not doing anything because it's not setting anything inside of `_DataGridRow`. The Style needs to be passed to `TableRow` or `TableRowCell` to make it work. But I would rather remove Style and implement `Cursor` attribute on `TableRow` or `TableRowCell`. You can see the similar code on `CheckEdit` component.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor
```diff
@@ -24,7 +24,7 @@
                         }
                         else
                         {
-                            <TableHeaderCell Style="cursor: pointer" Clicked="@(() => OnSortClicked( column ) )" width="@column.Width">
+                            <TableHeaderCell Style="@(Sortable ? "cursor: pointer" : "")" Clicked="@(() => OnSortClicked( column ) )" width="@column.Width">
```
stsrki: I would also add `Cursor` attribute to this one.

Source/Extensions/Blazorise.DataGrid/_DataGridDetailRow.razor
```diff
@@ -1,6 +1,6 @@
 ﻿@typeparam TItem
 @inherits _BaseDataGridDetailRow<TItem>
-<TableRow>
+<TableRow style="background-color: unset;">
```
stsrki: This one is a bit tricky if we want it to be abstracted. Let's leave it as it is and I can get beck to it sometimes later. So for now it's OK.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor
stsrki: Also. `RowSelectable` should be true by default instead of false.

Herdo: > So, I tried the PR and this line is not working as expected. First, the `Style `is not doing anything because it's not setting anything inside of `_DataGridRow`. The Style needs to be passed to `TableRow` or `TableRowCell` to make it work.

Indeed, `Style` is the wrong property here, you're right - thanks.
`TableRow` or `TableRowCell` - good question. Cell would be better, in case anybody wants to override the cursor for a specific cell?

> But I would rather remove Style and implement `Cursor` attribute on `TableRow` or `TableRowCell`. You can see the similar code on `CheckEdit` component.

Sounds good to me. So we'd pass `_DataGridRow.Cursor` the cursor it should show here. Is there an enum for the cursors? Or want me to pass the strings? `_DataGridRow.razor` would pass it along to the `TableRow` or `TableRowCell` then.

> Also. `RowSelectable` should be true by default instead of false.

Do you mean the `?? false ?` part of my changes?
If so, my thoughts were that I don't want to introduce a "breaking change" for the cursor in the `DataGrid` component.

### Current behavior:
_RowSelectable is `null`_: normal cursor
_RowSelectable returns `false`_: normal cursor
_RowSelectable returns `true`_: normal cursor

### New behavior (non-breaking in this PR):
_RowSelectable is `null`_: normal cursor
_RowSelectable returns `false`_: normal cursor
_RowSelectable returns `true`_: pointer cursor

### Regarding your comment (if I got that right, "breaking change"):
**_RowSelectable is `null`_: pointer cursor** ("breaking change")
_RowSelectable returns `false`_: normal cursor
_RowSelectable returns `true`_: pointer cursor

Source/Extensions/Blazorise.DataGrid/DataGrid.razor
Herdo: Will do!

Source/Extensions/Blazorise.DataGrid/_DataGridDetailRow.razor
Herdo: So leave the `unset` or remove it? 😄 

Yeah, I guess it would depend on the style provider used. 🤔 
This is probably more of a hotfix to remove the annoying flashing when hovering a "detail row", than a real fix.

Herdo:
@stsrki I'll make some changes tomorrow 👍 

On a side note, a thing that I wanted to fix but couldn't figure out directly:
When using nested `DataGrid` components (`DataGrid` in a `RowDetailTemplate`), all the properties (/styles) derive to the nested `DataGrid`.
E.g. I specify a parent `DataGrid.IsStriped = true`, the `DataGrid.IsStriped` inside the `RowDetailTemplate` will be false, however depending on the style provider (e.g. Bootstrap4), the child `DataGrid` will also have striped rows.
Any idea how we could tackle this? Want me to open an issue for that? 😁

Source/Extensions/Blazorise.DataGrid/DataGrid.razor
stsrki: You can see the Cursor enum in here: https://github.com/stsrki/Blazorise/blob/4b0917bad042eb863640c247bdea4134fafdc816/Source/Blazorise/Enums.cs#L1099

I know it's a little more work there but it's the right way to do :)

Regarding the RowSelectable. It's already working as a selectable even if the RowSelectable is null. So I think it you can go with:

RowSelectable is null: pointer cursor ("not really a breaking change")
RowSelectable returns false: normal cursor
RowSelectable returns true: pointer cursor

Source/Extensions/Blazorise.DataGrid/_DataGridDetailRow.razor
stsrki: Just leave it for now. I will do it sometimes later. Just open new issue so I can keep track of it and forget about it.

stsrki:
I guess the problem with stripped child data-grid is more of a Bootstrap issue. Probably there is some css rule where it inherits the parent style. Will need to investigate more.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor
stsrki: Just to let you know. When you add cursor styles for table row/cell you add them to `[Blazorise.ProviderName]/Styles/providerName.scss`, [example here](https://github.com/stsrki/Blazorise/blob/4b0917bad042eb863640c247bdea4134fafdc816/Source/Blazorise.Bootstrap/Styles/blazorise.bootstrap.scss#L21)

Then for it to be compiled automatically you need to have VS extension [WebCompiler](https://marketplace.visualstudio.com/items?itemName=MadsKristensen.WebCompiler). That way it will compile it to the final css.

stsrki:
I have made some minor changes, but overall you made a great job. Thank you for all the effort.

PR #464, opened 2019-12-31
Sidebar items can be expanded by default when using the dynamic method

koryphaee:
When using the Sidebar extension I wanted to use the dynamic approach because it seemed more flexible. Then I found out that items are collapsed by default. I ended up having to reimplement the logic from Sidebar.razor to add this option.

This commit adds exactly that option. The SidebarItemInfo has a new property which will expand it and all sub-items. The toggling still works no matter in which state the items start.

I think Blazorise is a fantastic library and I look forward to contributing more.

### DIFF
--- a/.github/workflows/dotnetcore.yml
+++ b/.github/workflows/dotnetcore.yml
@@ -1,0 +1,17 @@
+name: .NET Core
+
+on: [push]
+
+jobs:
+  build:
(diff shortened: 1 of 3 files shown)


stsrki:
Looks simple and clean. Thanks for the contribution.

Issue #452, opened 2019-12-19
New Feature: Icon Picker


WillianGruber:
In my application I need to enable my users to choose an icon to custom documents. So, I developed an Icon Picker, and I'm sharing here so that @stsrki may add to Blazorise if the component proves itself useful. My best bet is that it would be an Extension Component.
I'm sharing it here as I am using it now (with FontAwesome Icons), so probably it would need to be refactored to add other icon providers.
Feel free to modify as you see fit.

### How to use it
```
<h4>Simple Icon Picker</h4>
<IconPicker @bind-Icon="Icon"></IconPicker>
<h6>Chosen icon: @Icon</h6>
@code {
    protected string Icon { get; set; }
}
```

### Results:
![image](https://user-images.githubusercontent.com/56033692/71213969-a317bc00-2293-11ea-8048-7a4610ea8b9d.png)
![image](https://user-images.githubusercontent.com/56033692/71214310-bf1b5d80-2293-11ea-8011-aa367b78fd60.png)
![image](https://user-images.githubusercontent.com/56033692/71214334-d22e2d80-2293-11ea-8393-6293485fd26d.png)
![image](https://user-images.githubusercontent.com/56033692/71214361-e1ad7680-2293-11ea-8e36-851b37fb6579.png)
![image](https://user-images.githubusercontent.com/56033692/71214481-318c3d80-2294-11ea-94e8-0a39a688abed.png)

### Parameters:
**Icon** The selected formatted icon. You can pass an icon here and it will highlight as selected in modal, too, including parsing options.
**AllowFlip** Enables the options for Flipping horizontally, vertically, and both. Default: true
**AllowRotation** Enables the options for rotating 90, 180 and 270 degrees. Default: true
**AllowSize** Enables the size options. Default: true
**AllowSizeMinimum** Sets a minimum in the size options, so that the user cannot select a smaller size. Default: IconSize.ExtraSmall
**AllowSizeMaximum** Sets a macimum in the size options, so that the user cannot select a larger size. Default: IconSize.x10
**ButtonSize** The size of the button used to open the modal. Default: ButtonSize.None
**ButtonColor** The color of the button used to open the modal. Default: Color.None
**ButtonTooltipText** The tooltip of the button used to open the modal. Default: "Select an Icon"
**ButtonTooltipPlacement** The placement of tooltip of the button used to open the modal. Default: Placement.Bottom
**ModalTitle** The header title of the modal component. Default: "Select an Icon"
[...]

PR #503, opened 2020-01-13
Performance improv  only render modals child content if modal is opened

WillianGruber:
Hi @stsrki ,
I think it is not necessary to render the content of a modal if it is closed. I hope this also helps fix the lag in #452 .
Alternativelly, I'm rendering the iconpicker only of the modal is open.


stsrki:
I will need to test this before merging. There can be problems when you remove component from rendering tree with `if `statements. When the component is removed from the tree the usual life-cycle methods will not be fired and it will skip some required init or dispose steps.

stsrki:
@WillianGruber I just tested this and as I suspected, if we remove the modal content the usual life-cycle will not happen. In this case for example `ModalBackdrop` will dispose every time will close modal but it will NOT unregistered itself from java-script as it's reference will be null.

I think without major rewriting this PR will not work.

WillianGruber:
I see. Sorry about that.

stsrki:
There are so much tricks with Blazor that sometimes I wonder if it's all worth it: 😅

PR #504, opened 2020-01-13
Fix for InvalidOperationException in BaseNumericEdit

Herdo:
Replaced switch-case of `BaseNumericEdit.FormatValueAsString` with generic method usage of BindConverter (`BindConverter.FormatValue<T>`).

### DIFF
--- a/Source/Blazorise/NumericEdit.razor.cs
+++ b/Source/Blazorise/NumericEdit.razor.cs
@@ -85,23 +85,7 @@ namespace Blazorise
 
         protected override string FormatValueAsString( TValue value )
         {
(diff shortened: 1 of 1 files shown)


stsrki:
Thank you for the PR but unfortunately it doesn't' work for unsigned types. Also I already fixed this in another one: https://github.com/stsrki/Blazorise/pull/497

My PR also supports unsigned types which `BindConverter.FormatValue<T>` does not support.

Herdo:
Okay, glad to hear that!
Too bad the generic one doesn't work. Sucks to implement each type.

PR #519, opened 2020-01-22
Implemented DataGrid Multi-Sort.

ricardoromaobr:
Implemented a multi-sort is necessary to change the  ReadData handle to  take the multi-columns to sort.

### DIFF
--- a/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
+++ b/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
@@ -5,6 +5,7 @@ using System.Collections.Generic;
 using System.ComponentModel;
 using System.Linq;
 using System.Linq.Expressions;
+using System.Reflection;
 using System.Threading.Tasks;
 using Blazorise.DataGrid.Utils;
 using Microsoft.AspNetCore.Components;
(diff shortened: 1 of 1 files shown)


ricardoromaobr:
I comment in the pull request that will be necessary to adjust the ReadData handle. But no, is necessary do nothing. Because it columns will have   yours  directions already configured.  Sotting is made :)

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -379,15 +374,28 @@ private void FilterData( IQueryable<TItem> query )
...
-                        query = query.OrderBy( item => sortByColumn.GetValue( item ) );
-                }
-
+                var isFirtSorting = true;
+                foreach ( var sortByColumn in sortByColumns )
+                {
+                    if ( isFirtSorting )
+                    {
+                        if ( sortByColumn.Direction == SortDirection.Ascending )
+                            query = query.OrderBy( x => GetField<TItem>( sortByColumn.Field ).Compile().DynamicInvoke( x ) );
```
stsrki: I would like this to be precompiled and not compiled on every sort.

stsrki:
@ricardoromaobr I have made some changes. Mainly I removed the need for recompile of expression tree. Please test on your side before I merge.

ricardoromaobr:
Ok, I test and Works. :)

PR #615, opened 2020-02-25
Add DisplayGridColumns parameter

robalexclark:
Add DisplayGridColumns parameter so that columns can be hidden on the grid but still available in the editor.

The datagrid now takes its list of columns from GridDisplayColumns which filters any columns with DisplayGridColumns=false.

### DIFF
--- a/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
+++ b/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
@@ -465,6 +465,11 @@ namespace Blazorise.DataGrid
         protected IEnumerable<BaseDataGridColumn<TItem>> EditableColumns => Columns.Where( x => x.ColumnType != DataGridColumnType.Command && x.Editable );
 
         /// <summary>
+        /// Gets only columns that are available for display in the grid.
+        /// </summary>
+        protected IEnumerable<BaseDataGridColumn<TItem>> GridDisplayColumns => Columns.Where( x => x.ColumnType == DataGridColumnType.Command || x.DisplayInGrid );
+
+        /// <summary>
         /// Returns true if <see cref="Data"/> is safe to modify.
         /// </summary>
         protected bool CanInsertNewItem => Editable && Data is ICollection<TItem>;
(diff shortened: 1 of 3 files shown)


Source/Extensions/Blazorise.DataGrid/DataGridColumn.razor.cs
```diff
@@ -141,6 +141,11 @@ internal void SetValue( TItem item, object value )
         /// </summary>
         [Parameter] public bool Filterable { get; set; } = true;
 
+        /// <summary>
+        /// Gets or sets whether the column is visible in the grid
+        /// </summary>
+        [Parameter] public bool DisplayInGrid { get; set; } = true;
```
stsrki: Rename to `VisibleInGrid`

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -464,6 +464,11 @@ public Task SelectRow( TItem item )
         /// </summary>
         protected IEnumerable<BaseDataGridColumn<TItem>> EditableColumns => Columns.Where( x => x.ColumnType != DataGridColumnType.Command && x.Editable );
 
+        /// <summary>
+        /// Gets only columns that are available for display in the grid.
+        /// </summary>
+        protected IEnumerable<BaseDataGridColumn<TItem>> GridDisplayColumns => Columns.Where( x => x.ColumnType == DataGridColumnType.Command || x.DisplayInGrid );
```
stsrki: Rename to `VisibleColumns`?

robalexclark:
Ok, I've renamed the properties to fit in better with the other properties and pushed my changes as well as changed the target branch to dev090. There is a nasty conflict on DataGrid.razor.cs so it might be easier to  keep your version and manually add in the following block below Editable columns (sorry)

```
        /// <summary>
        /// Gets only columns that are visible in the grid.
        /// </summary>
        protected IEnumerable<BaseDataGridColumn<TItem>> VisibleColumns => Columns.Where( x => x.ColumnType == DataGridColumnType.Command || x.VisibleInGrid );
```

stsrki:
@robalexclark No problem. I will add the feature manually. Anyways thank you for taking time for this PR!

stsrki:
Merged with #730

Issue #617, opened 2020-02-26
DataGrid Row Double click


vmakharashvili:
I need to make double click on DataGrid row. I'm trying to access to table tr tag to bind dblClick event, but it's not possible. Please add this feature to the DataGrid component.

stsrki:
Do you have a use-case for this feature? Also a proposal API would be great! tnx

vmakharashvili:
We had a table which had double-click event on the row. The table looked like this:
`<Table IsHoverable="true" IsBordered="true">
        <TableHeader ThemeContrast="ThemeContrast.Light">
            <TableRow>
                <TableHeaderCell Class="text-left">Name</TableHeaderCell>
                <TableHeaderCell Class="text-left">IBAN</TableHeaderCell>
                <TableHeaderCell Class="text-left">Currency</TableHeaderCell>
                <TableHeaderCell Class="text-left">Balance</TableHeaderCell>
                <TableHeaderCell Class="text-left">Priority</TableHeaderCell>
                <TableHeaderCell Class="text-left">Id</TableHeaderCell>
                <Blazorise.TableHeaderCell></Blazorise.TableHeaderCell>
            </TableRow>
        </TableHeader>
        <TableBody>
            @foreach (var account in _accounts)
            {
                <tr @ondblclick="(() => OnAccountRowClick(account))">
                    <TableRowCell Class="text-left">@account.Name</TableRowCell>
                    <TableRowCell Class="text-left">@account.AccountNumber</TableRowCell>
                    <TableRowCell Class="text-left">@account.Ccy</TableRowCell>
                    <TableRowCell Class="text-right">@account.Balance.ToString("N")</TableRowCell>
                    <TableRowCell Class="text-left">@account.Priority</TableRowCell>
                    <TableRowCell Class="text-left">@account.Id</TableRowCell>
                    <TableRowCell Class="text-center">
                        <Button Color="Color.Primary" IsOutline="true" @onclick="(() => OnAccountRowClick(account))">
                            <i class="oi oi-eye" aria-hidden="true"></i>
                        </Button>
                    </TableRowCell>
                </tr>
            }
        </TableBody>
    </Table>`
There was a double-click on table row. 
For other reason we changed the table into DataGrid. Here we just have DataGridColumn. There is no notion about table-row. How can I make table row click events in DataGrid?

stsrki:
[...]

PR #619, opened 2020-02-27
feat: click and double click events on DataGrid row

stsrki:
#617

### DIFF
--- a/Source/Blazorise/TableHeaderCell.razor.cs
+++ b/Source/Blazorise/TableHeaderCell.razor.cs
@@ -25,17 +25,23 @@ namespace Blazorise
             base.BuildClasses( builder );
         }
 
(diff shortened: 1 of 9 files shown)


GilShalit:
Hi, how does this work?  How does one access the row?

stsrki:
@GilShalit You have `RowClicked` and `RowDoubleClicked` events. Both are having an Item as an event attribute field.

eg. 

```cshtml
RowClicked="@OnRowClicked"`
```

and

```cs
Task OnRowClicked( DataGridRowMouseEventArgs<SomeModel> eventArgs )
{
  var item = eventArgs.Item; // access item here

  return Task.CompletedTask;
}
```

GilShalit:
Ah neat! now I can get into Edit mode with 

DataGridRef.Edit(item);

PR #621, opened 2020-02-28
Update Autocomplete.razor.cs

gmiseria:
I think could be usefull to handle the Tab key as a selection on the AutoComplete control. If you use only Enter and you are inside a form you will launch a submit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,3 +211,18 @@ public class Startup
     }
 }
 ```
+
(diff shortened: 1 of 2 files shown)


stsrki:
@gmiseria Thanks for the PR. Regarding the enter key when inside of a Form, I think that can be handled by preventDefault.

PR #635, opened 2020-03-11
Add DataGrid FilterTemplate

vmakharashvili:
Fieature Request: 633

I added filter template to DataGrid.


stsrki:
Hi, I have looked at your changes and I think this could work. I don't allow any commits to master directly so can you switch to **dev090** branch instead? After that I will test and merge if everything goes well. tnx

vmakharashvili:
Sure, I will.

Issue #633, opened 2020-03-09
DataGridColumn Filter Template


vmakharashvili:
In DataGrid there is a very useful filter textboxes. I use them really well as server-side filtering. However, there is an issue: this filter is only string type and when we need int, Date, multiple-choice or other type filters, this filter becomes useless.

I think there will be good idea to have a filter-template inside DataGridColumn. This will give us ability to have our own custom filter which will be super-useful.

For example:
```cshtml
<DataGrid TItem="Customer" Data="@_customers" PageSize="CustomerService.PageState[Page.Customers].PageSize" Filterable="true"
          RowDoubleClicked="((e) => OnCustomerRowClick((e.Item as Customer)))" Narrow="true" Hoverable="true"
          ReadData="(e => ReadData(e))">
    <DataGridColumn TItem="Customer" Field="@nameof(Customer.Id)" Caption="Id" Sortable="true" Filterable="true" Filter="@Filter("Id")">
        <DataGridColumnFilterTemplate>
            ... some code
        </DataGridColumnFilterTemplate>
    </DataGridColumn>
    </DataGridNumericColumn>
</DataGrid>
```

stsrki:
I tried that already #331

The problem is in how Blazor works when building the render-tree and I was unable to do anything. To make this work I will either have to rebuild DataGrid internals or do it something like this #394 

I will have to decide after I have more time to investigate.

vmakharashvili:
I downloaded the source and created the FilterTemplate in my local computer. It works fine.
How can I contribute to the source code?

PR #639, opened 2020-03-11
Dev090 datagrid filter template

vmakharashvili:
Feature request #633


stsrki:
Hi, I have made some modification to your PR and now I cannot push changes to your repository. I'm getting this error:

>  ! [remote rejected] HEAD -> dev090-datagrid-filter-template (refusing to allow an OAuth App to create or update workflow `.github/workflows/dotnetcore.yml` without `workflow` scope)
> error: failed to push some refs to 'https://github.com/vmakharashvili/Blazorise'

stsrki:
Removing github from Windows Credential fixed the problem. 🤦‍♂

Your code worked but I noticed some possible memory leaks. Rendering tree always create new `FilterContext` which means the new event handler would be created also for every change. I moved everything to DataGridColumn and created `FilterContext` once on initialize along with event handler. Unsubscribe will be called on Dispose, Also modified the tests slightly. 

Anyways, great work. It is a simple and elegant solution to our problem. Thank you!

PR #692, opened 2020-04-07
Is inline

8VAid8:
Blazor is supposed to has "is-inline" class, not "inline" - https://bulma.io/documentation/modifiers/responsive-helpers/


stsrki:
Hi, looks good, but...

Working branch is `dev090`. So can you switch to that one instead of `master`?

A lot has changed on dev090 so probably there will be some conflictd. Maybe it would be easier to create new PR?

tnx

8VAid8:
Yea, i'll create a new PR.

PR #694, opened 2020-04-07
"inline" class has been replaced by "is-inline"

8VAid8:
Blazor is supposed to has "is-inline" class, not "inline" - https://bulma.io/documentation/modifiers/responsive-helpers/

Also replaced that class for radio buttons.


stsrki:
All done. Thank you!

PR #699, opened 2020-04-08
Fixed minor issues in RadioGroup tests so that they pass.

DarthPedro:
The code:

`        WaitAssert.Equal( "false", () => radioG.GetAttribute( "checked" ) );
`

were all failing because the attribute for checked is not false, it is actually null  when radio buttons are unchecked.


stsrki:
Who would have think null value is the reason for failing 🤦‍♂ 

I have just fixed some minor formatting by running _Format Document_(Ctrl+E, D), that's something I always do for every file.

Anyways great work, thank you for taking time to fix the issue!!

Issue #569, opened 2020-02-04
[Bug] Scroll problem with sidebar


WillianGruber:
**Describe the bug**
When having a large sidebar, the scroll on it only happens when the page scroll is in the botton.

**To Reproduce**
Go to https://materialdemo.blazorise.com/tests/forms.
Make sure not all the sidebar items are displayed (for me, ToDo List is not displayed, there is no space for it)
Scroll down with the mouse over the sidebar. The page scrolls down. After the page hits the botton of scroll position, sidebar begins to scroll down.

**Expected behavior**
When the mouse is over the sidebar, it should scroll independently from the page.

stsrki:
I'm already familiar with this issue .I guess I can't ignore it any more :)

stsrki:
Fixed with #713 thanks to @MitchellNZ

PR #713, opened 2020-04-09
fix: add back sticky header & scrollable sidebar

MitchellNZ:
- Removed comment that was blocking sticky/fixed header from working
- Added styles for scrollable sidebar

### DIFF
--- a/Source/Blazorise/Styles/_layout.scss
+++ b/Source/Blazorise/Styles/_layout.scss
@@ -31,7 +31,7 @@
     color: rgba(0, 0, 0, 0.65);
     /*background: #001529;*/
     /*line-height: 64px;*/
-    /*&-fixed {
+    &-fixed {
         position: sticky;
         z-index: 1;
         top: 0;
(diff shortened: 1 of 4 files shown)


stsrki:
Just a simple solution :) Thanks you!

This also fixes #569

Source/Blazorise/Styles/_layout.scss
```diff
@@ -31,7 +31,7 @@
     color: rgba(0, 0, 0, 0.65);
     /*background: #001529;*/
     /*line-height: 64px;*/
-    /*&-fixed {
+    &-fixed {
```
MitchellNZ, line 34: @stsrki this is the change I am referring to that I added back.
It looked like you mistakenly commented this line when you were removing `line-height`.

Without this, there was no more `.b-layout-header-fixed` generated in the CSS, which broke the `fixed` header.

stsrki: Yes you're right, sorry about that. Since it's now back I can proceed with merge.

MitchellNZ: Great! Sounds good to me.
Thanks again.

PR #723, opened 2020-04-13
Fix: b-body-layout & Material SCSS

### LINKED ISSUE #722: .b-body-layout is not working Bulma
'.b-body-layout' is messing with Bulma layout content. It needs to be disabled.

MitchellNZ:
- Moved `.b-body-layout` styles into all providers except Bulma (fixes #722)
- Refactored Material `CSS` into `SCSS` files


stsrki:
You've done even more work than needed. Looks good. Thank you!

PR #726, opened 2020-04-13
Added DateEdit tests

DarthPedro:
- Added unit tests and mock for DateEdit component.
- Added component level tests for DateEdit as well.
- Disabled 2 Min-Max tests because they did not work as expected - may be bug in DateEdit component.


stsrki:
I tried the tests and they're failing. It seems the reason is different date formats on your and mine machine. 

For example on test `SetDateTime()` you expect:

`<span id=""date-event-initially-undefined-result"">5/3/1970 12:00:00 AM</span>`

and on my machine I get the result

`<span id="date-event-initially-undefined-result">3.5.1970. 0:00:00</span>`

DarthPedro:
Ah, forgot about the possible different date formats. I will fix the tests.

DarthPedro:
I fixed the date formatting issues with the component test expected value strings. Please give it a try.

stsrki:
Everything works now. Good job as always!

PR #727, opened 2020-04-13
Fix slider initial value ignoring max

### LINKED ISSUE #677: Slider component not initializing as expected with values over 100.
**Describe the bug**
When you set Slider.Value to a number greater than 100 during initial binding, it gets trimmed to 100 (even when the Max is much higher).

**To Reproduce**
1. Create a Slider with the following:
`<Slider TValue="int" Min="10" Max="1000" Value="900" Step="5" Class="slider">Cool slider</Slider>`

2. Click F5 and start your app.
3. Notice the slider thumb is displayed at a lower value (looks like about where 100 would be). But it's Value property is correct.
4. If you use the slider, it will move up past that value and up to the Max.

**Expected behavior**
The slider should start about 90% through the slider (near the right end). But it stays at about 100 (about 10%).

**Additional context**
Note: This works fine if the value range is between 1 through 100, like:
<Slider TValue="int" Min="1" Max="100" Value="90" Step="5" Class="slider">Cool slider</Slider>
When you launch the app now, the slider is at the right position (90).

MitchellNZ:
- Simple fix for `Slider` initial value ignoring `Max` on Bootstrap provider
- Add demo/test for `Slider` having large `Max`

Fixes #677

### DIFF
--- a/Demos/Blazorise.Demo/Pages/Tests/FormsPage.razor
+++ b/Demos/Blazorise.Demo/Pages/Tests/FormsPage.razor
@@ -133,6 +133,7 @@
                     <FieldLabel ColumnSize="ColumnSize.Is2">Range slider</FieldLabel>
                     <FieldBody ColumnSize="ColumnSize.Is10">
                         <Slider TValue="decimal" Value="25m" Max="100m" />
+                        <Slider TValue="int" Min="10" Max="1000" Value="800" Step="5" />
                     </FieldBody>
                 </Field>
                 <Field Horizontal="true">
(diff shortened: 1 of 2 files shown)


stsrki:
Wait, placing `min` and `max` attributes before `value` has fixed the issue?! I think this is the bigger problem that should be sent to Blazor team.

Anyways thank you for fixing this!

@DarthPedro can you do some unit tests to cover Slider?

MitchellNZ:
@starki yeah, pretty silly issue!

But as far as I could tell, this was only an issue for the Bootstrap provider..

It was working fine in Ant Design before I did this change.

stsrki:
Ant Design has a completely custom implementation and is not using `input`  element.

MitchellNZ:
Oh I see, I didn’t check for custom implementation there. My bad.

But it’s working now for all now!

stsrki:
Yes, not it's working for all providers.

PR #736, opened 2020-04-14
[AntDesign] fix: ant card image/cover

MitchellNZ:
Since Ant Design has the concept of a `Cover` element for `Card` which is structured like the following:

```
<div class="ant-card-cover">
  <img ... />
</div>
```

This class `.ant-card-cover` applies additional styling to the child elements.
These are missed in Blazorise.

Since Blazorise has `CardImage` as its equivalent (without the parent class) I have added the additional CSS directly to the `` element that has the `.ant-card-cover` class.

Example of Ant Design CSS can be found [here.](https://github.com/ant-design/ant-design/blob/master/components/card/style/index.less#L132)

**Edit:**
Moved this CSS once I saw it is partially covered in the `ThemeGenerator` (still learning about Blazorise code structure!)

I think that the `Top` and `Bottom` `img` elements of the `Card` should follow Ant Design and always have radius that matches the `Card`.

However, I have left the `options?.ImageTopRadius` untouched, this can be used to override the default I am suggesting.

### DIFF
--- a/Source/Blazorise.AntDesign/AntDesignThemeGenerator.cs
+++ b/Source/Blazorise.AntDesign/AntDesignThemeGenerator.cs
@@ -443,6 +443,16 @@ namespace Blazorise.AntDesign
                 .Append( $"border-radius: {GetBorderRadius( theme, options?.BorderRadius, Var( ThemeVariables.BorderRadius ) )};" )
                 .AppendLine( "}" );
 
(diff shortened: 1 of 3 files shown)


stsrki:
I think you managed to learn Blazorise structure quite well. 👍 

PR is working, thank you!

PR #737, opened 2020-04-15
[AntDesign] fix: progress bar

MitchellNZ:
There was an issue with the Ant Design progress bar background taking up too much vertical space:
image

I have fixed it to the following: 
image

**Edit:**
Additionally, I have also added better styling for the `ProgressBar` inner text on Ant Design.

Was:
image

Now:
image

However, since this is non-standard of Ant Design, I have put this styling in a new "Blazorise" specific class `.b-ant-progress-text`


stsrki:
It's now working better than ever. Thank you once again!

PR #740, opened 2020-04-15
 Created NuGet build script that triggers on new tag creation with the build version number, like: 0.9.1-preview

DarthPedro:
- build script that triggers on tag creation.
- gets the version number from the repo tag. Tag must be just the version number: 0.9.0-preview.
- builds projects and runs unit tests
- packages library projects into NuGet packages: Gets package descriptive information from content in csproj file.
- uploads artifacts to build page (zip file with all of the nukpgs)


stsrki:
I can see that echo is printing valid version number. eg. `0.9.16-preview`

But when I go to the build result to see the version inside of built DLLs it still says 0.9.0. Either I'm looking at the wrong place or `dotnet build` is ignoring `/p:Version=$(git describe --tags $(git rev-list --tags --max-count=1))`

stsrki:
Can you add me as a contributor to your repository? So I can create new tags for testing.

DarthPedro:
Did changing the build switches from /p to -p work? The online articles I found specifically used /p for the version parameter setting.

stsrki:
We'll find out soon. I was waiting for you to give me access rights. It's building..

.github/workflows/build-nuget.yml
```diff
@@ -34,7 +36,7 @@ jobs:
       run: dotnet publish --configuration Release --no-restore --no-build
 
     - name: Pack - Release
-      run: dotnet pack --configuration Release --no-restore -o ./Packages /p:PackageVersion=$(git describe --tags $(git rev-list --tags --max-count=1))
+      run: dotnet pack --configuration Release --no-restore -o ./Packages /p:Version=$VERSION_TAG
```
DarthPedro, line 39: Getting the right nuget package version requires the PackageVersion property.

stsrki:
I think this is now ready to be merged. There was a lot of commits while testing for both of us so I'm not sure if I just squash everything or do we create new clean PR? It's your call.

Oh yeah, I have also added last job to push nugets to test feed. Currently it's failing because it's missing api key from on your repo. I have already test push on my own dummy repo so the job should pass.

DarthPedro:
Great work! You found the magic combination.
We should just squash the PR commit. There's been so much back and forth, I'm afraid I will miss something recreating a new PR.

stsrki:
I can go sleep in peace now 😅

stsrki:
Thank you for all the work so far. You even forced me to learn GitHub actions in the process :)

Issue #627, opened 2020-03-04
DataGrid Filter box placement


vmakharashvili:
**Describe the bug**
Issue is related to filter use. 
When I don't want to use the first column filter and want the second one, The filter box renders under the first column.
`  <DataGrid TItem="WeatherForecast" Data="@forecasts" Filterable="true">
        <DataGridColumn TItem="WeatherForecast" Field="@nameof(WeatherForecast.Date)" Caption="Date" Sortable="true" 
                        Filterable="false"></DataGridColumn>
        <DataGridColumn TItem="WeatherForecast" Field="@nameof(WeatherForecast.TemperatureC)" Caption="Temperature C" 
                        Filterable="true"></DataGridColumn>
    </DataGrid>`

In browser filter box renders under "Data" column.
![image](https://user-images.githubusercontent.com/16862175/75857379-9543f280-5e0f-11ea-8ae5-80169333582a.png)

stsrki:
It's a bug definitely. And a stupid mistake on my end also..

When generating the filter controls I do this

```cshtml
@foreach ( var column in Columns )
{
    @if ( !column.Filterable )
        continue;
    //
}
```

It should actually be an empty cell instead of continue. I will fix this for the next preview.

PR #742, opened 2020-04-16
fix: insert empty cell when DataGridColumn Filterable is False

MitchellNZ:
Fixes issue in #627 as per @stsrki suggestion.

Changed demo project `DataGrid` page `Childrens` column to show as example.


stsrki:
Thank you!

PR #748, opened 2020-04-16
fix: Ant Desgin TableRowHeader

MitchellNZ:
One last Ant Design fix before **0.9.0** from me!

Ant Design does not seem to support `th` tags in the `tbody` element rows (`TableRowHeader` component).

It was looking like:
image

I have updated them to be `td` elements for Ant Design and created a custom CSS class for `bold`.
image


stsrki:
Thanks, everythings good. I just renamed the class to be more in-line with ant design.

MitchellNZ:
@stsrki No worries.

My original idea with adding the `.b-` prefix was so we could quickly tell what are Blazorise extension classes vs official Ant Design ones.

But I’m happy to follow your guidance!

PR #749, opened 2020-04-17
refactor: Collapse component away from Card dependency

MitchellNZ:
Creating a PR for this in **dev090** - but its up to you @stsrki if you want this in **0.9.0** or not!

**Summary**
Refactored `Accordion` component to work independently from the `Card` component.

**Reasoning**

- This will be more extendable to new `ClassProviders` (such as Ant Design) that do not rely on the `Card` component for their `Collapse`/`Accordion` elements.
- Ant Design was not working well with this system due to having different CSS classes for the `Collapse` elements - I'm sure there will be more providers that are like this too.

**Docs/Testing**

- Update the `accordion` docs to reflect changes
- Manually tested with all `ClassProviders`

**Notes**
@stsrki could you please take a look at the `CollapseContent` element. 
I was torn between nesting the `Body` (like I have done) vs creating a new `CollapseContentBody` element that would need to be manually nested like:
```
<CollapseContent>
  <CollapseContentBody>
    text goes here
  </CollapseContentBody>
</CollapseContent>
```
I didn't do it like this example above because it appears like this element always exists in each framework and is only really used as a sub-container. 
Seemed over-kill to make the developer **always** do this too.

**Example**
Ant Design before:
image

Ant Design after:
image

All other providers are unchanged (in terms of their UI).


stsrki:
I made some changes:

- renamed CollapseContent to CollapseBody, as Body is used on most components
- named Collapse Visible parameter, to know where it comes from

All in all, I must say you have learned Blazorise internals really well. Thank you!

MitchellNZ:
Great! Looks good.
I was actually thinking of renaming it to `ContentBody` also haha.

Glad I can help a little, Blazorise is very nice to work with, you’ve done a fantastic job designing it!

stsrki:
And I was just thinking of leaving it as `CollapseContent` haha.

Thanks, it's good to know people are happy with Blazorise.

MitchellNZ:
Well then... haha.

I think either is good anyway.

I ended up with`CollapseContent` because:
- It follows Ant Design (where my personal experience is)
- The `Content` element could one day have its on own `Header` and `Body` (kinda already has its own `Body`)
- Blazorise has a`ModalContent` with a `ModalBody`

But `CollapseBody`, like you said, is inline with other existing Blazorise components.. like `CardBody` which this was modelled off.

PR #759, opened 2020-04-18
Fixing Modal Size CSS

mwolford:
I found a bug in 0.9 where setting the ModalSize is missing the CSS class prefix for bootstrap.  This lead to getting `xl` not `modal-xl` when setting to ExtraLarge.


stsrki:
Thank you!

PR #765, opened 2020-04-19
Moved all E2E tests to bUnit

DarthPedro:
- migrated the remaining E2E tests to bUnit component tests.
- almost all of the same tests are covered... there are a couple that I'm waiting for help from bUnit team to give me a solution.
- these tests now run in ~6 seconds rather than 4 minutes on Selenium.
- test code coverage goes up from ~10% to 26% with additional tests.

Note: I did not delete the current E2ETests project, but if this all gets accepted, I would recommend removing that project.


.github/workflows/build-nuget.yml
```diff
@@ -3,8 +3,8 @@ name: NuGet Build
 on:
   create:
     tags: 0.9.*, 1.*
-    branches-ignore: '*'
-
+    branches-ignore:
+      - 'features/**'
```
MitchellNZ, line 7: Could you please also add an ignore for like `dev***-**`?
Trying to cover cases like `dev090-my-new-feature` (as recommending formatting in the contributing docs).

MitchellNZ: However, maybe the docs should be updated to create these features under the feature branch.. 
like `features/dev090-my-new-feature`

DarthPedro: The standard way to do this in Git repos is to use features/dev090-my-new-feature. We should follow that way of doing it. I will update the contributing doc to say that. Thanks.

MitchellNZ: Typically, the `features/` prefix is only really used in the _Gitflow Workflow_ and not so much in the _Forking Workflow_ (like on GitHub).
But I'm happy to go with whatever makes it easiest!

DarthPedro: That's the standard we use a work, even though we don't use Gitflow... I can make either way work, just need a decision.

MitchellNZ: I also use that approach at work (but with the [Gitflow Workflow](https://www.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow)), so I am also comfortable with it too :)

The decision is for @stsrki! I have just been using the approach he outlined in the Contributing docs.

Also worth mentioning, while we are talking about this topic, another quite common approach I have seen is using `{username}/my-new-feature` as can be seen on the [aspnetcore GitHub](https://github.com/dotnet/aspnetcore/branches/active) branches

stsrki: Well, it all depends on the project. Naming branches with `features` prefix make sense, but I don't think it's good for this one. With Blazorise we have well defined milestones and branches are named after that, `dev088, dev090, dev091`, etc. That's why I want all feature and bug-fix branches to have prefixes named after the branch where it goes to. It's easier to follow.

I also noticed problem with `build-nuget.yml`. It fires every time new branch is created instead of only on new tags. I tried to fix it but there are still some issues. During the day I will see what can be done.

Anyways, for now naming stays the same, have branches with prefix dev{version}.

stsrki:
Everything looks good. Don't see any issues so far.

Let E2E tests stay for awhile. Until we figure how to handle java-script part,

.github/workflows/build-nuget.yml
DarthPedro: I will look at updating the script for the dev{version} pattern.

PR #786, opened 2020-04-22
Additional fixes for responsive Layout

MitchellNZ:
@stsrki until [this CSS issue](https://github.com/w3c/csswg-drafts/issues/865) is fixed, we cannot use `position: sticky` header **and** `overflow-x: hidden` on `Layout` elements. `overflow-x: hidden` is required for responsive to work well within the `Layout` elements.

Since responsiveness is more important.. I have made workaround for the `sticky` header by using `fixed` and calculating the offset of the `Sidebar` (in `_layout.scss`).

This isn't perfect for all situations and would be better if the offset calculation was tied to the `margin` of the `Sidebar` (but this requires more JSInterop knowledge than I have!)

**Also**, removed the now un-need `.b-body-layout`.. which is a win.

### DIFF
--- a/Demos/Blazorise.Demo/Layouts/MainLayout.razor
+++ b/Demos/Blazorise.Demo/Layouts/MainLayout.razor
@@ -170,7 +170,7 @@
                 </BarMenu>
             </Bar>
         </LayoutHeader>
-        <LayoutContent Padding="Padding.Is4.OnX">
+        <LayoutContent Padding="Padding.Is4.OnX" Style="margin-top: 64px">
             @Body
         </LayoutContent>
     </Layout>
--- a/Source/Blazorise.AntDesign/Styles/_layout.scss
+++ b/Source/Blazorise.AntDesign/Styles/_layout.scss
@@ -1,7 +1,3 @@
 ﻿.b-layout-header {
     line-height: 64px;
 }
-
-.b-body-layout {
-    overflow-x: hidden;
-}
--- a/Source/Blazorise.Bootstrap/Styles/_layout.scss
+++ b/Source/Blazorise.Bootstrap/Styles/_layout.scss
@@ -1,3 +1,1 @@
-﻿.b-body-layout {
-    overflow-x: hidden;
-}
+﻿
--- a/Source/Blazorise.Frolic/Styles/_layout.scss
+++ b/Source/Blazorise.Frolic/Styles/_layout.scss
@@ -1,3 +1,1 @@
-﻿.b-body-layout {
-    overflow-x: hidden;
-}
+﻿
--- a/Source/Blazorise.Material/Styles/_layout.scss
+++ b/Source/Blazorise.Material/Styles/_layout.scss
@@ -1,3 +1,1 @@
-﻿.b-body-layout {
-    overflow-x: hidden;
-}
+﻿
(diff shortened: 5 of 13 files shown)


stsrki:
The issue with `position:sticky` is reported more than 3 years ago. I'm not sure it will be ever fixed :)

The layout now works although it's not pretty. It makes a lot of assumptions like `margin-top: 64px;` and `width: calc(100% - 220px);`. I need to think this through.

We have two ways:

1. write these two styles cases in Demo css.
2. use css variables

  - `margin-top: var(--layout-needs-offset-from-top, 64px);`
  - `width: calc(100% - var(--sidebar-is-placed-so-move-the-bar, 220px));`

** Funny names on purpose :)

MitchellNZ:
Haha yeah I’m not so sure it will be fixed either.. 😅

I agree, I was originally going to make CSS variables to control this and build them into the `ThemeGenerator`..
But I wasn’t sure if that was the right approach?

I haven’t looked into the `ThemeGenerator` much yet.. so my understanding is quite low,
but it looks like it could work well with my upcoming Bar changes by doing it that way too!

stsrki:
I plan to refactor theme generator as it became a little messy lately. So maybe it's better to write the styles into demo css for now. And for 0.9.2 will do the necessary steps to make it better.

MitchellNZ:
Okay, sure thing!

But by doing that, it will break anyone using they `Layout` components with a `Sidebar` and `Fixed` header.. and only that combination 😅

So probably quite a low use case! But perhaps worth mentioning in the release notes with what custom offsetting CSS to write (this calc stuff) to workaround until it’s fixed.

stsrki:
Or just leave it like this and people can always override CSS for their own use-cases...

MitchellNZ:
That’s probably the best option, seeing as `64px` and `220px` are Blazorise defaults 👍

I’m happy to update the `Layout` docs tomorrow explaining that!

stsrki:
OK we're going with this. I'm going to merge it and later when you write notes in docs, just create new PR. tnx

MitchellNZ:
Also, as a final note! 

Since the `Layout` components were modelled off Ant Design, it’s worth noting they also have the same problem and **do not** support the the combination of: `Sider` + `Fixed` header.

It seems people create their own fixed sub-headers within Ant Design by using their `Affix` component (as far as I can tell)

stsrki:
Maybe we can also have our own version of Affix?

MitchellNZ:
I think that would be good!

I was looking into it today, there is quite a lot of JavaScript to handling all the positioning 😅

stsrki:
So, I removed the `margin-top` from css and specified it in demo on Style property. There was already special rule for Bar(`Style="height: 64px;"`) so I think this is OK.

For Sidebar I created new variable and used that one in case it's needed.

MitchellNZ:
Awesome! I think this is a good solution 👍

Also, is the responsive table working well for you in the Bootstrap provider now?

stsrki:
Every provider works except for Bulma. I just started to look at it.

MitchellNZ:
Great! Just let me know if there is anything further I can do to help.

Otherwise I’ll continue with the `Bar` changes tomorrow 👍

stsrki:
No problem. See you tomorrow, and thank you!

Issue #831, opened 2020-04-24
Ant Card missing margins/padding


WillianGruber:
After switching from Material to Ant, all the margins and paddings in my project messed up.
Seems like Ant isn't implementing them.

Comparing the demos from material and ant:
Material:
![image](https://user-images.githubusercontent.com/56033692/80237129-94bb2180-8632-11ea-8ed0-e3a8c3e4b356.png)
Ant:
![image](https://user-images.githubusercontent.com/56033692/80237169-a6042e00-8632-11ea-9834-6eaa6827b64b.png)

MitchellNZ:
Unfortunately, Ant Design doesn't apply spacing automatically to the `Rows` and `Columns` like Bootstrap and other frameworks do, they use a [Grid Gutter system](https://ant.design/components/grid/#components-grid-demo-gutter).

You can read about how to use this with Blazorise in the docs [here](https://blazorise.com/docs/components/grid/#gutter).
This was added in v0.9 and does support the other providers too.

But you are right, the demo projects do not reflect this well.

WillianGruber:
I see, but having switchable providers with minimal effort in migration as a concept for Blazorise, it would be nice if this could be made. Maybe the component in Ant side could apply it to the underlying markup as attribute splatting? I know it is not recommended by ant developers, but I think this option would fit Blazorise, what do you think @stsrki ?

MitchellNZ:
@WillianGruber yes, I do understand where you are coming from, and it would be a matter of just adding a default `Gutter` or CSS override to the Ant Design provider. 
We would just have to assume some sort of acceptable default.

However, by adding:
```
<Row Gutter="(20, 0)">
    ...
</Row>
```

It looks like this in Ant Design:
![image](https://user-images.githubusercontent.com/6387861/80271330-63665280-8713-11ea-9801-105c5e03bd82.png)

And this in Material:
![image](https://user-images.githubusercontent.com/6387861/80271338-71b46e80-8713-11ea-8f49-eae3498485f9.png)

Which provides a more consistent experience overall! 

So there is also the argument that these should always be set if you want a consistent experience between providers.

But I also like the idea of providing a default to Ant Design if none is set, it makes it that extra bit easier! :)

Decision is over to @stsrki

stsrki:
I think the best option is to add default margins to Card so that they are consistent with other providers. This is how we already do it for other components.

[...]

PR #833, opened 2020-04-25
fix: Ant Design default row gutter to match Bootstrap

MitchellNZ:
Added `Row`/`Column` guttering to Ant Design as a default that matches Bootstrap/Material.

Addresses #831


stsrki:
Thanks, I'm in the middle of refactoring so I will test it later,

MitchellNZ:
No worries!

stsrki:
Default grid margins broke the form spacing. AntDesign by default does not have many spacings for form columns so I had to add them. I don't see any issue now except for column offset on forms.

image

I will merge and we can return later to fix offset.

Issue #792, opened 2020-04-22
error with charts after redrawing


lucamorelli:
**Describe the bug**
I copied the chart sample in a new page, and after pressing the redraw button a few times (sometime 3, sometimes more) I obtain this error in the log console

```
ncaught TypeError: Cannot read property 'transition' of null
    at ni.transition (chart.js@2.8.0:7)
    at ni.draw (chart.js@2.8.0:7)
    at ni.render (chart.js@2.8.0:7)
    at Object.callback (chart.js@2.8.0:7)
    at Object.advance (chart.js@2.8.0:7)
    at Object.startDigest (chart.js@2.8.0:7)
    at chart.js@2.8.0:7
Same problem updating to version 2.9.3 of chart.js

```
**To Reproduce**
1. copy the source code from https://blazorise.com/docs/extensions/chart/
2. press the redraw button over and over for a few times

stsrki:
I believe think this is the issues for live charts? There is a problem with live charts when you go with mouse over the chart point. I'm already aware of the this but will need more time to debug.

aunabidi:
@stsrki  I just upgraded to Blazorise 0.9 and am seeing this issue on a page where I have 2 charts (not live). I initialize both charts in OnAfterRenderAsync firstRender with blank data and redraw them once more when the data is available. I am getting the "Cannot read property 'transition' of null" error on page load.

stsrki:
@aunabidi Just to make sure. Can you refresh your browser by pressing Crtl+F5? It tends to use cached js files.

aunabidi:
@stsrki I tried that as well. The error is in chart.js@2.8.0:7

aunabidi:
@stsrki  I have tried a bunch of different things but I can't get my charts to render. Any ideas?

stsrki:
Really not sure. For ordinary charts everything works on my side. Can you guys create a small project and push it to github where I can debug it?

aunabidi:
It works fine with static data. I am calling an API asynchronously to populate the data and the labels which is causing the issues I think. Have you tried rendering dynamic data and labels on a chart? 

I will try to get you a test github project this weekend. Thanks a lot for you help!

aunabidi:
@stsrki I have created a test github project where I recreated the issue. On the index page if you click on redraw a couple of times you will see that it throws the error message in the console. This is happening even without calling an API for dynamic data.

https://github.com/aunabidi/BlazoriseTest

stsrki:
[...]

PR #881, opened 2020-05-13
fix: new apis for updating chart

stsrki:
#792 error with charts after redrawing

This PR adds additional APIs to handle the chart update.

- AddDatasetsAndUpdate()
- AddLabelsDatasetsAndUpdate()

The problem with separate calls to chart with Clear(), AddLabel(), AddDataset() and Update() is that chartjs for some reason will raise an exception sometimes before update is called. Since it is random I cannot find the true cause. So I introduce this new APIs that will merge several calls into one operation.

### DIFF
--- a/Demos/Blazorise.Demo/Pages/Tests/ChartsPage.razor
+++ b/Demos/Blazorise.Demo/Pages/Tests/ChartsPage.razor
@@ -120,11 +120,7 @@
     {
         await chart.Clear();
 
-        await chart.AddLabel( Labels );
-
-        await chart.AddDataSet( getDataSet() );
-
-        await chart.Update();
+        await chart.AddLabelsDatasetsAndUpdate( Labels, getDataSet() );
     }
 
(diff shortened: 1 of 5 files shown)


pmagnus2:
This fix works for one dataset. However, if we have say 5 lines displayed in one LineChart then we need multiple calls to AddDataSetsAndUpdate() and the same error occurs again. Any ideas?

stsrki:
@pmagnus2 Can you open a new issue with the steps and code to reproduce the error?

pmagnus2:
I will see if I can strip out a small example by the weekend. For now, the quick fix is to disable animation in the LineChartOptions by adding:
`Animation =new Animation() { Duration=0}`

pmagnus2:
@stsrki Was not able to reproduce the error in a fresh project. Could be some other issue in my code. Will revert with new issue if it can be consistently reproduced.

PR #896, opened 2020-05-14
Dev091 datagriddetail columnspan fix

xxCruciblexx:
Content of detail row wasn't spanning entire width of the detail row for me without a command column being specified.  Reversing subtraction logic created the same effect when command column was specified.  Decided to remove it to resolve the issue.  
I'm probably doing something wrong and have never tried creating a pull request before so feel free to straighten me out and tell me what I should do :) and if it's better to ask on gitter or just open an issue that's cool just let me know.

### DIFF
--- a/Source/Extensions/Blazorise.DataGrid/_DataGridDetailRow.razor.cs
+++ b/Source/Extensions/Blazorise.DataGrid/_DataGridDetailRow.razor.cs
@@ -24,7 +24,7 @@ namespace Blazorise.DataGrid
             => Columns.Any( x => x.ColumnType == DataGridColumnType.Command );
 
         protected int ColumnSpan
-            => Columns.Count;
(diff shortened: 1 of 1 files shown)


stsrki:
Leaving column span to all columns still not working for all scenarios. For example of you unchecked **editable** than there will be row leaking outside of table

image

I have made some changes but before I merge can you confirm it works for you also?

xxCruciblexx:
Ahh I didn't try that combination when testing.  Yep that works for me.  I will be sure to play around with the demo next time instead of just my app to try and catch things like that.  Btw, thanks so much for all your work on this project, it really is impressive what your doing.  I'm so excited about this and Blazor in general, and want to help if I can.

stsrki:
Well, this PR confirms that I have also not checked all combinations :) Hopefully this covers it now.

Thank you for the help!

Issue #36, opened 2019-02-05
Expand list of Icon types/names


stsrki:

stsrki:
[Material.x.FontAwesome.Icons.xlsx](https://github.com/stsrki/Blazorise/files/3913361/Material.x.FontAwesome.Icons.xlsx)

List if icons, done by @WillianGruber

WillianGruber:
Remember it's only the free icons. Many material icons exist only in paid version of FA.

stsrki:
I don't have much experience with paid icons. Maybe there can be a provider included in Blazorise that can handle paid icons via credentials or whatever the requirements are.

PR #912, opened 2020-05-21
feat: expand list of icons

stsrki:
#36 Expand list of Icon types/names


Source/Extensions/Blazorise.Icons.FontAwesome/FontAwesomeIcons.cs
```diff
@@ -0,0 +1,1011 @@
...
+        public static readonly string Edit = "edit";
+        public static readonly string Egg = "egg";
+        public static readonly string Eject = "eject";
+        public static readonly string EllipsisH = "ellipsis-h";
+        public static readonly string EllipsisV = "ellipsis-v";
+        public static readonly string Envelope = "envelope";
+        public static readonly string EnvelopeOpen = "envelope-open";
+        public static readonly string EnvelopeOpenText = "envelope-open-text";
+        public static readonly string EnvelopeSquare = "envelope-square";
+        public static readonly string Equals = "equals";
```
WillianGruber, line 307: this should have the '''new''' keyword

stsrki: I didn't notice that. Thank you!

WillianGruber: In my generator, I used reflection so that if any member of ```object``` needs to be redefined, it includes the ```new``` keyword

stsrki: I have fixed the generator so the `new `keyword should now work.

stsrki:
@WillianGruber I have used your excel file to load missing icons and now the total number of mapped icons is little above 300.

PR #928, opened 2020-05-25
Check if component is rendered before calling JSRunner in Dispose()

### LINKED ISSUE #923: Don't call JSRunner during prerendering
When using ASP.NET Core hosted **Blazor WebAssembly** app it's possible to make use of **server prerendering**. With this technique first user request is responded to by server instantaneously with full page html. This is useful for SEO and fast content delivery while app is still downloading in the background. Since prerendering happens only on the server with no connection to the browser, it's impossible to call any JS Interop. When `OnAfterRenderAsync()` is used it's no issue since it's not called when prerendering, however Blazorise also calls JSRunner in `Dispose()` (as can be [seen on Button](https://github.com/stsrki/Blazorise/blob/7d206fa2f2c1808e81e799186ebcb5a55e14c8d6/Source/Blazorise/Button.razor.cs#L84)) and it's missing check whether component is indeed rendered in browser.

This will result in the following exception when `<Blazorise.Button>` is server prerendered:
> System.AggregateException: Exceptions were encountered while disposing components. (JavaScript interop calls cannot be issued during server-side prerendering, because the page has not yet loaded in the browser. Prerendered components must wrap any JavaScript interop calls in conditional logic to ensure those interop calls are not attempted during prerendering.)

**Sample GitHub repo:** https://github.com/Xeevis/BlazorisePrerendering

### Steps to reproduce the behavior:

1. Create new Blazor WebAssembly 3.2.0 hosted project
```powershell
dotnet new blazorwasm -o BlazorApp1 --hosted
```
2. Integrate Blazorise [according to documentation](https://blazorise.com/docs/usage/bulma/) in both Client and Server
3. Create new _Host razor page in Server
```powershell
dotnet new page -n _Host -o BlazorApp1/Server/Pages -np
```
4. Copy contents of **_Client/wwwroot/index.html_** into **_Server/Pages/\_Host.cshtml_** and replace `Loadng...` with prerendered component
```html
@page
@addTagHelper *, Microsoft.AspNetCore.Mvc.TagHelpers
(...)
<body>
    <app><component type="typeof(BlazorApp1.Client.App)" render-mode="ServerPrerendered" /></app>
```
5. In **Server/Startup.cs** change fallback from _**index.html**_ to _**/_Host**_ page
```csharp
//endpoints.MapFallbackToFile("index.html");
endpoints.MapFallbackToPage("/_Host");
```
6. Open _**Client/Pages/Counter.razor**_ and replace button element with Blazorise variant
`<Blazorise.Button Color="Blazorise.Color.Primary" @onclick="IncrementCount">Click me</Blazorise.Button>`
7. Run the Server
```powershell
dotnet run -p BlazorApp1/Server
```
8. Go to [https://localhost:5001/counter](https://localhost:5001/counter) (may be necessary to clear cache for prerendering to kick in)

**Expected behavior**
Don't attempt to call **JSRunner** in `Dispose()` if component isn't browser rendered.

Xeevis:
Fixes #923


stsrki:
By the look of it, it seems good. I'm in the middle of some larger task so I will test it when I finish.  Thank you for the contribution!

Xeevis:
Sure, no problem. This code works for me well so far and I've made it based on this piece of documentation: [Detect when the app is prerendering](https://docs.microsoft.com/en-us/aspnet/core/blazor/lifecycle?view=aspnetcore-3.1#detect-when-the-app-is-prerendering)

stsrki:
I have tested it and everything works just fine. Also I have renamed IsRendered to Rendered.

Thank you once again!

PR #940, opened 2020-05-29
Add RichTextEdit based on QuillJS

njannink:
Here my initial PR for a RichTextEdit for Blazorise.

I would like to get some initial feedback about the direction and if it fits in the framework


WillianGruber:
Will it be in a different package? Like charts, snackbar, sidebar, etc?

stsrki:
@njannink I don't allow any PR into master directly. Can you switch to dev091?

It's pretty late here so I don't have much time, here are some things I notice:

- Move RichTextEdit code into `.razor.cs` code behind and please follow code-style guidelines
- The same for other components
- Why RichTextEditToolbarButton actions must be defined by the users? What are the benefits?

I will need to finish some work before I can test this, until then please switch branch as that is the biggest issue now. 

And thanks your for the effort, this is pretty big PR!

njannink:
@WillianGruber yes its in a seperate project/packaged

@stsrki I will update the code to comply to the rest of the styles

The `RichTextEditToolbar` elements are optional and are for customizing the RTE toolbar like is possible with QuillJS. The most simple usages of the RichTextEditor is

```
<RichTextEditor ContentChanged="@OnContentChanged" />
```

PR #947, opened 2020-06-01
Add flask icon to Font Awesome

xxCruciblexx:


njannink:
Maybe also add the material icon:

https://material.io/resources/icons/?search=flask&icon=science&style=baseline

stsrki:
@xxCruciblexx As @njannink has already mentioned. You also need to add it to dictionary for material provider. Also don't forget to add it to `FontAwesomeIcons.cs` and `MaterialIcons.cs`

xxCruciblexx:
Thanks for the feedback. My bad, wasn't using that project in my solution. I should be smart and use the full Blazorise solution from now on when trying to make changes. I added them. `FontAwesomeIcons.cs` already had flask in case your wondering why it's not in the list. Let me know if I missed anything else.

stsrki:
I just fixed sorting of icon names as I realized they were messed up.

Thank you!

PR #948, opened 2020-06-01
DataGrid aggregate column shift fix

xxCruciblexx:
Aggregate row cells were shifted one cell to the right if the DataGrid was set to not editable but had a command column specified. Used logic from DataGridRow.


stsrki:
Tested and it's working! Thank you!

Issue #763, opened 2020-04-19
Option to make vertical Bar


stsrki:
Refactor Bar component so it can be used as a sidebar.

Linked issue: #567 

@MitchellNZ I'm assigning this to you as you already plan to work on it.

MitchellNZ:
Other PRs I will try look into as part of this:
- [ ] #6 Refactor Sidebar project
- [ ] #190 smaller side bar
- [ ] #350 Sidebar width and Bar height (**possibly**)
- [ ] #567 Implement Bar Color as a specific parameter in BarThemeOptions
- [ ] #766 Toggle Animation not working with Bar component

stsrki:
Sidebar refactor could be too much to handle but if you can pull it off it would be great.

stsrki:
@MitchellNZ  I will close this one. For any issue still left just create separate ticket(s).

PR #954, opened 2020-06-03
Feature: Vertical Bar options

MitchellNZ:
Related to #763.
Does not address all issues linked to it. I have marked the ones that it does address.
Will most likely make a follow up PR to fix outstanding issues.

Looking for feedback and some assistance @stsrki!

### Outstanding work
**Theming**
_Breakpints:_
- [x] Generate Breakpoints CSS in ThemeGenerator (this will replace _breakpoints.scss)

_Bar:_

- [x] Implement ThemeGenerator for sizing
- [x] Implement ThemeGenerator for colours (background, selected item, etc)
- [x] Selected item should have the Theme background (this should be done like GenerateDropdownStyles)

_Documentation:_

- [x] Bar `Vertical` modes.
- [x] Bar `Collapse` modes.
- [x] Bar `NavigationBreakpoint`.
- [x] New `BarIcon`.
- [x] How to create Right Align bar (see below).

### Known issues
_Ant Design Bar:_
- [x] Bug: Selected/Active Item not highlighting
- [ ] Bug: `BarMode.VerticalPopout` does not work
- [x] Bug: Selected/Active Item not highlighting
- [x] Bug: Fix Right Align Bar (needs CSS)
- [x] Bug: Fix clicking links on small bar (need to move icon inside link or something)
- [x] Bug: Hidden bar does not work
- [x] Bug: Small bar does not work

### Future features
- Dynamically created (code created) Bar
- Third level dropdown menu (this works for inline, but not small or popout)
- Support for Light/Dark themes with AntDesign

_Note: I will open an issue (or multiple issues) to track these "future features" once merged_

### Additional notes for docs
**_[ADDED]_ How to: Right align sidebar**
- Put `LayoutSider` containing bar after `Layout`
- Put `RightAligned="true"` on all `BarDropdownMenu` elements within the right bar

**_[FUTURE]_ How to: Merge BarBrand with topbar**
- Add `box-shadow: none;` style to vertical bar
- Add `bg-primary` class (or equiv) to vertical `BarBrand`

_Note: I will add this to the docs at a later time.. its a bit hacky and I want to find a better way_

### Summary
Let me know what you think @stsrki! You should have access to make changes - if not, let me know.
Otherwise some comments would be good and I can fix them up myself :) 

If something in this list doesn't make sense to you, let me know and I can explain more.. these are just copied from my personal notes.
I will continue working on the known issues too!


Source/Blazorise/Styles/_breakpoints.scss
```diff
@@ -0,0 +1,29 @@
+﻿body:before {
```
stsrki, line 1: I cannot seem to find where body:before is used for Bar. Is there a way for this to be custom class instead, like ".b-body-bar" or something?

MitchellNZ: This code is actually used for the new `BreakpointAdapter` I made.
This adapter allows the component to run code (like the `Bar` needs to) on breakpoint changing.

You can see in `blazorise.js` there is a method (`getBreakpoint`) that calls `window.getComputedStyle(document.body, ':before')` to figure out what the current breakpoint is (this is automatically called by the `window.reszie` event that is configured).

The reason it is on the body tag is because this `BreakpointAdapter` can be used to by any component that needs to run code on breakpoint change (and putting it on `body` doesn't rely on anything else), and I thought it might be useful for the future! 

These are some examples of what I modeled the pattern after: [here](https://www.lullabot.com/articles/importing-css-breakpoints-into-javascript), and [here.](https://gomakethings.com/the-easy-way-to-manage-css-breakpoints-in-javascript/)

MitchellNZ: Also - I was suggestion in my PR that these are generated in the `ThemeGenerator` so they can be easily configured to match a new provider or by Blazorise users

stsrki: I've been thinking of going to do theming. Starting with dark and light theme for Bar and the going with colors.

One thing I'm not sure how to handle. How to override bootstrap, bulma or antdesign breakpoints? We cannot just override them as there are just too many classes that are using them.

Source/Blazorise/Styles/_bar.scss
```diff
@@ -0,0 +1,405 @@
...
+        height: var(--b-sidebar-brand-height, 64px);
+    }
+
+    .b-bar-item {
+        margin: auto;
+        flex-grow: 1;
+        min-height: 40px;
+
+        .b-bar-icon {
+            font-size: 18px;
```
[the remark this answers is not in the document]
MitchellNZ: Yeah. Happy to use `rem`, I'm more comfortable with `px` so I just default to that haha.
Will update any `font-sizes` to start with.

stsrki:
This is what happened with Bulma breakpoints. Similar is for other providers but now as big as this.

image

It can be seen in chrome media query inspector. Just to be clear, this is completely on my fault,! It's something that needs to be taken care sooner or later.

Source/Blazorise/Styles/_breakpoints.scss
MitchellNZ: In regards to the theming, in `_bar.scss` you will see a comment called **Theming** which has `.bar-bar-dark` and `.b-bar-light` under it.. I kept everything relating to Theming wrapped up in those 2.
My plan was to replace everything inside those with CSS generated from the `ThemeGenerator` (as well as the `:root` stuff at the top for sizing).

Yeah, I'm not sure if it will be possible to override them. But with the moving the `_breakpoint.scss` into the `ThemeGenerator`, at least it means it will be easy for us to match them per provider (and perhaps just not make them overrideable by Blazorise users since it will cause too many issues)

Source/Blazorise/Styles/_bar.scss
```diff
@@ -0,0 +1,405 @@
...
+
+    .b-bar-mobile-toggle {
+        right: 20px;
+        margin: auto;
+        display: none;
+    }
+    /* 
+        Theming
+    */
+    &.b-bar-dark {
```
MitchellNZ, line 185: @stsrki these are the two sections I was referring to that I think should be generated in the `ThemeGenerator`.
` &.b-bar-dark` and `&.b-bar-light` + their children.

As well as `:root` at the top of this file

MitchellNZ:
> This is what happened with Bulma breakpoints. Similar is for other providers but now as big as this.
> 
> image
> 
> It can be seen in chrome media query inspector. Just to be clear, this is completely on my fault,! It's something that needs to be taken care sooner or later.

@stsrki Is this picture referring to the collapsing of sidebar without the toggle?

If so, there is an inconsistency between the `BreakpointAdapter` and `media` based breakpoints that I have noticed too.. (forgot to add it to my bug list actually).

I am unsure if its a bug from my new `BreakpointAdapter` or the existing `Breakpoint` functionality 🤔

stsrki:
> > This is what happened with Bulma breakpoints. Similar is for other providers but now as big as this.
> > image
> > It can be seen in chrome media query inspector. Just to be clear, this is completely on my fault,! It's something that needs to be taken care sooner or later.
> 
> @stsrki Is this picture referring to the collapsing of sidebar without the toggle?
> 
> If so, there is an inconsistency between the `BreakpointAdapter` and `media` based breakpoints that I have noticed too.. (forgot to add it to my bug list actually).
> 
> I am unsure if its a bug from my new `BreakpointAdapter` or the existing `Breakpoint` functionality 🤔

No it is actually from too many different media breakpoints. There are slight difference in defining the breakpoint between Blazorise, Bulma, Snackbar and/or any other component. They do not follow predefined provider breakpoints. But that is just one problem. Bigest problem is that we don't have solution to this

|           | XS    | SM    | MD     | LG     | XL     |
|-----------|-------|-------|--------|--------|--------|
| Bootstrap | <576  | >=576 | >=720  | >=960  | >=1140 |
| Bulma     | <=768 | >=769 | >=1024 | >=1216 | >=1408 |

As you can see by default breakpoints are diferent between providers. And even if we introduce new breakpoints through the theme, these will still be present. As far as I know they can not be removed from css.

stsrki:
PS. maybe we can open new ticket for this conversation?

Demos/Blazorise.Demo/Layouts/MainLayout.razor
```diff
@@ -4,106 +4,126 @@
...
+                    <BarStart>
+                        <BarLabel>Main</BarLabel>
+                        <BarItem>
+                            <BarLink To="">
+                                <BarIcon IconName="IconName.Dashboard" />
+                                Dashboard
+                            </BarLink>
+                        </BarItem>
+                        <BarItem>
+                            <BarDropdown Open="true">
```
MitchellNZ, line 26: @stsrki probably worth mentioning..

As we had previously talked about, I was having a lot of trouble getting `bind-Visible` to work well with this component (due to a combination of the `BarItem` hook method **and** the custom component overrides resetting the `bind` state **and** the internal `Toggle()` method the `BarDropdownToggle` calls).

So I just added this `Open="true"` to indicate a default state. 
Obviously, this means a functionality loss now with the developer not being able to control the `Toggle` state.. but the performance and reliability are much better.

stsrki: I might look into it as I don't like having two APIs for almost the same thing. It can lead to confusion by the users.

MitchellNZ: @stsrki okay - that would be great!
I was thinking the same, so I disabled `Visible` from being `[Parameter]` to save confusion :)

stsrki:
I will do ThemeGenerator sizing next. Please let me know if you need any more help.

Source/Blazorise/BarDropdown.razor.cs
```diff
@@ -29,15 +29,24 @@ protected override void BuildClasses( ClassBuilder builder )
...
+        {
+            // This is needed for the two-way binding to work properly.
+            // Otherwise the internal value would not be set in the right order.
+            if ( parameters.TryGetValue<bool>( nameof( Visible ), out var newVisible ) )
+            {
+                store.Visible = newVisible;
+            }
 
-            base.OnInitialized();
+            return base.SetParametersAsync( parameters );
```
MitchellNZ: @stsrki Thank you for this! I couldn't figure this part out!

Will this no longer be needed once https://github.com/dotnet/aspnetcore/pull/19642 is addressed?

stsrki: No, component activator solves another problem. This is more of how Blazor works with parameters.

Actually, I had a similar problem with `TextEdit `a year ago so I just figured it might work.

Source/Blazorise/Link.razor.cs
```diff
@@ -88,6 +88,7 @@ protected override void OnParametersSet()
 
             absoluteUri = To == null ? string.Empty : NavigationManager.ToAbsoluteUri( To ).AbsoluteUri;
             active = ShouldMatch( NavigationManager.Uri );
+            DirtyClasses();
```
MitchellNZ, line 91: @stsrki I had to add this line to get the `active` class appearing on the `Link` when its within the `BarDropdownItem` in AntDesgin (and only for AntDesgin!)  I cannot figure out why this happening.. 🤔
Appears to be some sort of race condition with this method (`OnParametersSet`) and `OnLocationChanged`?
Seems like this may cause performance issues.. Maybe you could take a quick look, if you get a chance?

stsrki: Maybe we can override Link for AntDesign? I can try it later.

stsrki:
The PR is really big but I think this should now be ready for merge. The only thing left is `Bug: BarMode.VerticalPopout does not work` and I'm not sure what is wrong with that. Maybe I'm looking at the wrong thing?

Anyways I will wait until you give me green light.

stsrki:
These can wait until 0.9.2

- Dynamically created (code created) Bar
- Third level dropdown menu (this works for inline, but not small or popout)
- Support for Light/Dark themes with AntDesign

We can open separate issue for each one.

MitchellNZ:
The bug `BarMode.VerticalPopout does not work` is only for the `AntDesign` provider.
Its annoyingly quite difficult to get it working correctly! (Due to more playing around with custom CSS).

I think its probably best to merge this as it is and I'll create a follow up issue with this bug + the other things I plan on fixing for next release. There are some things linked to the original issue (https://github.com/stsrki/Blazorise/issues/763) I haven't had chance to fix as part of this either.

I don't want to be holding up this release anymore than I already have!
Does that sound good with you?

stsrki:
That sounds great. Honestly this is more than I could imagine, you really did a great job with Bar component and I cannot thank you enough times.

So yeah, I will merge this now. And since I'm little stuck with other work I will not be able to finish release notes for 0.9.1 as fast as I thought. If you happen to do any other small task before release just create new PR. Otherwise we push anything left to 0.9.2.

And once again, THANK YOU!

PR #957, opened 2020-06-04
Fix documentation for some `Is***` (boolean) properties.

StefH:
And I wonder if this documentation for the **FieldLabel** is valid because I don't see these properties in the source code?

image


stsrki:
Thank you for PR. I guess the mentioned properties are leftovers from old Blazorise code when you had to specifically tell it what is the context of label. They can safely be removed from documentation.

StefH:
Ok. I will remove these also.

stsrki:
Thank you!

PR #958, opened 2020-06-06
Add ChartColor.FromHexString() method

StefH:


stsrki:
Hi, can you switch to `dev091` branch? I don't allow PRs into master except for documentation changes.

Also can you explain the reason for the feature and how to use it? Thanks!

StefH:
1. Branch switched

2. I've created a new helper class `HTMLColorCodeParser` which can parse hexcodes and code based on string.

3. Code can be used like:
``` cs
var color = ChartColor.FromHTMLColorCode("AABBCC");
```

or

``` cs
var color = ChartColor.FromHTMLColorCode("AshGray");
```

4. I could not find where to add unit tests.

PR #968, opened 2020-06-10
EmptyTemplate and LoadingTemplate

DDDenisSobek:
Developer has opportunity for setting an empty template or a loading template.
This opportunity is helpfull, because if you load data from REST services for example, then your user can see whether grid is empty or your page is loading.

### DIFF
--- a/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
+++ b/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
@@ -293,9 +293,12 @@ namespace Blazorise.DataGrid
 
         #region Filtering
 
-        protected Task HandleReadData()
+        protected async Task HandleReadData()
         {
-            return ReadData.InvokeAsync( new DataGridReadDataEventArgs<TItem>( CurrentPage, PageSize, Columns ) );
+            IsLoading = true;
+            await ReadData.InvokeAsync( new DataGridReadDataEventArgs<TItem>( CurrentPage, PageSize, Columns ) );
+            IsLoading = false;
+            StateHasChanged();
         }
 
         protected Task OnSortClicked( DataGridColumn<TItem> column )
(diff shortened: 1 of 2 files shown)


Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -668,6 +704,8 @@ protected IEnumerable<TItem> DisplayData
         /// </summary>
         [Parameter] public RenderFragment NextPageButtonTemplate { get; set; }
 
+        [Parameter] public bool IsLoading { get; set; }
```
stsrki, line 707: Maybe `IsLoading` doesn't need to be Parameter?

stsrki:
@DDDenisSobek Nice work!

Can you write an usage example? Here or even better in the documentation? https://github.com/stsrki/Blazorise/blob/master/CONTRIBUTING.md

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
DDDenisSobek: Yeah, i think, i remove the annotation and make it protected.

DDDenisSobek:
> 
> 
> @DDDenisSobek Nice work!
> 
> Can you write an usage example? Here or even better in the documentation? https://github.com/stsrki/Blazorise/blob/master/CONTRIBUTING.md

I will do my best. :)

mnelvers:
Is there a way to set these up so that the text from both the LoadingTemplate and the EmptyTemplate don't show up at the same time?  In the attached image below you can see that both the "Loading..." text from the LoadingTemplate and the "No groups..." text from the EmptyTemplate are display at the same time.

Capture

DDDenisSobek:
> 
> 
> Is there a way to set these up so that the text from both the LoadingTemplate and the EmptyTemplate don't show up at the same time? In the attached image below you can see that both the "Loading..." text from the LoadingTemplate and the "No groups..." text from the EmptyTemplate are display at the same time.
> 
> Capture

Should be fixed at version 0.9.1.1.
#1051 
#1032
Which version do you have? @mnelvers

mnelvers:
We have 0.9.1.  I'll let the team know about the upgrade.  Thanks.

stsrki:
Actually #1051 will in the next patch, 0.9.1.2. Probably in the next few days.

PR #985, opened 2020-06-15
ChartJS : convert Options object to Dictionary to fix null values issue

StefH:


stsrki:
I tried it and it works quite well. But to make it more production ready I think there should be some improvements:

- Naming of JSON fields must be `camelCase`. Currently they are the same as in C# code(eg `ResponsiveAnimationDuration` should be `responsiveAnimationDuration`)
- Maybe even have naming as additional parameter when converting
- Make use of `[DataMember( EmitDefaultValue = false )]` attribute. Currently everything that is not null will be included in dictionary
- Maybe use `DataMember` attribute for naming override, `[DataMember( Name = "test" )] public bool Display { get; set; } = true;`

Source/Blazorise/Utils/Converters.cs
```diff
@@ -1,156 +1,227 @@
...
+                return null;
+            }
+
+            var dictionary = new Dictionary<string, object>();
+
+            foreach ( PropertyDescriptor property in TypeDescriptor.GetProperties( source ) )
+            {
+                var value = property.GetValue( source );
+
+                if ( value != null )
```
stsrki, line 49: Check for `[DataMember( EmitDefaultValue = false )]` attribute here.

Source/Blazorise/Utils/Converters.cs
```diff
@@ -1,156 +1,227 @@
...
+                {
+                    var type = value.GetType();
+
+                    if ( IsSimpleType( type ) )
+                    {
+                        dictionary.Add( property.Name, value );
```
stsrki, line 55: Make use of `[DataMember( Name = "test" )]`, or change the name to camelCase.

StefH: 1. Yes
2. And I'll use camelCasing (Do you already have some c# to do that ? Or should I just add some code in Utils?)

stsrki: Just add one. Maybe this? https://stackoverflow.com/questions/42310727/convert-string-to-camelcase-from-titlecase-c-sharp

Source/Blazorise/Utils/Converters.cs
```diff
@@ -1,156 +1,227 @@
...
+using System.Linq;
+#endregion
+
+namespace Blazorise.Utils
+{
+    public static class Converters
+    {
+        #region Constants
+
+        private static readonly Type[] SimpleTypes =
```
stsrki, line 15: What about nullable types?

stsrki: Also, byte, short, int, long, etc.

Source/Blazorise/Utils/Converters.cs
```diff
@@ -1,156 +1,227 @@
...
+
+            var dictionary = new Dictionary<string, object>();
+
+            foreach ( PropertyDescriptor property in TypeDescriptor.GetProperties( source ) )
+            {
+                var value = property.GetValue( source );
+
+                if ( value != null )
+                {
+                    var type = value.GetType();
```
stsrki, line 51: Try this for Nullable check `var type = Nullable.GetUnderlyingType( value.GetType() ) ?? value.GetType();`

Source/Blazorise/Utils/Converters.cs
StefH: `type.IsPrimitive` in the method **IsSimpleType**  takes care of byte, short, int, long etc.

stsrki:
I have tested and everything seems fine now. Basically you created a basic JSON serializer :)

Thank you for all the work so far. You did a great job!

PR #987, opened 2020-06-16
TreeView component extension

robalexclark:
I've attempted to merge in my treeview component into Blazorise. Its not complete yet because the css classes that I added to Blazorise.scss and compiled use webcompiler never make it into the _content/blazorise.css class. Can you help me understand why?

I guess I will also need to add into the various IClassProvider implemtations for the different providers for TreeView,TreeViewHidden etc. but I need to get the basic version working first.


stsrki:
To compile sccs files you need to have Web Compiler for VS installed: https://marketplace.visualstudio.com/items?itemName=MadsKristensen.WebCompiler

Also, based on the code you already have I think it's best to create new project for TreeView extension instead of having it in the core Blazorise project.

Source/Extensions/Blazorise.TreeView/TreeView.razor.cs
```diff
@@ -0,0 +1,80 @@
...
+
+        [Parameter] public Func<TNode, IEnumerable<TNode>> SetChildNodes { get; set; }
+
+        [Parameter] public IList<TNode> ExpandedNodes { get; set; } = new List<TNode>();
+        [Parameter] public EventCallback<IList<TNode>> ExpandedNodesChanged { get; set; }
+
+        [Parameter] public string ExpandNodeIconClass { get; set; } = "far fa-plus-square cursor-pointer";
+        [Parameter] public string CollapseNodeIconClass { get; set; } = "far fa-minus-square cursor-pointer";
+        [Parameter] public string NodeTitleClass { get; set; } = "p-1 cursor-pointer";
+        [Parameter] public string NodeTitleSelectedClass { get; set; } = "bg-primary text-white";
```
stsrki, line 70: Probably make use of Blazorise `BackgroundColor` and `TextColor` in IClassProvider

Source/Extensions/Blazorise.TreeView/Blazorise.TreeView.csproj
```diff
@@ -0,0 +1,27 @@
...
+
+  <Import Project="..\..\..\Build\Blazorise.props" />
+
+  <PropertyGroup>
+    <PackageId>Blazorise.TreeView</PackageId>
+    <PackageLicenseFile>LICENSE.md</PackageLicenseFile>
+    <PackageIcon>Blazorise.png</PackageIcon>
+    <PackageProjectUrl>https://blazorise.com/</PackageProjectUrl>
+    <RepositoryUrl>https://github.com/stsrki/Blazorise</RepositoryUrl>
+    <PackageTags>blazorise blazor components treeview</PackageTags>
```
stsrki, line 11: Add yourself as author

`<Authors>%Authors%;Niek Jannink</Authors>`

Source/Extensions/Blazorise.TreeView/TreeView.razor
```diff
@@ -0,0 +1,39 @@
...
+            @if ( hasChildren == true )
+            {
+                <span class="tree-view-icon" @onclick="@(() => OnToggleNode(node, !nodeExpanded))">
+                    <i class="@(nodeExpanded ? CollapseNodeIconClass : ExpandNodeIconClass)"></i>
+                </span>
+            }
+
+            <div class="tree-view-title">
+                <span class="@NodeTitleClass @(nodeSelected ? NodeTitleSelectedClass : String.Empty)" @onclick="@(() => OnSelectNode(node))">
+                    @TextContent( node )
```
stsrki, line 21: Maybe better name is `NodeTemplate` or `NodeContent`?

Source/Extensions/Blazorise.TreeView/TreeView.razor
```diff
@@ -0,0 +1,39 @@
...
+            {
+                <TreeView Nodes="SetChildNodes(node)"
+                          TextContent="TextContent"
+                          SetChildNodes="SetChildNodes"
+                          ExpandedNodes="ExpandedNodes"
+                          ExpandedNodesChanged="ExpandedNodesChanged"
+                          SelectedNode="SelectedNode"
+                          SelectedNodeChanged="SelectedNodeChanged"
+                          Visible="nodeExpanded"
+                          HasChildNodes="HasChildNodes" />
```
stsrki, line 35: Can `HasChildNodes` be detected from data without specifying it explicitly?

robalexclark:
I've made as many changes as I can but I am still a bit lost in understanding how the FluentSpacing and Background.Primary classes work - are these something that need building for each provider?

Source/Extensions/Blazorise.TreeView/TreeView.razor
robalexclark: Unfortunately not (at least I could not separate them), The TreeView is (probably) unique in that it is recursively created within itself, and the data class that is provided (TNode) can be anything, the only way I know of passing the child nodes, visibility and whether the child nodes themselves contain child nodes is by referencing them in the component. Really the users would probably never set the visibility of a treeview themselves, but I still need a Visibility parameter hooked up to nodeExpanded (so the child treeviews display when the node is expanded).

stsrki:
> I've made as many changes as I can but I am still a bit lost in understanding how the FluentSpacing and Background.Primary classes work - are these something that need building for each provider?

Sometimes it's hard for me to realize not everyone understand what's going on behind the scenes in Blazorise. I should have probably explain it better :)

Anyways I have made some changes so you can now see from example. It's still not ideal because I wanted to use class builder for node but `bool nodeSelected = node.Equals( SelectedNode );` is giving me trouble.

Source/Extensions/Blazorise.TreeView/TreeView.razor
stsrki: Maybe one way is to have internal component named `_TreeView` that will have those parameters and have `TreeView` as a public component that is a wrapper around `_TreeView`.

Similar concept I have for DataGrid. Also that way we can have CascadingParameter(s) like SelectedNode.

stsrki:
I have made some refactoring, cleaned API and updated the documentation. It should now be ready for merge. When you got time please check one more time on your end if everything works. tnx

robalexclark:
Thanks for cleaning up - it is much better now! To be honest I think I probably got out of my depth with the way it was before, as a treeview contained a treeview which contained a treeview etc.

I have checked that it all works, made some final changes to docs and added a test page description.

stsrki:
I can see there can be more work and improvements, specifically add support for validations, but for now I think it is ready for merge. We can always add modifications later.

Thank you for all your the time and work you put into this. I really appreciate it!

PR #1010, opened 2020-06-24
docs: Fix TreeView case sensitive navigation issue

MitchellNZ:


stsrki:
Not sure how I missed that. Thank you!

PR #1021, opened 2020-06-24
Update Material demo JS to latest version

MitchellNZ:
The Material demo application had out-of-date JavaScript which was causing animations to not work correctly.


stsrki:
I see you used JS files from **djibe** material fork. I guess it's OK since I plan to move to it anyways. We still need to use his CSS files but only when they're finished.

Thank you!

MitchellNZ:
Oh I see! Sorry, I thought you were already using his CSS too.

But it seems to work well with his updated JS :)

stsrki:
Well there is not much of JS using by the Blazorise anyways so I guess it will not hurt us :)

PR #1027, opened 2020-06-26
fix: Server-Side pre-rendering issue with Bar calling JSInterop

### LINKED ISSUE #1026: Bar issue with Server-Side pre-rendering
Due to a JSInterop call in [Bar.razor.cs](https://github.com/stsrki/Blazorise/blob/master/Source/Blazorise/Bar.razor.cs#L48) `OnInitializedAsync`, it is causing server-side pre-rendering to crash the application.

MitchellNZ:
@stsrki opening PR as a patch for `dev091`. 
I can change the target branch if thats not what you want.

Fixes #1026

### DIFF
--- a/Source/Blazorise/Bar.razor.cs
+++ b/Source/Blazorise/Bar.razor.cs
@@ -12,7 +12,9 @@ namespace Blazorise
     {
         #region Members
 
-        // Used to keep track of the breakpoint state for this component.
+        /// <summary>
+        /// Used to keep track of the breakpoint state for this component.
+        /// </summary>
         private bool isBroken;
 
         private Breakpoint breakpoint = Breakpoint.None;
@@ -60,6 +62,7 @@ namespace Blazorise
                     // This needs to be run to set the inital state, RegisterBreakpointComponent and OnBreakpoint will handle
                     // additional changes to responsive breakpoints from there.
                     isBroken = BreakpointActivatorAdapter.IsBroken( Breakpoint, await JSRunner.GetBreakpoint() );
+
                     if ( Visible == isBroken )
                     {
                         Visible = !isBroken;
--- a/Source/Blazorise/Bar.razor.cs
+++ b/Source/Blazorise/Bar.razor.cs
@@ -1,8 +1,4 @@
 ﻿#region Using directives
-using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Net.Http.Headers;
 using System.Threading.Tasks;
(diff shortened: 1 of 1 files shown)


stsrki:
Looks good. The only issue I have is that by the time OnFirstAfterRenderAsync is executed and you change Visible state, possibly it would need to redraw by calling StateHasChanged.

stsrki:
Oh yeah, I switched to new branch. From now **dev** branches are for new stuff, while **sup** will be used for support of published version(s).

MitchellNZ:
> Looks good. The only issue I have is that by the time OnFirstAfterRenderAsync is executed and you change Visible state, possibly it would need to redraw by calling StateHasChanged.

I thought this too, but it seemed work fine when I tested it (possibly due to the `VisibleChanged` being bound)?
Maybe you could please double check this before merging!

stsrki:
I would if I would know what to test 😅. This is your area.

For example what is the usage of lastBrokenState?

One optimization I think will be good is to have if the Visible need to be changed and then refresh

```
if (  Visible != something )
{
    Visible = something;
    StateHasChanged();
}
```

MitchellNZ:
😅
The purpose of this code is control the initial state (collapsed or not) depending on the breakpoint that has been defined.
It should only run on the initial render, and `RegisterBreakpointComponent` will take care of the breakpoint state from there.

I think I will rename `lastBrokenState` to `isBroken`..

It has two simple rules:
- If the Bar "is broken" (isBroken= true) then we want it to collapse (Visible = false).
- If the Bar "is not broken" (isBroken= false) then we want it to be shown (Visible = true).

So I think the following code will fit with your suggestion of adding `StateHasChanged()`:
```
isBroken = BreakpointActivatorAdapter.IsBroken( Breakpoint, await JSRunner.GetBreakpoint() );
if ( Visible == isBroken )
{
    Visible = !isBroken;
    StateHasChanged();
}
```
I will update it!

stsrki:
Maybe the naming should be `collapsed` or something in a way?

Also it could be good to write a comment to explain what it do and why it's used for...

MitchellNZ:
To me, `isBroken` is a little bit more clear, since it is tracking the current breakpoint state.
But I am happy to change it further if you think otherwise :)

I have updated with some comments too.. its a difficult thing to explain, so I hope it makes sense 😅

stsrki:
I think it's OK now. The comments are for us in the future so we know what's going on :)

Good work as always, Thank you!

PR #1032, opened 2020-06-26
Fixes for DataGrid manual data read behavior

smokedlinq:
A couple fixes for the following scenarios:

- New form isn't shown when the data is empty and the edit mode is set to Form
- The empty template and loading template both show when loading and empty
- Hide the pager when the data is empty or while loading data
- Fixed the scenario when manual data read is being used and a new item is added the view isn't refreshed respecting the sorting, filtering, and paging
- Fixed the scenario when manual data read is being used and an item is deleted on a page and the page becomes empty

### DIFF
--- a/Source/Extensions/Blazorise.DataGrid/DataGrid.razor
+++ b/Source/Extensions/Blazorise.DataGrid/DataGrid.razor
@@ -77,9 +77,9 @@
             }
         </TableHeader>
         <TableBody>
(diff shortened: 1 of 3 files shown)


stsrki:
I made some modifications. Can you please re-test before I merge?

smokedlinq:
Looks good @stsrki thank you

stsrki:
No, thank you! :)

PR #1051, opened 2020-06-30
Fix problems with empty and loading template

DDDenisSobek:
Problems:
- Sizing empty or loading template is moving columns of table and is difficult
- When empty or loading template is visible, you can not add new entry
- It is possible, that empty and loading template are visible together

Fix problems with empty and loading template:
- add a full span column row
- empty and loading has full span column row and are moved to table body
- edit row will be visible even if empty or loading template are visible
- empty template will not be visible if loading template is visible


stsrki:
There was already some fixes for this with #1032 

Can you switch to branch `sup091`? There might be some conflicts.

DDDenisSobek:
> 
> 
> There was already some fixes for this with #1032
> 
> Can you switch to branch `sup091`? There might be some conflicts.

Ok.

DDDenisSobek:
@stsrki Do it looks fine for you?

stsrki:
Looks good so far! 

Can I ask you to add some loading simulation to the demo app? Something like a button **Simulate loading** and then it will make use of this changes that you have created. 

Button can be added here:

image

DDDenisSobek:
> 
> 
> Looks good so far!
> 
> Can I ask you to add some loading simulation to the demo app? Something like a button **Simulate loading** and then it will make use of this changes that you have created.
> 
> Button can be added here:
> 
> image

It tried my best for a simulation, but i had to make some restrictions for simulation.
Sorting, editing and filtering has to be disabled, because otherwise i have to implenent sorting, editing and filtering at DataGridPage.

stsrki:
Thanks, I will check it later after work.

stsrki:
I reverted simulation code to previous version as I felt it would complicate things too much. Regarding the loading fixes, they look fine. Thank you for the help!

PR #1055, opened 2020-06-30
Option to hide the toggle icon in a dropdown

njannink:
Option to hide the arrow icon of a dropdown(toggle).


WillianGruber:
Could you also implement it in the ThemeDropdown options? Directly specifying in the component should take precedence over theme, as usual. If not specified, falls back to the value in theme.

WillianGruber:
Could you also implement it in the ThemeDropdown options? Directly specifying in the component should take precedence over theme, as usual. If not specified, falls back to the value in theme.

njannink:
@WillianGruber added the theming support

stsrki:
I made some optimizations 

- renamed hidden css classname to be more in line with bootstrap naming convention. 
- made new method in class provider for icon toggle visibility
- renamed ToggleIconVisible property

PS. I really like snytax for getting the default value based on theme settings `ToggleIconVisible.GetValueOrDefault( Theme?.DropdownOptions?.ToggleIconVisible ?? true )`. Nice and elegant!

njannink:
I think it would be better if the option / theme types where `struct ` and not a `class` then you wouldn't need the null operations (?) and you could write it as:

`ToggleIconVisible.GetValueOrDefault( Theme.DropdownOptions.ToggleIconVisible )`

But that should be a different PR / refactoring

stsrki:
Theme is not created by default so it should be able to be null. If it would be struct then there would be always some default values that would override the properties on component.

stsrki:
Merged! Thank you!

Issue #857, opened 2020-05-03
Datagrid pager


SandroRiz:
If you have, let's say 35 items in your grid and pageSize is 10, you correctly present 4 link for the 4 pages
1 - 2 - 3 - 4
The same if you are on page 2 or 3, but if you are on the last page, you present only 3 links ( 2- 3- 4)
Why ?

Addition: how can I translate the terms Prev, Next, etc and do a templating of some aspect (eg. style of current page, etc.)

Thanks

stsrki:
The reason is that `MaxPaginationLinks` is set to 5. So when you're on page 1 you can go from 1 to 5 page links. When you're on page 4 for example you will still need see max of 5 links but they are counted with offset to the left and to the right, 2-3-4-5-6, with 4 in the middle.

Now it seems there is a slight miscalculation as it didn't take in account maximum number of pages. That is a bug.

SandroRiz:
Thank you.
What about the possibility to edit the pager? I read in the 0.9 release notes 
_Added some new templates for previous and next pagination buttons.
New styling properties to allow custom classes and styles for rows, cells, header and filter._

but I don't find any documentation how to change

stsrki:
That is just to option to replace prev and next links with your own markup. For page numbers there is still no alternate styling available.

robalexclark:
@SandroRiz  You can always override classes in your own css, for example to change the size (padding) of the pager boxes:

```
.page-link {
    padding: .3rem .6rem;
}

```

omar0402:
How can I edit the css class for a paging component that it´s part of a DataGrid component?

robalexclark:
There are probably lots of ways to override the default css provided by blazorise but in the past I have created an overriding css class .page-link {...} in my main site.css file and set the desired padding etc there.

omar0402:
@robalexclark 
Yeah, I noticed that, but I'm talking about the pagination component that you invoke on the dataGrid with the attribute *ShowPager*, I want to center it, but it's applying a div tag with the "justify-content-between" class by default.

PR #1077, opened 2020-07-04
DataGrid default value setter for new items

### LINKED ISSUE #934: DataGrid: Define default value for input on RowInserting
I would like to be able to define a default value for an input field when creating a new record using the DataGrid NewCommand. 

Context: We have a list of "config groups", each of which has a list of "configs". We have a view populating DataGrids for each of these "config groups", allowing for CRUD operations on each of their "configs". When we want to add a new "config" to a "config group", we need context on what "config group" this "config" belongs to.

Problem: I cannot seem to get a NumericEdit template that will pre-populate the CellValue with the Id of the corresponding "config group". This is what I have right now, and its obviously not working:

```cshtml
<DataGridNumericColumn TItem="StitchApplicationSettingModel" Field="@nameof(StitchApplicationSettingModel.GroupIdIfExists)" Displayable="false" Caption="Group Id" Editable="true">
                                            <EditTemplate>
                                                <NumericEdit TValue="int" Value="@((int)( settingGroup.Id ))" ValueExpression="@(() => (settingGroup.Id))" ValueChanged="@(v => ( (CellEditContext)context ).CellValue = v)" />
                                            </EditTemplate>
                                        </DataGridNumericColumn>
```

I can see the value in the NumericEdit field when creating a new record, but when the RowInserting callback is triggered, its CancellableRowChange Value for GroupIdIfExists is always the type default value of 0 (since it is an int). If I manually edit the field, the ValueChanged callback does update the CellValue, but I don't want to rely on the user to modify this field just to update the CellValue that gets passed to the RowInserting callback.

I am not sure what the correct solution for this is... Maybe just provide a DefaultValue parameter that also updates the CellValue when the input field is initialized?

### LINKED ISSUE #949: issue with overriding NewCommand button in Grid
I apologize if this isn't a bug, but I am new to Github so if there is a better place to post this type of thing, please let me know. Bottom line is I am having trouble working with the new button in the grid control. When I click the button, I have a Date Entered column where I would like to default it's value to today's date. I have been unable to find documentation that shows how to "extend" the functionality of that button and write my own code to default the values of the "new" row that is temporarily added when I click the new button.
Another approach I looked at is using EditTemplate. I figure if the value is that default value of 1/1/0001 12:00:00 AM, I could just default it to DateTime.Now.ToString. 

Documentation says we can get or set cell data use CellEditContext, but the below has no effect. I don't get an error either. Its like code isn't there. I still get the default date of 1/1/0001.

                    <DataGridColumn TItem="ProjectLogViewModel" Field="@nameof(ProjectLogViewModel.DateEntered)" Caption="Date" Width="200">
                        <EditTemplate>
                            @{
                                //var DateEntered = ((string)(((CellEditContext)context).CellValue));
                                //DateEntered = DateTime.Now.ToString();

                                ((CellEditContext)context).CellValue = DateTime.Now.ToString();

                            }
                            <TextEdit Text="@((string)(((CellEditContext)context).CellValue))"></TextEdit>
                        </EditTemplate>
                    </DataGridColumn>

Thanks,
D-J

AjaxSantos:
- Add DefaultValue Property to DataGridColumn
- Add SetDefaultFromEditableColumns Methode to DataGrid
- Use SetDefaultFromEditableColumns by DataGrid.OnNewCommand

Resolves #934 
Resolves #949 

@stsrki Should i use DefaultValue also for not editable columns?


Source/Extensions/Blazorise.DataGrid/DataGridColumn.cs
```diff
@@ -262,6 +262,8 @@ public string FormatDisplayValue( TItem item )
         /// </summary>
         [Parameter] public RenderFragment<CellEditContext> EditTemplate { get; set; }
 
+        [Parameter] public object DefaultValue { get; set; }
```
stsrki, line 265: I had in mind for this to be function so it can be set dynamically, like `Func<TItem, object>`.

stsrki:
Don't see a reason to have default value for display values. Just make it for editable fields.

Source/Extensions/Blazorise.DataGrid/DataGridColumn.cs
AjaxSantos: True. I will change it. :)

AjaxSantos: When we use `Func<TItem, object>`, user could set properties on them self.
Then we could move it to DataGrid and use `Func<TItem>` with name of SetDefaultValues.
Like this:
```
<DataGrid ... SetDefaultValues="employee => {
  employee.Salary = 100.0;
  employee.IsActive = true;
}"
```

stsrki: It might be good idea to have `Func<TItem>` on DataGrid but maybe implementing it could be tricky. Try and see if i'ts not too complicated.

Otherwise just have `Func<object>` on DataGridColumn. It's way easier I think.

AjaxSantos: How do you think about actual version?

stsrki: Seems good now. Still not sure about the name `SetDefaultValues`. Maybe `NewItemDefaultValue` would better indicate the meaning of it? Or do you have better name in mind?

Other than that, the code is nice and clean.

AjaxSantos: Hmm, difficult. `SetDefaultValues` call, what the action should do it. `NewItemDefaultValue` call, when action is executed. Maybe `NewItemDefaultSetter`?

stsrki: I like `NewItemDefaultSetter`. We can go with that name.

stsrki:
@AjaxSantos  I must say you did a great job with this one. Thank you!

stsrki:
PS. since you're on a role with changes to DataGrid, wanna have a look at #857 ? 😅

AjaxSantos:
I take a look on it. :)

Issue #998, opened 2020-06-19
Refactor the TextChanged mode for TextEdit and NumericEdit to be configured for each instance


intiharj:
Currently, one can toggle the Blazorise.Options.ChangeTextOnKeyPress value in Statup.ConfigureSerices to control the way all TextEdit and NumericEdit instances behave in an application.  For server-side applications, we usually want this set to false to support a user rapidly typing long text strings.  However, there are common cases where we want the UI to update on each key stroke ( vs waiting for a control to lose focus).  Examples are: update search results as the user types; update a price calculator as the user tweaks a numeric value.  Perhaps this would also be valuable on Slider as well.

Suggested implementation would be something like a TextChangedMode or BindValueMode property on TextEdit and NumericEdit of type Blazorise.TextChangedMode  ( enum TextChangedMode { OnLostFocus = 0, OnInput = 1 }

stsrki:
Thank you for writing the ticket. I just realized there are `KeyDown`, `KeyPress` and `KeyUp` events on `TextEdit`. Numeric still lacks similar events. Hopefully that will work for you for now, until I add the requested feature for all input components.

intiharj:
I tried the KeyX events.  While they tell me what key the user hit, I don't see a way to access the total string that the user sees.  I was able to work around this by subclassing NumericEdit and flipping Options.ChangeTextOnKeyPress to true in MyNumericEdit.OnInitialized().

PR #1096, opened 2020-07-06
Added ChangeTextOnKeyPress on input fields to be able to override global settings

stsrki:
#998 Refactor the TextChanged mode for TextEdit and NumericEdit to be configured for each instance


WillianGruber:
Should numeric edit include that too?

Issue #663, opened 2020-03-26
Switch Component needs Color Property


vmakharashvili:
**Switch Component**
Switch component doesn't have a Color property, which is very useful.

**I'd love to have**
```html
<Switch TValue="bool" Color="Color.Primary">Some Text...</Switch>
```

PR #1097, opened 2020-07-06
Add colors to Switch component

AjaxSantos:
Resovles  #663

@stsrki I would need your help for the coloring the shadow by clicking on switch. Or could you explane me how i can change it?


Source/Blazorise/Switch.razor.cs
```diff
@@ -37,6 +40,20 @@ protected override void BuildClasses( ClassBuilder builder )
 
         protected override string TrueValueName => "true";
 
+        [Parameter]
+        public Color Color
+        {
+            get
```
stsrki, line 46: Do get as one line `get { return color; }`. It's just a convention in Blazorise :)

Source/Blazorise.AntDesign/Styles/_switch.scss
```diff
@@ -0,0 +1,37 @@
+﻿@import "variables";
+
+.ant-switch-primary.ant-switch-checked {
```
stsrki, line 3: It's better to use scss loop instead of setting each color individualy

```scss
@each $name, $value in $theme-colors {
    $color: $value;
    $color-invert: color-yiq($value);

    &.something-#{$name} {
        background-color: $color;
    }
}
```

Source/Blazorise/Switch.razor.cs
```diff
@@ -19,13 +19,16 @@ public partial class Switch<TValue> : BaseCheckComponent<TValue>
...
+        private Color color = Color.None;
+
         #endregion
 
         #region Methods
 
         protected override void BuildClasses( ClassBuilder builder )
         {
             builder.Append( ClassProvider.Switch() );
+            builder.Append( ClassProvider.SwitchColor( color ), Color != Color.None );
```
stsrki, line 31: Change to `ClassProvider.SwitchColor( Color )`

stsrki:
From what I can see you're following naming convention for each provider. I like that. So far there's not any big issue.

Source/Blazorise.AntDesign/Styles/_switch.scss
```diff
@@ -0,0 +1,37 @@
...
+	background-color: $primary !important;
+}
+
+.ant-switch-secondary.ant-switch-checked {
+	background-color: $secondary !important;
```
stsrki, line 8: Try not to use `!important` whenever you can. It can be problematic to override color in user apps.

stsrki:
Shadows are changed with pseudo elements on `<label>`, at least in bootstrap.

image

stsrki:
Also, there is `GenerateInputCheckEditStyles` in ThemeGenerator that is used to change switch color. It should probably be moved to a separate method since now it need to support multiple colors. Similar to buttons.

Source/Blazorise.AntDesign/Styles/_switch.scss
AjaxSantos: Problem is, that color is elsewhere defined and i do not know where and whether i must delete it.
Like here:
grafik
Last color is my color.

stsrki: It's probably set inside of theme generator.

AjaxSantos:
> 
> 
> Also, there is `GenerateInputCheckEditStyles` in ThemeGenerator that is used to change switch color. It should probably be moved to a separate method since now it need to support multiple colors. Similar to buttons.

I guess, i do not understand, what do you mean. Sorry. :(

stsrki:
No worries, I can do that part :)

AjaxSantos:
@stsrki Coloring should be completed - expect changes for ThemeGenerator.

stsrki:
> @stsrki Coloring should be completed - expect changes for ThemeGenerator.

Not sure I understood the last message :) Expect changes from you or from me? I'm perfectly fine to do it.

Demos/Blazorise.Demo.Material/wwwroot/css/material/material.css
```diff
@@ -3099,10 +3099,6 @@ fieldset:disabled a.btn { pointer-events: none; }
 
     .custom-control-input.focus ~ .custom-control-label::before, .custom-control-input:active ~ .custom-control-label::before { opacity: 0.12; -webkit-transform: scale(1, 1) translateZ(0); transform: scale(1, 1) translateZ(0); }
 
-    .custom-control-input:checked ~ .custom-control-label::after { color: #ff4081; }
```
stsrki, line 3102: This should not be removed. Users would still have their own version of material.css so it would not behave the same as in the demo app.

AjaxSantos:
> 
> 
> > @stsrki Coloring should be completed - expect changes for ThemeGenerator.
> 
> Not sure I understood the last message :) Expect changes from you or from me? I'm perfectly fine to do it.

Acutal color of switch will not be shown without !important.
If i understand you right, it could be a problem with the ThemeGenerator (conservation at Source/Blazorise.AntDesign/Styles/_switch.scss), but i do not know how to fix it.
>
>
> Also, there is GenerateInputCheckEditStyles in ThemeGenerator that is used to change switch color. It should probably be moved to a separate method since now it need to support multiple colors. Similar to buttons.
And here i can not suggest, what code should be programmed.

stsrki:
It's OK really, I just asked so that both of us don't work on the same thing. I can continue from now on.

AjaxSantos:
I will look your work, so that i can next time do it self.
Thanks. :)

stsrki:
Done

- Implemented theming for all providers
- Fixed switch styles for Bulma
- Added switch to eFrolic
- No need for `!important`
- Added new demo page for form elements

This should now be ready. Please just check on your side to see if there is any issue left.

AjaxSantos:
@stsrki It is nice to see, how theming works in Blazorise. Also it is cool to see, how sass works, because i have not worked before with sass. I think next time i can do it self. Thank you.

stsrki:
@AjaxSantos Scss is nice to work with, but it have a small learning curve. Nothing too hard.

As of theming, it's not perfect but it works good enough. One thing I don't like about it is the all hard-coding of styles. I have some prototypes for a new theming generator, but it will take some time until I finish it.

stsrki:
@AjaxSantos Did you test updated version? I would merge it after you confirm it's OK.

AjaxSantos:
Oh, i have not thought, that i should check it. xD But i checked it yesterday, it workes fine for me. :)

stsrki:
Going to be merged then :)

stsrki:
@AjaxSantos Thank you for this PR. I know how much work there was, I really appreciated it!

PR #1098, opened 2020-07-07
fix: Resolved Bar issue relating to recent Layout Improvements

MitchellNZ:
`.b-bar-menu` still had `height: calc(100% - var(--b-vertical-bar-brand-height, 64px));`
This was causing the Bar to break when you created it _without_ a `BarBrand`.

Noticed when further testing new layout changes for "header on top" setup.

Also fixed issue with AntDesign bar.


stsrki:
Looking good. Thank you!

PR #1110, opened 2020-07-08
Adding somekind of lazy loading to the tree view

NoOneKnows92:
It's just one line of code, but it helped me to lower the initial loading time. By only loading the children if the root node is expanded.


stsrki:
Looks good. Thank you!

PR #1146, opened 2020-07-15
Theme Builder

WillianGruber:
Theme Page contains all theme properties released so far.
    Jumbotron and Switch are not yet released.
    Reworked App.razor to initialize all properties.
    Reworked MainLayout to not include logic (SOC), and only point to Theme Page.
    As the previous gradient/border radius/color were basically obsolete, I deleted them.


Demos/Apps/ThemeApp/ThemeCustomizer.razor.cs
```diff
@@ -0,0 +1,16 @@
+﻿using Blazorise;
+using Microsoft.AspNetCore.Components;
+
+namespace ThemeApp
+{
+    public partial class ThemeCustomizer : ComponentBase
+    {
+        private string SelectedTab { get; set; } = "General";
```
stsrki, line 8: Move properties to end of class and enclose everything with regions. Example of code style https://github.com/stsrki/Blazorise/blob/master/Source/Blazorise/Badge.razor.cs

Demos/Blazorise.Demo/Layouts/MainLayout.razor
```diff
@@ -165,35 +170,8 @@
                     </BarStart>
                     <BarEnd>
                         <BarItem>
-                            <BarDropdown>
```
stsrki, line 168: Not sure about removing this from the menu. I use it to quickly test theme generator when I change something. I think it's best to return this and then add additional button to open theme builder.

Demos/Apps/ThemeApp/ColorTransparencyEdit.razor.cs
```diff
@@ -0,0 +1,12 @@
+﻿using Microsoft.AspNetCore.Components;
+
+namespace ThemeApp
+{
+    public partial class ColorTransparencyEdit : ComponentBase
+    {
+        private string _color;
```
stsrki, line 7: Also follow the same formatting style.

Demos/Blazorise.Demo/Layouts/MainLayout.razor
WillianGruber: Couldn't that be made in the new page too? I thought it was redundant then. But if you insist, I'll return it.

Demos/Apps/ThemeApp/ColorTransparencyEdit.razor
```diff
@@ -0,0 +1,16 @@
+﻿<Field Horizontal="true">
+    <FieldLabel ColumnSize="ColumnSize.IsHalf">@Label</FieldLabel>
+    <FieldBody ColumnSize="ColumnSize.IsHalf">
+        <Row>
+            <Column ColumnSize="ColumnSize.Is3">
+                <ColorEdit Color="@Color.Substring(0,7)"
+                           ColorChanged="@(x=>{ Color = x + (Color.Length == 9 ? Color.Substring(7, 2) : string.Empty); })" />
```
AjaxSantos, line 7: Hey, if you make an own method for this and give parameter a nice name, it will be prettier than now.

Demos/Blazorise.Demo/Layouts/MainLayout.razor
stsrki: Having it on a new page means I would need to navigate for every change. This is useful for quick theme setup

Demos/Apps/ThemeApp/ColorTransparencyEdit.razor
stsrki: I agree. And it would be easier to maintain.

Demos/Blazorise.Demo/Layouts/MainLayout.razor
WillianGruber: That's what the instant preview card is for, but it is not ready. Anyway, I'll undo the deletion of that item

Demos/Apps/ThemeApp/ColorTransparencyEdit.razor
WillianGruber: Done

Demos/Blazorise.Demo/Layouts/MainLayout.razor
WillianGruber: Done

Demos/Apps/ThemeApp/ColorTransparencyEdit.razor.cs
WillianGruber: done

Demos/Apps/ThemeApp/ThemeCustomizer.razor.cs
WillianGruber: Done

stsrki:
I'm testing now. Since you're basically initializing all theme options you should also set some default colors so it doesn't look different from usual app. See the image(text is too bright, and brand is white, I guess the same is for light theme), 

image

You can find default Bar colors in `Source\Blazorise\Styles\_bar.scss`.

stsrki:
I can see a lot more work to improve theme builder and also theme provider. For example we have default values for some options but they are not good for all frameworks. So that should be separated and maybe have something like `IDefaultThemeOptionsProvider`.

Also I think every option should be nullable so if the option is null it will not be applied. Currently because of this a lot of components are broken and look weird.

So yeah, a lot more work is going to be done here...

WillianGruber:
About the default colors, I agree. It's just I'm still using sidebar, and didn't test with the new vertical bar.
Some other topics to adjust are the max/min/step of sliders, I did what I thought was best.

WillianGruber:
> Also I think every option should be nullable so if the option is null it will not be applied. Currently because of this a lot of components are broken and look weird.

How should this be done? A switch for every property?

stsrki:
You did a great job so far I must admit. I think I can proceed from now on as I will need to figure how to best handle all of the options and maybe there will be some refactoring going on. It's not going to happen any time soon so you're free to modify whatever until I start work. Probably at the end of 0.9.2 cycle or on 0.9.3.

stsrki:
PS. Also @MitchellNZ is doing a lot of work on Bar component so we need to wait until he finish.

stsrki:
> How should this be done? A switch for every property?

Not sure jet. I will need to think through.

WillianGruber:
> 
> 
> I'm testing now. Since you're basically initializing all theme options you should also set some default colors so it doesn't look different from usual app. See the image(text is too bright, and brand is white, I guess the same is for light theme),
> 
> image
> 
> You can find default Bar colors in `Source\Blazorise\Styles\_bar.scss`.

I missed the default alpha values for Bar. Maybe it is better now.

stsrki:
I must say theme builder is coming up just fine :)

Luk164:
This is looking great! Forget dark and light themes, this way users can style their apps in any way they like! Can I request a settings object that can then be stored in an EFCore database? That way it will be easy to save each users settings no matter what device he uses. (Or we can just save it in cookies instead, less hassle)

WillianGruber:
I am saving it in mongodb, maybe @stsrki can implement that when he refactors this.

Luk164:
> I am saving it in mongodb, maybe @stsrki can implement that when he refactors this.

Maybe a single standardized object/structure with [Serializable] annotation? So that it can be just added as a one-way reference to user and EFCore will take care of the rest.

stsrki:
I guess `Theme` object can be used just for that. If needed I can add the `[Serializable]` attribute.

Luk164:
> I guess `Theme` object can be used just for that. If needed I can add the `[Serializable]` attribute.

`[Serializable]` For those who want to serialise it and `{ get; set; }` on all relevant variables for those who want to use Entity Framework. Also `[NotMapped]` for values the EF should ignore.

stsrki:
@WillianGruber A lot of things have changed in Blazorise since this PR and it would be almost impossible to adjust all the conflict and changes. And soon I plan to do more work on theme providers and its optimizations. After which I will probably do the theme builder, inspired by this PR, just with the new added changed.

So I will close this now. Thank you for all the work and it will not go in vain. Just a different route :)

WillianGruber:
No problems, I'm anxious to see your version then!

Issue #1131, opened 2020-07-12
Property unifications


stsrki:
#900, #1087 

Currently input components like TextEdit, NumericEdit, Select, Check etc. are using `Size` enum to define the size.
```cs
public enum Size
{
    None,
    ExtraSmall,
    Small,
    Medium,
    Large,
    ExtraLarge,
}
```

But the problem is that most of the frameworks only support Small, Medium and Large implementations on input element. Also some components like Check and Radio doesn't support any sizing and since they are inherited from the same BaseInput class they also have the Size property.

When I started Blazorise I used Size for everything and obviously now is the time to clean the API. 

There are few ways to proceed.

1. Introduce new enum named `InputSize` that will have only the most supported values, `Small` and `Large`. This is the easiest way as most of the API will stay the same.
2. For every component group add new size enum, like InputSize(for TextEdit, DateEdit, etc.), CheckSize, SelectSize, etc.
3. Leave `Size` as it is and then implement missing sizes in CSS and theme generator. This one requires a lot more work but I'm not sure if it's the right time to do it.

So for now I think best is to go with first choice. Clean the API, unify all sizes across the components and then in one of the next milestone(0.9.4 or 0.9.5) add more options like ExtraSmall and ExtraLarge.

@MitchellNZ and @WillianGruber I would like your opinion before I start on this. tnx

stsrki:
Note for me:

Sizing of Check component https://mdbootstrap.com/snippets/jquery/mdbootstrap/820849#css-tab-view

MitchellNZ:
This sounds like a good plan to me @stsrki 
I really like the idea of unifying attributes where it makes sense, #1087 is a good example!

Also, I'm happy to help with this where I can, if you need it

stsrki:
Yeah I think it's best to just make it simpler now. I started doing sizing of different components but it would just take too long. So it's better to just clear the API instead. And the entire plan for 0.9.x to 1.0 milestone is exactly that, to clear the API so onxe we reach the 1.0 we can say that it's nice and stable.

stsrki:
One thing I cannot decide on which is better. For example for typography components like Text, Paragraph, Heading etc. Is it better to have properties:

**1. Color, Alignment, Transform, Weight**

```cshtml
<Paragraph Color="TextColor.Primary" Alignment="TextAlignment.Right"></Paragraph>
```

OR

**2. TextColor, TextAlignment, TextTransform, TextWeight**

```cshtml
[...]

Issue #1063, opened 2020-07-01
<Heading Size=HeadinSize.IsSomething> generate Non semantic Html


StephaneJourdois:
Hi,

Strange behavior for the Heading Component witch generate non semantic HTML tag 

**Describe the bug**
The Heading component generate a `<p class="h1">` (or else) instead of `<h1>` tag

**To Reproduce**
```
<Heading Size="HeadingSize.Is1">
     My title H1
</Heading>
```
**Expected behavior**
`<h1>My title H1</h1>`

**Screenshots**
![image](https://user-images.githubusercontent.com/22419870/86258471-27d88080-bbbb-11ea-9ca6-f4dd866494bf.png)

Thanks for your great job !

Stephane

stsrki:
`<p class="h1">` is still valid markup. See https://getbootstrap.com/docs/4.5/content/typography/#headings

njannink:
Problem is that certain html interperters and scrapers don't interpret that very well. Eg html to markdown would transform a `<h2>Heading 2</h2>` to `##Heading 2##` and would transform the `<p class="h1">` to plain text. So I agree with @StephaneJourdois that actual `<h*>` tags should be generated.

stsrki:
Fixed. Going to be available in the next preview!

PR #1147, opened 2020-07-15
WIP: Property unification and custom input sizes

stsrki:
#1131 Property unifications
#1063 Heading generate Non semantic Html
#1204 Implement custom Check for Bulma provider

## Breaking changes

- Renamed Heading `TextColor` property to Color
- Removed enums `ButtonSize` and `ButtonsSize` and replaced them with `Size`

## New features

- Implemented custom Switch sizes: `xs, sm, md, lg, xl`
- Implemented sizes for input (Edit) fields, including the Select
- ButtonSize expanded with `xs,  md, xl` sizes
- Created custom Check and Radio for Bulma and eFrolic providers


stsrki:
@MitchellNZ I will merge this PR since there are more things needed to be worked on. At the end of milestone we can see if there is anything left to fix. And also see if the new sizes are consistent across the provider, eg. Button is same as Dropdown, and it is on same level as Input. That kind of things.

PR #1148, opened 2020-07-16
Add IE11 support

tzengshinfu:
1.Replace arrow function with anonymous fuction in whole JS files.
2.Replace const/let keyword with var in whole JS files.
3.Replace template literals with concatenated strings in whole JS files.
4.Add script tag for IE11 in _Home.cshtml.


stsrki:
I appreciate the effort, really, but I don't have any plan to support IE. Especially since Microsoft doesn't support it anyway https://docs.microsoft.com/en-us/aspnet/core/blazor/supported-platforms?view=aspnetcore-3.1. Blazorise must work for all scenarios out of the box, without some special conditions for IE.

Also regarding the arrow functions, const/let keyword and other stuff in java-script, I plan to actually convert everything to never versions of JS that also supports classes and all new features. Maybe ES5 or newer. And supporting two versions of JS code would be a real pain...

I'm really sorry but I cannot accept this PR.

tzengshinfu:
It’s okay. I understand. We need to go to the future. Thanks for your great job.

PR #1227, opened 2020-08-08
Generic Typing for SelectList bound value

DumberDrummer:
- Changed "SelectedValue" type and all other necessary type references to a generic type
- Updated Documentation and example pages to match
- Using just "object" threw errors regarding the eventcallbacks when trying to bind datatypes that were not objects 
- Allows binding of, say, EF Entity properties directly to select list output. 

I have some other changes (improvements?) that exist in my custom version of this component, but I have not been able to get selenium or jekyll to work on windows according to the contributing instructions, and so I decided to limit to a small change for my first contribution.

Thank you for all your hard work, I hope this small bit helps!


stsrki:
Tested and it's working fine. Thank you!

PR #1234, opened 2020-08-10
Validation not being removed from Validations

peterlentine:
**Bug**
There is a bug preventing a validation from being removed from validations.
**Change**
Implemented IDisposable in the Validation class and adding a call to NotifyValidationRemoved when dispose is called.

### DIFF
--- a/Source/Blazorise/Validation.razor.cs
+++ b/Source/Blazorise/Validation.razor.cs
@@ -18,7 +18,7 @@ namespace Blazorise
     /// <summary>
     /// Container for input component that can check for different kind of validations.
     /// </summary>
-    public partial class Validation : ComponentBase, IValidation
+    public partial class Validation : ComponentBase, IValidation, IDisposable
     {
         #region Members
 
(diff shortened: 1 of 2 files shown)


stsrki:
Hey @peterlentine, is it OK if I switch this feature to dev092 instead of sup091? I think it fits better to be in next release.

peterlentine:
> Hey @peterlentine, is it OK if I switch this feature to dev092 instead of sup091? I think it fits better to be in next release.

I've only tested this against dev091 and sup091. I'd have to test against of dev092.

peterlentine:
@stsrki I've tried to test my change against dev092. Unfortunately even without my change on dev092 I'm encountering a separate issue. The validators do not trigger.  I'm not sure what the issue is. If we merge this change into sup091 will it end up making it into the next release?

peterlentine:
@stsrki Not sure what the issue is with dev092, but validation are not working at all for me there. I'm fine with this merging into dev092 in hope that it will be resolved by the time it is released. Are you cool with that?

stsrki:
I tried to see if there is any changes on validation components but there is none. So I have no idea why it wouldn't work. From what I can see there are not that many changes in this PR. 

I will switch branch and see how it goes.

stsrki:
Or maybe not :/

There were some conflicts after switch so I reverted to sup091. Maybe it's better to close this PR and create new one from dev092 from scratch?

stsrki:
Closing in favor of #1269

PR #1235, opened 2020-08-11
Ant design select component improvements

akusmin:
1. Supported states Loading, Allow Clear, Disabled 
2. Clear selection
3. Put label (not value) as a selected item
4. Adjusted component styles


Source/Blazorise.AntDesign/Select.razor.cs
```diff
@@ -178,14 +215,68 @@ static IEnumerable<TValue> ParseMultipleValues( string csv )
...
+                ? items[SelectedValue]
+                : null
+                : null;
+            }
+        }
+        /// <summary>
+        /// Gets or sets values comparer.
+        /// </summary>
+        [Parameter]
+        public IEqualityComparer<TValue> EqualityComparer
```
stsrki, line 257: Can you please tell me what is the reason for EqualityComparer as a parameter in AntDesign since it doesn't exist in base Select component? AntDesign Select is just a custom implementation of the base component and it shoud not be used directly.

Source/Blazorise.AntDesign/Select.razor.cs
```diff
@@ -168,6 +180,31 @@ static IEnumerable<TValue> ParseMultipleValues( string csv )
             }
         }
 
+        private void SetComparer( IEqualityComparer<TValue> comparer )
+        {
+            var _items = new Dictionary<TValue, RenderFragment>( comparer );
```
stsrki, line 185: Maybe just clear member `items` and apply new values directly? Also please remove underscore from var names. tnx

stsrki:
The new features are working really good. Can you just cover the questions I made you before merging? tnx

Source/Blazorise.AntDesign/Select.razor.cs
akusmin: Idea was to use not primitive for Value values but it is not supported by Blazor. So default comparer is enough. I'll remove  this parameter.

stsrki:
I get this error when I navigate to validations page in Demo app

image

akusmin:
> I get this error when I navigate to validations page in Demo app
> 
> image

Should be fixed

akusmin:
@stsrki hi. is there any updates. can you see any issues? will you approve?

stsrki:
I cannot build and run AntDesign demo projects. This is what I get

image

Try setting the **Blazorise.Demo.AntDesign.Server** as startup project and try on your side.

Source/Blazorise/Select.razor.cs
```diff
@@ -209,6 +211,25 @@ public bool Multiple
...
+            {
+                loading = value;
+                Disabled = value;
+            }
+        }
+
+        /// <summary>
+        /// Gets or sets loading property.
+        /// </summary>
+        [Parameter] public bool AllowClear { get; set; }
```
stsrki, line 231: I'm not sure about this property. It's only used on AntDesign and nowhere else. It's best to either remove it or implement the same behavior for all providers.

PR #1242, opened 2020-08-12
Added support for DateTimeOffset conversions

smokedlinq:
We experienced an issue with a Select component being bound to DateTimeOffset and the Convert.ChangeType not supporting it natively, much like Guid. So, I added DateTimeOffset support like the Guid type.


stsrki:
Hi @smokedlinq , can you also add some test to cover new datatype?

smokedlinq:
Sure thing, let me know if you want me to tweak this.

stsrki:
Looking good. 

I just realized you're on sup091 branch. Can you please switch your work to dev092?

smokedlinq:
Going to push this change into the dev092 branch as a different PR since I don't think I can change my source ...

PR #1252, opened 2020-08-13
Adds in focus events for TextEdit

### LINKED ISSUE #1243: Handle OnFocus Events on Inputs
I am trying to update classes on focus in and focus out. The reason for this is because my company has current css styles I cannot change and would like to apply them to Blazorise.

I would like to see all the Focus events supported listed here:
https://docs.microsoft.com/en-us/aspnet/core/blazor/components/event-handling?view=aspnetcore-3.1
`@onfocus @onfocusin @onfocusout`

### Additional context
I am happy to make a PR for this if requested.

Kronos11:
Simply adds in the 

- `@onfocus`
- `@onfocusin`
- `@onfocusout`

events to TextEdit

Fixes #1243


Kronos11:
@stsrki Is this the right approach? It's working on my own branch so just wanted to verify this is the approach you are thinking?

Source/Blazorise/TextEdit.razor.cs
```diff
@@ -1,129 +1,141 @@
...
-        /// <summary>
-        /// The size attribute specifies the visible width, in characters, of an <input> element.
-        /// </summary>
-        /// <see cref="https://www.w3schools.com/tags/att_input_size.asp"/>
-        [Parameter] public int? VisibleCharacters { get; set; }
-
-        #endregion
-    }
-}
+﻿#region Using directives
```
stsrki, line 1: I'm trying to figure why entire `TextEdit.razor.cs` file is marked as changed. You only added three properties...

stsrki:
@Kronos11 Yeah, the PR look good. That's what I was thinking to do. At least until the major rewrite somewhere in the future :)

Source/Blazorise/TextEdit.razor.cs
Kronos11: I'm developing in vscode on a mac, maybe thats it?

stsrki: I tried formatting it. Nothing changes. Maybe it switched to different file encoding?

Kronos11: when I look at it in vscode it just says that I only added those 3 properties, I don't see any whitespace differences.
Encoding is UTF-8 with BOM

stsrki: The same for me, I just checked. UTF-8 with BOM.

Kronos11:
Anything else required here? I can also work on adding it to some other components

stsrki:
@Kronos11 The PR looks good. I need to run it and merge but always something else distract me. If you want to add new things I can wait with merge.

Kronos11:
I can add them to all razor components with an input box if that's ok with you?
Add the parameters to BaseInputComponent?

stsrki:
Good job in PR. Thank you!

PR #1261, opened 2020-08-16
[DO NOT MERGE YET] Color added to radiogroups

andelacolak:


andelacolak:
@stsrki  I've added custom colors to radio button group. Will I add other features such as outline, small and large buttons, block, disabled and so on?

stsrki:
First, you need to switch to `dev092` branch instead of `master`.

Next, I'm still not sure if the best thing is to have just one `Color` property on. I think it's better to have two separate color properties:

1. `ActiveColor` - this one will show for checked Radio buttons
2. `DefaultColor` or juzt `Color` - that will be applied for all inactive buttons

I didn't tried it still, will do it after you switch branches. From what I can see the code should work in current state. Also after bootstrap is fixed, the same needs to be done for other providers.

stsrki:
@andelacolak Any progress on this PR?

PR #1269, opened 2020-08-19
Remove a validation from the parent validation set when it is disposed

Dedac:
For scenarios with a dynamic set of validations, we want to ensure a validation is removed if it has been disposed, so the validation set doesn't keep checking for it, and holding on to it in memory.

This is taken from #1234 and moved to this dev092 branch as mentioned [here](https://github.com/stsrki/Blazorise/pull/1234#issuecomment-675088852)

@peterlentine

### DIFF
--- a/Source/Blazorise/Validation.razor.cs
+++ b/Source/Blazorise/Validation.razor.cs
@@ -81,7 +81,7 @@ namespace Blazorise
                 // To avoid leaking memory, it's important to detach any event handlers in Dispose()
                 ParentValidations.ValidatingAll -= OnValidatingAll;
                 ParentValidations.ClearingAll -= OnClearingAll;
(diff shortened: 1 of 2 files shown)


stsrki:
This is basically the same PR as in #1234 ?

stsrki:
Who takes the credit for the work now? 😅

Dedac:
Yeah, I was talking to @peterlentine, and he asked me to recreate it for him, as you had mentioned

stsrki:
Ah, OK. In that case I will mention both of you in the release notes for this PR. And I will close the old PR.

Thanks for your help btw :)

Dedac:
please do

stsrki:
@Dedac and @peterlentine, it took me a little longer but I finally tested it. It's working good. Thanks for the help!

Issue #1045, opened 2020-06-29
Bootstrap 4.5.0 - ColumnSize="ColumnSize.Is12.OnMobile"  still use 'col-xs-12' (deprecated)


Remy-Be:
Hello :) 

**Describe the bug**
When I use the Columsize.Is12.onMobile property the generated css class is 'col-xs-12', but this one is no longer used in the current version of bootstrap (v4.5.0).

**To Reproduce**
 ```
<Column ColumnSize="ColumnSize.Is3.OnDesktop.Is4.OnTablet.Is12.OnMobile">
        <div style="border:1px solid red;">Box col-xs-12</div>
</Column>
```

**Expected behavior**
Is it possible to replace it with the class 'col-12' to avoid display problems on mobile?

**Screenshots**
Bug :
![image](https://user-images.githubusercontent.com/16755574/86020511-4b1ef680-ba28-11ea-8ddc-16f21f4ca3c4.png)
Expected :
![image](https://user-images.githubusercontent.com/16755574/86020424-2e82be80-ba28-11ea-8240-44c8769a787e.png)

Thank you !

Regards
Rémy

stsrki:
I have no idea how I missed that!? Thank you for letting me know. It will be fixed!

Remy-Be:
You're welcome ! Thank you :)

Issue #1042, opened 2020-06-29
Vertical Bar Improvements (part 1)


MitchellNZ:
Improvements for Vertical Bar modes:

Bugs:
- [x] Toggle button breaks on mobile
- [x] Horizontal Bar expand/collapse not working (see #1091) 
- [x] Bar `100vh;` causes scroll bar in Firefox and Edge (pre-chromium) (see #1092)
- [x] Problem with vertical BarItem longer text (see #1093)
- [x] When you click on item on popup menu, the menu should close. Now it stays open and you need to click outside of it to close it
- [x] When the screen is too short, `VerticalSmall` Bar dropdown menus create an extra scroll bar (even when hidden)
- [x] BarDropdown is not working in Bulma (see #1114)
- [x] `BarIcon` does not work with attributes correctly (such as `Display`, `Class`, `Style`)
- [x] `BarDropdownMenu` popup customisable width in `ThemeBarOptions` (see #1152)
- [x] `BarDropdownToggle` component hover/click does not work over chevron (see #1180)

Features:
- [x] `BarToggler` support 
- [x] Support for multi-level `BarDropdownMenu` (this works for inline, but not small or popout)
- [x] Add bar height (for top bar) to `ThemeProvider`

Documentation:
- [x] Documentation for `BarToggler`
- [x] Documentation for nested `BarDropdownMenus`
- [ ] Documentation for `BarHeight` (horizontal bar)
- [x] Summary comments in code for new parameters (e.g. `Bar` parameter for `BarToggler`, etc)

stsrki:
![image](https://user-images.githubusercontent.com/900302/86060348-d4541e80-ba64-11ea-91bc-d1831af17c1a.png)

MitchellNZ:
_Note for self:_

> When` the screen is too short, VerticalSmall Bar dropdown menus create an extra scroll bar (even when hidden)

This is partially fixed.
Can be fully fixed after #1137

MitchellNZ:
_Note for self:_

>  Support for multi-level BarDropdownMenu

Currently only supported 1 nested level of dropdown.
Will follow this up after 0.9.2.

MitchellNZ:
Issues outstanding after PR part 1 (#1287)

- [ ] AntDesign `BarToggler` fixes (including CSS)
- [ ] AntDesign nested `BarDropdownMenus`
- [ ] Probably some other AntDesign stuff I have missed...
- [ ] Documentation for `BarToggler`
[...]

Issue #1024, opened 2020-06-25
Docs 0.9.2


stsrki:
## Breaking Changes

- [x] Update to .Net 5
- [x] Removed `ModalSize.None`
- [x] `<Divider>` property `Type` renamed to `DividerType?`
- [x] ChartOptions renamed
  - `AxeTicks` to `AxisTicks`
  - `AxeGridLines` to `AxisGridLines`
  - `AxeScaleLabel` to `AxisScaleLabel`
  - `AxeMinorTick` to `AxisMinorTick`
  - `AxeMajorTick` to `AxisMajorTick`
- [x] #1147 Property unification and custom input sizes
- [x] #1227 Generic Typing for SelectList bound value

## Features

- [x] #841 RichTextEdit based on QuillJS
- [x] #1042 Vertical Bar Improvements (part 1)
- [x] #1271 Added the possibility to change the page size and the position of the pager.
- [x] #1371 Reinitialize EditContext on Validations
- [x] #1266 Link button

### Optimizations

- [ ] #1115 Delay the text entered in TextEdit when ChangeTextOnKeyPress is true
- [ ] #663 Add color for Switch component

### DataGrid

- [ ] Add `PageButtonTemplate `
- [ ] Add `CaptionTemplate` to column header
- [ ] `NewItemDefaultSetter` for new items
- [ ] `PopupColumnSize` for sizing edit fields
- [ ] `PopupSize` for sizing edit dialog
- [x] #601 Validation on DataGridColumn
- [x] #306 Validation in datagrid
- [ ] #1106 DataGrid access to FilteredData

### Others

- [ ] #917 Allow Popup Size to be Set for DataGrid edit/new Modals
- [ ] #918 DataGrid. Title property for modal dialog
- [ ] #1064 DataGrid component editing mode needs to be customized
- [x] #1061 Alert visibility two-way binding
- [ ] #1023 Add `Stacked` to chart options
- [ ] #929 Theme Color Alpha
- [ ] #870 DividerType in Theme DividerOptions
- [x] #934 DataGrid: Define default value for input on RowInserting
- [x] #949 Overriding NewCommand button in Grid
- [x] #998 Refactor the TextChanged mode for TextEdit and NumericEdit to be configured for each instance
- [x] #982 Is it possible to put an image in the datagrid Header instead of a string for the caption? similar to the newcommand in the datagridcommandcolumn.
- [x] #1089 Layout Improvements
- [x] #1084 Whole layout overlay
[...]

PR #1287, opened 2020-09-02
Vertical Bar Improvements [WIP]

MitchellNZ:
Fixes issues & features described in #1045 
Any remaining unchecked issues & features there will be carried over into a new issue and later release.
The work left for this PR is described below.

***Known Outstanding Work***
- [ ] AntDesign `BarToggler` fixes (including CSS)
- [ ] AntDesign nested `BarDropdownMenus`
- [ ] Probably some other AntDesign stuff I have missed...
- [ ] Documentation for `BarToggler`
- [ ] Documentation for nested  `BarDropdownMenus`
- [ ] Documentation for `BarHeight` (horizontal bar) 
- [ ] Summary comments in code for new parameters (e.g. `Bar` parameter for `BarToggler`)

@stsrki as mentioned this is unfinished, but if you would like to merge for a preview version, it _should_ hold up fine! (Just make sure you give it a test yourself first 😅)

If you do that, I will start on a new PR for remaining fixes I've outlined above.
If you find anything else I've missed, just add it to this PR or we can start a new issue if you merge it.


Source/Blazorise/BarIcon.razor
```diff
@@ -1,13 +1,15 @@
 ﻿@inherits BaseComponent
-@if (!HasCustomRegistration)
+@if ( !HasCustomRegistration )
```
stsrki, line 2: I'm thinking if `BarIcon `is needed as a separate component. Maybe it's good to implement `@( IconName as RenderFragment )` on an existing `Icon`? 

That would leave us with custom `b-bar-icon` class, and how to detect it needs to be used. Maybe see if Icon is placed inside `BarItem`?

Source/Blazorise/EventHandlers.cs
```diff
@@ -15,4 +17,10 @@ namespace Blazorise
     public delegate void ClearAllValidationsEventHandler();
 
     public delegate void ValidationsStatusChangedEventHandler( ValidationsStatusChangedEventArgs e );
+
+    [EventHandler( "onmouseenter", typeof( MouseEventArgs ), true, true )]
+    [EventHandler( "onmouseleave", typeof( MouseEventArgs ), true, true )]
+    public static class EventHandlers
```
stsrki, line 23: What is the usage of this class?

Source/Blazorise/Stores/BarItemStore.cs
```diff
@@ -27,7 +28,8 @@ public override int GetHashCode()
             // end up having the same hash code for e.g. two instances where one has only noCache set and the other
             // only noStore.
             var result = Active.GetHashCode()
-                ^ ( Disabled.GetHashCode() << 1 ); // increase shift by one for every bool field
+                ^ ( Disabled.GetHashCode() << 1 )
+                ^ ( BarVisible.GetHashCode() << 1 ); // increase shift by one for every bool field
```
stsrki, line 32: Every other bit must be raised by one. In this case it should be 2.

stsrki (commented):
My Visual Studio instance is corrupted from last update so I could not review it from there. Instead I had to use github. Hopefully I didn't miss something...

Overall the PR looks good so far. Here are some issues I detected so far.

Source/Blazorise/EventHandlers.cs
MitchellNZ: I think its mostly for Intellisense.. But I needed support for `onmouseenter` and `onmouseleave`.
Found the solution here: https://github.com/dotnet/aspnetcore/issues/13104

Source/Blazorise/BarIcon.razor
MitchellNZ: I agree! My only issues is that the it needs to apply this class inside any `Bar` entity (`BarItem`, `BarLink`, `BarDropdownToggle`, etc).

As far as I'm aware, `[CascadingParameter]` only allows to flow to immediate children? (So I couldn't have `[CascadingParameter] BarItem` because of this).

Do you know of any other way I can achieve this? 😅

Source/Blazorise/EventHandlers.cs
stsrki: First time I hear about this solution haha. In that case just write a comment describing what is used for.

Source/Blazorise/BarIcon.razor
stsrki: Well, for every BarLink, BarDropdownToggle, etc. there is always placed in at least one BarItem. So I think it doesn't matter how many cascaded BarItem's there is as long as there is one, the classname will be applied.

stsrki: Or just it needs to know that it is placed inside of `Bar`, and that's it. What do you think?

stsrki: Do we transfer this question to another PR with other issues?

MitchellNZ: > Or just it needs to know that it is placed inside of Bar, and that's it. What do you think?

I like this idea, just not sure how to do it other than cascading down everything 🤔

stsrki: Maybe do the `<CascadingValue Value="Bar" Name="Bar">` this one time? 😅

stsrki: Actually, just having `[CascadingParameter] protected Bar ParentBar { get; set; }` is enough. All we need to know it's not null.

MitchellNZ:
> I merged this PR and for the rest of issues with AD you can open new PR. Also when you got some time, could you make note of all the changes so far in the #1024 ?

All changes I have made for this are in #1042.

Would you like me to individually list them? 

Or just list it as:
- [ ] Vertical Bar Improvements #1042

stsrki:
List them as individual items. And if there are any breaking changes also list them on first paragraph. This is mainly so I can keep track of things when writing documentation and release notes for new version.

PR #1297, opened 2020-09-05
Update Slider value when dragged by mouse

Herdo:
The slide currently doesn't update directly when it's dragged around by the mouse.
It's only updated when the drag stops / mouse button signals up.

To caputure mouse input directly, the `oninput` needs to be handled as well (see stackoverflow: [onchange event on input type=range is not triggering in firefox while dragging](https://stackoverflow.com/a/19067260/2132796)).

### DIFF
--- a/Source/Blazorise/Slider.razor
+++ b/Source/Blazorise/Slider.razor
@@ -2,7 +2,7 @@
 @inherits Blazorise.BaseInputComponent<TValue>
 @if ( !HasCustomRegistration )
 {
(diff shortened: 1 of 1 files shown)


stsrki:
It could be problematic for large pages with a lot of controls. The same issue can sometimes be visible for TextEdit when typing too fast. So maybe it would be good to implement the same behavior for `oninput` event on Slider?

Herdo:
Same as `BaseTextInput` - to handle `oninput` and `onchange` separately? I can see where you come from, of course.
But with splitting it and only triggering on one of those, it wouldn't provide full functionality, as described in the stackoverflow answer.
Or are you referring to a different behavior of `BaseTextInput`?

stsrki:
Gonna accept this as it is not a breaking change. And in case there is an issues with slow response we can always optimize it.

@Herdo Thanks for your work!

PR #1310, opened 2020-09-16
Upgrade to .Net5.0 RC1

stsrki:


maliming:
hi @stsrki 

Do you have the plan to publish the RC version of Blazorise packages to nuget?

stsrki:
Hi, it's already published on MyGet. I don't like to publish previews on NuGet because then a lot more people would download it easier and the biggest problem is that live documentation is not in sync with previews. So it would lead to the confusion I think. Getting it from MyGet instead at least you know that it's "nightly" build so you can expect it to be a little off. That being said, every preview is pretty stable so that part is not the problem :)

maliming:
Thanks, @stsrki  I forgot the myget! : )

PR #1325, opened 2020-09-22
Fix for Displayable column in inline edit mode

Jinjinov:
Fix for https://github.com/stsrki/Blazorise/issues/1223 

DataGridColumn Displayable="false" shows column in DataGridEditMode.Inline edit mode


stsrki:
Thank you for taking the time to fix this!

Issue #393, opened 2019-11-29
Resizeable columns (drag-to-resize) in DataGrid.


stsrki:

Jinjinov:
I really need this feature, but I can't find much about resizing in Blazor - I only found this:

https://github.com/dotnet/aspnetcore/issues/14132#issuecomment-533069194

This comment is 1 year old... is there anything new in Blazor since then?

Can I help somehow? (right now I am learning Blazor, but I have 18 years of experience with C#, WinForms, WPF)

stsrki:
It will take a while until I commit to this feature.. If you could do it and create a PR I would be glad to help.

Jinjinov:
After some googling it seems like there is no way to do this in pure C# (or perhaps I don't know what to search for).

It seems like JS interop is the only way to do this - but my knowledge of JavaScript is very limited...

I will take a look, but I can't promise anything...

Jinjinov:
I took a quick look and this is what I found out:

You can use CSS resize property (https://www.w3schools.com/cssref/css3_pr_resize.asp) but it only works with a resize handler in the bottom right corner and it can't be changed.

There are a few alternatives, but they require additional JavaScript:

https://stackoverflow.com/questions/12194469/best-way-to-do-a-split-pane-in-html
https://stackoverflow.com/questions/46931103/making-a-dragbar-to-resize-divs-inside-css-grids

Or, you can use JavaScript libraries:

https://split.js.org/ 
https://jqueryui.com/resizable/
https://github.com/RickStrahl/jquery-resizable

I didn't find a "splitter" component in Bootstrap, but I found one in Metro:

https://metroui.org.ua/splitter.html

Here are some library comparisons:

https://www.w3tweaks.com/8-css-re-sizable-panes.html
https://www.digitaldesignjournal.com/jquery-resizable-div/

I don't like to include libraries, because everything becomes bloated, so I would choose plain JS from the stackoverflow.

What do you think?

stsrki:
I tend to use vanilla JS as much as possible so I would choose that path in this case.

From the links provided splitting is not that complicated, I have even done it in some of my projects. But it might be difficult to add it to DataGrid. or it might be easy. We'll both know once the work starts :)

Jinjinov:
I agree - the use vanilla JavaScript is optimal.

[...]

PR #1328, opened 2020-09-23
Resizable DataGrid preview / prototype - need feedback

Jinjinov:
added blazorise.datagrid.css and blazorise.datagrid.js to Blazorise.DataGrid/wwwroot
added @ref="@ElementRef" to Blazorise/Table.razor
added @ref="@table" to Extensions/Blazorise.DataGrid/DataGrid.razor
added [Parameter] public bool Resizable { get; set; } to Extensions/Blazorise.DataGrid/DataGrid.razor.cs
modified Extensions/Blazorise.DataGrid/DataGrid.razor.cs to invoke blazorise.datagrid.js in OnAfterRenderAsync

I have successfully added https://htmldom.dev/resize-columns-of-a-table/ to Blazorise and it works.

You were right, it is very simple - if you know Blazor, JavaScript and Blazorise - but I don't, so this was quite hard :)

Can you give me some feedback, did I do something wrong, should I change anything?


stsrki:
From a quick glance, it looks clean and simple and I like it. Currently, I'm too busy with work so I will be able to test later in the day.

Jinjinov:
in `blazorise.datagrid.js` I replaced

      const cols = table.querySelectorAll('th');

with

     const cols = table.rows[0].querySelectorAll('th');

because if the table is `Filterable` it has a second row of `TableHeaderCell` and therefore `th` and the resizable js runs 2x

stsrki:
`table.rows[0]` could throw an exception if there is no row in the grid. So it's best to add a check before.

Jinjinov:
like this?

        if (table.rows == null || table.rows.length == 0)
            return;

        const cols = table.rows[0].querySelectorAll('th');

stsrki:
Yes but then the problem is when there is no data when DataGrid is initialized it will not initialize column resizing since you're selecting 'th'. And 'th' does not exist until at least one row is present.

🤔

Jinjinov:
I just tried it with an empty table, no data. I used Chrome dev tools to debug js:

If `ShowCaptions` and `Filterable` are false, then `table.rows.length` is 0
If `ShowCaptions` or `Filterable` is true, then `table.rows.length` is 1
If `ShowCaptions` and `Filterable` are true, then `table.rows.length` is 2

Jinjinov:
in blazorise.datagrid.js I replaced

      resizer.style.height = `${table.offsetHeight}px`;

with

      resizer.style.height = `${col.offsetHeight}px`;

to make the table only resizable form the header

This fixes 2 problems:

#1 - If ShowCaptions and Filterable are false, and table.rows.length is 0 - then the table is not resizable - this should be mentioned in the documentation :)

#2 - the div resizer created by js had the height of the table, but it didn't adjust it's height when you added / removed / filtered rows - it was not responsive

stsrki:
It's working OK now for Bootstrap and Material, but for other providers, it doesn't work. Probably because they have a different table structure or different class names.

stsrki:
I have a new `Splitter` component in works that might be used for this feature. It's based on the same code but with some modifications to make it more flexible. So once I finish it you can try to implement it here?

Jinjinov:
Of course, I can _try_ :)

David-Moreira:
Hello, 

I'm interested in this feature. @Jinjinov do you have the time / plans to end the implementation?

stsrki:
It would be great if this feature was finished as it's really nice to have. As I remember it, there were some issues with dragging where in some cases **mouse up** was not detected and it would still be able to resize even when the mouse button was not pressed.

Also, it was not working for all providers, only for bootstrap and material.

Jinjinov:
@David-Moreira We are using this feature in production for 3 months now and the customer didn't have any complaints about it. Unfortunately I don't have the time to implement this for other providers.

David-Moreira:
> @David-Moreira We are using this feature in production for 3 months now and the customer didn't have any complaints about it. Unfortunately I don't have the time to implement this for other providers.

I see. 
Thank you for the response. I might want to pick up where you've left then when I have the time. Kinda of unknown territory for me, but I'm guessing you can throw some tips if needed. Thanks.

Jinjinov:
@David-Moreira to start, see https://github.com/stsrki/Blazorise/issues/393#issuecomment-696206065

David-Moreira:
> @David-Moreira to start, see [#393 (comment)](https://github.com/stsrki/Blazorise/issues/393#issuecomment-696206065)

I see thanks! 
Interesting how you've chosen to go with vanila js. I also agree with it. But the thing is, sometimes we are just reinventing the wheel, and specialized libraries already exist with most of the common problems solved... Which means that with a new implementation the possibility of eventually finding a problem which most likely is already solved in a specialized library is quite high in my opinion. Then again, I'm not sure how complex or not is the resize api and if there are any pitfalls.

Most likely wouldn't bring a direct dependency with another lib unless there are actually really complex stuff and strski agrees ofc, but at least gonna take a look at some libs for their core implementation. 

Thanks for the tips. I'll let you guys know when I start to take a look at this.

Jinjinov:
@David-Moreira I agree that we shouldn't reinvent the wheel when it comes to complex problems, but when you read https://htmldom.dev/resize-columns-of-a-table/ you will see that writing less than 50 lines of code does not warrant using a specialized library.

David-Moreira:
Hello guys,
I was just about to set up work on this. What's the best strategy to work on this PR? 
Should I clone @Jinjinov's branch and submit changes/PR there?

stsrki:
@David-Moreira It's best to just open a new branch from dev093 and copy @Jinjinov work. There are too many changes to merge here.

stsrki:
Done with #1885 which was inspired by this PR. Thank you @Jinjinov !

Issue #967, opened 2020-06-10
DataGrid - Is it possible to force single-column sorting?


Enritski:
The new ability to sort the GridView by multiple columns is great but, sometimes, it would be good to force sorting to be by a single column at a time.  
Currently, if I click on the first column heading, it sorts by that column as expected.  If I then decide to sort by column 2 instead, clicking the second column heading adds an additional level of sorting so that the rows are sorted by the first columns and THEN by the second.  I know that I can click the first column heading a further twice to reverse and then remove the sorting on column 1, leaving only the second column sort in place but that is not very intuitive for my users.  Is there any way to force the sort to be by one column at a time?
Ideally, I would like there to be an option to turn on / off multi-column sorting but, for now, I just want to have single-column sorting.  Another option might be to follow the approach as used in the DevExpress Blazor GridView, where you have to hold down the shift key to have multi-column sorting.

thuhaile:
is this feature currently in the 0.9.2 preview?

stsrki:
@thuhaile No, it still needs to be done. It's scheduled for 0.9.2 release

PR #1356, opened 2020-10-03
Added sort mode to data grid

NoOneKnows92:
Addressed the feature request of #967. "It ain't much, but it's honest work" ;)

### DIFF
--- a/Demos/Blazorise.Demo/Pages/Tests/DataGridPage.razor
+++ b/Demos/Blazorise.Demo/Pages/Tests/DataGridPage.razor
@@ -59,6 +59,17 @@
                     </Field>
                     <Field>
                         <FieldLabel>
+                            Sort mode
+                        </FieldLabel>
+                        <FieldBody>
+                            <Select @bind-SelectedValue="@sortMode">
+                                <SelectItem Value="@DataGridSortMode.Single">Single</SelectItem>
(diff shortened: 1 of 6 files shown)


stsrki:
Hi, it's working OK for the most part but there are some inconsistencies. For example, if you are in multiple mode and you have sorted by multiple columns, if you switch to single-mode those same columns will stay sorted. So to solve it I think it's best to have two different states for sorting, one for single and one for multi sort. 

Like this

```
DataGridColumn<TItem> sortBySingleColumn
IList<DataGridColumn<TItem>> sortByMultipleColumns
```

or

```
Dictionary<DataGridSortMode,IList<DataGridColumn<TItem>>> sortByColumns
```

I think second options is better and easier to implement. What do you think?

NoOneKnows92:
I don't think that the most use cases would suspect that the grid safes the sorting state, if the sort mode is been switched. I agree that it's a bug when you switch the sort mode and more than one column being sorted. Maybe all columns should be reseted except the last sorted column.

To address the safe state of the sorting i would allow the user to set a default sorting state, which is currently not possible. Maybe it should also give a `SortingContext`, which the programmer can attach or change from outside of the grid if wanted.

Beside than that i would prefere your second approach, because it's more future safe if more sorting modes should be implemented.

stsrki:
I've managed to add support fol holding the sort states for both modes so I think this will be good enough. Thanks for your help :)

Issue #1000, opened 2020-06-20
DataGrid TextAlignment does not affect caption


imtrobin:
Hi reporting,

when I use TextAlignment.Center on DataGridColumn, the values are centre, but the Captions are still left aligned.

0.9.0

PR #1357, opened 2020-10-03
TextAlign also for the table header of data grid

NoOneKnows92:
Added the possibility to also align the table header of the data grid if wanted, which can be seen as a fix to #1000 maybe.


stsrki:
I'm getting errors `AlignHeader` does not exist. Looking at the code I think it's not needed and probably you left it by accident?

NoOneKnows92:
I added this property to fit the current documentation where it says that the property `TextAlignment` only aligns the display cell and if someone wants to align the header alongside the body it needs to be set over the property `AlignHeader`. This would be my understanding of the current documentation and how it can be enhanced without leaving someone behind.

stsrki:
Instead of using AlignHeader I have added new HeaderTextAlignment property. This way we have more freedom to align the header independently of the cell. Also, I created a new TableHeaderCellTextAlignment method so that it doesn't use the same code as in the cell.

All in all, good job and thank you :)

PR #1375, opened 2020-10-13
fix 1280 - bad author names in nuget package

njannink:
Fix nuget author names

Issue: [1280](https://github.com/stsrki/Blazorise/issues/1280)


stsrki:
Thank you!

PR #1395, opened 2020-10-19
Update _DataGridRowEdit.razor

veyond:
Fix cell style and cell class are not apply in edit mode


stsrki:
Hi, can you do this change on the `dev092 `branch instead of on `master`?

Issue #852, opened 2020-05-01
DataGrid EditTemplate get row data


njannink:
At the moment it is only possible to get the edited value through the context in the edit template of a datagrid column. But in some cases I need the entire object to be able to show the right editor and or disable something. Now I have to make a dummy property on the Model that exposes the entire object.

Would be nice if next to `@context` there would also be something like `@rowdata` or `@model` accessible in the EditTemplate

AntonTurba:
I second @njannink's request.  I've been trying to throw together a drop-down column for a prototype Blazor project at work. 
Since I have the drop-down setup inside the EditTemplate, it would be preferable to also store the state/current selected value inside my row object.

Luk164:
@stsrki I have encountered a similar problem. We already have `@context` that is made from a specified field of a object, can the object itself be made available? Like for example the `@model` @njannink mentioned?

David-Moreira:
+1
Have also encountered this situation. Also ended up going with some workaround to grab the data, but seems like it would make sense to have the row/record data available to do some column logic based on the entire row/record.
It seems to me, that each one of the Template's provided in the DataGridColumn, has a different "behaviour". One is exposing a context, another a record, etc...

![image](https://user-images.githubusercontent.com/22283161/99151670-b9d9b500-2694-11eb-9be4-0679385f706d.png)

So a suggestion: they could be normalized behaving the same, exposing their own context with the row record available.
DisplayTemplate -> Display**Context** (provide row record)
FilterTemplate -> Filter**Context** (provide row record)
EditTemplate -> Edit**Context** (provide row record)
CaptionTemplate -> Caption**Context** (provide row record)

Luk164:
@David-Moreira That looks even better. The PR I made is much simpler because I was afraid to make any breaking changes, so I just added the row record value into the `CellEditContext`. BTW the behaviour you proposed is pretty much the same as the WPF and UWP systems use for their datagrids.

David-Moreira:
Nice, did not know that was the system on those data grids, I'm more of a web kinda guy! 
And It's not like everything is created perfectly the first time.
Sometimes breaking changes are just needed. 😆
Let's see what's @stsrki take on this.

stsrki:
[...]

PR #1465, opened 2020-11-14
Add model to EditTemplate context

Luk164:
- This commit adds model to edit template and sets it in InitEditItem inside of DataGrid.razor.cs file. This allows access to parent model of the edited value.

- Related to issue #852

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -244,3 +244,4 @@ ModelManifest.xml
 # jekyll
 _site
 Gemfile.lock
+*.dt
--- a/.gitignore
+++ b/.gitignore
@@ -244,4 +244,3 @@ ModelManifest.xml
 # jekyll
 _site
 Gemfile.lock
-*.dt
--- a/Source/Extensions/Blazorise.DataGrid/CellEditContext.cs
+++ b/Source/Extensions/Blazorise.DataGrid/CellEditContext.cs
@@ -14,5 +14,11 @@ namespace Blazorise.DataGrid
         /// Gets or sets the editor value.
         /// </summary>
         public object CellValue { get; set; }
+
+        /// <summary>
+        /// Gets or sets the parent model of the edited CellValue - to be used in decision making
+        /// TODO: Check if init would not be better than set
+        /// </summary>
+        public object Model { get; set; }
     }
 }
--- a/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
+++ b/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
@@ -186,7 +186,7 @@ namespace Blazorise.DataGrid
 
         protected void OnNewCommand()
         {
-            var newItem = CreateNewItem();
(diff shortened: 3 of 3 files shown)


Luk164:
@stsrki The discussion is getting kinda off-topic in the other thread so I would like to continue our little chat here if you dont mind.

Luk164:
@stsrki Any news? Did you find a better solution? I am technically waiting for this because one of my features requires it.

stsrki:
I think I will go with this idea. While maybe it's not the best it is the easiest to implement and to use. But, instead of make it an `object` type I would make it a generic type as that would be better than cast every time.

Ps. Can you please base your work on dev093 branch rather than master?

Luk164:
@stsrki Sure, I will change it right away. And sorry for the master thing, 093 did not exist yet when I created this and 092 was freshly merged.

Luk164:
@stsrki I tried adding the types and everything broke, maybe try merging and adding it yourself? It seems the types cascade through huge amount of classes and I am not sure how everything works.

stsrki:
Thanks, I will try to do it over the weekend if I have enough time. I'm currently doing some major refactoring/optimizations of Blazorise internals.

Luk164:
@stsrki Once you are done, could you create a preview release? I would like to start implementing and testing some stuff and I don't want to keep switching around the libraries and nugets.

stsrki:
Thats the plan :)

stsrki:
Resolved with #1503. I added a new `CellEditContext<TItem>` with `Item` field that inherits from `CellEditContext`. That way there would not be any breaking changes in case someone casts context to `CellEditContext`.

I will close this PR. Also, I have added your PR to the release notes.

Luk164:
@stsrki Can you also create that preview release nuget please? All I see at the moment on nuget repo is latest-stable (0.9.2), no previews.

stsrki:
Sure. I will release it tomorrow during the day. I still need to finish some small things.

Luk164:
@stsrki I can see the 0.9.2.1 release, but no preview channel for 0.9.3 (And yes I enabled the include prerelase)

stsrki:
@Luk164 Sorry, it took me longer than expected to finish everything. It's now published on MyGet

Luk164:
@stsrki Gotcha, just wanted to make sure. I still do not see it but it probably just needs some time to get through nuget system. I will reply again as soon as I see it appear. Thanks

Luk164:
@stsrki Still no preview channel:

stsrki:
It's on MyGet https://github.com/stsrki/Blazorise#try-preview

Luk164:
@stsrki Thx, I missed that one. Is there any reason why you don't just create a preview channel on nuget?

stsrki:
Having it on NuGet would make it harder for me since live documentation is not in sync with the preview. And there can be a lot of breaking changes between versions. This way when you get it from MyGet at least you know what you can expect.

Luk164:
@stsrki You literally have to tick a box to even see those so I don't think it would be that big of a problem, but it's your choice

Issue #1464, opened 2020-11-13
Provide new object/objectFactory for DataGrid


Luk164:
**Is your feature request related to a problem? Please describe.**
I do not see any way to provide a new object for dataGrid to use for new command. It tries to create one itself but that is not enough for me. I know I can modify it before it is inserted, but that is hacky at best.

**Describe the solution you'd like**
Instead I would like to create a fresh object myself and give it to dataGrid to populate with fresh values. Alternatively (and this does not exclude the previous approach) I would like to be able to provide an object factory to dataGrid instead,

### Additional context
I have a fairly complex entityFramework object with many built in safeties that are causing trouble when dataGrid starts assigning values that have not been properly initialized, and I do not want to create an entirely separate object just for it to save the values and then copy them manually into properly prepared new object.

stsrki:
Have you tried `NewItemDefaultSetter`?

https://blazorise.com/docs/extensions/datagrid/#newitemdefaultsetter

Luk164:
@stsrki Oh my god I am blind, thx for the tip.

Luk164:
@stsrki Weird, for some reason intelliSense is not picking it up. Though that is probably a fault on my side, the latest update screwed my IDE up a little.

stsrki:
I just released the new version 0.9.2 :)

Luk164:
@stsrki Explains a lot :)

stsrki:
Don't forget to read about breaking changes :) https://blazorise.com/news/release-notes/092/

Luk164:
@stsrki For sure, have a coffee on me

stsrki:
Thank you 😉

Luk164:
Well it turns out I still need to be able to provide a new object factory it would seem. Specifically I use lazy-loading proxies in my DB. Considering your code just creates a new object of the required type, this would have two benefits:

- We could use objects w/o a parameterless constructor
- We could use proxies instead of objects themselves

Luk164:
@stsrki Do you think this could be added? It does share some functionality with the NewItemDefaultSetter but it is not the same. Maybe I could completely replace the reference it gives me for setting but then we have wasted resources creating an object from scratch just to be thrown out.

PR #1477, opened 2020-11-18
NewItemFactory parameter added

Luk164:
- I added a parameter for NewItemFactory. This will be used instead of the default creator if specified (not null).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -244,3 +244,4 @@ ModelManifest.xml
 # jekyll
 _site
 Gemfile.lock
+*.dt
--- a/.gitignore
+++ b/.gitignore
@@ -244,4 +244,3 @@ ModelManifest.xml
 # jekyll
 _site
 Gemfile.lock
-*.dt
--- a/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
+++ b/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
@@ -186,7 +186,7 @@ namespace Blazorise.DataGrid
 
         protected void OnNewCommand()
         {
-            var newItem = CreateNewItem();
+            TItem newItem = NewItemFactory != null ? NewItemFactory.Invoke() : CreateNewItem();
 
             NewItemDefaultSetter?.Invoke( newItem );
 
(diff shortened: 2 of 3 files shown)


Luk164:
Related to issue #1464

.gitignore
```diff
@@ -243,4 +243,5 @@ ModelManifest.xml
 
 # jekyll
 _site
-Gemfile.lock
\ No newline at end of file
+Gemfile.lock
+*.dt
```
stsrki, line 247: Is this added by mistake?

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -958,6 +958,11 @@ protected IEnumerable<TItem> DisplayData
         /// </summary>
         [Parameter] public Action<TItem> NewItemDefaultSetter { get; set; }
 
+        /// <summary>
+        /// Function that, if set, is called to create new instance of an item. If left null a default constructor will be used.
+        /// </summary>
+        [Parameter] public Func<TItem> NewItemFactory { get; set; }
```
stsrki, line 964: Maybe name it `NewItemCreator`? A _factory_ kinda points to a similar design pattern but it doesn't feel like it in this case.

.gitignore
Luk164: Yeah, sorry. That is from a time tracking plugin. Feel free to discard.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
Luk164: @stsrki Maybe... Factory normally creates more than one type so you may be right. It does not matter to me in the end as long as I can specify it.

Luk164: Also I I am not sure if we should pass a parameter with the method. It is possible to pass up to 16 params but that would require a LOT of overloads so a single parameter which holds a class with other parameters would be the preferred approach.

Luk164: Changed to creator in new commit.

.gitignore
Luk164: Removed in next commit. No idea Where the GemFile.lock came from so I left it there.

PR #1478, opened 2020-11-19
Dev092 datagrid multiselection

David-Moreira:
Hello @stsrki 
Hope you've been well.
Here's the multiselect implementation I envision. Could you take a look? :)

**Notes:** 
- You talked about exposing a IReadOnlyList with the selected rows. 
I don't think we can actually do this with Blazor parameters can we? The same thing would apply to SelectedRow? No point in it being settable outside of the Grid?

- To implement the top checkbox that selects or unselects all at a time, I had to do some "tricks" with flags to propagate the changes to the rest of the checkboxes, any idea of, how we could do this differently?
Ended up implementing the same behaviour as CommandColumns for clarity to the user. Can provide a multiselect column, which has a default look, and can also provide a template, otherwise multiselect also works without having a multiselect column.

- I noticed the following:
When you use your codebehind .razor.cs
Why do you prefer to inherit over using as a partial? Any special reason?
For example:
_DataGridRowEdit(_DataGridRowEdit.razor) inherits _BaseDataGridRowEdit (_DataGridRowEdit.razor.cs)
Why not just partial class _DataGridRowEdit (_DataGridRowEdit.razor) and partial class _BaseDataGridRowEdit (_DataGridRowEdit.razor.cs)
Just wondering if that's a personal preference, or there's a technical aspect to it. :)


stsrki:
For every changed file that I open in VS I get:

image

stsrki:
@David-Moreira  Can you please base you branch on dev093 instead of master? That would make it easier for me to update, and then merge later. Tnx

David-Moreira:
> For every changed file that I open in VS I get:
> 
> image

Thanks for fixing!

David-Moreira:
> @David-Moreira Can you please base you branch on dev093 instead of master? That would make it easier for me to update, and then merge later. Tnx

Sure! 👍

David-Moreira:
Done.

stsrki:
> Done.

Thanks

stsrki:
There are now some conflicts after I merged the new feature to the branch. Can you please resolve them within your PR so I can continue with the review? Thanks

David-Moreira:
> There are now some conflicts after I merged the new feature to the branch. Can you please resolve them within your PR so I can continue with the review? Thanks

Sure, I'll take a look as soon as I have the time and fix, Thanks.

David-Moreira:
Hello @stsrki ,
Merged and resolved.
Already noticeable perf difference on the datagrid demo. 👍

stsrki:
Thanks, I will do a review as soon as possible.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -871,11 +933,29 @@ protected IEnumerable<TItem> DisplayData
...
 
+        /// <summary>
+        /// Gets or sets currently selected rows.
+        /// </summary>
+        [Parameter] public List<TItem> SelectedRows { get; set; }
+
+        /// <summary>
+        /// Enables multiple selection
+        /// </summary>
+        [Parameter] public bool MultiSelect { get; set; }
```
stsrki, line 944: Might be a good idea to have it as enum `DataGridSelectionMode`

- Single
- Multiple

`public bool SelectionMode { get; set; }`

Source/Extensions/Blazorise.DataGrid/_DataGridRow.razor.cs
```diff
@@ -55,13 +58,31 @@ protected internal async Task HandleClick( BLMouseEventArgs eventArgs )
...
             }
             else
             {
                 await Selected.InvokeAsync( Item );
             }
-        }
+
+            if ( ParentDataGrid.MultiSelect )
+            {
+                if ( multiSelect is object )
```
stsrki, line 76: Why do you check for object here? `multiSelect is object`

Source/Extensions/Blazorise.DataGrid/_DataGridRow.razor.cs
```diff
@@ -88,14 +109,19 @@ protected internal Task OnCancelCommand()
...
+
         #endregion
 
         #region Properties
 
         /// <summary>
         /// Indicates if the row is selected.
         /// </summary>
-        protected bool IsSelected => ParentDataGrid.EditState == DataGridEditState.None && (object)ParentDataGrid.SelectedRow == (object)Item;
+        protected bool IsSelected => ( ParentDataGrid.EditState == DataGridEditState.None && (object)ParentDataGrid.SelectedRow == (object)Item ) || ( ParentDataGrid.SelectedRows is object && ParentDataGrid.SelectedRows.Any( x => (object)( x ) == (object)Item ) );
```
stsrki, line 124: When you check for equality, use extension method `IsEqual`

Source/Extensions/Blazorise.DataGrid/_DataGridRow.razor.cs
```diff
@@ -173,6 +199,11 @@ protected internal Task OnCancelCommand()
         [Parameter] public EventCallback Cancel { get; set; }
 
         /// <summary>
+        /// Activates the multi select command.
+        /// </summary>
+        [Parameter] public EventCallback<(bool, TItem)> MultiSelect { get; set; }
```
stsrki, line 204: Instead of tupple maybe better would be to have a struct or class object?

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -127,6 +138,12 @@ internal void Hook( DataGridAggregate<TItem> aggregate )
 
         protected override Task OnAfterRenderAsync( bool firstRender )
         {
+            if ( !MultiSelect && SelectedRows is object )
```
stsrki, line 141: Also checking `SelectedRows is object`.

stsrki:
PS. All of my latest changes are now included in the PR so it was a little difficult to navigate through your changes 😅

stsrki:
PPS. Looking good so far,

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
David-Moreira: Good Idea. I'll make that change.

Source/Extensions/Blazorise.DataGrid/_DataGridRow.razor.cs
David-Moreira: It's a null check. Same as multiselect != null, just easier to read.
multiselect should be null if checkbox column functionality isn't being used.

Source/Extensions/Blazorise.DataGrid/_DataGridRow.razor.cs
David-Moreira: Agreed. Will make that change.
I implemented the equality as it already was implemented here: 
`(object)ParentDataGrid.SelectedRow == (object)Item;`

Source/Extensions/Blazorise.DataGrid/_DataGridRow.razor.cs
David-Moreira: I like value tuples for simple things like this. 
I guess a struct is fine here, if you're looking for a bit more clarity. I'll make that change as requested.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
David-Moreira: Same thing as before. It's a null check. 
It's kind of a personal preference. Would you like that I change it to != null since that's your standard way of null checking.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -185,7 +202,7 @@ protected Task OnRowDoubleClickedCommand( DataGridRowMouseEventArgs<TItem> event
 
         protected void OnNewCommand()
         {
-            TItem newItem = NewItemCreator != null ? NewItemCreator.Invoke() : CreateNewItem();
+            var newItem = CreateNewItem();
```
David-Moreira: I just noticed. 
Seems like I might have messed up the merge here. I'll fix it.
(It's strange that auto-merge did not consider it.)

Source/Extensions/Blazorise.DataGrid/_DataGridRow.razor.cs
stsrki: Its not so common to see the null check like that.

I think better is to have it like ussual(`name != null`) or in new c# 9

```
if (name is null) { }

if (name is not null) { }
```

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -958,11 +1038,6 @@ protected IEnumerable<TItem> DisplayData
         /// </summary>
         [Parameter] public Action<TItem> NewItemDefaultSetter { get; set; }
 
-        /// <summary>
-        /// Function that, if set, is called to create new instance of an item. If left null a default constructor will be used.
```
David-Moreira: Same, messed up merge. I'll fix it.

David-Moreira:
> PPS. Looking good so far,

Great. I'll make the requested changes in the meantime.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -958,11 +1038,6 @@ protected IEnumerable<TItem> DisplayData
         /// </summary>
         [Parameter] public Action<TItem> NewItemDefaultSetter { get; set; }
 
-        /// <summary>
-        /// Function that, if set, is called to create new instance of an item. If left null a default constructor will be used.
-        /// </summary>
-        [Parameter] public Func<TItem> NewItemCreator { get; set; }
```
David-Moreira: Same, messed up merge. I'll fix it.

Source/Extensions/Blazorise.DataGrid/_DataGridRow.razor.cs
David-Moreira: I'll fix it by using the same convention as you've been already using. `name != null`
Sincerely, between 
`name != null`
&&
`if (name is not null) { }`
I still like the old way the best... they're both "different then" kind of assertions, either way.

David-Moreira:
@stsrki here's the following changes:
- Merge fix that had squashed NewItemFactory parameter added (#1477) commit.
- Changes from review && suggestions.
  - Introduced DataGridSelectionMode enum. Single / Multiple
  - Changed `is object` null check to the standard `!= null` 
  - Changed equality comparison to use IsEquals extension
  - Introduced MultiSelectEventArgs that informs whether the row has been selected or not, instead of ValueTuple.

stsrki:
I think this is almost ready to be merged. Can you write some documentation in https://github.com/stsrki/Blazorise/blob/dev093/docs/_docs/extensions/datagrid.md.

And if you have some time left please add a section "## Features" and write something about the multi-select mode. I think this deserves a mention for your work 😉 https://github.com/stsrki/Blazorise/blob/dev093/docs/_posts/2020-12-15-release-notes-093.md

stsrki:
@David-Moreira I merged it as I'm limited with time before I release the new preview. For the documentation please open new PR:

Excellent work btw. You really did a great job. Thank you!

David-Moreira:
@stsrki  Thanks for the cleanup. looks good. Great!
As for the auto format. Do you know of anything like editorconfig, that can set an auto format solution wide? As my auto-format definitions are different then yours. Actually I had to disable it, since they're different.

Yea, I'll do a new PR with the documentation and release notes section when I have the time. Thank you.
I can also help you with the multi select support. Since I'll be needing it myself.
I should also be making use of the new feature on preview version. So I'll let you know if I find anything's missing or a bug,

**Note:** 
I noticed you've removed:
```
            if ( !MultiSelect && SelectedRows != null )	
            {	
                SelectedRows = null;	
                InvokeAsync( () => StateHasChanged() );	
            }
```
I'm pretty sure I had it in there for when the user turns the DataGrid from Multiple back to Single Selection Mode, the SelectedRows goes back to the original null state.
I'll retest this use case and let you know.

stsrki:
There is already `.editorconfig` in the solution folder. Not sure why it doesn't pick it up.

I removed 

```
if ( !MultiSelect && SelectedRows != null )	
            {	
                SelectedRows = null;	
                InvokeAsync( () => StateHasChanged() );	
            }
```

because I wanted to have SelectedRows unchanged when modes are switched. Now it's working as expected.

PR #1503, opened 2020-11-22
Added Item to CellEditContext

### LINKED ISSUE #852: DataGrid EditTemplate get row data

stsrki:
resolve #852 DataGrid EditTemplate get row data

resolve #1465

### DIFF
--- a/Demos/Blazorise.Demo/Pages/Tests/DataGridPage.razor
+++ b/Demos/Blazorise.Demo/Pages/Tests/DataGridPage.razor
@@ -177,7 +177,7 @@
                                 }
                             </DisplayTemplate>
                             <EditTemplate>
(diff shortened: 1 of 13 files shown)


David-Moreira:
@stsrki Great! Thank you for this!
I noticed the following gramatical errors in bold if you'd like to correct. :)
        /// <summary>
        /// Gets the reference to the model that is **curerntly** in edit mode.
        /// </summary>
        /// <remarks>
        /// Note that this model is used only for reading and you should never update it directly or any
        /// of it's field members. For writing the edited value **ou** must use <see cref="CellEditContext.CellValue"/>.

stsrki:
@David-Moreira Fixed. Thank you!

PR #1505, opened 2020-11-23
Add missing <property>Change EventCallbacks

czf:


stsrki:
Hi @czf, thank you for your work on this but unfortunately, I cannot accept it. The reason is that it will not work. For `Changed` events to work there must be a way to invoke them and they must be invoked programmatically. Blazor unfortunately doesn't work in a way that they will be called without much more work. And even if we somehow manage to do it, we can have more problems with parameters synchronization and refreshing when too much event callback are involved.

PR #1525, opened 2020-11-24
LoadingTemplate RenderFragment for buttons

nielscil:
Possibility to add LoadingContent RenderFragment to a button for more customization.


stsrki:
Hi @nielscil, I tried your code. And while it works, I don't see much reason for it. It seems like overcomplicating something that is already possible with just a few lines of code.

**Example**

```cshtml
<Button Color="Color.Primary" Clicked="ShowLoading">
    @if ( _isloading )
    {
        @("This is a custom loading template")
    }
    else
    {
        @("Click to load")
    }
</Button>

@code{
    private bool _isloading;

    private async Task ShowLoading()
    {
        _isloading = true;

        await Task.Delay( TimeSpan.FromSeconds( 3 ) );

        _isloading = false;
    }
}
```

nielscil:
Hi @stsrki, thanks for your reply. While your example is ofcourse a very simple solution, I would prefer to make the loading construction reusable. As a developer I could wrap this example of yours in a own component but then I would be missing all the other properties of the button (like color, clicked etc). To me the solution provided in de pull request is less complicated then making some kind of decorator with all the available properties of the button.

stsrki:
It's not stopping you from doing that. 

1. If you want to create a wrapper around Button and you don't want to add all the properties you can just use attribute splatting and pass all the parameters down to the Button: https://www.meziantou.net/capturing-unmatched-attributes-and-attribute-splatting-in-blazor.htm

2. Or you can create a Loading component and use it like:

```cshtml
<Button Color="Color.Primary" Clicked="ShowLoading">
    <Loading Active="@_isloading">
        // add content here or create separate rendering templates
    </Loading>
</Button>
```

3. Another solution is to inherit from Button and then register it with DI. This way you can implement your own logic however you want. This is available in the newest 0.9.3 preview

stsrki:
I've been thinking some more about it and it might not be such bad an idea. Please check my latest changes and check if everything is OK.

Also, I would like it if you could also mention this in the button documentation. Thanks.

nielscil:
HI, thanks for reconsidering. I've checked your changes and they look fine by me but maybe we should also rename the method for providing default template from `ProvideDefaultLoadingContent` to `ProvideDefaultLoadingTemplate`. 

I'm going to check how to add this to the documentation this afternoon.

nielscil:
I've added the documentation. Can you please check if you agree with the changes.

nielscil:
@stsrki Do I need to do anything more before you can merge the pull request?

stsrki:
@nielscil I just wanted to think it through before I merge. It's good. 

Thank you!

PR #1526, opened 2020-11-24
Chart :  Update argument from AddLabelsDatasetsAndUpdate to support object (multi-line labels)

StefH:
Chart :  Update argument from AddLabelsDatasetsAndUpdate to support object (multi-line labels)


stsrki:
Merged, Thanks!

PS. for future reference, please work on `dev093`. I only allowed this merge now because it was a small change.

PR #1567, opened 2020-12-02
Repeater component

njannink:
See issue [1562](https://github.com/stsrki/Blazorise/issues/1562)


njannink:
@stsrki found some threading issues when writing the unit tests so dont merge this yet

njannink:
This should be more or less it. Would appreciate your oppinion @stsrki and @MitchellNZ

njannink:
If you agree with adding this component I will add a demo page and documentation

stsrki:
Code wise everything looks good.

Now I would like a real-world example to see what are the real benefits of Repeater and why it is better instead of `@foreach`. So demo and documentation pages would be great :)

njannink:
Ok that is something I will do tomorrow.

Btw why would you remove the nullable directives? It's good practice to start supporting nullable so thats why I added it to this file. Gradually we can then support nullable in the entire project. Like how Microsoft did it

stsrki:
I'm not a fan of nullable directives to tell you the truth. For reference types I want always them to be null so why do I need to mark them as nullable? For value types, we already had that so I'm fine with that. The only real benefit is that it warns you of the potential missing null check, but to make it possible you need to mark EVERYTHING as nullable. It's a pain for me really. We use it on one project where I work and I hate it! Maybe I'm just too old school 😅

njannink:
Problem is that we are making a library. Consumers have the option to turn on nullable checks. But since we don't support it they don't have this possibility. And I agree for UI components its a bit tedious, because generally all parameters are optional by design.

stsrki:
Looking good now. I made some small changes but overall the component is pretty good. Hopefully, it will be usable for people :)

Thank you, and good job! 💪

njannink:
I will be one of the people using it. And we can use it ourselfs aswell for adding the `INotifyPropertyChanged` support to various controls. Thx for your additional input and good we got it included quickly

PR #1569, opened 2020-12-03
Add PageSizeChanged

DDDenisSobek:
I added PageSizeChanged, because if i set PageSize (because of user costumized start page size for example) in manual read mode, PageSize will not be changed at chaging over control because of overwriting from set page size.
So PageSizeChanged give me a chance to change my page size variable for data grid.

Example:
That are changings in DataGridPage.
DataGrid_1
DataGrid_2
DataGrid_3
If you remove setting PageSizeChanged and try to change page size over control, you can see the problem.


stsrki:
Only bugfixes go into the **sup**port branches, and all-new features go into the dev09x branch. I think it's best to move this into dev093

PR #1570, opened 2020-12-03
DataGrid: EmptyCellTemplate

paulosdfigueiredo:
- Implemented EmptyCellTemplate to allow customization of all cells that are null. Updated the docs and added an EmptyCellTemplate to Datagrid Demo.
- EmptyCellTemplate is only applied if DisplayTemplate is not defined for that column.

### DIFF
--- a/Demos/Blazorise.Demo/Pages/Tests/DataGridPage.razor
+++ b/Demos/Blazorise.Demo/Pages/Tests/DataGridPage.razor
@@ -234,6 +234,9 @@
                             @("Create Employee")
                         }
                     </PopupTitleTemplate>
+                    <EmptyCellTemplate>
+                        <Text Style="opacity: .5;">--</Text>
+                    </EmptyCellTemplate>
                 </DataGrid>
             </CardBody>
         </Card>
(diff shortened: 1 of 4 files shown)


stsrki:
What happens for non-nullable types like `int` or `bool`? They are by default `0` or `false` so checking for `null `does not apply for them. Should the EmptyCellTemplate be visible for them also?

paulosdfigueiredo:
I would say that non-nullable won't show EmptyCell. 0 and false are values that you may want to render. anyway you have the DisplayTemplate for those cases where you want different behavior.

stsrki:
I think `EmptyCellTemplate` would make more sense to be defined inside of `DataGridColumn`. That would allow for more customization in case each column would want to be different.

stsrki:
🤔 Now that I think about it more, a `DisplayTemplate ` could be used for more control over column `null` cells. So it's probably all good.

stsrki:
@paulosdfigueiredo I made some changes but overall it's a good PR. Thank you!

paulosdfigueiredo:
I had the same thoughts about adding it to columns, but has you said, there's already DisplayTemplate for those. EmptyCellTemplate should be used exactly when you want this behaviour in more than one column, without having to go through them all.

I decided not to add an Item context, but i agree that's a good thing as it adds more customization possibilities.

Keep up the good work!

PR #1577, opened 2020-12-04
Component files structure

### LINKED ISSUE #1576: Restructure files and components
Move all components into **Components** folder.

Also separate all enums, events arguments, utilities etc. into sešarate files

stsrki:
resolve #1576 

This PR changes the structure of the files in the Blazorise project. Since they are grown too much it was hard to scroll and navigate. With this change, every component is now grouped into a subfolder. Also, utilities, models, enums, event arguments, and other objects are also moved to their own files.

**Example in the solution explorer:**

image

image

@MitchellNZ I would really appreciate your thinking on this.


MitchellNZ:
@stsrki I think this is much better!
Much easier to navigate, and I completely agree with the structure 😁

PR #1579, opened 2020-12-05
Add PageSizeChanged

DDDenisSobek:
I added PageSizeChanged, because if i set PageSize (because of user costumized start page size for example) in manual read mode, PageSize will not be changed at chaging over control because of overwriting from set page size.
So PageSizeChanged give me a chance to change my page size variable for data grid.


stsrki:
Should I add this feature? I think it would be a lot faster.

DDDenisSobek:
Would be nice. Sorry. :(

stsrki:
No problem. Thank you for helping :)

Issue #1240, opened 2020-08-12
How to customize paging button in datagrid


NicolasFVS:
**Is your feature request related to a problem? Please describe.**
I want to customize paging button and i want to use localisation for them in my datagrid

### Additional context
![image](https://user-images.githubusercontent.com/69345012/90017331-aa208e00-dcab-11ea-8476-2cf17df3699e.png)

How to have text in French for example?

Thank-you for your answer

stsrki:
In preview version of Blazorise you can use these templates to customize pagination buttons

- FirstPageButtonTemplate
- LastPageButtonTemplate
- PreviousPageButtonTemplate
- NextPageButtonTemplate
- PageButtonTemplate

NicolasFVS:
Thank-you I will try it

NicolasFVS:
Can you give us an example, we don't know how to use this kind of template?

Thank-you

stsrki:
Here you go :), It's just a standard usage of render templates in Blazor https://docs.microsoft.com/en-us/aspnet/core/blazor/components/templated-components?view=aspnetcore-3.1

```cshtml
<DataGrid TItem="Employee"
            Data="@dataModels"
            ... >
    <PreviousPageButtonTemplate>
        <span style="color: blue; font-size: 2rem;">
            Get back
        </span>
    </PreviousPageButtonTemplate>
    <PageButtonTemplate>
        <span style="color: red; font-size: 2rem;">
            @context.PageNumer
        </span>
    </PageButtonTemplate>
    <NextPageButtonTemplate>
        <span style="color: green; font-size: 2rem;">
            Go forward
        </span>
    </NextPageButtonTemplate>
</DataGrid>
```

Notice that only `PageButtonTemplate` have `@context`, and it's because it needs to have page number for each button.

NicolasFVS:
I thought thats it works like that, but in my project it doesn't recognize PageButtonTemplate. any idea?

![image](https://user-images.githubusercontent.com/69345012/90250535-7759cf80-de3c-11ea-9894-2696ca4c3e67.png)

stsrki:
You probably missed when I mention it's only available in preview. You can get it by following https://github.com/stsrki/Blazorise#try-preview

SimonAndrews:
Would it be possible to provide away to allow full override of the pagination display via a <PagerTemplate> or similar to the other templates in the DataGrid? and provide a context that has the current page and either total records or total pages.

[...]

PR #1580, opened 2020-12-05
Datagrid: ButtonRow

David-Moreira:
Hello @stsrki,

Can you take a brief look? 
I've exposed the ButtonRowTemplate fragment with the CommandContext API, tell me if you like this approach of exposing the same CommandContext as in the CommandContextRow.

I'm not sure if this is possible, but if the new ButtonRowContext could be initialized by Datagrid and then shared by the ButtonRow as well as the new pager we envision, that would be neat.

PS: I can't seem to make my auto format work with the same configurations as you... that sucks...


stsrki:
Hi, I've made large changes in the Blazorise structure and now there are some conflicts in this PR. Can you resolve them before I review it? tnx

David-Moreira:
Hello @stsrki 
merged.

stsrki:
I'm looking at the PR and I noticed some things

- First top pagination is now centered and it looks strange

image

- Clicking on Edit button does nothing
- Yeah, formating is really broken now 😅

Overall, it seems like a good approach, especially by exposing an API through the context. But it feels strange to have named `ButtonRowTemplate` that is inserted into pagination row.

David-Moreira:
- First top pagination is now centered and it looks strange
Nice catch, gotta fix that

- Clicking on Edit button does nothing
No row selected... so it does nothing... ideally there could be a message informing no row is selected?

- Yeah, formating is really broken now
Anything else, other than the top pagination?

- But it feels strange to have named ButtonRowTemplate that is inserted into pagination row.
It does feel strange... thats why I had suggested a Pager Component instead...

Eitherway, feels good to you if I go with this approach shown by the draft?

David-Moreira:
@stsrki

stsrki:
@David-Moreira Sorry I didn't answer before. Generally, I think we're on the right track. You can continue with this approach and we can later clean API if it needs cleaning.

David-Moreira:
@stsrki 
Implemented.
- Clicking on Edit button does nothing
Ended up leaving this to the developer to handle (If we implement something opinated like a message: "No record selected", that can be something the developer does not want in certain sceneries). 
In the demo, we took the approach to show an implementation where the Edit/Delete button will be disabled if no record is selected.

- First top pagination is now centered and it looks strange
Both top and bottom paginations now behave the same. Meaning that if both are selected, there will be two Button Rows.

David-Moreira:
Hello @stsrki,
Merged latest changes and I've added the missing ClearFilter button.
I noticed you mentioned this PR in an issue #1240 while speaking of a pager template, but that isn't what this PR does. At least, not yet, not on this PR.
This PR adds a Button Row Template, so at least the users can further add buttons to the grid, on the pager section, with additional functionality, in addition to having access to the default: New/Edit/Delete/Clear Filter internal commands.
image

After we finish ButtonRow Template, then we would be going with the Pager Template, that further extends upon this, by providing a place where all the functionality from current Pagination and the functionality we are now providing in ButtonRowTemplate is exposed so the developer can customize it himself.

I know you're probably too busy, but any idea when this can be reviewed and merged in into a preview?
I personally need this feature, I understand if this can't be done soon, it's just so I can manage expectations and possibly use this branch temporarly so I can already have access to the feature.

Keep up the good work! Thanks!

stsrki:
Thanks for merging all the changes. So far I was too busy with implementing localization so I wasn't able to do anything else. Now that PR is updated I will review it before doing anything else. Probably later when I get home.

David-Moreira:
@stsrki Fine! ^^ Thank you.
When doing a brief test of the latest merge I noticed that the Large Data option stopped working correctly, at least when first selecting the Large Data option, meaning that probably something is not quite right with the ReadData functionality on the latest dev093. 
So when I select the Larga Data option, it does not trigger ReadData, but when I click Clearfilter for example, it triggers the ReadData and everything seems normal from there.

(Also made sure to checkout dev093 and test it, just so I confirm it wasn't any of this branch changes, problem remains)
I'm not sure if this is intended. I suspect it has to do with commit:
1ec264801af2d2fa4fc822f8cf7680d11caadb5d	Paulo Figueiredo <paulosdfigueiredo@gmail.com>	04/01/2021 23:35:41 +00:00	Moved paginationContext's events subscription from constructor to OnAfterRender.

But I don't quite have the time to be debugging or rollingback to figure it out at the moment. If you could take a look, that would be great.

paulosdfigueiredo:
Hi,

Are you sure that it was working before? 

HandleReadData is only invoked after first render if ReadData.HasDelegate. There is nothing to trigger HandleReadData automatically when ReadData parameter is set/changed, so it only loads the data on any event that invokes it (like the clearfilter you mentioned).

David-Moreira:
Yes pretty sure that I tested it before pulling dev093 and it was working in the way that enabling Large Data would promptly fill the data. However that doesn't mean that it was filling it from ReadData. 

I might just have deduced that the problem is on the wrong place, might not be the ReadData but something that changed on the demo...or somewhere else... 
Like I said I ended up not having time to debug / rollback to test it.

stsrki:
Tried it. It looks good to me so far.

I think the next option would be to remove(hide) the current command buttons once `ButtonRowTemplate `is defined. Either that or have some kind of enum as a parameter, like `DataGridCommandsMode { Default, ButtonRow }`. What do you think?

stsrki:
One thing I noticed. If you write any filter any click **Clear filter** it will reset Datagrid but it will leave the text in the filter input. And it seems this is the old problem. Behaving the same on live demo.

David-Moreira:
> Tried it. It looks good to me so far.
> 
> I think the next option would be to remove(hide) the current command buttons once `ButtonRowTemplate `is defined. Either that or have some kind of enum as a parameter, like `DataGridCommandsMode { Default, ButtonRow }`. What do you think?

Not sure, I follow.
So at the moment, we have this:
```                    
                     <ButtonRowTemplate>
                        @if (showButtonRow)
                        {
                            <Button Color="Color.Success" Clicked="@context.NewCommand.Clicked">New</Button>
                            <Button Color="Color.Primary" Disabled="@(selectedEmployee is null)" Clicked="context.EditCommand.Clicked">Edit</Button>
                            <Button Color="Color.Danger" Disabled="@(selectedEmployee is null)" Clicked="context.DeleteCommand.Clicked">Delete</Button>
                            <Button Color="Color.Link" Clicked="@context.ClearFilterCommand.Clicked">Clear Filter</Button>
                        }
                    </ButtonRowTemplate>
```

What you are suggesting is this?

`                  <ButtonRowTemplate CommandMode="ButtonRow"></ButtonRowTemplate>`

Which internally would generate pretty much the same as in the example above?

David-Moreira:
> One thing I noticed. If you write any filter any click **Clear filter** it will reset Datagrid but it will leave the text in the filter input. And it seems this is the old problem. Behaving the same on live demo.

Seems like the bug was already there, but I can take a look and submit a separate PR with the fix.

David-Moreira:
@stsrki 
For the clear filter bug: #1719
As for the other "bug" I reported. I decided to revert to the commit:

0cb9981d1121f2d5beb70c1dddbd8add1c1f0a91	Mladen Macanovic <mladen.macanovic@gmail.com>	24/12/2020 10:34:18 +00:00	Large data mode added to DataGrid demo

Which has the exact same behaviour.
If when you load the page, you don't interact with the grid and select Large Data the grid goes empty. Which seems to be actually the correct behaviour since you've first provided an empty collection, however, in my opinion there should be a way to properly call ReadData when the switch is made from "Auto" to "Manual". although I don't think that will be a much used use-case eitherway.

If when you load the page, interact with the grid pagination, and select Large Data, then the Grid keeps data... seems like it goes back to page 1 calls page changed and therefore ReadData, which lead me to raise incorrect assumptions the first time.

Sorry about that @paulosdfigueiredo  👍

paulosdfigueiredo:
No problem, we're all here to make things work :relaxed:

That problem you are reporting in demo it's not related with this PR, that is how DataGrid's ManualMode is implemented.

In my opinion to solve that problem, DataGrid's HandleReadData should be public, and if someone decides to change ReadData after first render, they should manually invoke it.

David-Moreira:
> No problem, we're all here to make things work ☺️
> 
> That problem you are reporting in demo it's not related with this PR, that is how DataGrid's ManualMode is implemented.
> 
> In my opinion to solve that problem, DataGrid's HandleReadData should be public, and if someone decides to change ReadData after first render, they should manually invoke it.

Yea that works or maybe something more general like a refresh command that according to the data grid definitions knows how it should refresh the grid.

David-Moreira:
> > Tried it. It looks good to me so far.
> > I think the next option would be to remove(hide) the current command buttons once `ButtonRowTemplate `is defined. Either that or have some kind of enum as a parameter, like `DataGridCommandsMode { Default, ButtonRow }`. What do you think?
> 
> Not sure, I follow.
> So at the moment, we have this:
> 
> ```
>                      <ButtonRowTemplate>
>                         @if (showButtonRow)
>                         {
>                             <Button Color="Color.Success" Clicked="@context.NewCommand.Clicked">New</Button>
>                             <Button Color="Color.Primary" Disabled="@(selectedEmployee is null)" Clicked="context.EditCommand.Clicked">Edit</Button>
>                             <Button Color="Color.Danger" Disabled="@(selectedEmployee is null)" Clicked="context.DeleteCommand.Clicked">Delete</Button>
>                             <Button Color="Color.Link" Clicked="@context.ClearFilterCommand.Clicked">Clear Filter</Button>
>                         }
>                     </ButtonRowTemplate>
> ```
> 
> What you are suggesting is this?
> 
> ` <ButtonRowTemplate CommandMode="ButtonRow"></ButtonRowTemplate>`
> 
> Which internally would generate pretty much the same as in the example above?

I kept thinking on what you had said. I understand now. You want to be able to switch between the existing Datagridcommand columns and the button row using an enum. 

So if both are defined, we could just use the enum to switch between the two. That seems fine to me but there should always be the option to have both active, since the developer might want to use the data command columns for the create edit delete but still use the button row to provide additional functionality like import export refresh some special filter... Etc

Does not seem that important of a feature to me, but useful, in my case at least I will be able to keep the old command buttons defined in the Datagrid, introduce button row, and just use the enum to disable the other ones instead of deleting. 

So default - does nothing
Commands - shows command column buttons but hides button row
Button row - shows button row but hides command column buttons.

I will get on it soon, I might have time tomorrow.

stsrki:
Yes, that was exactly what I was thinking

David-Moreira:
@stsrki Done!
Please take a look! :)

Source/Extensions/Blazorise.DataGrid/_DataGridRowEdit.razor
```diff
@@ -92,14 +92,14 @@ else if ( EditMode == DataGridEditMode.Inline )
         {
             @if ( column.ColumnType == DataGridColumnType.Command )
             {
-                @if ( ParentDataGrid?.Editable == true )
+                @if ( ParentDataGrid?.ShouldRenderCommands() == true )
```
stsrki, line 95: I don't ParentDataGrid will ever be null, so having a nullable check is not needed here.

stsrki:
Looking good to me.

There is something that I should have thought through more. The current **default** is too strange and it's a breaking change in fact. I think better would be:

- Default - show only commands(later a default can be controller with some global settings if needed)
- Commands - specifically show only commands
- ButtonRow - specifically show only button row

What do you think?

Source/Extensions/Blazorise.DataGrid/_DataGridRowEdit.razor
David-Moreira: Agreed. I was just maintaining existing code logic. That happens in more places.

stsrki: I can clean it on other places :)

David-Moreira:
> Looking good to me.
> 
> There is something that I should have thought through more. The current **default** is too strange and it's a breaking change in fact. I think better would be:
> 
> * Default - show only commands(later a default can be controller with some global settings if needed)
> * Commands - specifically show only commands
> * ButtonRow - specifically show only button row
> 
> What do you think?

There definitely needs to be an option that enables both. I might want to keep commands for their current functionality. And additionally use button row to implement my own functionality... Import, export, custom filters etc.. 

I don't see why the default can't do exactly that. Generally you don't expect that the default of something has specific restrictions...it should be more open ended?

Your saying it looks strange because we're currently implementing the crud functionality on both places is that it?

We could customise the demo so that on:

default uses button row for the new button and clear filter and commands for just edit and delete.

On commands shows every command

On button row shows every command on the button row and additionally some custom feature so that the developer understand he can add custom functionality there.

The demo would look better like that.

Source/Extensions/Blazorise.DataGrid/_DataGridRowEdit.razor
David-Moreira: Yea, sure there might even be a very very micro performance gain there.

David-Moreira:
@stsrki  Additionally why do you consider it a breaking change? If people don't provide the button row template everything works as before...?

stsrki:
@David-Moreira You're right. I forgot that it won't show the button row if you don't define a template 🤦‍♂️

David-Moreira:
@stsrki that's fine...😅
Would you like me to do something else on this PR? Want me to clean up the null conditions on the parentdatagrid? 
Or to take me up on the offer to make the suggested changes on the demo, or is it just fine like this?

stsrki:
Just document a new template and add a usage example. If you want you can clean the null conditions but it's not a priority.

After that, I will review it for the last time to make sure there are no issues, but overall I think it's already good.

stsrki:
PS. make it in the demo so that not both modes are visible by default. It can be confusing with all the buttons doing the same thing.

David-Moreira:
@stsrki 
Requested docs + removal of ParentDataGrid null condition.

David-Moreira:
Hello @stsrki ,
hope you've been well. :)

I know you're probably quite busy, but I'd like to request, if you can review this any time soon. 
If it all possible, I'd like to use this soon on the next preview.. :)

Thanks in advance!

stsrki:
It took us some time but it finally paid off :) You did a great job with this PR and I'm sure it will be of great use for a lot of people.

Thank you for all your support so far!

David-Moreira:
No problem! ^^ That's what's great about open source. And you have a great project here. 

Next up should be fully customizing the pager. I'll see when I have time to pick it up. I was also interested on working on finish the resizing datagrid... so one of those eventually ^^

PR #1612, opened 2020-12-16
Added events EditorFocus and EditorBlur for RichTextEdit

tumam:
Added events EditorFocus and EditorBlur for RichTextEdit


stsrki:
Looks good to me. Thank you!

PR #1648, opened 2020-12-28
Multiselect inlineedit bugfix

David-Moreira:
Two bug fixes related to multi select:
- While doing testing, noticed that when a multi select column exists and the GridEditMode is Inline, we have to tell DataGrid, not to treat the multiselect column as a normal column, as it was breaking the New/Edit functionality.
- Also ended up noticing another minor needed alteration in DataGridAggregationRow.


stsrki:
LGTM

Thanks!

PR #1652, opened 2020-12-29
The MaskEdit Start implementation

ricardoromaobr:
The start MaskEdit component implementation. I know it need be better and better.  But if the idea is accept the community can start to help finish. 

at the moment: 

````html
    <MaskEdit EditMask="aaa-999-XXX"  />
````

````html
    <MaskEdit  isCurrency= "true" />
````

Characters accept in the EditMask property: 

a - just a letter will be accept
9 - only digits [0-9]
X - Any character

And the Currency mode. Still need some things, but already works.

### CODE REVIEW

**Comment by stsrki** [comment]
I'm trying the new component and so far it works well, for the most part. There are some improvements that would make it even better.

1. When you start writing (Zip number for example), and you write **12345-6**, the dash is added automatically. But if I hit backspace to clear **6** I expect to also remove the dash. Instead, I need to hit backspace once again to clear the dash or space. Expected: **12345**. Currently, it is **12345-**
2. When you write a large currency number it resets to **zero** every time
3. Instead of `bool IsCurrency` I think the better option is to have `string Currency`. If defined, we know that currency mode is on, and also we know what is the currency symbol.
4. I'm thinking is it better to implement this on `TextEdit` with an additional parameter `EditMask` and/or `Currency`, instead of having it as a separate component ?! But we can do this at the end after all other stuff is done.

Btw, you did a good job so far! I might continue to work on this after I finish some other things.

**Comment by stsrki** [comment]
PS. This might be better options for edit mask

- `a` - Represents an alpha character (A-Z,a-z)
- `9` - Represents a numeric character (0-9)
- `*` Represents an alphanumeric character (A-Z,a-z,0-9)

**Comment by ricardoromaobr** [comment]
Yes. I think so. 
I will try to fix some this issues.

And string Currency, in this case we need to use the provider to the currency that was defined?
The user would need to put more than one input one for Brazil and other for US, for example. 
I think that DotNet has a default value to use culture.
Ok. I will look some things here. 
I'm care of my 2 kids. :)

**Comment by stsrki** [comment]
Cool, take it slow :)

**Comment by ricardoromaobr** [comment]
The currency reset because I use an integer type, and when the value became large it overflowed. :) I think this is just to change the type to long.

**Comment by stsrki** [comment]
It's better to have it as `decimal`. It makes more sense for currency.

**Comment by ricardoromaobr** [comment]
I think to add a property decimal CurrencyValue, that already has a decimal value of the input value, be used. So the user doesnt need make conversion to use it. But if it will be integrate on TextEdit I not know if is a good idea?

**Comment by ricardoromaobr** [comment]
I use a integer as intermediate repository just to format the value. Because decimal remove the less significant zero on decimal part. I started with decimal and change to integer. Because a take each part separated the the integral part and decimal part. and format each part separated and  concatenate  it after to result. 

Btw, it will be a value only when it were a currency define. Maybe as nullable type decimal?.

**Comment by ricardoromaobr** [comment]
Ok, this little list is made. :)

**Comment by ricardoromaobr** [comment]
If I remember TextEdit already use EditMask but in the diferent way. 
There is a different semantic in the use. 
And use the javaScript to do the mask.
The use of the EditMask property on TextEdit is  a little different.

**Comment by ricardoromaobr** [comment]
Is it for now. 
I not created a property  decimal CurrencyValue  yet. What do you think? 
What I'm thinking is:
```csharp
    // after edited and Currency was defined 
    decimal itemPrice =  ItemPriceMaskEdit.CurrencyValue;

```

stsrki:
Thanks, I will check the code as soon as possible and let you know.

ricardoromaobr:
Ok. I look in issues and I think that must  of the relate problem will be covered. 
And  is not necessary to be leading with regex express for simple things like I see in the issues.
 :)

stsrki:
@ricardoromaobr Sure. I'm currently in the middle of doing some larger work on Blazorise so after I finish I will have more time to test and review your code.

stsrki:
I might try modifying your code and move it into TextEdit and hopefully ditch the javascript 🤞

ricardoromaobr:
:), will be very nice! There are some friends that are waiting for it. Here, we are waiting the new release would be a new preview to use this new features we have severals TextEdit to be masked.

stsrki:
@ricardoromaobr Sorry for taking this long with this feature. The problem is that it needs a lot more work to be done to make it stable enough. So I think it's best to take another approach and instead of doing everything in C# I will make use of an external JS library that is already proven. In the 095 milestone, we learned a lot to make use of JS modules without making it complicated to the users so this is the best way I think.

I have made it for the 1.0 milestone and so I will close this now.

PR #1673, opened 2021-01-06
Minor improvement to EditContextValidator

janhegner:
* EditContextValidator.TryGetValidatableProperty now respects
  ErrorMessageResourceName
* EditContextValidator.ErrorMessageResourceName now works with
  custom ErrorMessage-attribute
* fixed some typos in comments

I encountered two different issues with the Blazorise validator when used together with DataAnnotations:
1. If the ErrorMessageResourceName property is set in one of the DataAnnotations, the error message must be null. Therefore the EditContextValidator must leave the ErrorMessage property untouched.
See here: https://docs.microsoft.com/en-us/dotnet/api/system.componentmodel.dataannotations.validationattribute.errormessagestring?view=net-5.0#remarks
(Remarks section)
2. If the DataAnnotation contains a custom Error-String in the data model. E.g. 
```
[StringLength(50, ErrorMessage = "The {0} must be at least {2} and at max {1} characters long.",
            MinimumLength = 2)]
public string LastName { get; set; } = "";
```
Then the ValidationContext.ErrorMessage is initialized with the custom message by default.
Therefore the EditContextValidation-class should also perform its parameter and format string retrieval magic.


stsrki:
I have tested your change and so far there is no problem with validations. Hopefully, I didn't miss something :)

In any case, thank you for the PR, and sorry it took so long to merge!

janhegner:
Thank you for merging...  no need to apologize.

PR #1692, opened 2021-01-11
Add OpenReadStream for IFileEntry and FileEdit

### LINKED ISSUE #1681: Add OpenReadStream to FileEntry
Would it be possible to add an `OpenReadStream()` to `FileEntry` so the File can be streamed to the Application rather than read in all at once? This would mirror the API found in the [`InputFile`](https://docs.microsoft.com/en-us/aspnet/core/blazor/file-uploads?view=aspnetcore-5.0#inputfile-component) component in Blazor 5.0.

Meberem:
Based largely on the implementation from AspNet with some small tweaks for Blazorise. 

It does require the `System.IO.Pipelines` package which might be unwarranted at this point.

Closes #1681

### DIFF
--- a/Source/Blazorise/Blazorise.csproj
+++ b/Source/Blazorise/Blazorise.csproj
@@ -16,4 +16,8 @@
     <None Include="..\..\NuGet\Blazorise.png" Pack="true" Visible="false" PackagePath="" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Include="System.IO.Pipelines" Version="5.0.0" />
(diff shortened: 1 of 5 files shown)


stsrki:
Looks good. Although I'm not sure about adding `System.IO.Pipeline` package. I want to keep the packages only to basic. Any change you can do this with regular IOs?

Also, you created PR for `master` instead of `dev093`, where all new features are going. I keep the master clean until the milestone is done. Can you move your changes to a `dev093` branch?

Once again, good job :)

PR #1694, opened 2021-01-12
Add OpenReadStream for IFileEntry and FileEdit

### LINKED ISSUE #1681: Add OpenReadStream to FileEntry

Meberem:
Based largely on the implementation from AspNet.

Not quite sure what happened with the last PR. After rebasing onto dev093 there is no need for System.IO.Pipelines!

Closes #1681


stsrki:
There was a major refactor on dev093 so you probably had problems with that.

Anyways, good work. I will review it as soon as possible.

stsrki:
@Meberem 

I'm testing it now but I get the `Object reference not set to an instance of an object.` error.

Here is my code:

```cs
private static int OneMb = 1024 * 1024;
async Task OnFileChanged( FileChangedEventArgs e )
{
    try
    {
        var file = e.Files.FirstOrDefault();

        var buffer = new byte[OneMb];
        using ( var bufferedStream = new BufferedStream( file.OpenReadStream( long.MaxValue ), OneMb ) )
        {
            int readCount = 0;
            int readBytes;
            while ( ( readBytes = await bufferedStream.ReadAsync( buffer, 0, OneMb ) ) > 0 )
            {
                Console.WriteLine( $"Read:{readCount++} {readBytes / (double)OneMb} MB" );
                // Do work on the first 1MB of data
            }
        }
    }
    catch ( Exception exc )
    {
        Console.WriteLine( exc.Message );
    }
    finally
    {
        this.StateHasChanged();
    }
}
```

Maybe I'm using it wrong?

Meberem:
Hey, sorry for the delay. I just tried your brnach in my setup and I also get a NullReferenceException. I note however this seems to be from the [`TextLocalizer`](https://github.com/stsrki/Blazorise/blob/dev093/Source/Blazorise/Localization/TextLocalizer.cs#L101) service, or rather its localizerService not having a `SelectedCulture`. 

image

I may also have my setup wrong too but I just have a plain Blazor Server app with 
```
services
    .AddBlazorise()
    .AddBootstrapProviders()
```

On a slight aside, would it be possible to setup the Blazor Server Demos so it can be debugged from Visual Studio? I found that breakpoints weren't being hit because the Referenced WebAssembly Version had setup WebAssembly debugging.

stsrki:
Hmm, that's strange. Localization should work automatically. There's no need to add anything else that you already don't have.

Also, debugging should also work. I use it all the time for WebAssembly and for Server-side apps.

stsrki:
@Meberem I found what the problem is with NullReferenceException. `I didn't specify the public CultureInfo SelectedCulture { get; private set; } = CultureInfo.DefaultThreadCurrentUICulture;` and so SelectedCulture was null when it was used. The demo worked because I have set it on the main layout 😅

It will be fixed!

stsrki:
Still having error

image

Meberem:
I pulled the latest but I had to change your change to 
```
public CultureInfo SelectedCulture { get; private set; } = CultureInfo.DefaultThreadCurrentUICulture ?? CultureInfo.CurrentUICulture;
```

as apparently my `CultureInfo.DefaultThreadCurrentUICulture` is always `null`, not quite sure why that would be the case.

As for the null ref, it is to do with the `var file = e.Files.FirstOrDefault();`, after the file is uploaded, the handler is called again with an empty `Files` property, so the fix in my sample looked like this
```diff
async Task OnFileChanged( FileChangedEventArgs e )
{
    try
    {
        var file = e.Files.FirstOrDefault();
+       if (file == null)
+       {
+          return;
+       }

        var buffer = new byte[OneMb];
        using ( var bufferedStream = new BufferedStream( file.OpenReadStream( long.MaxValue ), OneMb ) )
        {
            int readCount = 0;
            int readBytes;
            while ( ( readBytes = await bufferedStream.ReadAsync( buffer, 0, OneMb ) ) > 0 )
            {
                Console.WriteLine( $"Read:{readCount++} {readBytes / (double)OneMb} MB" );
                // Do work on the first 1MB of data
            }
        }
    }
    catch ( Exception exc )
    {
        Console.WriteLine( exc.Message );
    }
    finally
    {
        this.StateHasChanged();
    }
}
```

stsrki:
`public CultureInfo SelectedCulture { get; private set; } = CultureInfo.DefaultThreadCurrentUICulture ?? CultureInfo.CurrentUICulture;` That makes more sense actually.

OK, so we need to prevent it from calling the changed event a second time.

stsrki:
I investigated more and it's working as expected. The Changed event is called a second time after the upload is finished with `await fileEdit.UpdateFileEndedAsync( fileEntry, offset == fileEntry.Size );`. And that is OK I think.

So, it seems PR is now ready to be merged :) 

Can you just write a simple example in the file edit docs?

Meberem:
@stsrki let me know if you want any changes to the documentation :)

Thanks!

stsrki:
Looking good now. Excellent work! Thank you for all the help so far :)

PR #1702, opened 2021-01-13
Close Alerts and Modals if CloseButton is within them

Meberem:
Updating docs to reflect this change.

I hope you don't mind this kind of change. I saw that ParentModal and ParentAlert was available but not being used so thought this change would be a small quality of life improvment for this project.

I also wonder what the prefered format should be for listing components in the documentation. In Modal components are listed with brackets i.e. <Field> but in the Alert component they are listed without i.e. AlertDescription. Would you be able to offer any guidance on this?


stsrki:
Hmm in theory I like it. But, it can lead to some nasty side-effect. What if we don't want to close modal automatically. In fact, I already know it will introduce a lot of "bugs" on some of my client's projects. Auto-close would skip current closing validation where close logic was under the clicked event handler.

stsrki:
I will need to think this through. We already have `Closing` handler of a Modal that can be used to not close the modal until it's ready. The same logic can be applied on Alert. But either way, it's a breaking change with a lot of manual work required for every project.

Meberem:
I did see that there was the `Closing` handler on the Modal, which this is correctly used/called with the `Close` method. I can add this to the `Alert` component if so desired, but I do understand how this could be a problem with existing clients.

I remember using [SetCompatibilityVersion](https://docs.microsoft.com/en-us/aspnet/core/mvc/compatibility-version?view=aspnetcore-2.2) for ASP MVC project. Do you think it could be possible to adopt something similar here? Where the new default behaviour is opt in?

stsrki:
It could be done with something like a compatibility version, or similar. But that feels like a hack to me.

stsrki:
@Meberem I modified the behavior of CloseButton.

- Added AutoClose parameter
- AutoClose is true by default
- Added AutoCloseParent to global option so that we can disable the new feature
- Fixed demo and docs examples
- Added tests

Please test one more time and see if this is OK now. Thanks

Meberem:
Sorry it took a while, but yes all looks good from here. Thank you for the update

PR #1708, opened 2021-01-15
Localization support

### LINKED ISSUE #1445: Allow translation for elements with text
Hi, First many thanks for your work is this beautiful nuget package. I use it in 2 projects and it makes me win a lot of time.

### Missing feature 

I'm a French developer, and some component like File or DataGrid have some text that comes by default (browse for the file, pagination text for DataGrid).

Do you have already a solution to translate the text or planned to make on update to allow change the text of the components ?

### Needed solution

If you don't already have the solution, the solution is to be able to change the text or to update the template to be able to change the text.

### Screenshots

image

image

stsrki:
resolves #1445 Allow translation for elements with text


AlexTeixeira (commented):
Seems ok from my point of view. If it works using the current browser or cookie culture it's perfect. I'm ready to test and give feedback

stsrki:
Sure, I will release it soon. Just need to finish something before.

PR #1716, opened 2021-01-16
Material DateEdit - Years and Month selectable

andreatosato:
Please change DateEdit Years and Month and enable selection.
I think that select years and months is very key feature.
I change JS for enable it by design.
I see this docs for [datepicker](http://daemonite.github.io/material/docs/4.1/material/pickers/)

PR #1717, opened 2021-01-17
Material DateEdit - Years and Month selectable 

andreatosato:
Please change DateEdit Years and Month and enable selection.
I think that select years and months is very key feature.
I change JS for enable it by design.
I see this docs for [datepicker](http://daemonite.github.io/material/docs/4.1/material/pickers/) 

Change branch -> #1716


stsrki:
You should check-out your feature branch from dev093 and not from the master 😅

PR #1718, opened 2021-01-17
Material DateEdit - Years and Month selectable

andreatosato:
Please change DateEdit Years and Month and enable selection.
I think that select years and months is very key feature.
I change JS for enable it by design.
I see this docs for [datepicker](http://daemonite.github.io/material/docs/4.1/material/pickers/)


stsrki:
It was a bumpy ride but it's looking good now :). Tested and it's working as expected. Thank you!

BTW, for future reference please look at [CONTRIBUTING](https://github.com/stsrki/Blazorise/blob/master/CONTRIBUTING.md) guide 💪

PR #1719, opened 2021-01-17
Clear Filter Bug Fix: Filter Template usage

David-Moreira:
@stsrki The Clear Filter bugfix.
There's a couple of things here.
The bug only happens when using the Filter Template. I noticed the Filter Template takes in a FilterContext from a DataColumn.FilterContext property. So on the Demo both the First Name and City Column Filters have this bug.
I don't know why there exists a Filter and a FilterContext property on the DataColumn but I'm sure there's a good reason...

- The FilterContext.TriggerFilterChange wasn't modifying the filter's current SearchValue with the new Value.
- On the demo itself, both the inputs weren't correctly bound to the value.
- I made the FilterTemplate take in the Filter property instead of the FilterContext property which seems to always have the correct updated value across the code.


stsrki:
I tested it and it's working well now. To be honest I don't remember why there are two filter contexts used. I think it was done by one of the community members and it seems I didn't notice it while I was reviewing it. I will investigate and see if I can remove the internal filter context. #1720

In any case. Thank you for a quick fix. You were really productive today 💪

PR #1729, opened 2021-01-19
Dutch localized resources added

nielscil:
Added dutch translation in resources for datagrid, carousel and file edit.


stsrki:
Good work, thank you!

PR #1735, opened 2021-01-21
BarLink clicked event fired

Morasiu:
I've added a simple `Clicked` event firing.

As you can see I've had to add `TestNavigationManager` class to Tests. It was because the `Link` component has `NavigationManager` injected using `@Inject` and there wasn't any registered. My implementation of `TestNavigationManager` was based on Microsoft implementation in Blazor test. You can find it [here](https://github.com/dotnet/aspnetcore/blob/2588e246e5ca469603d3253225a5dcecda45a1cd/src/Components/WebAssembly/WebAssembly.Authentication/test/WebAssemblyAuthenticationServiceCollectionExtensionsTests.cs#L440).


stsrki:
I did some more testings and modifications and I think it's now safe to merge. Since `Link` component is now handling clicked event I removed the ordinal `<a>` elements. Hopefully, it will not make problems 🤞

Anyways, good job, and thank you for your help!

PR #1742, opened 2021-01-22
Trigger DataGrid reload mode externally

### LINKED ISSUE #1732: DataGrid: trigger ReadData mode externally
I'm trying to have a button, that calls a method loading the data grid with data. While loading I want the loading template to show. This isn't possible however because ReadData cannot be triggered from the outside, it only works when I do filtering or pagination on the grid - because the DataGrid code recognizes these actions. It seems like a basic feature enough to be able to have a button for loading the grid with data, right?

I'm not sure how to implement it, but some way to reload the datagrid so that ReadData is triggered again (for example this happens on first load of a webpage).

stsrki:
resolves #1732

### DIFF
--- a/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
+++ b/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
@@ -419,6 +419,24 @@ namespace Blazorise.DataGrid
 
         #region Filtering
 
+        /// <summary>
+        /// Triggers the reload of the <see cref="DataGrid{TItem}"/> data.
+        /// </summary>
+        /// <returns>Returns the awaitable task.</returns>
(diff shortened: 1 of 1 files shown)


lrequena:
Nice one! Will this one be on next preview release?

stsrki:
@lrequena Yes it will.

jensengbg-jakob-reesalu:
Very neat! Thanks for this one! I came back to this as I needed to rework the way I present loading. Turns out my set up will not work well with using the loading template after all. However, the .Reload() method allowed me to remove a method for handling button click and reloading grid in a very simple way, so still a win!

elbadryd:
It seems that the loading template does not display when Reload is invoked or ReadData is invoked after the initial load. Is that expected?

PR #1747, opened 2021-01-25
Dev093 numeric edit up down

d-oit:
WIP: CSS / UnitTest

Bugfix: The old code didn't check for Min / Max value on input a new number. 

  <FieldLabel ColumnSize="ColumnSize.IsFull.OnTablet.Is2.OnDesktop">Number with up/down</FieldLabel>
        <FieldBody ColumnSize="ColumnSize.IsFull.OnTablet.Is10.OnDesktop">
            <NumericEdit TValue="decimal?" ShowUpDown="true" Min="1" Max="42" />
        </FieldBody>

CSS:
unsatisfied with the group addon would like to create the possible to change the icons / positon / background. Couldn't found a perfect style for the right border (not round - background color -> which one for all frameworks?, button click without border color)

@stsrki Would like to use a css that work for all UI Frameworks. DxExpress and the other link that you found use are derived from bootstrap classes. 

You could change it if you have already have a better solution. 

For iOS you have to set the right int / decimal pattern:
https://css-tricks.com/finger-friendly-numerical-inputs-with-inputmode/#ios-solution-use-the-pattern-attribute-on-a-text-input

Is this too special for the framework?


stsrki:
Hi @d-oit 

Please follow our contribute guidelines https://github.com/stsrki/Blazorise/blob/master/CONTRIBUTING.md

PR #1754, opened 2021-01-27
Add chart data/label Shift and Pop

vflame:
Adds API to remove oldest label/data to create event driven "streaming".


stsrki:
@vflame Looking good. Thank you for this PR!

PR #1775, opened 2021-02-01
Fix bug when ButtonRowTemplate is null

stsrki:
When RenderFragment is passed as a parameter and it is `null` it will raise an exception. This PR fixes it.

@David-Moreira I noticed it only on production 😅


David-Moreira:
Oops! Nice catch! 😅
Thanks for fixing. :P

stsrki:
It was a nasty one and I watched it on time 😄

PR #1789, opened 2021-02-04
DataGrid: fix problem with nested fields

### LINKED ISSUE #1788: DataGrid problems with nested fields
**Describe the bug**
I'm having problems with nested properties after update to preview8. 

I get an ArgumentException with the message "Cannot detect the member of {item.Type}".

**To Reproduce**
To reproduce it in demo i added a new property to Employee.

```
public class Employee
{
	public Boss ReportsTo { get; set; }
}

public class Boss
{
}
```
And then added new column to datagrid

`<DataGridColumn TItem="Employee" Field="ReportsTo.FirstName" Caption="Reports To" Editable="false" />`

**Expected behavior**
Should work.

**Screenshots**
image

**Additional context**
I suspect i found the problem https://github.com/stsrki/Blazorise/commit/0b7f43b1359d70170969b2c8c43a00bdfa3d40d9#r46727285

paulosdfigueiredo:
fixes #1788

### DIFF
--- a/Source/Extensions/Blazorise.DataGrid/Utils/FunctionCompiler.cs
+++ b/Source/Extensions/Blazorise.DataGrid/Utils/FunctionCompiler.cs
@@ -30,7 +30,7 @@ namespace Blazorise.DataGrid.Utils
 
             Expression field = null;
 
-            MemberInfo memberInfo = GetSafeMember( item.Type, propertyOrFieldName );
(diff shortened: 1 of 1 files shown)


stsrki:
Tested and it works. Including the original issue with interfaces.

@paulosdfigueiredo Thank you for the quick fix!

PR #1793, opened 2021-02-04
Portuguese localized resources added

paulosdfigueiredo:
Added portuguese translation in resources for datagrid, carousel and file edit.


stsrki:
Can you just change files encoding to UTF-8? They are currently set to ANSI. I could probably do it but I don't know what are the special characters in Portuguese.

paulosdfigueiredo:
Is it good? All i did was "Save as... -> Save with encoding" in visual studio.

stsrki:
I had some work to finish before reviewing this PR. Looking good now. Thank you!

PR #1805, opened 2021-02-06
Improved DataGrid Localization for de-DE

janhegner:
Some improvements to the german localization.

In detail:
Sparen -> Speichern:  Sparen is not the best word. You would use it in the context of save money.
Stornieren -> Abbrechen:  Stornieren is not the best word. You would use it in the context of cancelling a hotel booking
First/Last: Anfang/Ende. Not literal translation, but common.
Prev/Next: Zurück/Weiter. Also not the literal translation, but common translation in this context

items vs rows: I would suggest a consistent naming. I chose a consisting naming for the German localization. I hope that's ok.

"{0} - {1} von {2} Zeilen" vs "Zeilen {0} - {1} von {2}"   <= both correct but the latter sounds nicer to the German native speaker.


stsrki:
Looking good. Thank you!

PR #1806, opened 2021-02-07
Text localizer fallback to neutral culture

janhegner:
At first, I just couldn't make Localizations for the DataGrid work in my project. After some debugging, I realized that:
1. My application supported the culture "de" instead of "de-DE"
2. Both internet browsers on my system (Firefox & Chromium) were requesting pages with the neutral culture ("de") instead of the specific culture.

The TextLocalizer.GetTranslations() method searchs for an exact match of the culture and only the specific translations are available in Blazorise.

This pull request aims to improve the handling of specific and neutral cultures.

I think the Localization should be changed in the following manner:
1. The default translation resources in Blazorise should be the neutral ones. Translations for specific cultures can also be added if necessary. 
2. Localization requests of a specific culture (e.g. de-AT) should fallback to their respective neutral culture ("de") if a specific translation is not available. The current fallback is the invariant culture (english)

Both changes were implemented in this pull request


stsrki:
My idea was more that we need to explicitly define every culture. But after some thinking on your PR, I think you're right. It makes more sense to have neutral cultures and then override them with specific, based on need.

Thank you for your help!

PR #1881, opened 2021-02-19
RTE - Queue requests for after render and support for content updates

njannink:
Fix for RichTextEdit issue [1752](https://github.com/stsrki/Blazorise/issues/1752)

### DIFF
--- a/Source/Blazorise/Base/BaseComponent.cs
+++ b/Source/Blazorise/Base/BaseComponent.cs
@@ -86,30 +86,6 @@ namespace Blazorise
             executeAfterRendereQueue.Enqueue( action );
         }
 
-        /// <summary>
-        /// Executes given action after the rendering is done.
-        /// </summary>
-        /// <remarks>Don't await this on the UI thread, because that will cause a deadlock.</remarks>
-        protected async Task<T> ExecuteAfterRender<T>( Func<Task<T>> action )
-        {
-            var source = new TaskCompletionSource<T>();
-
-            ExecuteAfterRender( async () =>
-            {
-                try
-                {
-                    var result = await action();
-                    source.TrySetResult( result );
-                }
-                catch ( Exception e )
-                {
-                    source.TrySetException( e );
-                }
-            } );
-
-            return await source.Task.ConfigureAwait( false );
-        }
-
         protected override async Task OnAfterRenderAsync( bool firstRender )
         {
             Rendered = true;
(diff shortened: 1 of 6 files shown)


Source/Blazorise/Base/BaseComponent.cs
```diff
@@ -86,6 +86,30 @@ protected void ExecuteAfterRender( Func<Task> action )
             executeAfterRendereQueue.Enqueue( action );
         }
 
+        /// <summary>
+        /// Executes given action after the rendering is done.
+        /// </summary>
+        /// <remarks>Don't await this on the UI thread, because that will cause a deadlock.</remarks>
+        protected async Task<T> ExecuteAfterRender<T>( Func<Task<T>> action )
```
stsrki, line 93: Can you explain why this implementation of ExecuteAfterRender belongs in the root component? Seems like a specific code that should be better somewhere in RTE project.

njannink: Its a generic method not specific for RTE only. My preference would be to make it an extension method, but because `ExecuteAfterRender` is `protected` thats not possible. I'm fine with also keeping this in the RTE only and then you can always move it to the base component when you need it

stsrki: Calling existing `ExecuteAfterRender` didn't work?

njannink: No because I need async return value. So that's why I work with a `TaskCompletionSource<TResult>`

stsrki: Ah, that's why. Since `ExecuteAfterRender` is `protected` then it's best to create a new `BaseRichTextEditComponent` where you will inherit from `BaseComponent` and then move your implementation of `ExecuteAfterRender`. That way we will leave the root class clean. Is that OK?

njannink: Ok thats fine by me. We should convince Microsoft on protected extension methods :)

stsrki: It is already possible to define friendly assemblies for `internal` methods :)

```cs
[assembly: InternalsVisibleTo( "Blazorise.AntDesign" )]
[assembly: InternalsVisibleTo( "Blazorise.Bootstrap" )]
[assembly: InternalsVisibleTo( "Blazorise.Bulma" )]
[assembly: InternalsVisibleTo( "Blazorise.Frolic" )]
[assembly: InternalsVisibleTo( "Blazorise.Material" )]
```

stsrki:
So I removed `initialized` flag and just called `ExecuteAfterRender` for every operation. It works fine. You can test again after the merge and fix it if needed.

Anyhow, great job. I think now the extension will be a lot safer. Thank you!

PR #1884, opened 2021-02-20
Don't run validation if input is disabled

### LINKED ISSUE #1882: Validation is still being called when Disabled=true
Validation is still being called when Disabled=true. Is there a way to skip validation if a component is disabled? Thanks.

stsrki:
fixes #1882 Validation is still being called when Disabled=true

### DIFF
--- a/Source/Blazorise/Components/Validation/IValidationInput.cs
+++ b/Source/Blazorise/Components/Validation/IValidationInput.cs
@@ -16,5 +16,10 @@ namespace Blazorise
         /// value types. For example a Select component can have single-value and multi-value.
         /// </remarks>
         object ValidationValue { get; }
+
+        /// <summary>
+        /// Returns true if input is disabled.
+        /// </summary>
+        bool Disabled { get; }
     }
 }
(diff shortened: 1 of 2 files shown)


David-Moreira:
@stsrki 
Wow I actually had a use case, where this was helpfull, I'm using the disabled field as a placeholder for the data that is selected somewhere else... it just started breaking that validation... xD Any ideias for workarounds?

stsrki:
Hmm, so you want validation on a disabled field?

stsrki:
Can you make it Readonly?

David-Moreira:
It's basically something like this.
The disabled is usefull for the layout and automatically disabling of the field, and together with the validation stuff, I quickly bind selected stuff from Datagrid into a model with some simple frontend validation.
image

stsrki:
I understand the scenario. But disabled inputs should not participate in validation.

Look at https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled#constraint_validation

David-Moreira:
I see, Thanks! I'm just trying to figure out an easy workaround, so I don't have to be doing big changes now. :P

stsrki:
Like I said, just add `ReadOnly` parameter and adjust CSS to look like a disabled input. It's how usually is done.

David-Moreira:
I'll do that, thanks for the tip.

PR #1886, opened 2021-02-20
Datagrid: Introduced async / await on missing events. Introduced a Cancellation Token to issue a cancellation on page changed / page size changed

David-Moreira:
Hello @stsrki 
Here's the changes I have suggested. Including the idea of a Cancellation token to be able to issue cancellations.

You can test it on the DataGrid Demo Page it now has a random delay on ReadData (when Large Data is selected). If you switch between pages or page sizes, it should always only display the latest operation whereas before, it would update the UI with all the queued operations.

Just let me know if you agree with this idea or not.

### DIFF
--- a/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
+++ b/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
@@ -169,7 +169,7 @@ namespace Blazorise.DataGrid
                     {
                         await InvokeAsync( StateHasChanged );
                     }
-                    
+
                 } );
 
                 paginationContext.SubscribeOnPageChanged( async currentPage =>
@@ -453,7 +453,7 @@ namespace Blazorise.DataGrid
 
             if ( ManualReadMode )
             {
-                return InvokeAsync( () => HandleReadData ( CancellationToken.None ) );
+                return InvokeAsync( () => HandleReadData( CancellationToken.None ) );
             }
             else
             {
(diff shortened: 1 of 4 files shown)


Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -192,8 +192,12 @@ protected override async Task OnAfterRenderAsync( bool firstRender )
                 } );
 
                 if ( ManualReadMode )
+                {
                     await HandleReadData( CancellationToken.None );
 
+                    return;
```
stsrki, line 198: The return is needed here as it was not working well without it. I remember there was a problem a few months ago.

Please test once again to confirm I didn't break it.

stsrki:
It's working nice. And the code is well structured. I just added one comment so please check it before I merge.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
David-Moreira: That's hella strange. Are you sure the problem had to actually do with that? It's fine with me as long as it works. I'm testing and looks good to me.

David-Moreira:
Looks good to me.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
stsrki: I can't find the actual merge that fixed it. I'm really not sure right now.. 

Just in case let's leave it for now. All it does will skip the addition call to StateHasChanged after HandleReadData finishes.

David-Moreira: Maybe if before StateHasChanged was called without InvokeAsync?
Anyway, I was just figuring out it may not be fine, it skips the base.OnAfterRenderAsync which might have logic... and I was just doing the AfterRenderQueue on the base... hmmm

PR #1891, opened 2021-02-21
Data grid check all indeterminate state

### LINKED ISSUE #1777: DataGrid Check All - indeterminate state
Might be a good feature to have an **indeterminate** state for **check-all** checkbox when not everything is selected.

image

```cshtml
<Check Indeterminate="true"></Check>
```

@David-Moreira What do you think?

David-Moreira:
Closes #1777

### DIFF
--- a/Source/Extensions/Blazorise.DataGrid/DataGrid.razor
+++ b/Source/Extensions/Blazorise.DataGrid/DataGrid.razor
@@ -32,7 +32,7 @@
                             @if ( MultiSelect )
                             {
                                 <TableHeaderCell Class="@column.HeaderCellClass" Style="@column.BuildHeaderCellStyle()" TextAlignment="@column.HeaderTextAlignment">
-                                    <_DataGridMultiSelectAll TItem="TItem" MultiSelectAll="OnMultiSelectAll"></_DataGridMultiSelectAll>
+                                    <_DataGridMultiSelectAll TItem="TItem" MultiSelectAll="@OnMultiSelectAll" IsIndeterminate="@IsMultiSelectAllIndeterminate"></_DataGridMultiSelectAll>
                                 </TableHeaderCell>
                             }
                         }
--- a/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
+++ b/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
@@ -101,6 +101,12 @@ namespace Blazorise.DataGrid
         /// </summary>
         internal bool MultiSelect => ( SelectionMode == DataGridSelectionMode.Multiple );
 
+
+        /// <summary>
+        /// Checks if the MultiSelectAll is indeterminate, meaning that only some of the current view rows are selected.
+        /// </summary>
+        private bool IsMultiSelectAllIndeterminate => ( MultiSelect && ( this.SelectedRows?.Any() ?? false) && !(this.viewData.Except(this.SelectedRows).Count() == 0) );
+
         #endregion
 
         #region Constructors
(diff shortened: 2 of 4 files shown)


stsrki:
It's working OK but still has room to improve.

The problem is with pagination. It only works for the current page. If you check all boxes and navigate to another page you lose the state and `Check all` becomes indeterminate, while row checkboxes are unchecked.

I'm not sure how we handle this. Do we keek the internal state of all checked items?

David-Moreira:
Oh I didn't test that use case. 
That seems like flawed logic. There's a flag at the Datagrid cs that controls the indeterminate flag based on the currently selected items. It must be checking that there are items checked "globally", but not all from the current page so it think it's indeterminate. It should only check if there are items checked in the current page not globally.

You can fix it if you want, otherwise, 
I'll see if I have time later today.

David-Moreira:
Also it seems to me that you'd like to keep the checkboxes state when changing page. I'm more sure  if it makes sense.
I'd you select item 1, 3 and 5 on page 1. When you change to page 2 you don't expect that it now automatically selects item 6, 8 and 10.

stsrki:
No, but you would expect to be all checked on each page if you clicked on `Check All`.

But then again there is another problem. If ReadData is used we have no way of knowing what is or what can be checked. There can be millions of rows on the backend.

Maybe this feature was not a good idea 😅

David-Moreira:
Actually I think that's wrong strski.

The user only expects that it selects only what he's seeing when he clicks select all and that's how it was implemented. It selects all records on the current page. Not the whole freaking data set!!! 😳 At least, that's how I have always seen these type of select all feature working on grids. With that in mind, it makes no difference to Readdata since it selects the available page. 
Of course that if you have read data on, change page and come back the state will be lost, because you will have refresh the data. But that is to be expected when the data is dynamic, the data may no longer be the same so you can't really be saving states...

I don't see anything wrong with this feature strski.

paulosdfigueiredo:
I agree with @David-Moreira, select all should only select visible items.

If people want to select more, they should choose more items per page or an option with "all items" and then select all.

stsrki:
Personally, I was always doing it the was that _select all_ would mean select all rows. No matter what. But considering I was doing win apps it is an entirely different world :)

I'm reading and googling more into this and there is no silver bullet here. No one can agree on the _right_ way. So I think we can go with your solution. Most of the people tend to go that way from what I can see. And if anyone wants it different then he/she can do their own implementation.

Just one thing before I merge. When you navigate to page 2, reset the Check all to the false state. Currently, it stays indeterminate even though nothing is checked on a new page.

David-Moreira:
I see. Well I kinda develop alot of Web apps 😆 

Like I said that's flawed logic on the indeterminate flag on Datagrid cs. Bug that has to be fixed. If you'd like to quickly fix it, that's fine. Otherwise I don't have time yet maybe in 6h or so.

stsrki:
I tested it once again before the merge and it's working OK now. 🤔 strange...

stsrki:
I merged it anyways. We can always feel with it later if there are complaints.

I'm thanking you so much lately that I start to lose a way of saying it 😅 Thank you!

David-Moreira:
I'll test it later. And submit a new PR if I detect it.
No problem at all, and **NO, Thank you @stsrki** for providing and supporting this awesome open source project. 

This is the beauty of it, we all contribute to a "common good", and it makes our life so much darn easier when implementing stuff to have these type of libraries!

stsrki:
I'm well aware this is open-source and that anyone can contribute. But not everyone does it, and those who are are basically giving their own personal time to the community. I don't expect anyone to do it, but when they do, it really feels nice :)

PR #1896, opened 2021-02-22
Datagrid multi selection fixes

David-Moreira:
You're probably getting tired of my PRs right about now... xD

Ok so the indeterminate flag was wrong when changing to an empty page just like you had tested it.
When fixing that, also the check all wasn't correctly updating with the currently DisplayData that was selected when switching back and forth between already visited pages.

Also made the logic a bit clearer by removing the expression body in favor of a simple method.

DisplayData was changed from Protected to Internal as I needed to access this information, that should be okay, but just tell me otherwise.


stsrki:
Don't worry. Your PRs are always welcome :)

Made some small optimizations but it was already working nicely. Thank you once again!

PR #1905, opened 2021-02-23
RTE - Fix custom config and add docs + example

njannink:
See also discussion [1901](https://github.com/stsrki/Blazorise/discussions/1901)


stsrki:
I get an error when I run the demo

image

njannink:
Ahh damn. I will just comment that part it out because it's an example how to load custom modules

stsrki:
It's now working. Thank you once again!

PR #1916, opened 2021-02-25
BaseAfterRenderComponent Null Bug Fix

David-Moreira:
So this wasn't easy to reproduce... But I had it happen.
image

You basically have to switch between pages/components really fast and it may happen.
I could reproduce it often when clicking these pages on the Demo: 
Nested Forms / Nested Validations and Misc really fast, like within a second.

I could then faithfully reproduce by adding this code:
`            ExecuteAfterRender( () => Task.Delay( 2500 ) );`
```
 protected override async Task OnAfterRenderAsync( bool firstRender )
        {
            Rendered = true;
            ExecuteAfterRender( () => Task.Delay( 2500 ) );
            if ( executeAfterRenderQueue?.Count > 0 )
            {
                while ( executeAfterRenderQueue.Count > 0 )
                {
                    var action = executeAfterRenderQueue.Dequeue();

                    await action();
                }
            }

            await base.OnAfterRenderAsync( firstRender );
        }
```

**So Dispose runs while AfterRender is still running.** And then suddently the collection is null. Due to the async nature of blazor, we should either add a null check to the while loop or remove this null that I introduced when I refactored the component.
My guess is, that this could be more dangerous on environments with more latency. 

I removed the null and it's okay.

This one's unrelated, but happened while testing this fix, happened just once clicking really fast on the pages at the bottom, but I can't seem to reproduce it anymore:
image

### DIFF
--- a/Source/Blazorise/Base/BaseAfterRenderComponent.cs
+++ b/Source/Blazorise/Base/BaseAfterRenderComponent.cs
@@ -70,7 +70,6 @@ namespace Blazorise.Base
                 if ( disposing )
                 {
                     executeAfterRenderQueue?.Clear();
-                    executeAfterRenderQueue = null;
                 }
             }
         }


stsrki:
Good catch. There are similar use cases through the codebase so I should probably re-check them all.

Thank you!

PR #1918, opened 2021-02-25
Datagrid: Is pager visible when button row is on and grid is Empty with an EmptyTemplate set

### LINKED ISSUE #1913: Datagrid Empty : No pager
Hello @stsrki 
I was finally finishing up migrating the buttons to ButtonRow.

But I noticed that when the Datagrid is empty, the pager doesn't render which kinda makes sense, but that also means that buttons that can have logic that makes sense even if the grid is empty, will not render... :'( 

Think we can add something to enable pager even when empty? It does make sense now that devs can configure buttons on the pager.
image

David-Moreira:
Closes #1913 

I was thinking if it made sense to show when loading, maybe the user could have a button to cancel the ongoing loading... but I left that as it was for now.


stsrki:
@David-Moreira You will get the special mention in the release notes! :)

PR #1930, opened 2021-02-27
Datagrid Resize : Ignore clicks from resizer triggering sort or other click events.

### LINKED ISSUE #1928: DataGrid - Resize triggers sort
**Describe the bug**
When DataGrid is resizable, clicking on the "resize line", the column on the left triggers Sort.

This happens with both ResizeModes, even if we click it out of the Header with ResizeMode.Column.

**To Reproduce**
1. Open Demo
2. With resize and sort enabled, click/drag "resize line" to resize the column.

David-Moreira:
Closes #1928 

So here I was thinking the fix would be as easy as to preventPropagation on the resizable click... Oh damn, I was so wrong...
Well that did solve most of it, but resizing and leaving the mouse on top of the adjacent parent column, would still trigger the column mouseup event and corresponding click.

As soon as mouseup event is triggered there's no way to stop the click event. So I've had to try a few things to avoid triggering the mouseup event when resizing on top of the column, 
- I tried setting the pointer-Events to none, which actually worked, but then the mouse would loose the resize look... no good.
- There was also a way to wrap the column itself in a div and stop the click from propagating below, but that would be changing the table's HTML, which could eventually mess people's css or expected markup... so no good...
				
I ended up just tracking the dates of when the mousedown and mouseup from resizing were triggered and then do logic to cancel the click event, if I determined it came from the resizer.

I also noticed the init was running more then once, duplicating resizers unnecessarily, so I removed an interop and still added a check to the js.

### DIFF
--- a/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
+++ b/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
@@ -1022,14 +1022,8 @@ namespace Blazorise.DataGrid
 
                 resizable = value;
 
-                if ( resizable )
-                {
-                    ExecuteAfterRender( () => InitResizable().AsTask() );
-                }
-                else
-                {
+                if ( !resizable )
                     ExecuteAfterRender( () => DestroyResizable().AsTask() );
-                }
             }
         }
 
--- a/Source/Extensions/Blazorise.DataGrid/wwwroot/blazorise.datagrid.css
+++ b/Source/Extensions/Blazorise.DataGrid/wwwroot/blazorise.datagrid.css
@@ -13,5 +13,10 @@
 }
 
     .b-datagrid-resizer:hover, .b-datagrid-resizing {
+        cursor: col-resize !important;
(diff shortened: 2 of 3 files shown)


Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -1022,14 +1022,8 @@ public bool Resizable
 
                 resizable = value;
 
-                if ( resizable )
-                {
-                    ExecuteAfterRender( () => InitResizable().AsTask() );
-                }
-                else
-                {
+                if ( !resizable )
```
stsrki, line 1025: Why did you removed InitResizable? Wouldn't that make if not work if Resizable is changed in runtime?

stsrki: Ignore, I debugged it and noticed `await RecalculateResize();` in `OnAfterRenderAsync`.

stsrki:
It's a little strange to use Date, but it works so I will merge. Thanks :)

Issue #994, opened 2020-06-18
Improvements on TreeView


stsrki:
Features for next version of TreeView

- [ ] Validation support
- [ ] Editing of node
- [ ] EditTemplate for custom editing
- [ ] Expand/Collapse doesn't work on mobile(see comment bellow)

@robalexclark

FrancisValade:
It would be nice to have a way to customise the icons for opening and closing a node. For example, I would like to replace the plus and minus icons for a right and down arrow.

nightroman:
@stsrki Not sure if it's the right place to ask/suggest.

It's rather hard to expand and collapse `TreeView` nodes on mobile phone. It works 50/50, I'd say.
Maybe the suggested above custom (larger) icons may help, if implemented.

stsrki:
I just tested on mobile. You're right, something is wrong with click or tap events. I need to "click" several times to make it expand or collapse. Weird...

nightroman:
If this needs a separate issue please let me know.

stsrki:
No need, I have updated the original ticket with new item.

nightroman:
One more minor defect (?) - on hovering over expand/collapse icons the mouse pointer does not change from the "arrow" to "hand" (click me convention).

CMontani89:
Is there a way to expand all Nodes by default on page load?

robalexclark:
To expand all nodes I would loop through and add all nodes to an IList<T> expandedNodes and then set that to @bind-expandednodes e.g.

  ```
<TreeView Nodes="holdings"
                              @bind-SelectedNode="selectedNode"
                              @bind-ExpandedNodes="expandedNodes"
                              GetChildNodes="@(item => item.Children)"
                              HasChildNodes="@(item => item.Children?.Any() == true)">
                        <NodeContent>
                            @if (context.HoldingType == "Holding Unit")
                            {
                                <Icon Name="IconName.Building" />
                            }
                            else if (context.HoldingType == "Room")
                            {
                                <Icon Name="@("fa-door-closed")" />
                            }
                            else if (context.HoldingType == "Cage")
                            {
                                <Icon Name="@("fa-th")" />
                            }

                            @context.Text
                        </NodeContent>
                    </TreeView>
```

Jinjinov:
> It would be nice to have a way to customise the icons for opening and closing a node. For example, I would like to replace the plus and minus icons for a right and down arrow.

Any progress on this?

[...]

PR #1932, opened 2021-02-27
TreeView Selected Class: Added Selected Node Class

### LINKED ISSUE #1803: TreeView: SelectedClass
Hello @stsrki 
I've also started using the Tree view for a use case I have.
I noticed that the selected class is already put there by Blazorise, In the case of bootstrap, bg-primary. 
Would be nice to just have a SelectedClass to set my own freely. Otherwise I have to endup overriding bg-primary.

David-Moreira:
Closes #1803

Part of #994

### DIFF
--- a/Demos/Blazorise.Demo/Pages/Tests/TreeViewPage.razor
+++ b/Demos/Blazorise.Demo/Pages/Tests/TreeViewPage.razor
@@ -13,7 +13,9 @@
                           @bind-SelectedNode="selectedNode"
                           @bind-ExpandedNodes="ExpandedNodes"
                           GetChildNodes="@(item => item.Children)"
-                          HasChildNodes="@(item => item.Children?.Any() == true)">
+                          HasChildNodes="@(item => item.Children?.Any() == true)"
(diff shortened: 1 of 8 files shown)


stsrki:
Might be a good idea to also have `SelectedNodeStyle`. What do you think?

David-Moreira:
You're right. I'll see if I have time later today to add it. 
I was also thinking if it made sense to pass in the context so people can display class or style according to additional logic. I think you do that on some places in Datagrid and it was already useful to me personally.

stsrki:
Yeah it passed through my mind. You can add the context if you have the time.

David-Moreira:
Yea, we better do that right away so we don't need to do breaking changes in the future.
I'll base myself on these two. These are great and enable to extend to have extra relevant properties in the future.
```
        [Parameter] public Action<TItem, DataGridRowStyling> RowStyling { get; set; }
        [Parameter] public Action<TItem, DataGridRowStyling> SelectedRowStyling { get; set; }
```

David-Moreira:
@stsrki Done.

stsrki:
LGTM!

Good work as usual. Thank you!

PR #1935, opened 2021-02-27
Add EditOrder to DataGridColumn

AjaxSantos:
Add EditOrder to DataGridColumn.
Fixed problems with order in inline edit mode.

### DIFF
--- a/Source/Extensions/Blazorise.DataGrid/DataGridColumn.cs
+++ b/Source/Extensions/Blazorise.DataGrid/DataGridColumn.cs
@@ -263,7 +263,7 @@ namespace Blazorise.DataGrid
         /// <summary>
         /// Gets or sets where column will be displayed on edit row/popup.
         /// </summary>
-        [Parameter] public int EditOrder { get; set; }
+        [Parameter] public int? EditOrder { get; set; }
 
         /// <summary>
         /// Allows the cell values to be entered while the grid is in the new-item state.
--- a/Source/Extensions/Blazorise.DataGrid/DataGridColumn.cs
+++ b/Source/Extensions/Blazorise.DataGrid/DataGridColumn.cs
@@ -210,6 +210,10 @@ namespace Blazorise.DataGrid
         /// </summary>
         public virtual DataGridColumnType ColumnType { get; } = DataGridColumnType.Text;
 
+        public bool IsMultiSelectColumn => ColumnType == DataGridColumnType.MultiSelect;
+
+        public bool IsCommandColumn => ColumnType == DataGridColumnType.Command;
+
         /// <summary>
         /// Gets or sets the column's display caption.
         /// </summary>
@@ -255,6 +259,11 @@ namespace Blazorise.DataGrid
         /// Gets or sets where column will be displayed on a grid.
         /// </summary>
         [Parameter] public int DisplayOrder { get; set; }
+
+        /// <summary>
+        /// Gets or sets where column will be displayed on edit row/popup.
+        /// </summary>
+        [Parameter] public int EditOrder { get; set; }
 
         /// <summary>
         /// Allows the cell values to be entered while the grid is in the new-item state.
--- a/Source/Extensions/Blazorise.DataGrid/_DataGridModal.razor
+++ b/Source/Extensions/Blazorise.DataGrid/_DataGridModal.razor
@@ -26,22 +26,16 @@
                     <ValidationSummary Label="@ParentDataGrid.ValidationsSummaryLabel" />
                 }
                 <Fields>
-                    @foreach ( var column in Columns )
+                    @foreach ( var column in OrderedEditableColumns )
                     {
-                        @if ( column.ExcludeFromEdit )
-                            continue;
-
-                        @if ( column.CellValueIsEditable )
-                        {
-                            <Field ColumnSize="@column.PopupFieldColumnSize" Padding="Blazorise.Padding.Is2.OnX.Is2.OnY">
-                                <FieldLabel>
-                                    @column.Caption
-                                </FieldLabel>
-                                <FieldBody>
-                                    <_DataGridCell TItem="TItem" Column="@column" Item="@EditItem" CellEditContext="@EditItemCellValues[column.ElementId]" ShowValidationFeedback="@ParentDataGrid.ShowValidationFeedback" />
-                                </FieldBody>
-                            </Field>
-                        }
+                        <Field ColumnSize="@column.PopupFieldColumnSize" Padding="Blazorise.Padding.Is2.OnX.Is2.OnY">
+                            <FieldLabel>
+                                @column.Caption
+                            </FieldLabel>
+                            <FieldBody>
+                                <_DataGridCell TItem="TItem" Column="@column" Item="@EditItem" CellEditContext="@EditItemCellValues[column.ElementId]" ShowValidationFeedback="@ParentDataGrid.ShowValidationFeedback" />
+                            </FieldBody>
+                        </Field>
                     }
                 </Fields>
             </Validations>
--- a/Source/Extensions/Blazorise.DataGrid/_DataGridModal.razor.cs
+++ b/Source/Extensions/Blazorise.DataGrid/_DataGridModal.razor.cs
@@ -1,6 +1,7 @@
 ﻿#region Using directives
 using System;
 using System.Collections.Generic;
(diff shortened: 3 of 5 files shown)


David-Moreira:
Hello,
Oh I just started using DisplayOrder myself, I didn't know Edit would not take that into account. 
I expect to be able to set it once and not having to worry about having to go set on other places... potentially duplicating it. 

In example: 
```
<Col> DisplayOrder = 2 EditOrder =2</Col>
<Col> DisplayOrder = 3 EditOrder =3</Col>
<Col> DisplayOrder = 4 EditOrder =4</Col>
```
Maybe a more general Order could be available?
```
<Col> Order = 2 </Col>
<Col> Order = 3 </Col>
<Col> Order = 4 </Col>
```

And is there actually any use case where it makes sense to have this?? Specially inline edit... seems like it'll just make it all out of sync...
**Random example:** 
```
<Col> DisplayOrder = 1 EditOrder = 5</Col>
<Col> DisplayOrder = 5 EditOrder = 1</Col>
<Col> DisplayOrder = 2 EditOrder = 6</Col>
```

AjaxSantos:
@David-Moreira I think, that display order has to be taken for inline mode.
And because of using:
If user can change display order of columns, but edit order should not change.

David-Moreira:
@stsrki I think the DisplayOrder feature itself has to be reviewed
Some simple tests:
image
image
image


@AjaxSantos did you consider **Inline New** in your PR?

Should the Order also propagate to Form and Popup for uniformity? I see an order on the table, and then suddently the Form/Popup has another Order.

David-Moreira:
@AjaxSantos did you consider Inline New in your PR?

Should the Order also propagate to Form and Popup for uniformity? I see an order on the table, and then suddently the Form/Popup has another Order.

AjaxSantos:
I do not think so.
Example:
I do not interesst in gender or childrens, when i search for customers.
But if i want to add or edit customer, it would make sense, if gender stands before first and last name and after these childrens.

stsrki:
I think it's better to have both EditOrder and DisplayOrder. The more control, the better.

David-Moreira:
I don't fully disagree. I just expect that if I change the display order. Inline edit & inline new remains in sync with the display order. Since inline edit uses the exact same header columns from the display columns / order. 

So if I use inline edit and I change the display order, now I have to duplicate the order on the edit order to make it consice. Seems quite redundant. 

For form & poup its a good feature. More control yes.

AjaxSantos:
For inline display order will be used. Look on _DataGridRowEdit.razor.

stsrki:
Inline edit is still an edit mode. And not all columns will be editable. So it makes sense to also have it ordered in a different order than display.

David-Moreira:
I don't think you guys are getting my point
Here's a screenshot of the current behaviour:
I changed the display order of Gender/Childrens and DOB. 
Shouldn't that be automatically in sync with the DisplayOrder? This makes no sense at all to me.
image

stsrki:
@David-Moreira Yes, you're right. It's my mistake. Yeah, in that case, EditOrder should not be used for inline edit mode.

David-Moreira:
Can we ask @AjaxSantos to make the changes according to what we've just discussed then? And to not forget about the inline edit new?

AjaxSantos:
> I don't think you guys are getting my point
> Here's a screenshot of the current behaviour:
> I changed the display order of Gender/Childrens and DOB.
> Shouldn't that be automatically in sync with the DisplayOrder? This makes no sense at all to me.
> image
> 
> To solve this, now you'd have to do this for every single column:
> Same EditOrder as DisplayOrder
> And that's redudant, because InlineEdit always takes advantage of the header columns. Inline Edit would need to have it's own header columns if you wanted to have different DisplayOrder vs EditOrder.
> 
> ```
>   <DataGridDateColumn DisplayOrder="-1" EditOrder="-1" TItem="Employee" Field="@nameof( Employee.DateOfBirth )" DisplayFormat="{0:dd.MM.yyyy}" Caption="Date Of Birth" Editable="true" />
>   <DataGridNumericColumn DisplayOrder="-2" EditOrder="-2" TItem="Employee" Field="@nameof( Employee.Childrens )" Caption="Childrens" Editable="true" Filterable="false" />
>   <DataGridSelectColumn DisplayOrder="-3" EditOrder="-3" TItem="Employee" Field="@nameof( Employee.Gender )" Caption="Gender" Editable="true">
> ```
> 
> Edit: Like I've said before, for the other modes, it does make sense.

I can not reproduce your problem.
image
image
image

David-Moreira:
Of course you can't... you used the redudant way...
```
 <DataGridDateColumn DisplayOrder="-1" EditOrder="-1* TItem="Employee" Field="@nameof( Employee.DateOfBirth )" DisplayFormat="{0:dd.MM.yyyy}" Caption="Date Of Birth" Editable="true" />
 <DataGridNumericColumn DisplayOrder="-2" EditOrder="-2" TItem="Employee" Field="@nameof( Employee.Childrens )" Caption="Childrens" Editable="true" Filterable="false" />
 <DataGridSelectColumn DisplayOrder="-3" EditOrder="-3" TItem="Employee" Field="@nameof( Employee.Gender )" Caption="Gender" Editable="true">
```
The inline edit should work **without** **`EditOrder`** being set. Why? Because currently the corresponding edit labels, are the ones from the **Display Header Columns**. So currently you will never have an EditOrder different then the DisplayOrder  when on Inline Edit.

AjaxSantos:
DisplayOrder wins against EditOrder by inline mode:
problem_with_edit_order
problem_with_edit_order_0
problem_with_edit_order_1
problem_with_edit_order_2
problem_with_edit_order_3

AjaxSantos:
> Of course you can't... you used the redudant way...
> 
> ```
>  <DataGridDateColumn DisplayOrder="-1" EditOrder="-1* TItem="Employee" Field="@nameof( Employee.DateOfBirth )" DisplayFormat="{0:dd.MM.yyyy}" Caption="Date Of Birth" Editable="true" />
>  <DataGridNumericColumn DisplayOrder="-2" EditOrder="-2" TItem="Employee" Field="@nameof( Employee.Childrens )" Caption="Childrens" Editable="true" Filterable="false" />
>  <DataGridSelectColumn DisplayOrder="-3" EditOrder="-3" TItem="Employee" Field="@nameof( Employee.Gender )" Caption="Gender" Editable="true">
> ```
> 
> The inline edit should work **without** **`EditOrder`** being set. Why? Because currently the corresponding edit labels, are the ones from the **Display Header Columns**. So currently you will never have an EditOrder different then the DisplayOrder when on Inline Edit.
> So:
> 
> ```
>  <DataGridDateColumn DisplayOrder="-1" TItem="Employee" Field="@nameof( Employee.DateOfBirth )" DisplayFormat="{0:dd.MM.yyyy}" Caption="Date Of Birth" Editable="true" />
>  <DataGridNumericColumn DisplayOrder="-2" TItem="Employee" Field="@nameof( Employee.Childrens )" Caption="Childrens" Editable="true" Filterable="false" />
>  <DataGridSelectColumn DisplayOrder="-3" TItem="Employee" Field="@nameof( Employee.Gender )" Caption="Gender" Editable="true">
> ```
> 
> Should automatically display Inline Edit correctly, without explicitly setting the EditOrder.

So you want, that if no EditOrder set, you get display order by editing?

David-Moreira:
Exactly!! But ONLY on **INLINE EDIT**, since inline edit does make use of the display columns as a way to identify which columns are you editing.

On the other two modes: Form/Popup it makes total sense. And an additional improvement on those two, would be to: if EditOrder isn't set, it takes in the DisplayOrder instead for uniformity.

AjaxSantos:
> Exactly!! But ONLY on **INLINE EDIT**, since inline edit does make use of the display columns as a way to identify which columns are you editing.
That is actual happen.
image
image
image
> 
> On the other two modes: Form/Popup it makes total sense. And an additional improvement on those two, would be to: if EditOrder isn't set, it takes in the DisplayOrder instead for uniformity.
Can i try.

David-Moreira:
Oh, if you already done that. That's cool. 👌 
What about the new Button? Is it working the same also?

AjaxSantos:
For new, it works also.

stsrki:
Looking good now. I've just added a rule to use DisplayOrder if EditOrder is not used.

@AjaxSantos Thank you!

PR #1937, opened 2021-02-28
DataGrid - Added  SortDirectionTemplate to Column

paulosdfigueiredo:
Added SortDirectionTemplate to DataGridColumn to allow custom icon for sort direction.

Also fixed the cursor when a row is not sortable.


paulosdfigueiredo:
Although I think it would be better to add an entire column as context, like the CaptionTemplate

`public RenderFragment<DataGridColumn<TItem>> CaptionTemplate { get; set; }`

 i just insert the SortDirection, because Column.CurrentDirection is internal and i don't know if i should change that.

paulosdfigueiredo:
I don't know how i closed this, probably missclicked 😅.

stsrki:
> Although I think it would be better to add an entire column as context, like the CaptionTemplate
> 
> `public RenderFragment<DataGridColumn<TItem>> CaptionTemplate { get; set; }`
> 
> i just insert the SortDirection, because Column.CurrentDirection is internal and i don't know if i should change that.

It's better this way. CurrentDirection is just for internal purposes and is not meant to be public.

stsrki:
Tested and working fine. Thank you!

Issue #1920, opened 2021-02-25
Blazor server - issue in component cleanup


Unlink:
**Describe the bug**
I tried to fix RichTextEdit disposal in https://github.com/stsrki/Blazorise/pull/1867
but I think that issue is bigger.

When we have a page with components that have some JS cleanup For example buttons. When the component is disposed, js function is called.
![obrázok](https://user-images.githubusercontent.com/9026197/109200020-5fc2b280-77a0-11eb-9b98-59744ff6a742.png)
This is ok in the case of for example navigation to another page. But if the user closes the tab, JSRuntime is not available anymore. This will cause task to be paused (or something else?). After some time all of those tasks are canceled (as we can see in case of RichTextEdit). This will cause slow down of blazor app (probably due to exception handling).

**To Reproduce**
1. Make a page with lots of buttons (500)
2. Refresh the page a few time
3. The app became very slow

I'm not sure how to fix it, maybe using IComponentContext.IsConnected?
https://github.com/dotnet/aspnetcore/issues/9613

stsrki:
I think you're right. This `async void` was an overlook on our side. From what I can see there is no reason why it cannot be made `ValueTask`.

@njannink What do you think? See the last comment for reference #1867

stsrki:
Guys, check #1921

njannink:
Looks good to me

Unlink:
Thanks, this is the solution for exceptions showing in the error log for RTE.
But I think that there is still an issue with all components being disposed after the connection closes.

In addition to my example
try to set JSInteropDefaultCallTimeout to for example 5 seconds (instead of 1 minute by default).
This will show slow down after 5 seconds of the first page reload.
```
services.AddServerSideBlazor().AddCircuitOptions(o =>
{
     o.JSInteropDefaultCallTimeout = TimeSpan.FromSeconds(5);
});
```

In VS I can confirm those exceptions
![obrázok](https://user-images.githubusercontent.com/9026197/109267688-40f90600-780a-11eb-96da-31377e531417.png)

They are not shown in the error log, which's good, but I think that they are slowing down the overall performance of the server.
If I increase Timeout, the slow down occurs later because those tasks are being canceled later.

stsrki:
@Unlink How do I reproduce the errors? I've added `o.JSInteropDefaultCallTimeout = TimeSpan.FromSeconds(5);` but still nothing shows.

stsrki:
[...]

PR #1941, opened 2021-03-01
IAsyncDisposable

stsrki:
The first prototype of using IAsyncDisposable to clean the component and call JSInterop.

The purpose of this PR is to resolve issues with #1920

### DIFF
--- a/Demos/Blazorise.Demo.Bootstrap.RC/Startup.cs
+++ b/Demos/Blazorise.Demo.Bootstrap.RC/Startup.cs
@@ -34,7 +34,8 @@ namespace Blazorise.Demo.Bootstrap.RC
             services.AddServerSideBlazor().AddHubOptions( ( o ) =>
             {
                 o.MaximumReceiveMessageSize = 1024 * 1024 * 100;
-            } ).AddCircuitOptions( o =>
+            } )
+            .AddCircuitOptions( o => // WARNING: Only for testing! Should be removed before merge!
             {
                 o.JSInteropDefaultCallTimeout = TimeSpan.FromSeconds( 5 );
             } );
--- a/Demos/Blazorise.Demo.Bootstrap.RC/Startup.cs
+++ b/Demos/Blazorise.Demo.Bootstrap.RC/Startup.cs
@@ -1,3 +1,4 @@
+using System;
 using Blazorise.Bootstrap;
 using Blazorise.Icons.FontAwesome;
 using Blazorise.RichTextEdit;
@@ -33,6 +34,9 @@ namespace Blazorise.Demo.Bootstrap.RC
             services.AddServerSideBlazor().AddHubOptions( ( o ) =>
             {
                 o.MaximumReceiveMessageSize = 1024 * 1024 * 100;
+            } ).AddCircuitOptions( o =>
+            {
+                o.JSInteropDefaultCallTimeout = TimeSpan.FromSeconds( 5 );
             } );
         }
 
--- a/Demos/Blazorise.Demo.Bootstrap.RC/Startup.cs
+++ b/Demos/Blazorise.Demo.Bootstrap.RC/Startup.cs
@@ -34,10 +34,6 @@ namespace Blazorise.Demo.Bootstrap.RC
             services.AddServerSideBlazor().AddHubOptions( ( o ) =>
             {
                 o.MaximumReceiveMessageSize = 1024 * 1024 * 100;
-            } )
-            .AddCircuitOptions( o => // WARNING: Only for testing! Should be removed before merge!
-            {
-                o.JSInteropDefaultCallTimeout = TimeSpan.FromSeconds( 5 );
             } );
         }
 
--- a/Source/Blazorise/Base/BaseAfterRenderComponent.cs
+++ b/Source/Blazorise/Base/BaseAfterRenderComponent.cs
@@ -7,7 +7,7 @@ using Microsoft.AspNetCore.Components;
 
 namespace Blazorise.Base
 {
-    public abstract class BaseAfterRenderComponent : ComponentBase, IAsyncDisposable
+    public abstract class BaseAfterRenderComponent : ComponentBase, IDisposable, IAsyncDisposable
     {
         #region Members
 
@@ -51,9 +51,10 @@ namespace Blazorise.Base
         }
 
         /// <inheritdoc/>
-        public ValueTask DisposeAsync()
+        public void Dispose()
         {
-            return DisposeAsync( true );
+            Dispose( true );
+            GC.SuppressFinalize( this );
         }
 
         /// <summary>
@@ -73,20 +74,33 @@ namespace Blazorise.Base
             }
         }
 
+        /// <inheritdoc/>
+        public ValueTask DisposeAsync()
+        {
+            return DisposeAsync( true );
+        }
+
         /// <summary>
         /// Releases the unmanaged resources used by the <see cref="BaseComponent"/> and optionally releases the managed resources.
         /// </summary>
         /// <param name="disposing">True if the component is in the disposing process.</param>
         protected virtual ValueTask DisposeAsync( bool disposing )
         {
-            if ( !AsyncDisposed )
+            try
             {
-                AsyncDisposed = true;
+                if ( !AsyncDisposed )
+                {
+                    AsyncDisposed = true;
 
-                Dispose( disposing );
+                    Dispose();
+                }
+
+                return default;
             }
-
-            return ValueTask.CompletedTask;
+            catch ( Exception exc )
+            {
+                return new ValueTask( Task.FromException( exc ) );
+            }
         }
 
         #endregion
--- a/Source/Blazorise/Base/BaseAfterRenderComponent.cs
+++ b/Source/Blazorise/Base/BaseAfterRenderComponent.cs
@@ -7,7 +7,7 @@ using Microsoft.AspNetCore.Components;
 
 namespace Blazorise.Base
 {
-    public abstract class BaseAfterRenderComponent : ComponentBase, IDisposable
+    public abstract class BaseAfterRenderComponent : ComponentBase, IAsyncDisposable
     {
         #region Members
 
@@ -51,10 +51,9 @@ namespace Blazorise.Base
         }
 
         /// <inheritdoc/>
-        public void Dispose()
+        public ValueTask DisposeAsync()
         {
-            Dispose( true );
-            GC.SuppressFinalize( this );
+            return DisposeAsync( true );
         }
 
         /// <summary>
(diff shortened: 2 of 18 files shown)


Demos/Blazorise.Demo.Bootstrap.RC/Startup.cs
```diff
@@ -33,6 +34,9 @@ public void ConfigureServices( IServiceCollection services )
             services.AddServerSideBlazor().AddHubOptions( ( o ) =>
             {
                 o.MaximumReceiveMessageSize = 1024 * 1024 * 100;
+            } ).AddCircuitOptions( o =>
+            {
+                o.JSInteropDefaultCallTimeout = TimeSpan.FromSeconds( 5 );
```
njannink, line 39: Maybe don't check this in. Its for testing only

stsrki: Yeah, I forgot to comment that will not be merged.

Source/Blazorise/Base/BaseAfterRenderComponent.cs
```diff
@@ -83,6 +98,11 @@ protected virtual void Dispose( bool disposing )
         /// </summary>
         protected bool Disposed { get; private set; }
```
njannink, line 99: Is this property still needed?

Source/Blazorise/Components/Bar/Bar.razor.cs
```diff
@@ -131,7 +143,7 @@ protected override void Dispose( bool disposing )
                     NavigationManager.LocationChanged -= OnLocationChanged;
```
njannink, line 143: I think its better to first unwire the events before starting the async operation. Goes for all dispose methods

Source/Blazorise/Base/BaseAfterRenderComponent.cs
stsrki: Yes, because there are still some components that are disposing synchronously.

Source/Blazorise/Components/Bar/Bar.razor.cs
stsrki: You're right. Gonna change that.

Source/Blazorise/Components/Bar/Bar.razor.cs
```diff
@@ -115,14 +115,26 @@ public Task OnBreakpoint( bool broken )
...
+                    }
+                    catch
+                    {
+                        if ( !task.IsCanceled )
+                        {
+                            throw;
+                        }
+                    }
 
                     DisposeDotNetObjectRef( dotNetObjectRef );
```
Unlink, line 139: I think that we should call `DisposeDotNetObjectRef( dotNetObjectRef ); ` always regardless of exception or not.
When there is an exception, disposal of NetObjectRef is not called.
I'm not sure, if we need to catch task canceled exception there, it looks like that blazor does its own way.

maybee sth. like this?
```
try
{
    await task;
}
finally 
{
    DisposeDotNetObjectRef( dotNetObjectRef );
}
```

njannink: Isn't the dispose of the NetObjectRef not required to unwire the javascript/c# bridge? So when the jsruntime times out isnt this all gone already?

David-Moreira:
I'm not sure what's going on with RTE as I don't know the component and current implementation, so to give my opinion i'll have to take some time to take a more concrete look then just these snippets.
I'll be taking a look at the multiselect pending PR in the meantime.

And doing testing with `o.JSInteropDefaultCallTimeout = TimeSpan.FromSeconds( 5 );` should be done sparingly in my opinion, that's not an actual real world use case... It's like setting an HTTPClient timeout to 5 seconds...

As for the NetObjectRef, yes it does need to be disposed, it keeps a hold to the reference in memory as far as I know.
I can't find the presentation from someone from microsoft speaking about that now, but there's this article, blazor-university has some good content: 
From reading the article, it seems it's the JSRuntime itself that keeps it holding.
https://blazor-university.com/javascript-interop/calling-dotnet-from-javascript/lifetimes-and-memory-leaks/

stsrki:
> o.JSInteropDefaultCallTimeout = TimeSpan.FromSeconds( 5 );

That is just for testing purposes in case you guys wanted to reproduce the issue. I will remove it before submitting it.

@njannink DotNetObjectReference should always be disposed of. Otherwise, a memory leak will happen.

Source/Blazorise/Components/Bar/Bar.razor.cs
stsrki: Do we want to just catch the exception and ignore it `try {} finally {}`, or do we still want to rethrow it `try {} catch { throw; } finally {}`?

If we ignore it then the app will hopefully go on and dispose of any other resources. In this case `DisposeDotNetObjectRef( dotNetObjectRef );`.

Source/Blazorise/Components/Bar/Bar.razor.cs
```diff
@@ -128,15 +128,10 @@ protected override async ValueTask DisposeAsync( bool disposing )
...
                         await task;
                     }
-                    catch
+                    finally
                     {
-                        if ( !task.IsCanceled )
-                        {
-                            throw;
-                        }
+                        DisposeDotNetObjectRef( dotNetObjectRef );
```
stsrki, line 133: @Unlink @njannink Like this maybe?

Unlink: Yes, that looks good.

stsrki:
I will move this PR into the next milestone. The reason is that I need to be sure it will not break anything before I merge and the current release is already done. And considering that the _problem_ is not that big, it's best to postpone.

David-Moreira:
Seems good to me, I've never used AsyncDisposable, but I'm guessing there should be benefits. :) So I was reading the Microsoft documentation a bit. And IAsyncDisposable can be complementary to IDisposable.
>It is typical when implementing the IAsyncDisposable interface that classes will also implement the IDisposable interface. A good implementation pattern of the IAsyncDisposable interface is to be prepared for either synchronous or asynchronous dispose.

Also I'm not sure if we actually need the implementation on blazorise, what is it trying to solve at the moment?:
https://docs.microsoft.com/en-us/dotnet/api/system.iasyncdisposable?view=net-5.0
>Implementing IAsyncDisposable
You might implement IAsyncDisposable in the following situations:

>When developing an asynchronous enumerator that owns unmanaged resources. Asynchronous enumerators are used with the C# 8.0 async streams feature. For more information about async streams, see Tutorial: Generate and consume async streams using C# 8.0 and .NET Core 3.0.

>When your class owns unmanaged resources and releasing them requires a resource-intensive I/O operation, such as flushing the contents of an intermediate buffer into a file or sending a packet over a network to close a connection.



https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-disposeasync
Also not sure why this was removed:
`GC.SuppressFinalize(this)`
>The DisposeAsync() method
The public parameterless DisposeAsync() method is called implicitly in an await using statement, and its purpose is to free unmanaged resources, perform general cleanup, and **to indicate that the finalizer, if one is present, need not run**. Freeing the memory associated with a managed object is always the domain of the garbage collector. Because of this, it has a standard implementation:

Also here's an example implementation from one of microsoft's classes:
https://github.com/dotnet/corefx/blob/a7e2585c2a0748fc8105866dc3a25d076e5cfbc6/src/Common/src/CoreLib/System/IO/Stream.cs
https://github.com/dotnet/corefx/blob/a7e2585c2a0748fc8105866dc3a25d076e5cfbc6/src/Common/src/CoreLib/System/IO/BinaryWriter.cs

They seem to always wrap the async one in this try catch pattern:
```
public virtual ValueTask DisposeAsync()
{
    try
    {
        Dispose();
        return default;
    }
    catch (Exception exc)
    {
        return new ValueTask(Task.FromException(exc));
    }
}
```

stsrki:
Thank you @David-Moreira for taking the time and investigate it in detail. So, basically, the way Microsoft is doing is almost the same as I implemented. I just have to add `catch` block and revert `IDispose` along with `GC.SuppressFinalize(this)`. And we're good to go.

In case of something break we have the entire 0.9.4 milestone to fix it :)

stsrki:
(hopefully) Final changes committed.

David-Moreira:
I forgot to add,
the pattern that implicitly calls **IAsyncDipose** is this:
```
using System;
using System.Threading.Tasks;

class ExampleConfigureAwaitProgram
{
    static async Task Main()
    {
        var exampleAsyncDisposable = new ExampleAsyncDisposable();
        await using (exampleAsyncDisposable.ConfigureAwait(false))
        {
            // Interact with the exampleAsyncDisposable instance.
        }

        Console.ReadLine();
    }
}
```
https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-disposeasync

So I don't think we're currently using it. So take that into consideration now that you're supporting it. Although I only think the real big benefits would be on IO/network dispose operations.

njannink:
When a class doesnt have a finalizer then this does nothing. `GC.SuppressFinalize(this)` When it does have a finalizer it's good practice to add the `GC.SuppressFinalize(this)`

stsrki:
My thinking was that these are Blazor components and that it's mostly Blazorise classes where I don't use any finalized. But who knows, anyone can override them and add finalizer. So in the end it doesn't hurt to have `GC.SuppressFinalize(this)`.

stsrki:
> I forgot to add,
> the pattern that implicitly calls **IAsyncDipose** is this:
> 
> ```
> using System;
> using System.Threading.Tasks;
> 
> class ExampleConfigureAwaitProgram
> {
>     static async Task Main()
>     {
>         var exampleAsyncDisposable = new ExampleAsyncDisposable();
>         await using (exampleAsyncDisposable.ConfigureAwait(false))
>         {
>             // Interact with the exampleAsyncDisposable instance.
>         }
> 
>         Console.ReadLine();
>     }
> }
> ```
> 
> https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-disposeasync
> 
> So I don't think we're currently using it. So take that into consideration now that you're supporting it. Although I only think the real big benefits would be on IO/network dispose operations.

Hmm, and it seems I need to call `Dispose(false);` from DisposeAsync.

David-Moreira:
That depends on your implementation. 
Both Dispose and DisposeAsync should be doing pretty much the same in terms of cleaning up, the difference is that one is synchronous and the other async.

So, sometimes when calling DisposeAsync, if you don't really have an async dispose implementation, it might make sense to just call the standard Dispose method.

**Microsoft's Stream.cs** does that:
```
  public virtual ValueTask DisposeAsync()
  {
      try
      {
          Dispose();
          return default;
      }
      catch (Exception exc)
      {
          return new ValueTask(Task.FromException(exc));
      }
  }
```
However **Microsoft's BinaryWritter.cs**
does two different implementations since calling async, does actually have an async cleaning up `OutStream.FlushAsync()`:
 ```
public virtual ValueTask DisposeAsync()
{
    try
    {
        if (GetType() == typeof(BinaryWriter))
        {
            if (_leaveOpen)
            {
                return new ValueTask(OutStream.FlushAsync());
            }

            OutStream.Close();
        }
        else
        {
            // Since this is a derived BinaryWriter, delegate to whatever logic
            // the derived implementation already has in Dispose.
            Dispose();
        }

        return default;
    }
    catch (Exception exc)
    {
        return new ValueTask(Task.FromException(exc));
    }
}
```

stsrki:
@David-Moreira Well, I took the code from `Stream.cs` as an inspiration and implemented it in the same way, basically. You links might give me some new ways to optimize it but I think it is OK for now.

PR #1949, opened 2021-03-03
DataGrid: MultiSelectAll: BugFix When only one row, not unchecking the multi select all checkbox

### LINKED ISSUE #1947: Datagrid: MultiSelectAll not auto unselecting when 1 item unselected.
Hello @stsrki overall, so far so good. I only found this problem. Which I'll take a look on Blazorise later today. 

When only 1 item, it does not seem to be unselecting the multi select all, while it does work on the other cases.
Still not sure if it's my fault on the project I'm working on, or smthg on Blazorise Datagrid, I'll be first trying to replicate on Blazorise Datagrid later today.
image

David-Moreira:
Closes #1947 

I also noticed while testing, that clicking Delete triggers the RowClick, which means it will change SelectedRow / SelectedRows... This should only potentially be a problem with static data. But so far I haven't seen complaints about this which is strange. Not sure if we sould look at this in the future?

Personally, I just unselect SelectedRow myself when doing certain operations on the DataGrid. Since I use Grids that depend on the SelectedRow of other Grids.

### DIFF
--- a/Source/Extensions/Blazorise.DataGrid/_DataGridMultiSelectAll.razor.cs
+++ b/Source/Extensions/Blazorise.DataGrid/_DataGridMultiSelectAll.razor.cs
@@ -32,6 +32,10 @@ namespace Blazorise.DataGrid
                 if ( !hasSelectedRows || unselectedRows )
                     IsChecked = false;
             }
+            else
+            {
+                IsChecked = false;
+            }
 
             return base.OnParametersSetAsync();
         }
--- a/Source/Extensions/Blazorise.DataGrid/_DataGridMultiSelectAll.razor.cs
+++ b/Source/Extensions/Blazorise.DataGrid/_DataGridMultiSelectAll.razor.cs
@@ -19,23 +19,9 @@ namespace Blazorise.DataGrid
 
         protected override Task OnParametersSetAsync()
         {
-            var hasSelectedRows = ParentDataGrid.SelectedRows?.Any() ?? false;
-
-            if ( hasSelectedRows )
-            {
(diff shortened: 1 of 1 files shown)


Source/Extensions/Blazorise.DataGrid/_DataGridMultiSelectAll.razor.cs
```diff
@@ -19,23 +19,9 @@ internal Task IsCheckedChanged( bool e )
...
-                    IsChecked = true;
-
-                if ( !hasSelectedRows || unselectedRows )
-                    IsChecked = false;
-            }
-            else
-            {
-                IsChecked = false;
-            }
+            IsChecked = ( ParentDataGrid.SelectedRows?.Any() ?? false )
```
stsrki, line 22: Is this better?

stsrki:
We can handle Delete button click event and `stopPropagation` in 0.9.4. It's not a problem for now.

Source/Extensions/Blazorise.DataGrid/_DataGridMultiSelectAll.razor.cs
David-Moreira: Haha I was just talking about the flow of the feature in general. Not trimming down code. 
You can leave the new one, if the logic is exactly the same. :P

stsrki: Haha well, it doesn't hurt.

In case I'm not missing something I think the logic stays the same.

PR #1972, opened 2021-03-06
Optimize the IdGenerator class.

teo-tsirpanis:
The use of `string.Create` reduces the overhead of copying character buffers around, and eliminates the need for thread-local storage.

### DIFF
--- a/Source/Blazorise/Utilities/IDGenerator.cs
+++ b/Source/Blazorise/Utilities/IDGenerator.cs
@@ -1,5 +1,6 @@
 ﻿#region Using directives
 using System;
+using System.Buffers;
 using System.Threading;
 #endregion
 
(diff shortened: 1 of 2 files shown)


Source/Blazorise/Utilities/IDGenerator.cs
```diff
@@ -14,20 +15,22 @@ public sealed class IdGenerator : IIdGenerator
...
         #endregion
 
         #region Methods
 
-        private static string GenerateImpl( long id )
+        private static void GenerateImpl( Span<char> buffer, long id )
         {
-            var buffer = CharBufferThreadLocal.Value;
 
+            var Encode32Chars = "0123456789ABCDEFGHIJKLMNOPQRSTUV";
```
teo-tsirpanis, line 31: Placing the string in a local variable results in more optimized JITted code.

stsrki:
I just removed GenerateImplDelegate and passed GenerateImpl directly. 

Seems to work fine. Thank you for this optimization!

PS. for future reference, don't create PRs on master as I usually don't allow it. See here https://github.com/stsrki/Blazorise/blob/master/CONTRIBUTING.md

teo-tsirpanis:
> I just removed GenerateImplDelegate and passed GenerateImpl directly.

Didn't mention it but that's not the best idea performance-wise; it allocates a delegate every time `Generate` is called, as shown in its IL representation:

image

Something else you could do is remove `GenerateImpl` and move its code to a lambda passed directly to `string.Create` like this:

``` csharp
/// <summary>
/// Returns a random ID. e.g: <c>0HLH7Q6V92BQE</c>
/// </summary>
public string Generate
{
    get
    {
        var id = Interlocked.Increment( ref LastId );
        return string.Create( IdLength, id, (buffer, id) => {
            var Encode32Chars = "0123456789ABCDEFGHIJKLMNOPQRSTUV";
            // Accessing the last item in the beginning elides range checks for all the subsequent items.
            buffer[12] = Encode32Chars[(int)id & 31];
            buffer[0] = Encode32Chars[(int)( id >> 60 ) & 31];
            buffer[1] = Encode32Chars[(int)( id >> 55 ) & 31];
            buffer[2] = Encode32Chars[(int)( id >> 50 ) & 31];
            buffer[3] = Encode32Chars[(int)( id >> 45 ) & 31];
            buffer[4] = Encode32Chars[(int)( id >> 40 ) & 31];
            buffer[5] = Encode32Chars[(int)( id >> 35 ) & 31];
            buffer[6] = Encode32Chars[(int)( id >> 30 ) & 31];
            buffer[7] = Encode32Chars[(int)( id >> 25 ) & 31];
            buffer[8] = Encode32Chars[(int)( id >> 20 ) & 31];
            buffer[9] = Encode32Chars[(int)( id >> 15 ) & 31];
            buffer[10] = Encode32Chars[(int)( id >> 10 ) & 31];
            buffer[11] = Encode32Chars[(int)( id >> 5 ) & 31];
        });
    }
}
```

It will allocate a delegate only on the first time. You can see more about the allocation behavior of delegates at https://github.com/dotnet/roslyn/issues/5835.

And OK, I will keep the branch in mind.

stsrki:
I guess you know more about this than I :) In that case, just revert the delegate and I will merge it. Thanks

Issue #1408, opened 2020-10-25
Table with fixed header position


rad765:
Is it possible to have a Table with fixed TableHeader position while TableRows in TableBody is vertically scrolling`?

stsrki:
You can do it with CSS and it's fairly easy. Here is one example https://stackoverflow.com/questions/21168521/table-fixed-header-and-scrollable-body

rad765:
Thank you I will probably find somthing there, but that leaves me with a quesion. 
When is Blazorise the appropriate tool, and why shold we use it, if we easely can revert to html and CSS?
Sorry if this is a very stupid question, but I'am not very used to web-programming, Html and CSS, as I'm comming from the Xamarin Android world, so I saw Blazor an Blazorise as my entry to this world. (By the way I'm using the Material variant of Blazorise and found this page: [https://mdbootstrap.com/docs/jquery/tables/scroll/](https://mdbootstrap.com/docs/jquery/tables/scroll/) whitch contain mutch what I want.

stsrki:
Blazorise is an abstraction over different CSS frameworks so there is always going to be something missing that is specific to one framework. I try to create new features where they are most needed, but that takes time and resources. So until I cover most of it you will need to fall back to CSS. This is perfectly fine, as CSS is a perfect tool for customizing the look of your app. You should embrace it :)

And why should you use it? If you want to quickly prototype your app without building your own component, which takes time. Also, you can easily switch between frameworks if you desire.

rad765:
Thank you, You do a very great job. It is just beacuse I'm getteing lost in all the differnet tools in the web programming world, as this is a little new to me. I will continue to use Blazorise as far as I can :-)

PR #1977, opened 2021-03-07
DataGrid Virtualize

### LINKED ISSUE #1381: Virtualize option for Datagrid with .NET 5
With the new RC 2 candidate for NET 5, Blazor has the option for component virtualization.

Would it be possible to add support for a bool flag param in the datatable that enabled row virtualization?
I'm using Blazorise for some large projects at work that load and show a high number of rows to the user (500-8k).  

This request isn't a vital addition in my mind, paging has worked well enough for most cases, but if it's not a difficult change then I could see it adding some extra value to the datatable.

David-Moreira:
Hello @stsrki
Here's a very rough start version of the Virtualize on Datagrid. While it didn't take exactly much to get it working with the static data, and it's working quite fine with the filtering and commands, there's a crap ton of work to do with making it behave correctly with everything else.

You can track the progress here and give opinions or even help if you'd like.
Closes #1381

(Moved from dev093 to dev094)


David-Moreira:
Hello @stsrki 
so getting the layout right for this is killing me. And I'm tryiung to do it, before I start getting all the features of the datagrid to work properly with the Virtualize.

So we need to stick the TableHeader always on top.

**There's two major ways I found to do this:**
1 - Setting the outer div that wraps the table to have this style: 
      `overflow: scroll; height: 575px; overflow-anchor: none;`
enabling virtualize to work correctly and the scroll capability.

2 - On this version the only way I found to stick the header on top was applying a style to the table row header cell, **the th**. It won't work on thead or thread tr
```
    position: sticky;
    top: 0;
```
Additional stylying would be needed to keep the columns appearance equal to the rest of the table. And additionally we have two header rows which will make it so that we have to calculate the starting top position for the second header row... 

-------------------------------------

**Second way:**
1 - Set both the thead and tbody to display:block. This enables us to set the necessary style for virtualzie to work at the tbody level `overflow: scroll; height: 575px; overflow-anchor: none;`
2 - Now we'll loose the html tabular capabilities of keeping the tables dimensions in sync (display: table-row...), specially the columns width will be out of sync... javascript will be necessary to keep both widths in sync.

So I don't really like any of the options but at the moment, I'm not seeing any other solutions. From these two I think I like the second way the best.
I'd like to know your thoughts before proceeding with an implementation of this.

stsrki:
Trying it now. I can already see it will going to be a PITA.

- In-memory data works fine. 
- Pager needs to be hidden in virtualize mode, as it doesn't make sense in this mode
- Large data mode. How to even handle this? Do we load batches every time we scroll? What happens with previously loaded data when we scroll back up? Load again?

And that is just with a quick testing 😅

stsrki:
Regarding the header. Previously I've done it the first way. See https://github.com/stsrki/Blazorise/issues/1408

David-Moreira:
> Trying it now. I can already see it will going to be a PITA.
> 
> * In-memory data works fine.
> * Pager needs to be hidden in virtualize mode, as it doesn't make sense in this mode
> * Large data mode. How to even handle this? Do we load batches every time we scroll? What happens with previously loaded data when we scroll back up? Load again?
> 
> And that is just with a quick testing 😅

Yea. It's going to be a PITA. 😅 But like I said I'm trying to first get the layout right, before starting to work on making everything work correctly.
Virtualzie, just like Datagrid has both a provider for "static data" and "dynamic data", I haven't implemented the "dynamic data" yet, but it should be glueing the HandleReadData and the Virtualize dynamic provider together. It should work exactly as Datagrid HandleReadData, pagination and retriveing items on demand.

David-Moreira:
> Regarding the header. Previously I've done it the first way. See #1408
> 
> Edit* but we need to make it work natively with the Table component. So that all providers would work. I can add support for Table or TableHeader, maybe add parameter `bool Sticky` or something like that.

I don't think I've tried this one:
Thanks I'll test it and see if it's a better option then the one's I was trying.
```
table {
  width: 100%;
}
thead, tbody tr {
  display: table;
  width: 100%;
  table-layout: fixed;
}
tbody {
  display: block;
  overflow-y: auto;
  table-layout: fixed;
  max-height: 200px;
}
```

mathi82:
Hi Guys, 
Is there a reason why you are using table for the grid? 
When i started playing with making a grid, if found using div provided flexibility (maybe i was missing something). 

This is a screenshot of the blazorise grid (v0.93)
image

This is a screenshot of something i was working on. 
image

I think grid should look more like a grid than than a table. 

Other issues i noticed.

1. Multiple colum sorting doesn't indicate the sort order. 
2. Column resizing doesn't work properly. 
3. If Editable is selected, resize does not work at all. When grid is not editable, resize doesn't work properly. When i try to make the column wider, it actually shrinks the other columns. I believe this might be related to using a table. 
4. Can the column size be made smaller the largest item in the column? It doesn't look like it is possible.

David-Moreira:
Hello @mathi82 

@stsrki can answer the design decision... :P Personally I'm not a big fan either. But since Blazorise supports many providers, maybe that was easier to maintain compatibility/behaviour across them?

I don't really see anything noticeable in the screenshots... They are both tabular representations of data... The layout differs slightly, but that's a matter of stylying.
I have personally never really given a though about the Grid vs Table debate. But after taking a brief look at google, it seems like Table is more of a view oriented tabular structure while grid is more of a tabular structure that is oriented to provide additional functionality and interactability with the data, which Blazorise.Datagrid seems to be, even if it's using a Table behind the scenes.

1. That seems like a nice catch. Although I don't think it's quite common to see that in tables/grids? Maybe the one's Ive used just take the easy approach of only allowing a single column sorting, now that I was taking a look at a few of them.
2. Can you elaborate? Does it not work properly, or does it not work how you think it should work? And how do you think it should work?
3. I just tested on the Demo at: https://bootstrapdemo.blazorise.com/tests/datagrid It's working fine. Could you provide a repo? As for the resizing shrinking other columns, well it has to get the extra space somewhere? How do you think it should work? Could you provide an example, please?
4. Hmm what do you mean? For example if I have some word like: courageous I could shrink the column just enough that the word would have break to accomodate according to the column width, for example like this?
courag
eous
This does not seem very UI friendly? Was this what you meant?

stsrki:
I can answer the table questions, as @David-Moreira has already covered the rest.

The table is used because it is the most natural and the most _semantic_  way of displaying tabular data in the browser. While divs could be used, it would be a nightmare to make it work across different frameworks. Tables are natively supported by all of them.

mathi82:
Thanks for replying. 
Points 2&3 were supposed to a single item. 
i was testing the resize on the demo site. 
https://materialdemo.blazorise.com/tests/datagrid

Resize does work with editable selected, not sure why it didn't work the previous time. 
In the demo, there is a horizontal scrollbar, if i were to make one of the columns wider, i expect the entire table to become wider.
What actually happens is that the other column(s) shrink to give the extra space for the resizing column.

Regarding point 4, say my column is 100px, and the cell data is 150px for one of the row.
I'm not sure if it is possible to have the column at 100px, it seems like the column width is resized to fit the data.
Ideally it would be better to have ellipsis show that the text is larger than than column. 
This will prevent different row size based on the row content. 

The current version has the filters displayed below the header, shown at all times. 
Would it not be better to have the filter UI displayed only when the user clicks on a button.

David-Moreira:
@mathi82 Oh... you expect the columns to retain their width, but have the table width grow / overflow to accomodate the new dimensions. 
@stsrki is there a setting for that?
Also on the demo, that overflow (horizontal scrollbar) seems to come from the card div not the table itself.

4 / "column squishing": Ahhh I see. You expect this: image
I guess it might be helpful for people with lots of columns / data. Don't think it is supported at the moment, is it @stsrki ?

Filters: I kinda wondered that when I started using the Datagrid as its not very usual  when comparing with other datagrids, but it actually happens to be nice to rapididly be able to filter data without much fuzz. However I do think filter needs some more love, more filter options, maybe with and/or capabilities, and that might end up being the button we're used to see to apply various filtrs to a column, and I'm pretty sure that's in the works, right @stsrki here's the proposal for that: #1978

stsrki:
That's a lot of features to cover.

I suppose it would be a lot easier to do fixed column sizes with CSS than it would be to make it work with Blazorise. (eg. https://stackoverflow.com/questions/15668583/force-table-column-widths-to-always-be-fixed-regardless-of-contents). I mean, it can be done for sure, but not at the moment, considering all the other work.

I don't know if you guys are aware but there is a `Responsive="true"` flag to make the table scrollable for larger content.

As of filtering for #1978, I still not even planned for any milestone. Too much work at patching 0.9.3 and 0.9.4 is already full of tasks. 😅

mathi82:
@stsrki  i tried setting Responsive="true", it had no affect. 
@David-Moreira were you able to get the sticky header to work?

David-Moreira:
> @stsrki i tried setting Responsive="true", it had no affect.
> @David-Moreira were you able to get the sticky header to work?

@mathi82 I haven't gotten back to this yet. Both the solutions I presented worked, but were incomplete as we would have to do additional recalculations to keep everything looking the same.

I haven't tried the solution given by @stsrki yet. But it is looking like the most plausible of the three.

David-Moreira:
Well. I just tested this third option given by you @stsrki 
Same problem as the second option. Header columns go out of sync with body columns... This sticky header thingy is gonna be a bother... T_T let me know if  you have any other ideas. 
image

mathi82:
Hi @David-Moreira , 
Are the column header and cells respecting the requested column widths?

David-Moreira:
Hello @mathi82,
I only tried applying the given css and that's the result, I did no extra debugging as to why it is happening. Kinda got unmotivated not seeing it working right away xD 

Eitherway, since the feature isn't that forward, I think it's better to isolate the feature and do it as part of an extra PR and then merge it in here.
I'll work on it and submit a separate PR @stsrki is that okay?

@mathi82 With the separate branch I just started, this is how it is rendering with only applying the css again. Now time, to figure out why! 
image

stsrki:
@David-Moreira Seems like a vertical scrollbar is taking extra space and as a result, the column has a slight offset regarding the header.

mathi82:
@David-Moreira , @stsrki  I believe the reason for the offset is the columns widths are resized due to the vertical scrollbar.

David-Moreira:
Ooh, very nice catch guys! If I remember correctly there's an overflow option that does not take width, I think it's overlay, I'll test it later today once I get back to it.

David-Moreira:
> > If I remember correctly there's an overflow option that does not take width, I think it's overlay, I'll test it later today once I get back to it.
> 
> Be sure to check in firefox - there's some looooong debate about the correct behavior of overflow overlay - https://bugzilla.mozilla.org/show_bug.cgi?id=1521631 (and i don't know what the final resting state of it is)

Yep does not work on Firefox. See #2214

stsrki:
FYI. I move Blazorise to a new organization so make sure you update the git origin to `https://github.com/Megabit/Blazorise.git`

image

David-Moreira:
I see, thanks!

David-Moreira:
Hello @stsrki 
The css below interferes with fixed header on Datagrid:
image
What do you think about making it `!important` or adding more css specificity as it is important that is set to sticky for it to behave correctly.

stsrki:
It is always better ta have more css specifity then important flag. Its easier to override later.

I'm not at home now so not sure about the context. Is this regarding the table resize?

David-Moreira:
z-index probably should be set to higher by default. 
image
Didn't really find a common base z-index variable, and I found that you seem to just have used a concrete value in checkbox for example.
So setting it up to a concrete value of 10 for now.

Only variable I found was on bootstrap variables.
image

David-Moreira:
> It is always better ta have more css specifity then important flag. Its easier to override later.
> 
> I'm not at home now so not sure about the context. Is this regarding the table resize?

Nope. Applying Table Fixed Header feature to datagrid. The headers have the position set to relative by the datagrid instead of the position set by the Table Fixed header feature. Okay, Will do by adding css specificity instead. Thanks!

David-Moreira:
@stsrki need your opinion on the test im making on the latest commit.
The Virtualize component API uses a StartIndex as the parameter to fetch data. While we are using paging on datagrid, which is not 100% compatible.
**Example:**
Virtualize API requests:
 - StartIndex = 12 and 10 Items. 
   - We can't just request page 2 and 10 items, as we would be returning 2 already existing records and it messes it up. :P

Do you think we should add a new parameter to the ReadData with the VirtualizeStartIndex, or we should just straight up set up a new Functon to fetch data for the Virtualize Feature?

While I normally like a good separation of concerns, I don't think it can be that harmful to add it on **ReadData**, what do you think?
The closer to have our users to almost only activate a flag and immediatly have Virtualize working with almost no changes, would be pretty cool. Although it might be clearer if there's a VirtualizeReadData...

What's your opinion?

Source/Extensions/Blazorise.DataGrid/DataGrid.razor
```diff
@@ -124,10 +124,10 @@
             }
             @if (IsDisplayDataVisible)
             {
-                @if ( VirtualizeManualReadMode )
+                @if (VirtualizeManualReadMode)
                 {
 
-                    <Virtualize @ref="virtualizeRef" TItem="TItem" Context="item" ItemsProvider="HandleVirtualizeReadData">
+                    <Virtualize @ref="virtualizeRef" TItem="TItem" Context="item" ItemsProvider="HandleVirtualizeReadData" ItemSize="(float)63.75">
```
stsrki: Why exactly 63.75? Is this a magical number 😅? Might be a good idea to make it a parameter.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -592,10 +592,21 @@ protected async Task HandleReadData( CancellationToken cancellationToken )
 
         protected async ValueTask<ItemsProviderResult<TItem>> HandleVirtualizeReadData( ItemsProviderRequest request )
         {
-            CurrentPage = request.StartIndex + 1;
+            var itemIndex = request.StartIndex;
+            var requestItems = Math.Min( request.Count, TotalItems.Value - itemIndex );
+
+            PageSize = requestItems;
+            //TODO: Do we need to introduce a new API that takes in a startIndex?
```
stsrki, line 599: I'm not sure what this does or why we need a new API. Can you explain the background?

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -237,7 +237,7 @@ private async void OnPageSizeChanged( int pageSize )
 
             if ( ManualReadMode )
             {
-                await InvokeAsync( () => HandleReadData( paginationContext.CancellationTokenSource.Token ) );
+                await InvokeAsync( () => HandleReadData( -1, paginationContext.CancellationTokenSource.Token ) );
```
stsrki, line 240: Do we need to introduce -1?

stsrki:
> Do you think we should add a new parameter to the ReadData with the VirtualizeStartIndex, or we should just straight up set up a new Functon to fetch data for the Virtualize Feature?

I like the idea of keeping ReadData with new parameters more than introducing a new set of APIs. Today I plan to finish other work and I will test this PR so I know more about what happens exactly.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
David-Moreira: No we don't... I kinda just did random test code on that commit 😆
However it might be have to be the default as start idx 0 should be valid.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor
David-Moreira: Test code. I think I removed it. Most Virtualize parameters will go into a VirtualizeOptions class.

David-Moreira: Reason why I commit so often, is generally when I have to stop work, and I'm paranoid about loosing work locally and it not being saved remotely...

stsrki: I can relate. There were few times I got burned by failed disk 🔥

But still didn't learn my lessons and committing more😅

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
David-Moreira: Virtualize API has a delegate that tells you two parameters.
I want the data from this index forward and I want this quantity of data. 
You answer with the data and total data. So ReadData already does this and we can take advantage of it, by passing in the Virtualize request to read data, However instead of current page it asks for current index, and that is why we need a new parameter. We don't exactly need to expose a new api, that would be only for separation of concerns... Which again I really don't think we need to separate, was asking your opinion. 

So you can basically use pagesize for the quantity of data.
But have to introduce a new start index for the ReadData arguments so that the developer can use the start index instead of page size.

There might be left over code from testing with trying to make it work with page, but I don't think it will ever be possible to do it 100% from what I've tried.

stsrki: OK, I understand now. Yeah, the new parameter makes sense. Otherwise users would have to to calculate index each time and this way we can help them a little.

David-Moreira: And I was thinking... Since we already introduce start idx might be more clear to just introduce VirtualizeRequestCount instead of reusing page size, since we couldn't straight up get it working with the pagination arguments. 
Meaning VirtualizeRequestStartIndex and VirtualizeRequestCount are the two new ReadData arguments being proposed.and maybe an argument flag that tells if the ReadData call is a manual one or a Virtualize one which should help people building grids that support both options.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor
David-Moreira: Haha so you've felt it on your skin...! I think I have lost it only once, and it was quite a bit of work... However it was my fault, not the disk... 😆

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
stsrki: Makes sense. Just call them `VirtualizeStartIndex` and `VirtualizeCount` instead.

Maybe have an `enum ReadDataMode { Virtualize, Manual }`? Or if `VirtualizeStartIndex` and `VirtualizeCount` are not-null we can assume virtualize mode is on.

David-Moreira: I think I like the enum more then the nullable which can hide bugs, and the developer doesn't have to check for nullabikity which isn't clear and as readable / clean as using the enum. 
The enum makes it clear that the Datagrid wanted either a Virtualize or Manual request. Seems good to me.

David-Moreira: Also do the users outside of the data grid's internal api naming knowledge know that manual is the same as a paged request? Maybe the manual enum could be paging or page... Not sure.

stsrki: Yeah makes more sense to have enum. And naming it Paging instead of Manual is much better

stsrki:
Is it just me or is virtualize really slow? When I scroll it takes more than a second to see new items.

David-Moreira:
> Is it just me or is virtualize really slow? When I scroll it takes more than a second to see new items.

It calls a ReadData, it should be as slow as the data arrives. There's a task.delay 800 on ReadData to simulate a longer call. Is that what you are talking about? However good calls should have a 100/200 ms latency tops.

stsrki:
Yeah, making that delay smaller helps but it still feels weird when you scroll. Even in non-large-data mode. I think I might invest some time in optimizations and making the grid not rerender if it doesn't need to. It is a long-standing issue that I always postponed.

David-Moreira:
Hmm @stsrki I'm restesting, 
what browser did you use? Wasm or server? It does not seem that slow to me altough there's some tiny flickering if you scroll real fast.
And setting up a higher overscan value (requests more items then what's displayed to hide in html) also helps with this, which is also a VirtualizeOption that's going to be available.

Either way the re-rendering has to be investigated, but I'm finding it weird I don't seem to have the same exact issue as you.

stsrki:
I was testing on Chrome  90.0.4430.212, resolution 1920x1080.

---

With your latest changes, the flicker is less obvious and seems a little faster. 

There is one issue I noticed. Enable virtualize and large data mode. Disable virtualize. The pagination breaks, It says 5 items, but it shows 20+.

David-Moreira:
> I was testing on Chrome 90.0.4430.212, resolution 1920x1080.
> 
> With your latest changes, the flicker is less obvious and seems a little faster.
> 
> There is one issue I noticed. Enable virtualize and large data mode. Disable virtualize. The pagination breaks, It says 5 items, but it shows 20+.

One of many 😆 but thanks I'll add it to the list.

stsrki:
Before I go to 😴

What do you think we make demo DataGrid a little smaller? I think it it too crowded and less columns might behave and look better.

David-Moreira:
I guess it could have only the major features, and a couple of columns. Changing the buttons to icons...
It will look better. There's always the documentation for when someone is looking for a specific feature... 

Also we could just make the settings look nicer and hidden by default. Either having them behind an accordion or on a modal for example. 

However the page is kinda helpful for testing haha, maybe keep a test page just for us.

stsrki:
We can focus just on less columns and smaller buttons for now. Settings can stay as they are helpful for everyone.

David-Moreira:
When do you plan to release 0.9.4? I'm taking a 2 week vacation. 

So I don't know if I'll have more or less time to help with this. 😅

stsrki:
Don't worry. It will not be that soon, I still have ~20 issues left on backlog :)

David-Moreira:
Hello,
I augmented the Demo's DataGrid ReadData and VirtualizeReadData to have a simple filtering by basically copy pasting the inbuilt filtering system, so we can also test it out / see it working.
image

I noticed that the sorting is fine, but the search column info api, does not seem to have enough info about the filtering issued by the datagrid, is that intended? 
It's missing the CustomFilter and the DataGridFilterMethod. The FilterMethod, since it's general, shouldn't be hard for the developer to find out/grab out of datagrid's ref, but the customfilter was already defined column by column, maybe it should be available back here? I know this API is mostly for grabbing data out of a BD or a WebAPI, but the dev might want to apply that custom filter once the response data comes back or something, without having to take the hassle of somehow remap the delegate back again to each corresponding column or make a lookup to the datagrid's ref columns?

David-Moreira:
Also, about the performance, I made a simple change that it should now be better. I was reusing a component's lifecycle methods by inheritance which we did not need to do.
Still WASM looks slower then server. I know that they're doing considerable improvements to WASM again in net6.0, so if we eventually cannot improve it much further, we can eventually push the feature to 6.0.

Only other thing I noticed is that most grid actions trigger a double OnAfterRenderAsync, I don't think at the moment, it matters alot because we don't do too much work there. But if every other component is also doing it, it might be costly. But I haven't taken too much time to take an in depth look at it yet, just throwing it out there.

David-Moreira:
Done Few more fixes. Getting closer... 

**Notes:** 

**Resize** -> Can't see resize's border through the filter row... no idea why... So far tried messing with z-index, background color, etc.. Filter Row just overlaps the resizer border

**DetailRow / Aggregates -** Made them unsupported. They don't really function well with Virtualize ReadData.
Aggregate does not really make sense here, since Virtualize can load extra items into memory which are not yet visible as for DetailRow: (Virtualize ItemsProvider, calculates which index to request based on perceived row height, if the height changes, it requests incorrect data) 

**Edit Mode Form -** Made it unsupported. Does really work well with Virtualize ReadData. 
(Virtualize ItemsProvider, calculates which index to request based on perceived row height, if the height changes, it requests incorrect data)

**Filtering** is working okay but needs to somehow reset Virtualize ReadData.
i.e: I'm on index 1000 with no filters, and I add the filter name == "David". it will request for name == "David" and items starting on index 1000. It should go back to index 0.

David-Moreira:
I have been throwing various attempts at this one:
> "**Filtering** is working okay but needs to somehow reset Virtualize ReadData.
i.e: I'm on index 1000 with no filters, and I add the filter name == "David". it will request for name == "David" and items starting on index 1000. It should go back to index 0."


I can't find a way to reset the Virtualize. Do you have any ideas to help that could help solve this problem?
Only way I'm thinking is tinkering with the internals through reflection... which isn't pretty at all, even then, I'm not sure it's that easy. I'm also thinking if we should open an issue on Blazor. Seems like something that should be available on the Virtualize's API. A way to just straight up tell it to reset back to initial state.

stsrki:
Have you tried with this

```
Virtualize virtualize;
...
virtualize.RefreshDataAsync()
```

David-Moreira:
Ya, that's just to trigger a reload.

David-Moreira:
@stsrki 
I've figured it out with js interop. Can you help testing, see if you find any bugs we need to solve before releasing this?

**Current relevant notes:**

**Resize** -> Can't see resize's border through the filter row... no idea why... So far tried messing with z-index, background color, etc.. Filter Row just overlaps the resizer border

**DetailRow / Aggregates** - **Made them unsupported**. They don't really function well with Virtualize ReadData.
Aggregate does not really make sense here, since Virtualize can load extra items into memory which are not yet visible as for DetailRow: (Virtualize ItemsProvider, calculates which index to request based on perceived row height, if the height changes, it requests incorrect data)

**Edit Mode Form** - **Made it unsupported**, it sets the mode to inline. Does not really work well with Virtualize ReadData.
(Virtualize ItemsProvider, calculates which index to request based on perceived row height, if the height changes, it requests incorrect data)

stsrki:
> **Edit Mode Form** - **Made it unsupported**, it sets the mode to inline. Does not really work well with Virtualize ReadData.
> (Virtualize ItemsProvider, calculates which index to request based on perceived row height, if the height changes, it requests incorrect data)

Something is wrong with parameters when you created `_DataGridRowDisplay` component. You can force to show edit mode by doing:

1. Select Inline mode
2. Click on the Edit button
3. Go back and select Form mode while editing
4. It should show edit form

Is _DataGridRowDisplay really needed? Too many subcomponents can lead to poor performance. And it is just a few lines more.

David-Moreira:
I didnt get the instructions? 4. It should show edit form? The idea is to make it unsupported.

It was just to help with not having 3 big blocks of repeated code... Would need to do some benchmarking to figure out how it's possibly affecting the performance. but I'm sure it is, even if it's just some ms and adding more stuff to memory, I'll remove it than.

David-Moreira:
@stsrki

David-Moreira:
Like we've talked we are now missing the features that I wanted to make unsupported, due to Virtualize not handling different item sizes.
We'll attempt to make them supported:
- See if we can mess with the Virtualize internals to make them account for different height items and height changes dynamically.
- Create our own Virtualize component solving these issues.

David-Moreira:
Might be worth taking a look into, Variable height with React's Virtualize: https://developer.microsoft.com/en-us/fluentui#/controls/web/detailslist/variablerowheights

stsrki:
> Might be worth taking a look into, Variable height with React's Virtualize: https://developer.microsoft.com/en-us/fluentui#/controls/web/detailslist/variablerowheights

Meaning we would probably need to implement a completely new virtualize component?

David-Moreira:
> > Might be worth taking a look into, Variable height with React's Virtualize: https://developer.microsoft.com/en-us/fluentui#/controls/web/detailslist/variablerowheights
> 
> Meaning we would probably need to implement a completely new virtualize component?

Yea. I'm not sure about waiting on Microsoft... I don't think that would be nowhere near a priority for them...

Here's another example:
https://codepen.io/zupkode/pen/vYEPvrN?editors=1010

Still, in the meantime, if we can figure out a somewhat hacky way of handling the two current variable height use cases, that would at least suffice to get the feature out there. So we can work on a new one with more time.

stsrki:
The new "dirty prototype" works but there are several issues with the approach.

1. You're using strongly typed class-names, in this case, Bootstrap. We can solve it easily by sending class-names from ClassProvider, but there is a bigger problem:
2. Other providers can have different elements structures than in the _Bootstrap_ prototype.
3. When the form is visible, save and cancel buttons are fixed and half-visible:

image

---

I think the original version works well enough. We just need to handle the **scroll-to-top** of form. Scroll position and variable row sizes are going to be a little problem, It's not like users would scroll too much while editing. So for now just skip it. Make it work that once the form is visible it needs to align with the table-top.

We're returning the Datagrid JS file, right?

David-Moreira:
The points you mentioned 1,2,3 would be solved on the final version, yes. It's just a prototype I did to show without allocating too much time.

So in your opinion we should drop the prototype, is that it?
I just had another idea for another prototype, which might be better. Keep track of opened form or opened detail row and if user scrolls away they are removed. But I'm not sure if this would cause a flicker effect. 😥

We need the js file, yes. Or move the scroll to top or make a scroll function on the table.js. But I don't think we can "hack away" the dynamic item size without js

stsrki:
It is a lot better now. I like the scroll-to-top for the new items. Can you do the same for the edit? Once Edit is clicked top of the form should align with the table header. And once the Save/Cancel is clicked it should refocus to the original row. Now it stays few rows down.

stsrki:
Also, can you fix merge conflicts?

David-Moreira:
> It is a lot better now. I like the scroll-to-top for the new items. Can you do the same for the edit? Once Edit is clicked top of the form should align with the table header. And once the Save/Cancel is clicked it should refocus to the original row. Now it stays few rows down.

Good idea. I'll see what I can do.

David-Moreira:
> Also, can you fix merge conflicts?

More?? 😡😂

stsrki:
> > Also, can you fix merge conflicts?
> 
> More?? 😡😂

Haha it's only natural for long-term branches like this.

David-Moreira:
> It is a lot better now. I like the scroll-to-top for the new items. Can you do the same for the edit? Once Edit is clicked top of the form should align with the table header. And once the Save/Cancel is clicked it should refocus to the original row. Now it stays few rows down.

Recovered part of the script I was doing for the dirty prototype, and it looks pretty good like this in my opinion.
Take a look when you have the time.

David-Moreira:
Also I noticed `FixedHeader` for AntDesign is broken, which also breaks `Virtualize` on AntDesign.
But we should fix that on a separate PR. I'll take a look eventually.

Source/Extensions/Blazorise.DataGrid/Configuration/VirtualizeState.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Blazorise.DataGrid.Configuration
+{
+    internal struct VirtualizeState
```
stsrki, line 9: Better make it a `record` like we already have it in the core project. It is a lot easier to use.

David-Moreira: I don't see why a record would be better. It's a reference type used for value semantics. Which is not really applicable here? Not necessary to allocate memory? 
Structs still have their use.
Could you elaborate if I'm missing something?

stsrki: It's easier to compare, eg `state1 != state2` since it has all the Equal overloads already implemented. Also, it doesn't copy the entire object each time. I found it really good for our use-cases.

David-Moreira: Yep value semantics which are currently not needed. And it's internal so we can just change it when it actually makes sense to use the feature in my opinion... But I can change it right away, if you really do make a point in changing it and using records over structures despite both features having different use cases.

stsrki:
Did you notice this happens with the header second row when you scroll in edit mode?

image

stsrki:
I like the new Bogus data faker. But it takes too long to open the page(testet on WebAssembly). Can we cache the result or make it smaller?

David-Moreira:
> I like the new Bogus data faker. But it takes too long to open the page(testet on WebAssembly). Can we cache the result or make it smaller?

I always test on server side, which is super fast, I'll take a look at WebAssembly... but 1000 is such a low number, are we sure it's because of Bogus? I'll take a look.

David-Moreira:
> Did you notice this happens with the header second row when you scroll in edit mode?
> 
> image

I can't even notice it on the image, what's wrong??

stsrki:
> > Did you notice this happens with the header second row when you scroll in edit mode?
> > image
> 
> I can't even notice it on the image, what's wrong??

Go to the first row and editing a row, Then scroll down. YOu will see how the second header row also scrolls.

David-Moreira:
> > I always test on server side, which is super fast, I'll take a look at WebAssembly... but 1000 is such a low number, are we sure it's because of Bogus? I'll take a look.
> 
> It is. Mostly because of the `.RuleFor( x => x.Salaries, salaryFaker.Generate(3) )`. We only need it for `some` rows. Not all.

Did you just edit my own comment... xD
Either way can confirm...
WebAssembly : **7s**
image
vs
Serverside: **200ms**
image

I'll take a look at your suggested fix, but even so, that's 1000x3 => 3000 Shouldn't have such huge impact.

David-Moreira:
> Did you notice this happens with the header second row when you scroll in edit mode?
> 
> image

Actually broke it on #2622. The new resize event of #2622 is working just fine. But the Initialize has to be called at least once.
MY bad, was so focused on testing the resize I did not test scrolling down without risizing.

I'll submit a new PR with the fix, and then we merge back into this, and it should be back to normal.

stsrki:
Works good now 🎊

Now you can focus on any optimization you can think of. And fix some minor formatting here and there. Finally, it is almost ready 😃

David-Moreira:
I'll see what I can do. Including some reading. ^^ 
Found these best practices for webassembly:
https://docs.microsoft.com/en-us/aspnet/core/blazor/webassembly-performance-best-practices?view=aspnetcore-5.0

We do abuse parameters in our internals a bit. Might do some testing with that.
"**Don't receive too many parameters**
If a component repeats extremely often, for example hundreds or thousands of times, then bear in mind that the overhead of passing and receiving each parameter builds up.

**It's rare that too many parameters severely restricts performance**, but it can be a factor. For a <TableCell> component that **renders 1,000 times within a grid, each extra parameter passed to it could add around 15 ms** to the total rendering cost. If each cell accepted 10 parameters, parameter passing takes around 150 ms per component render and thus perhaps 150,000 ms (150 seconds) and on its own cause a laggy UI."

David-Moreira:
This section is also pretty good:
**Implement SetParametersAsync manually**

"One of the main aspects of the per-component rendering overhead is writing incoming parameter values to the [Parameter] properties. **The renderer has to use reflection to do this.** Even though this is somewhat optimized, **the absence of JIT support on the WebAssembly runtime** imposes limits.

In some extreme cases, **you may wish to avoid the reflection and implement your own parameter setting logic manually**. This may be applicable when:

You have a component that renders extremely often (for example, there are hundreds or thousands of copies of it in the UI).
It accepts many parameters.
You find that the overhead of receiving parameters has an observable impact on UI responsiveness."

stsrki:
We may try moving most of the parameters to State record object and then cascading it down to rows and cells. I would leave manual setting of parameters in SetParametersAsync as the last resort. And with cascading object it might not be needed.

Historicaly, I had to use parameters instead of cascade because back than cascade was non existing. And by the time it was introduced I didnt bother to refactor. Well, the time is come :)

David-Moreira:
Already got faster, by just refactoring the internal row parameters into the CascadingParameter.
I'll keep refactoring throughout the week.

stsrki:
Nice. And is just a simple change.

stsrki:
Something is broke now. If you try to click on the Edit button nothing happens. The edit state is not propagated down. If you click on something on a page that can trigger StateHasChanged then it will show an edit form. This is the reason why parameters were used before. But there is a way to fix it. Instead of cascading a DataGrid instance, we need to have a `record DataGridState` object that will be used to store all information about the grid. Having it as a record it will be fast, and Blazor will pick every change.

If you don't have much time, I will try to make something later in the day.

David-Moreira:
Don't worry that's probably just a simple statehaschabged call on the edit method. We can't forget eventcallbscks do it themselves 😊

David-Moreira:
Just took a look, that was the problem. Can you retest if you have time? 
But I do know what're you talking about. When changes become unknown between components. Not the problem here. 

And it's only normal that we're gonna have to retest the whole grid after refactoring haha

stsrki:
StateHasChanged is OK for now. What I was thinking we can make use of _state_ object that will be picked up by Blazor once it changes, without the need for us to explicitly call StateHasChanged.

David-Moreira:
> StateHasChanged is OK for now. What I was thinking we can make use of _state_ object that will be picked up by Blazor once it changes, without the need for us to explicitly call StateHasChanged.

Yes, we should do that. But where it makes sense... I don't think we need it just yet. And in this case the problem might have been exactly the same with the state object passing methods. (which does not make sense to me if the method is not volatile) 

Let's see how the rest of the refactoring goes.

David-Moreira:
Hello,
Done some more refactoring, looking good imo.
I did introduce the SetParametersAsync override on the most rendered components:_DataGridRow//_DataGridMultiSelect//_DataGridCommand and I think the benefits outweight the bother of manually updating these three component's parameters. But just let me know if you'd prefer not to.

I'm done for the day. Tomorrow I'll be focusing on testing what's done so far more thoroughly.
As for the State object I didn't really find a use yet. But I'm pretty sure we'll need it, or something similar to how Columns work, when we want to handle the state of multiple rows, like we've talked on the DetailRow thread. But I would say, for us to not worry with that complexity just yet.

David-Moreira:
Oh, another thing. I have found that this is pretty much called everytime. Which seems to trigger some re-calculations due to Dirty being set. I had tried a simple fix on the DetailRow PR, and it seemed to work fine. But I don't know if it's worth adding that risk to the release. What do you think?
image

Source/Extensions/Blazorise.DataGrid/DataGrid.razor
```diff
@@ -261,3 +207,33 @@
     @ChildContent
 </CascadingValue>
 
+@code {
+
+    private RenderFragment _Pagination => __builder =>
```
stsrki, line 212: Prefer member vars without underscore. Also, rename them to PaginationFragment, or PaginationRenderFragment, or something similar...

Source/Extensions/Blazorise.DataGrid/BaseDataGridColumn.cs
```diff
@@ -1,21 +1,15 @@
 ﻿#region Using directives
+
```
stsrki, line 2: I hate these spaces here 😂. CodeMaid did it

stsrki:
> Oh, another thing. I have found that this is pretty much called everytime. Which seems to trigger some re-calculations due to Dirty being set. I had tried a simple fix on the DetailRow PR, and it seemed to work fine. But I don't know if it's worth adding that risk to the release. What do you think?
> image

Maybe we can check if the sequence is changed? It might be faster to check then to set it dirty and rerender every time

Source/Extensions/Blazorise.DataGrid/BaseDataGridColumn.cs
David-Moreira: You're goddam picky! 😂

David-Moreira:
Fine. I'll reapply the same fix I used in the other branch... Basically the areEquals check and see how it behaves.

Source/Extensions/Blazorise.DataGrid/BaseDataGridColumn.cs
stsrki: Haha someone have to be. I worked on too many projects to know consistency is a must. Especially in a team

David-Moreira (commented):
Your OCD... didn't allow you to rest... :P

stsrki:
> > Clean extra empty lines
> 
> Your OCD... didn't allow you to rest... :P

You will get used to it 😂🤫

stsrki:
Clear Filter is not working. To reproduce:

1. Enable Virtualize and Large Data
2. Filter by using Email and write **Marga**

image

3. Click on Clear Filter and you get `Loading...` rows

image

David-Moreira:
Better just do the formatting on razor files, when we are about to merge... My preview version keeps unformatting it.

David-Moreira:
Hello @stsrki,
Looking good. You can take a look and merge.

Source/Extensions/Blazorise.DataGrid/_DataGridRow.razor.cs
```diff
@@ -47,12 +47,15 @@ public override Task SetParametersAsync( ParameterView parameters )
                     case nameof( Item ):
                         Item = (TItem)parameter.Value;
                         break;
+
```
stsrki: This formating adds too much noise into commits 😅

David-Moreira: I noticed && reverted, I re-fixed the settings, on latest commit.

stsrki: Ignore then :)

Source/Blazorise/Components/Table/Table.razor.cs
```diff
@@ -139,6 +139,21 @@ private async ValueTask RecalculateResize()
             }
         }
 
+        /// <summary>
+        /// If table has FixedHeader. 
+        /// Invokes javascript function moving the table scroll position to the provided pixels.
+        /// </summary>
+        /// <param name="pixels"></param>
+        /// <returns></returns>
+        public ValueTask FixedHeaderScrollTableTo(int pixels )
```
stsrki: This migth be usefull for regular table. Not just for fixed header. What do you think we enable it for all and rename it into ScrollTo

David-Moreira: I'm not on desktop now... But Didn't I implement it on table?

David-Moreira: Oh regular table... But does regular table have scrollbar? That should only be avaibake on fixed header?

stsrki: You're right. It's not working for the regular table. I have just tested it. 

We can try adding this feature https://stackoverflow.com/a/7853216/833106. I think it might be good for longer tables.

David-Moreira: Oh. I didn't know about that feature. That might be a good one yes. Let's try.
Might have both a ScrollToPx and Scroll To Element? 
I'll let you know how it goes once I try it.

stsrki: Yep, I have created a new ticket. Once i merge this you can try making it. Shouldn't be a problem I think.

stsrki:
Merging now. I think it is good now. In case there are issues found we will create new PRs.

PR #1982, opened 2021-03-08
Use a cached delegate in the IdGenerator class.

teo-tsirpanis:
A continuation of #1972.


stsrki:
This was fast. Thank you!

PR #2015, opened 2021-03-11
Datagrid : Resizable : Fix js trying to access inexisting element when Datagrid has already been disposed.

### LINKED ISSUE #2008: Datagrid: Resize feature: js should always check if exists even if halfway through work
Hello @stsrki 
I noticed this:
It's not that easy to reproduce, but has to do, with immediately closing a grid while it still is running the initial rendering process. I kinda only noticed because I have an app with various tabs(similar to the browser tabs) that each have grids, and I was closing them all quickly.

I have to investigate, if I remember correctly we do have a check at the begin of the JS, but it probably happens when it already is halfway through the work. So additional checks should be in place.
Microsoft.JSInterop.JSException: Cannot read property 'querySelectorAll' of null TypeError: Cannot read property 'querySelectorAll' of null     at Object.initResizable (-/_content/Blazorise.DataGrid/blazorise.datagrid.js:7:28)

David-Moreira:
Closes #2008 

this check seemed to be enough
```
if (table !== null) {
    cols = table.querySelectorAll('tr:first-child > th');
}
```
but I added whenever we're accessing the table, just in case.

Also removed 
`cols.forEach(x => x.classList.add('b-datagrid-resizing'));`
`cols.forEach(x => x.classList.remove('b-datagrid-resizing'));`
I have no idea why we had those in. Maybe commited by mistake.
Retested BS/Ant/Bulma/Material.

By the way have you noticed this on Ant? I don't really know the provider, but does not seem normal to me, for a dropdown to not show the selected value until clicked. Happens on other places on the demo with dropdowns.
image

### DIFF
--- a/Source/Extensions/Blazorise.DataGrid/wwwroot/blazorise.datagrid.js
+++ b/Source/Extensions/Blazorise.DataGrid/wwwroot/blazorise.datagrid.js
@@ -86,8 +86,6 @@
                 const mouseDownHandler = function (e) {
                     mouseDownDate = new Date();
 
-                    cols.forEach(x => x.classList.add('b-datagrid-resizing'));
-
                     // Get the current mouse position
                     x = e.clientX;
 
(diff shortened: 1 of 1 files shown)


stsrki:
The PR looks good. I will test it tomorrow in detail after I finish other work.

Yeah, I forgot about that issue with AntDesign. I will open a new ticket now. Thank

stsrki:
Thank you!

Issue #1997, opened 2021-03-09
Blazorise Material - First day of week


danypellerin:
The date picker with blazorise material starts on Monday while the typical first day of week is Sunday. It should at least be configurable if it does not start on the default first day of week.

stsrki:
Yes, I understand it completely. It may be that other date pickers are better at detecting current culture and then adjust the first data accordingly.

Any help is appreciated :)

PR #2018, opened 2021-03-11
Flatpickr support for material DateEdit and TimeEdit

### LINKED ISSUE #2023: Flatpickr for Material provider
Replace the curent buggy implementation of date picker with Flatpickr.

stsrki:
resolves #2023, #1997

The first version of Flatpickr for material DateEdit

**Done:**

- Date and DateTime modes are supported
- First day of the week can be defined
- Better handling of initial bound value for DateEdit and TimeEdit
- Refactor JSRunner methods

**Next:**

- Display format
- 24-hour format for TimeEdit
- Min and Max

@danypellerin can you also test it?


danypellerin:
First day of week is Sunday but the first day which is nice for me but the feature doesn't seem to work. I tried to set it on Monday and Wednesday and it remains Sunday to Saturday. Min and Max works fine.

PR #2031, opened 2021-03-14
Italian translation

andreatosato:
@devrushnet check this please.
@stsrki I think that all translations in italian is fixed.


stsrki:
Looking good. Thank you all!

PR #2083, opened 2021-03-23
Fix DataGrid nested field accesor

### LINKED ISSUE #2082: DataGrid editing nested fields
I have a DataGrid using in-line editing with an OnRowUpdated method. This had been working, but when I upgraded to 9.3 it stopped working. I updated to 9.3.2 today, but have the same problem. When I select the Save button it stays in Edit Mode. As far as I can tell,  I made all of the changes in the release notes. I narrowed it down to trying to change the value of a nested field.

I created the following test case. I set a break point in OnRowUpdated and it never gets called. If you change the Awarded column to Editable=false, the OnRowUpdated method will get called.

Thanks,

```cshtml
@page "/accountforms/test/"
@using Blazorise.DataGrid

<DataGrid  TItem="TestAccount" Data="@accounts" Sortable="true" PageSize="50" Editable="true"
          EditMode="DataGridEditMode.Inline" RowUpdated="@OnRowUpdated">
    <DataGridColumns>
        <DataGridCommandColumn TItem="TestAccount" Width="220px">
            <EditCommandTemplate>
                <Button Clicked="@context.Clicked" Class="btn-secondary mb-sm-1"><i class="fas fa-pen mr-1"></i>Edit</Button>
            </EditCommandTemplate>
            <DeleteCommandTemplate>
            </DeleteCommandTemplate>
            <SaveCommandTemplate>
                <Button Clicked="@context.Clicked" Class="btn-secondary mb-sm-1"><i class="fas fa-save mr-1"></i>Save</Button>
            </SaveCommandTemplate>
            <CancelCommandTemplate>
                <Button Clicked="@context.Clicked" Class="btn-secondary ml-md-1 ml-sm-0 mb-sm-1"><i class="fas fa-times mr-1"></i>Cancel</Button>
            </CancelCommandTemplate>
            <NewCommandTemplate>
            </NewCommandTemplate>
        </DataGridCommandColumn>
        <DataGridColumn TItem="TestAccount" Field="TimeAwarded.DisplayTime" Caption="Awarded" Sortable="true" Editable="true">
            <EditTemplate>
                <TextEdit Text="@((string)(((CellEditContext)context).CellValue))" TextChanged="@(v=>((CellEditContext)context).CellValue=v)" />
            </EditTemplate>
        </DataGridColumn>
        <DataGridColumn TItem="TestAccount" Field="@nameof(TestAccount.Rate)" Caption="Rate" DisplayFormat="{0:c}" Sortable="true" Editable="true">
            <EditTemplate>
                <NumericEdit TValue="double" Value="@((double)(((CellEditContext)context).CellValue))" ValueChanged="@(v=>((CellEditContext)context).CellValue=v)" />
            </EditTemplate>
        </DataGridColumn>
    </DataGridColumns>
</DataGrid>

@code {

    public class ViewTime
    {
        public int StoreTime { get; set; }

        public string DisplayTime { get; set; }
    }

    public class TestAccount
    {
        public ViewTime TimeAwarded { get; set; }

        public double Rate { get; set; }
    }

    List<TestAccount> accounts;

    {
        accounts = new List<TestAccount>();
        accounts.Add(new TestAccount()
        {
            TimeAwarded = new ViewTime() { StoreTime = 480, DisplayTime="8" },
            Rate = 15.5
        });
        accounts.Add(new TestAccount()
        {
            TimeAwarded = new ViewTime() { StoreTime = 960, DisplayTime = "16" },
            Rate = 17
        });
        accounts.Add(new TestAccount()
        {
            TimeAwarded = new ViewTime() { StoreTime = 6000, DisplayTime = "100" },
            Rate = 14
        });
    }

    private void OnRowUpdated(SavedRowItem<TestAccount, Dictionary<string, object>> employee)
    {
        string temp = employee.Item.TimeAwarded.DisplayTime;
    }
}
```

stsrki:
fixes #2082 DataGrid editing nested fields

### DIFF
--- a/Source/Extensions/Blazorise.DataGrid/Utils/FunctionCompiler.cs
+++ b/Source/Extensions/Blazorise.DataGrid/Utils/FunctionCompiler.cs
@@ -111,7 +111,7 @@ namespace Blazorise.DataGrid.Utils
 
             Expression subPropertyOrField = null;
 
-            MemberInfo memberInfo = GetSafeMember( item.Type, propertyOrFieldName );
+            MemberInfo memberInfo = GetSafeMember( item.Type, parts[0] );
 
             if ( memberInfo is PropertyInfo propertyInfo )
                 subPropertyOrField = Expression.Property( item, propertyInfo );


paulosdfigueiredo:
I remembered fixing something similar to this.

FunctionCompiler.cs has 2 mehtods that have code replication. 
```
private static Expression GetSafePropertyOrField( Expression item, string propertyOrFieldName )
private static Expression GetField( Expression item, string propertyOrFieldName )
```

I believe GetSafePropertyOrField could use GetField to improve maintainability. Is there any reason not to?

```
private static Expression GetSafePropertyOrField( Expression item, string propertyOrFieldName )
        {
            var field = GetField( item, propertyOrFieldName );

            // if the value type cannot be null there's no reason to check it for null
            if ( !IsNullable( field.Type ) )
                return field;

            // check if field is null
            return Expression.Condition( Expression.Equal( item, Expression.Constant( null ) ),
                Expression.Constant( null, field.Type ),
                field );
        }
```

stsrki:
I really don't remember what was the difference now but I guess it's there for a reason... 🤔

Issue #2103, opened 2021-03-25
Css-Cache for the ThemeGenerator


FStolte:
We recently started using the Blazorise Theme in one of our projects to get our custom primary color into the Blazorise css (for the Stepper component). Now, we noticed a performance impact of using the Theme which slowed down initial site loading from about 600ms to 1.7s. This is on Server-side Blazor, btw.

Since likely most projects will only use a few different Themes, or even just one, I think it might be useful to implement a cache for the ThemeGenerator. If the same Theme is passed several times to the generator, it should not be necessary to generate the css again every time. If this performance impact is not just on our side, then a cache could bring a nice performance boost to the framework.

Of course this only really helps Server-side Blazor, since Blazor Webassembly would have to generate each Theme for each user once, anyway.

stsrki:
It all depends on where the actual problem is. Does the generator takes too much time to build or does it takes slow to read HTML once the styles are loaded? I think I can easily cache the built styles. It's just a matter of few boolean flags. Hope it will help a little.

stsrki:
@FStolte Are you sure we need caching? I'm testing to see how many times Blazor could re-render ThemeProvider and re-generate the theme and I found it only happens once on the app start. Later when I navigate to any page it still doesn't re-generate. Unless I specifically call `ThemeHasChanged()`.

FStolte:
I think you might've misunderstood me. In a *single session*, the generation only happens once in the beginning as you say. But it happens once for *every* session, so for everyone accessing the page, the first load is delayed. I was talking about caching the css *across sessions* in a singleton service, so that it ideally only needs to happen once each time the app is deployed (for Server-side Blazor, of course).

stsrki:
Ahhh that. Well, it is working as expected. Remember, even if you have 100 users, one of those users can still choose to have a different theme color. That is, if you have a way to choosing a theme in your app.

The only thing I can do is to maybe move a generator into a service class that can be injected through the DI. That way you could re-register it as singleton maybe.

FStolte:
"even if you have 100 users, one of those users can still choose to have a different theme color"
That's why I was thinking of basically storing the css for every unique Theme if it was requested once, with an upper limit on how many Themes are cached totally. If the same Theme is requested twice, we don't need to re-generate it as long as it is still in the cache. So if 98 of those 100 users have the default theme and 2 picked custom ones, 97 users could still benefit from having the default theme cached.

[...]

PR #2153, opened 2021-04-03
Cache for the ThemeGenerator

FStolte:
This PR adds a ThemeCache for the ThemeGenerator. Whenever the generation of styles or variables for a Theme is triggered, the generator first checks if that Theme is already in the cache. If the Theme is found, the cached string is returned, otherwise it is generated and stored in the cache. If the size limit of the cache is reached, the last used Theme is removed from the cache.

The size of the cache is configurable via the BlazoriseOptions, and setting it to 0 practically deactivates the cache.

The effects of the cache are:
- For Blazor Webassembly, a Theme must only be generated the first time the user selects it. When switching between two Themes, there is a noticable delay the first time, while the change happens almost instantly after that.
- For Server-side Blazor, the benefit is even bigger since the cache is shared between all users. This means that the standard theme for a website must ideally only be generated once, which improves initial loading times for all but the first user accessing the site.

Implements #2103


stsrki:
Hey, @FStolte I'm looking through the PR now and it seems like a nice improvement. The only issue I have with it is that there are many manual works required to do in `Equals` and `GetHashCode`. I know this is the problem with reference types. Now I'm thinking maybe it would be a good idea to convert all theme options classes into `record` type. It would be a lot easier to maintain and a lot easier to compare.

The main issue with records is that it is a breaking change since users would need to use it differently by using the `with` keyword to change the theme options.

What do you think of the idea?

FStolte:
I like it 😁
I was having the same concern that the manual work for maintaining those methods might be a bit of a hassle, and a source of errors.
But I didn't even think about records as a solution for that. If you are ok with introducing a breaking change here, then I'll gladly adjust the implementation.

stsrki:
You can try changing small parts of code into records to see if there are any issue. Also see what it means to update theme options.

We can now do:

`Theme.ColorOptions.Primary = value;`

but with records, it might be a little different. So before you work on it too much please investigate it and let me know how it goes. If it makes sense, a breaking change would not be a problem :)

stsrki:
All theme options are now converted to `records` and I must say it works great. No need for any breaking change. I deliberately did not want them to be immutable (`get; init;`) records so that they are as easier to use as possible. The benefit now is that there is no need to manually define Equals and hash codes. It is all done by the framework.

@FStolte You can try one more time on your side and after you confirm I will merge it. Is that OK?

stsrki:
Made some small changed and I think it is now ready for merge. If you think it needs some fixes we can do it later with new PRs.

I just want to thank you and say that you made good work with this PR!

Issue #2165, opened 2021-04-06
Memory-leak on Server-side Blazor


FStolte:
We have been fighting with strange memory leaks in our Server-side Blazor app for the past weeks, but we could finally pinpoint the cause of the leak, and it lies within the custom ComponentActivator of Blazorise.

*TL;DR: The custom ComponentActivator needs to be scoped instead of singleton, and in long-term the ComponentActivator implementation probably needs to change.*

### Description

The issue is the following: When a transient service is created, normally the DI-container would just create the object, return it and then forget about it. However, when the service is also **disposable**, it needs to keep a reference so that it knows to dispose it when the scope is disposed.
So far, so good. Since the ComponentActivator is currently a singleton, the transient disposable services created by it will be tracked in the root scope of the DI-container, which means they cannot be garbage collected until the root scope is disposed. For Blazor Webassembly this means that all Blazorise UI-Components of one session can only be GCed when its circuit is terminated, which is already a bit inefficient but maybe still acceptable. Unfortunately, for Server-side Blazor this means that all Blazorise UI-Components of all users will remain in memory until the server is restarted. 😟

### How To Reproduce
1. Create a sample Server-side Blazor app with Blazorise
2. Add several Blazorise.Button to a page and reload that page several times
3. Put a breakpoint inside `ComponentActivator.CreateInstance()` and reload to trigger the method
4. Check the field `ServiceProvider._disposables` -> It contains several instances of Blazorise.Bootstrap.Button
5. Wait at least 3 minutes - after this time Blazor terminates a disconnected circuit

#### Expectation
After all circuits are terminated, no references to UI components should be remaining

#### Reality
Checking `ServiceProvider._disposables` again shows that all those references to Blazorise.Buttons are still there

### How to solve
We can mitigate the problem by making the ComponentActivator scoped. Then, Server-side Blazor would behave just as Blazor Webassembly, i.e. UI-components of one session are released once its circuit is terminated.
However, I think that in long-term there should be an improvement to the ComponentActivator, since we still keep all created components around until the end of the session, even if they have already been removed from the UI long ago.

stsrki:
Thank you for this very detailed report. I was able to reproduce it on my side and I will fix it for the next patch.

After that, I will see new ways to improve it. The thing is that all components have to be created as transient but as you already said, it can be a problem when they are disposable. Do you have any advice on how to make it better?

FStolte:
[...]

PR #2173, opened 2021-04-07
Store type mappings for ComponentActivator in Dictionary

FStolte:
This PR deals with #2165 and improves on #2167.

The component instantiation inside the ComponentActivator is now completely done by using `Activator.CreateInstance()` instead of the ServiceProvider. This circumvents the problem that the ServiceProvider holds references to disposable transient services and keeps them from being GCed in a timely manner.

To replace the ServiceProvider registrations of the provider-specific Blazorise components, a type dictionary was added to the ComponentActivator, mapping the default Blazorise types to the provider-specific ones.


stsrki:
It is working. But, there are several issues with this approach and it is not the way I would like it to go.

1. It is breaking change and in some way a regression of the API
2. It will not work for anyone else who implements their own IComponentActivator, see https://github.com/abpframework/abp/blob/48c52625f4c4df007f04d5ac6368b07411aa7521/framework/src/Volo.Abp.AspNetCore.Components/Volo/Abp/AspNetCore/Components/DependencyInjection/ServiceProviderComponentActivator.cs
3. Users are now forced to register their component at runtime instead of with DI at startup

I really appreciate your work and your help on this but I'm really not sure if I can accept this PR :(

FStolte:
Fair enough, I did not think of the Use Case that users might register their own component implementations with the DI-container to replace the default components. That would make this a breaking change, indeed. 🤔

So how about this: I provide this implementation as an alternate TypeMapComponentActivator, which is off by default and can be switched on by setting a flag in the BlazoriseOptions? That way, it would not be a breaking change and would allow everyone to decide for themselves whether they'd rather have better performance or more flexibility.

stsrki:
Yeah, I was thinking about that also but does it really make sense? It is a lot easier to create TypeMapComponentActivator in your own project and replace it within DI. The downside is that in case I register some new component in Blazorise you would also need to do the same on your side.

FStolte:
That is definitely an option for us. But I really wonder if we're the only ones hitting this issue? We have several users that are basically working with our app from 9 to 5, so even with the ComponentActivator being transient, the memory usage for their sessions increases over the day. As a result, the overall used memory is much higher in the afternoon than it was before our update to Blazorise 0.9.3 (while it is comparable in the morning with the same amount of active sessions).

I haven't tested the impact of my new implementation in our productive system yet, but I expect that it will save another ~30% of RAM usage.

FStolte:
@stsrki We have been running the TypeMapComponentActivator in our project for a while now and it works wonderfully. If you don't want to include it in Blazorise, I'd ask for another, smaller change that we could benefit from:

If you kept the component mapping in a Dictionary (as it is in this PR) and iterate through it in order to register the components in the ServiceCollection, then we could refer to that Dictionary in our custom ComponentActivator and would not need to maintain the mappings ourselves. If you're fine with that solution then I'll adjust this PR accordingly when I get around to it.

stsrki:
I might consider it. But I will need to see the code first before making the last call :)

stsrki:
Ouch, force-pushed 😅

image

What was the issue to make it?

FStolte:
I updated the PR, now it only moves the type mappings into Dictionaries which can be used in custom ComponentActivator implementations, like our TypeMapComponentActivator.

stsrki:
The idea looks OK although I would like it better if the list of the mapped components could be read with the same API. So instead of having `Config.AntDesignComponentMap` or `Config.BootstrapComponentMap` we could have just `Config.ComponentMap`.

Maybe have a `ComponentActivatorMap` with a dictionary of components? Or simply add a dictionary to a root Blazorise.Config?

Also, can you please open a new PR? I really hate to deal with force-push. In the past, I had problems with it and I just hate it, to say it mildly 😅

FStolte:
No Problem, I'll create a new PR then :)

Issue #1991, opened 2021-03-09
Fixes 0.9.3


stsrki:
## 0.9.3.10

- #2480 Pagination Items not aligned straight in DataGrid
- #2531 DataGrid.RowSelectable not firing when clicking the DataGridMultiSelectColumn checkbox directly
- #2539 Settings ProgressBar Max doesn't do anything
- #2413 DataGrid inline edit mode forgets text alignment settings
- #2533 DataGridMultiSelectColumn & Filtering Misbehavior
- #2549 DataGrid Selection Behavior on Switching Single / Multiple

## 0.9.3.9

- #2462 Time picker TimeChanged stopped firing
- #2474 Modal fails intermittently with InvalidOperationException: Sequence contains no elements error (#2475 by @GordonRudman)
- #2393 DataGrid Button row disappears on small screens

## 0.9.3.8

- #2437 Time Picker malfunction in 0.9.3.7
- #2435 Date component Max attribute not working as expected with InputMode.DateTime

## 0.9.3.7

- #2268 SetDeltaAsync API in Rich Text control does not work (#2286 by @njannink)
- #2324 Multi-select SelectedValues returns distinct values
- #2321 PR: set dotNetObjectRef to null on dispose
- #2280 Datagrid PageSize bugs
- #2351 24 time format of TimePicker
- #2413 DataGrid inline edit mode forgets text alignment settings
- #2374 EditText/Tab Underline - Theme
- #2395 Line charts throw error if dateset contains NaN data value

## 0.9.3.6

- #2262 Show zeros when DataGrid have no items
- #2253 Link-Type Button cannot be Disabled
- #2240 FileEdit filename text overflows the component
- #2252 FileEdit Validation

## 0.9.3.5

- #2182 Update fr.json
- #1083 Switch ReadOnly, Disabled unexpected behavior
- #1823 TextEdit FocusAsync "Event is already tracked" exception
- #2215 Modal is not centered when it is set to center
- #2211 ReadOnly NumericEdit still editable using arrow keys

## 0.9.3.4

- #2139 Vertical button group is not working
- #2146 DataGrid accents
- #2134 Material Modal behind Layoutheader
- #2122 Unable to select items using Autocomplete extension
- #2165 Memory-leak on Server-side Blazor
- #2155 Typography components should not default to the text-align left

## 0.9.3.3

- #2082 DataGrid editing nested fields
- #2044 JSException when viewed on Edge mobile app
[...]

PR #2182, opened 2021-04-08
Update fr.json

LaurieLavoie13:


imelnikova (approved):
Good catch. Indeed it did not make sense in French to have that last word "pagina", it looks to me like a typo.

stsrki:
Hi, thank you for fixing it. I would approve it but it should have been done on `sup093 `branch instead of the `master`. So it was quicker for me to just recreate it #2183

In any case, I have noted you as a contributor on release notes #1991

Thank you once again!

PS. for future reference https://github.com/stsrki/Blazorise/blob/master/CONTRIBUTING.md

PR #2187, opened 2021-04-10
Adds CustomFilter function on the DataGridColumn

### LINKED ISSUE #1966: Multisearch in the same column.
I am trying to make a column with a dropdown with checkboxes as a filter. The purpose is to filter specific words on the same column.

i've considered something like:
```
<FilterTemplate>
  <Dropdown>
      <Button>
          Items
      </Button>
      <DropdownMenu>
              @foreach (var item in items)
              {
                  <Check TValue="bool" Checked="@selectedStatus.Contains(item)" 
                      CheckedChanged="@((c) => { if (c)
                      {
                          selectedStatus.Add(item);
                      }
                      else
                      {
                          selectedStatus.Remove(item);
                      }

                          **context.TriggerMultiFilterChange(selectedStatus);**

                  })">@item</Check>
              }
      </DropdownMenu>
  </Dropdown>
</FilterTemplate>

@{    
List<string> selectedStatus = new List<string>();
List<string> items = new List<string>();
items.Add("CAR");
items.Add("BIKE");
items.Add("BUS");
items.Add("TRUCK");
items.Add("PLANE");
 }
```

Being `TriggerMultiFilterChange()` a method that let's you concatenate strings. So it can internally do something like:
SearchValue = a or b or c or ...
eg: `SearchValue = "['BIKE'],['TRUCK']";` or `SearchValue = "BIKE" || "TRUCK";`

### Additional context
image
Thanks!

### LINKED ISSUE #1978: Datagrid: Filter method by column // Enhancing Filter
Hello @stsrki 

I just got a requirement where certain columns that only have a handful of values, like for example:
- **active / inactive**

could have a filter dropdown to select which records to display. 

Taking the example above, If filtering by "**active**", the **Contains** `FilterMethod `isn't very effective as it will also display the "**Inactive**" records due to the contains method.
And can't really change the Filter Method to Equals, as it's a more general setting that affects every column.

What about having the feature on a column by column basis? This could also help with extending so that in the feature you can have a more complex filter feature, where you can also click a button and change the type of the filter being applied on each column.

**This here, is a feature that generally grids have and would be pretty cool:**
image

And any ideas on how I could implement this feature in the meantime?

stsrki:
resolves #1966 Multisearch in the same column
resolves #1978 Datagrid: Filter method by column // Enhancing Filter

I have added `CustomFilter` function on the DataGridColumn that can be used to override the internal filter method on per-column basis. Also, I have refactored the existing `SearchValue` from the `string` into an `object` so that it can accept any object type that later is used in the `CustomFilter` function. The downside is that it is a breaking change but I think it is not much of a problem since users only need to add `?.ToString()` to the existing SearchValue usages.

@David-Moreira What do you think? Do you have any ideas for improvements?

### DIFF
--- a/Demos/Blazorise.Demo/Pages/Tests/DataGridPage.razor
+++ b/Demos/Blazorise.Demo/Pages/Tests/DataGridPage.razor
@@ -208,11 +208,11 @@
                                 @($"Total emails: {context.Value}")
                             </DisplayTemplate>
                         </DataGridAggregate>
-                        <DataGridAggregate TItem="Employee" Field="@nameof( Employee.Salary )"                                             
+                        <DataGridAggregate TItem="Employee" Field="@nameof( Employee.Salary )"
                                            AggregationFunction=@((values,col)=>values.Where(v=>v.IsActive).Sum(v=>v.Salary))
                                            DisplayFormatProvider="@System.Globalization.CultureInfo.GetCultureInfo("fr-FR")">
                         </DataGridAggregate>
-                        <DataGridAggregate TItem="Employee" Field="@nameof( Employee.IsActive )" 
+                        <DataGridAggregate TItem="Employee" Field="@nameof( Employee.IsActive )"
                                            AggregationFunction=@(DataGridAggregate<Employee>.TrueCount) />
                     </DataGridAggregates>
                     <DataGridColumns>
@@ -251,7 +251,7 @@
                         <DataGridColumn TextAlignment="TextAlignment.Center" TItem="Employee" Field="@nameof( Employee.Id )" Caption="#" Sortable="false" Width="60px" />
                         <DataGridColumn TItem="Employee" Field="@nameof( Employee.FirstName )" Caption="First Name" Validator="@CheckFirstName" Editable="true">
                             <FilterTemplate>
-                                <TextEdit Placeholder="Search name" Text="@context.SearchValue" TextChanged="@(v=> context.TriggerFilterChange(v))" />
+                                <TextEdit Placeholder="Search name" Text="@context.SearchValue?.ToString()" TextChanged="@(v=> context.TriggerFilterChange(v))" />
                             </FilterTemplate>
                         </DataGridColumn>
                         <DataGridColumn TItem="Employee" Field="@nameof( Employee.LastName )" Caption="Last Name" Editable="true" ValidationPattern="[A-Za-z]{3}" />
@@ -262,7 +262,7 @@
                             </CaptionTemplate>
                             <FilterTemplate>
                                 @{
-                                    var selectedValue = @context.SearchValue ?? "*";
+                                    var selectedValue = @context.SearchValue?.ToString() ?? "*";
                                     <Select TValue="string" SelectedValue="@selectedValue" SelectedValueChanged="@(e => context.TriggerFilterChange(e == "*" ? "" : e.ToString()))">
                                         <SelectItem Value="@("*")">All</SelectItem>
                                         @foreach ( var item in dataModels )
@@ -280,7 +280,18 @@
                         </DataGridColumn>
                         <DataGridDateColumn TItem="Employee" Field="@nameof( Employee.DateOfBirth )" DisplayFormat="{0:dd.MM.yyyy}" Caption="Date Of Birth" Editable="true" />
                         <DataGridNumericColumn TItem="Employee" Field="@nameof( Employee.Childrens )" Caption="Childrens" Editable="true" Filterable="false" />
-                        <DataGridSelectColumn TItem="Employee" Field="@nameof( Employee.Gender )" Caption="Gender" Editable="true">
+                        <DataGridSelectColumn CustomFilter="@CustomGenderFilter" TItem="Employee" Field="@nameof( Employee.Gender )" Caption="Gender" Editable="true">
+                            <FilterTemplate>
+                                <Select TValue="string" SelectedValue="@selectedGenderFilter" SelectedValueChanged="@(value => { selectedGenderFilter = value; context.TriggerFilterChange( selectedGenderFilter ); })">
+                                    <SelectItem TValue="string" Value="@("*")">All</SelectItem>
+                                    <SelectItem TValue="string" Value="@("M")">Male</SelectItem>
+                                    <SelectItem TValue="string" Value="@("F")">Female</SelectItem>
+                                    <SelectItem TValue="string" Value="@("D")">Diverse</SelectItem>
+                                </Select>
+                            </FilterTemplate>
+                            @code{
+                                
+                            }
                             <DisplayTemplate>
                                 @{
                                     var gender = ( context as Employee )?.Gender;
--- a/Demos/Blazorise.Demo/Pages/Tests/DataGridPage.razor
+++ b/Demos/Blazorise.Demo/Pages/Tests/DataGridPage.razor
@@ -280,7 +280,7 @@
                         </DataGridColumn>
                         <DataGridDateColumn TItem="Employee" Field="@nameof( Employee.DateOfBirth )" DisplayFormat="{0:dd.MM.yyyy}" Caption="Date Of Birth" Editable="true" />
                         <DataGridNumericColumn TItem="Employee" Field="@nameof( Employee.Childrens )" Caption="Childrens" Editable="true" Filterable="false" />
-                        <DataGridSelectColumn CustomFilter="@CustomGenderFilter" TItem="Employee" Field="@nameof( Employee.Gender )" Caption="Gender" Editable="true">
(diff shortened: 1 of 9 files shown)


David-Moreira:
Hello @stsrki,
Great! I will checkout the branch and test later.
CustomFilter should give alot of freedom!
Personally I'd like to see inferance of the column type if that is at all possible. 

And in addition: 
With some contextual pre built in simple filters that you'd expect,
i.e: 
Column is string, you can filter by StartsWith, EndsWith, Contains, etc...
Column is number, you can filter by bigger then less then equals, etc...

David-Moreira:
Another example filter/search from a grid I've used before.
http://www.guriddo.net/demo/bootstrap/

Go to Searching/Filtering tab. I personally like the last one. The one with the magnifying glass icon.
image

stsrki:
Seems to me like we can already have the filtering that you propose. It is also named `CustomFilter`, and it is on the `DataGrid` level. It operates on the model.

image

Although it might be a good idea to also have a possibility to build search using the Expression tree 🤔

David-Moreira:
Yea, the developer can construct it himself if he wants with that CustomFilter. However, if it comes out of the box, seems like it would be pretty cool.
Not sure what the benefit of the Expression tree is here, what are you thinking?

We could also just stick to something similar to this one: Since that's more the way Blazorise.Datagrid does filtering.
image

stsrki:
With expression trees, it would be possible to do Where, And/Or, and any other operation. But I'm not sure how much benefit would that bring. Maybe only complexity. It is better to leave it as it is.

David-Moreira:
Yea, Expression Tree might be a bother to maintain.
Generally, implementations of this just take the "all" or "any" approach which simplifies the and / or filtering immensely, I've also found myself in a big dark hole when attempting to do that kind of stuff on something based on rules, and just ended going with the "all" or "any" approach.

Would you eventually consider taking in a PR of something similar to this? 
image

Although I still have Virtualize and ShiftClick feature to complete in the meantime... :P

stsrki:
I think with new CustomFilter on column this can be done by the user so I don't think we need to have it built-in with DataGrid.

David-Moreira:
That's not my point... I know the user can do it... As a user of a library I appreciate having features out of the box instead of reinventing the wheel myself... I know it's another feature to maintain but it seems like a good and common feature in grids... I would not characterize it as such an "extra" feature that it can be just left out and up to the developer to implement. 

For example, It's like if you exposed the new edit delete commands and then just said... Hey the developer can just implement the columns for this himself... And call the commands, I'm not gonna provide him with an already inbuilt way to speed it up. 

Anyway it's up to you as the maintainer. And these stuff are kinda subject to each own personal opinion. If you think it's fine like this. That's okay. I'm guessing we could see if more people eventually requesting the feature come up... And only then do it.

stsrki:
Yeah, I know, and I agree it would be nice to have. I'm just trying to think in terms of maintaining it which is not easy even at the current pace. Not to mention something that can be quite complex as combinations of filtering and building the graph.

We can talk more about the subject. Should be open a new ticket?

David-Moreira:
Yea I kind of figured. You have to think in terms of maintaining as you're a single person (and the ocasional community that helps) maintaining this whole library.

We can take it slow. I'd appreciate it, the ticket yes. I'll see if I can work and close the other stuff on my plate first and we can think about it.

David-Moreira:
Hello @stsrki so I was finally taking a more concrete look at the implementation. Works fine and flexible, just how we wanted.

In the following image you changed Func into a delegate. Although it's more descriptive, I'm pretty sure more and more developers nowadays know Func better then delegate and are more used to it as it's the newest more used feature, altough it works just the same, Func saves up having to declare the delegate. :P 
Just my opinion, I would personally gone with the Func. But I think this is a bit subjective.
image

I'd like to place some questions, if you'd care to answer them. :P
I was wondering why all this boxing? Was it a design decision not to do `DataGridColumn<TItem, TType>` ? Any use case in special? 
It seems that generics would be pretty good here. Helping with type inferance in Filtering and in your Value Getters?

Also why the change of SearchValue to an object? I'm pretty sure you'll be 99% of the time just casting it to string, is there a use case for it being an object?

Otherwise, looks good to me.

stsrki:
Delegate is more descriptive than regular Func. You can easily press F12 and see how it looks and what are the parameters. Func is just a generic delegate that doesn't mean anything on its own. So basically, it is just for more context.

I don't remember why exactly I chose to not add `TType` on DataGridColumn. It was long ago. But I think it was for more flexibility with passing the internal values around. It is a lot easier with `object` type.

I made the SearchValue as `object` instead of `string` because there was a feature request where the user had a use-case to search for item value inside of an **array**. The object can box any value, unlike the string. Also, we already have `TriggerFilterChange` that was needed to trigger grid refresh but it was accepting the string only. So it would be hard to serialize some complex types.

David-Moreira:
"Delegate is more descriptive than regular Func. You can easily press F12 and see how it looks and what are the parameters. Func is just a generic delegate that doesn't mean anything on its own. So basically, it is just for more context."

--> Agreed. However it's just that, yes. If we're talking about parameters it's even faster with Func... it's just right there... :P Also the property name should already be enough of a description. delegate is just for the sake of a named delegate I would say...Again.. it's kinda of opinionated.

Ah, so the searchValue is because of existing a TriggerFilterChange where we can actually pass something else then string. Normally Filtervalue in grids is just a primitive type, according to the column, since it normally comes from a text input on the UI.


Well, now that it's done I would not say, directly add the TType to the DataGridCol itself, since it's a freaking pain to go col by col declaring the TType. The TItem already is... :'( But we already have specific cols where we can infer types without the user declaring it.
**So would you consider something like this?** We could do a POC on a separate branch see how it goes. 
```
BaseCol<TItem,TType>
DataGridCol<TItem> : BaseCol<TItem, object>
DataGridCheckCol<TItem> : BaseCol<TItem, bool>
DataGridDateCol<TItem> : BaseCol<TItem, date>
DataGridNumericCol<TItem> : BaseCol<TItem, int or long>
```

We should be able to get the "generic power" where it makes sense and avoid boxing/unboxing. And still keep the current as it is. And we could also probably add a generic one, if people wanted to use it.
`DataGridTypeCol<TItem,TType> : BaseCol<TItem, TType>`

David-Moreira:
About this:
"Ah, so the searchValue is because of existing a TriggerFilterChange where we can actually pass something else then string. Normally Filtervalue in grids is just a primitive type, according to the column, since it normally comes from a text input on the UI."
Does it actually mesh well with the default filtering? If the user passes something else then a primitive type?

stsrki:
> About this:
> "Ah, so the searchValue is because of existing a TriggerFilterChange where we can actually pass something else then string. Normally Filtervalue in grids is just a primitive type, according to the column, since it normally comes from a text input on the UI."
> Does it actually mesh well with the default filtering? If the user passes something else then a primitive type?

Yes, it will work. As long as you cast it to the right type. It is in full control over the developer.

David-Moreira:
With your default filtering? The one from Datagrid? I'm not talking about the Custom one.

stsrki:
> BaseCol<TItem,TType>
DataGridCol<TItem> : BaseCol<TItem, object>
DataGridCheckCol<TItem> : BaseCol<TItem, bool>
DataGridDateCol<TItem> : BaseCol<TItem, date>
DataGridNumericCol<TItem> : BaseCol<TItem, int or long>

It might be worth trying. But I already see the problem. DataGridDateCol can be both `DateTime` and `DateTime?`, and `TimeSpan` and `TimeSpan?`. DataGridNumericCol can anything from, byte, float, int, long, and also nullable for any of them. How would we bind the values to those?

stsrki:
> With your default filtering? The one from Datagrid? I'm not talking about the Custom one.

Ah, that. yeah, it works fine. I tested it.

David-Moreira:
Honestly, not sure haha. I guess the Type of Cols had to be that precise.
DataGridDateCol and DataGridDateTimeCol, etc...
I wouldn't know what to do with numeric actually

> > BaseCol<TItem,TType>
> > DataGridCol : BaseCol<TItem, object>
> > DataGridCheckCol : BaseCol<TItem, bool>
> > DataGridDateCol : BaseCol<TItem, date>
> > DataGridNumericCol : BaseCol<TItem, int or long>
> 
> It might be worth trying. But I already see the problem. DataGridDateCol can be both `DateTime` and `DateTime?`, and `TimeSpan` and `TimeSpan?`. DataGridNumericCol can be anything from, byte, float, int, long, and also nullable for any of them. How would we bind the values to those?

Honestly, not sure haha, I'll get back to you on that one.
I could see DataGridDateCol  and DataGridDateTimeCol and DataGridTimeCol. But doing the same for the numeric one does not look right. It would actually need to have a TType restricted to those specific primitives.
I'll think about it, and do some digging how have other lbiraries went on about that and get back to you.

PR #2212, opened 2021-04-17
XML comments, code cleanup and syntax optimizations

stsrki:
With this PR I have done:

- Add XML comments to almost all classes and methods.
- Disabled comments only on some internal classes that have no need to be accessed from the user side.
- Added some missing events, (like Clicked) where I think they were missing
- Deleted some unused objects
- Optimized syntax based on VS recommendations

Also, I have skipped over the theme options until #2153 is done so that there are no conflicts. @FStolte if you don't have time to continue I might give some work on your PR?


FStolte:
@stsrki I'd probably be able to continue with the record idea for the ThemeOptions this weekend, but I'm not sure yet. So if you have time on your hand to work on it, then by all means do 😄

stsrki:
Ok np, I will try it. tnx :)

PR #2238, opened 2021-04-24
Russian Localization

SkiTles55:
Hello,
That is Russian localization for Carousel, FileEdit, and DataGrid


stsrki:
Looking good. Thank you!

PR #2285, opened 2021-05-01
RTE - SetDeltaAsync not working

njannink:
Fix for [2268](https://github.com/stsrki/Blazorise/issues/2268)


stsrki:
Can you recreate it on the `sup093` branch? It looks like a bug fix and no API has changed. So I can release it as part of next patch.

PR #2286, opened 2021-05-01
RTE - SetDeltaAsync not working

njannink:
Fix for [2268](https://github.com/stsrki/Blazorise/issues/2268)


stsrki:
Looking good now. Thank you!

PR #2340, opened 2021-05-10
Store type mappings for ComponentActivator in Dictionary

FStolte:
Type mappings are moved into a Dictionary for each provider, so that they can be used in custom ComponentActivator implementations.


stsrki:
LGTM now. Thank you!

PR #2364, opened 2021-05-15
Bulma GenerateAlertVariantStyles fix

Jimmys20:
Hello, I think I found a small bug in Bulma theme generator. Bulma alerts are styled like this `.notification.is-primary` and not like this `.notification-primary`. So the styles inside ThemeAlertOptions where not applied to the Alert component. Now it works.


stsrki:
Good catch. Looks good to me now.

Thank you!

Issue #1840, opened 2021-02-14
Add Selectable parameter to ListGroup


Jimmys20:
Hello, I think it would be nice if ListGroupItem could also generate an `<a>` or `<button>` element with the list-group-item-action  class. What I want to achieve is this kind of html https://getbootstrap.com/docs/4.6/components/list-group/#links-and-buttons. The main reason I would like this functionality is because list-group-item-action has special highlight styling when it's hovered or focused, a plain `<li class="list-group-item">` does not have any hover/focus feedback.

Maybe something like this:
`<ListGroupItem Type="ListGroupItemType.ListItem">item 1</ListGroupItem>`
would generate
`<li class="list-group-item">item 1</li>`
ListGroupItemType.ListItem is the default Type

`<ListGroupItem Type="ListGroupItemType.Button">item 1</ListGroupItem>`
would generate
`<button type="button" class="list-group-item list-group-item-action">item 1</button>`

`<ListGroupItem Type="ListGroupItemType.Link" To="https://blazorise.com/">item 1</ListGroupItem>`
would generate
`<a href="https://blazorise.com/" class="list-group-item list-group-item-action">item 1</a>`

Or maybe use the Mode property of ListGroup component like this:
```
<ListGroup Mode="ListGroupMode.Static">
    <ListGroupItem>item 1</ListGroupItem>
</ListGroup>
```
would generate
```
<ul class="list-group">
    <li class="list-group-item">item 1</li>
</ul>
```
--------------------------------------------
```
<ListGroup Mode="ListGroupMode.Selectable">
    <ListGroupItem>item 1</ListGroupItem>
</ListGroup>
```
would generate
```
<div class="list-group">
    <button type="button" class="list-group-item list-group-item-action">item 1</button>
</div>
```

and finally

```
<ListGroup Mode="ListGroupMode.Selectable">
    <ListGroupItem To="https://blazorise.com/">item 1</ListGroupItem>
</ListGroup>
```
would generate
```
<div class="list-group">
    <a href="https://blazorise.com/" class="list-group-item list-group-item-action">item 1</a>
</div>
```

I believe it makes more sense for a Selectable ListGroup to generate button instead of li elements. This would make a ListGroup with Mode="ListGroupMode.Selectable" actually look 'selectable' because now it looks like a static list without any hover/focus highlight.

I don't know if this can be applied to the other css frameworks that Blazorise supports.

stsrki:
[...]

PR #2365, opened 2021-05-15
Added Color property to ListGroupItem

Jimmys20:
Hello, I tried to add Color property to the ListGroupItem component ([https://getbootstrap.com/docs/4.6/components/list-group/#contextual-classes](https://getbootstrap.com/docs/4.6/components/list-group/#contextual-classes)). Also added support to style the colors from ThemeGenerator. So far bootstrap, material and bulma providers seem to work pretty well. Ant-design does not have classes for styling their list component, so I was not sure how to proceed.

When there is no Theme applied the default colors for material and bulma are too bright compared to bootstrap but that might not be a problem.
Untitled
There is no demo page for the ListGroup component so temporarily I added it to the Dashboard page for easier testing.

I did this mostly as an experiment. I like this project and I wanted to try to understand how it works internally. I believe looking at  other people's projects is a great way to learn new things. So if this is not a feature you wish to include in Blazorise that's fine by me. Otherwise feel free to review and request any changes needed.


stsrki:
So far I think you did a good job. Regarding the missing colors, not everyone will use a theme, so they must be added to the CSS. The colors are added under the `Styles/*.scss` and they should be automatically compiled with this extension https://marketplace.visualstudio.com/items?itemName=MadsKristensen.WebCompiler

You can look at `_jumbotron.scss` in the Bulma project for the simplest example.

Source/Blazorise/Interfaces/IClassProvider.cs
```diff
@@ -499,6 +499,8 @@ public interface IClassProvider
 
         string ListGroupItemDisabled();
 
+        string ListGroupItemColor( Color color );
```
stsrki, line 502: Don't forget to add implementation to all providers.

Jimmys20:
Added implementation to all providers (eFrolic is empty because it's being dropped) and added the missing colors. I think some more styles need to be added for when item is hovered (:hover) or selected (.active) because now it looks bad in some providers.

Source/Blazorise/Interfaces/IClassProvider.cs
Jimmys20: Ok

stsrki:
Yep, those little details like active and hover can be really hard to implement. But they're needed to make it right.

Based on bootstrap implementation(https://getbootstrap.com/docs/4.0/components/list-group/#contextual-classes), the hover should add a slight **darken** effect. And for active or selected state, it needs to be the current primary color. Also, the primary color should be the same as the theme primary color(if it is defined). Otherwise, it always falls back to one defined in Styles, and lastly to the one that came with the CSS framework.

A lot of work I know :)

Jimmys20:
There is a small inconsistency between Blazorise ListGroupItem and bootstrap list-group-item. In bootstrap when you want a selectable list item they suggest that you add the list-group-item-action class. This class adds the on-hover darken effect to the list item.

> Use `<a>`s or `<button>`s to create actionable list group items with hover, disabled, and active states by adding `.list-group-item-action`. We separate these pseudo-classes to ensure list groups made of non-interactive elements (like `<li>`s or `<div>`s) don’t provide a click or tap affordance.

This class also interacts with the contextual color classes and the active class e.g. `list-group-item-action list-group-item-success active` does not actually apply the primary color but a 'special' dark green color.

> Contextual classes also work with `.list-group-item-action`. Note the addition of the hover styles here not present in the previous example. Also supported is the `.active` state; apply it to indicate an active selection on a contextual list group item.

Untitled

Right now Blazorise doesn't apply the list-group-item-action class for selectable list items. This is related to issue I described some time ago [https://github.com/stsrki/Blazorise/issues/1840](https://github.com/stsrki/Blazorise/issues/1840).

I believe that other issue should be addressed first because list-group-item-action class affects the `:hover` and `.active` state styles.

stsrki:
#1840 is already scheduled for 0.9.4 so I can start working on it.

Regarding the selectable state for current `li` items, if it doesn't exist we can add it. I know it is not completely Bootstrap way, but I think it also makes sense to have it. Doesn't need to be `<a>` or `<button>` for that kind of effect.

Jimmys20:
So maybe simply introduce something like `builder.Append( ClassProvider.ListGroupItemAction(), ParentListGroupState.Mode == ListGroupMode.Selectable);` inside BuildClasses method, which will append the `list-group-item-action` class and then add some css like `.list-group-item-action { cursor: pointer; }` instead of using `<a>` or `<button>`? Or you mean to not use the `list-group-item-action` class at all and use custom styles for the different states?

stsrki:
No, you did explain it well. That seems like a simplest way to handle it. No worries, I will try to do it during the day and we'll see how it goes.

stsrki:
FYI. I move Blazorise to a new organization so make sure you update the git origin to `https://github.com/Megabit/Blazorise.git`

image

Jimmys20:
Hi, I tried to add theme implementation for the BootstrapThemeProvider. Can you take a look if I did it the correct way before I continue with other providers?

stsrki:
Will do. I'm currently fixing styles for the Bulma provider.

stsrki:
Bootstrap seems fine. I added the same thing for Bulma and AntDesign.

For Bulma, item colors are now working without a theme provider. The same thing is left to do for AntDesign.

stsrki:
I might be good to move list group examples from the dashboard page to a new separate listgroup page.

stsrki:
Wow, excellent job with the new list page. And you also included new flex and width utilities. 🔝

Jimmys20:
So what is still pending now for this pull-request?

stsrki:
Just need to test everything before the final merge.

stsrki:
All providers now work as expected.

@Jimmys20 It was a lot of work for a newcomer but you managed to do it. I'm impressed. Thank you for all the work so far!

PR #2366, opened 2021-05-15
Icelandic localization for DataGrid,Carousel and FileEdit

sturlath:
Just to give something small back ;-) This is the Icelandic localization for what I found in there. Hopefully I can do more in the future...


stsrki:
Thank you for the help. But could you just move changes to the dev094 branch?

sturlath:
Of course sorry! On it

PR #2379, opened 2021-05-18
Add Selectable parameter to ListGroup

### LINKED ISSUE #1840: Add Selectable parameter to ListGroup

stsrki:
resolves #1840

Done for all providers except for eFrolic since I plan to remove it.

@Jimmys20 I haven't done any items colors since that is already part of your PR. Please review this PR before I merge it. Thanks!

### DIFF
--- a/Source/Blazorise.AntDesign/AntDesignClassProvider.cs
+++ b/Source/Blazorise.AntDesign/AntDesignClassProvider.cs
@@ -547,6 +547,8 @@ namespace Blazorise.AntDesign
 
         public override string ListGroupItem() => "ant-list-item ant-list-item-no-flex";
 
+        public override string ListGroupItemSelectable() => "ant-list-item-actionable";
+
         public override string ListGroupItemActive() => Active();
 
         public override string ListGroupItemDisabled() => Disabled();
--- a/Source/Blazorise.AntDesign/Styles/_list.scss
+++ b/Source/Blazorise.AntDesign/Styles/_list.scss
@@ -2,3 +2,21 @@
     border-right: 0;
     border-left: 0;
 }
+
+.ant-list-item-actionable {
+    width: 100%;
+    color: $list-group-action-color;
+    text-align: inherit;
+
+    @include hover-focus() {
+        z-index: 1;
+        color: $list-group-action-hover-color;
+        text-decoration: none;
+        background-color: $list-group-hover-bg;
+    }
+
+    &:active {
+        color: $list-group-action-active-color;
+        background-color: $list-group-action-active-bg;
+    }
+}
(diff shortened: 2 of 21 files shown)


Jimmys20:
Wouldn't it be better to make use of the existing `ListGroupState.Mode` property instead of adding the new  property `ListGroupState.Selectable`?

stsrki:
How I missed that? I need to rest 😅

Wait for a little, going to fix it.

Jimmys20:
Looks good to me now! 👍

PR #2392, opened 2021-05-20
Add support for hidden default select item in SelectList component 

### LINKED ISSUE #2383: SelectList Placeholder
I was looking for a way to have a Placeholder Item in a SelectList. Basically I wanted to have Placeholder text show in the field if no value had been selected, but for it not to appear in the list.  I modified the Blazorise code to do this and was going to submit a pull request, but when I branched off dev094 I realised that something similar had been done to address #1401.  The difference was that I made the Placeholder SelectItem hidden so it did not appear in the list.

```
@if ( !String.IsNullOrEmpty(PlaceHolder) )
{
    <SelectItem hidden TValue="string" Value="String.Empty">@PlaceHolder</SelectItem>
}
```

I don't think it needs both DefaultItem and Placeholder, so maybe just the option to use the hidden attribute on the DefaultItem. Should hidden be added to SelectItem? I am happy to try to contribute this small change, as my first attempt at an open source contribution :-)

scottk4016:
Includes support for hidden SelectItem component
Fixed support for default SelectItem in AntDesign
resolves #2383


stsrki:
Looking good. I have no complaints whatsoever :)

Thank you!

PR #2432, opened 2021-06-01
Make Bulma provider classes public

Xeevis:
To be able to derive custom provider from existing Bulma provider.


stsrki:
LGTM. Thanks!

PR #2454, opened 2021-06-07
move BaseAfterRenderComponent to Blazorise namespace

Jimmys20:
Hello, all Blazorise base components are using `Blazorise` namespace except BaseAfterRenderComponent which uses `Blazorise.Base`. So, just a small refactor for consistency.


stsrki:
Cool, probably a small overlook from our side. Thank you for fixing it!

Issue #2474, opened 2021-06-14
BUG: Modal fails intermittently with InvalidOperationException: Sequence contains no elements error


GordonRudman:
**Describe the bug**
On occasion, somewhat randomly, but usually on the 3rd time the modal gets shown, the modal dialog fails with a InvalidOperationException: Sequence contains no elements error. FocusableComponents seems to get reset for some reason.

**To Reproduce**
It's an intermittent error, so I'm struggling to reproduce it reliably.

**Expected behavior**
No error to be thrown.

**Screenshots**
Here is the stacktrace of the error:
blazor.server.js:25 [2021-06-14T14:11:23.977Z] Error: System.AggregateException: One or more errors occurred. (Sequence contains no elements)
 ---> System.InvalidOperationException: Sequence contains no elements
   at System.Linq.ThrowHelper.ThrowNoElementsException()
   at System.Linq.Enumerable.First[TSource](IEnumerable`1 source)
   at Blazorise.Modal.<HandleVisibilityStyles>b__14_1()
   at Blazorise.Base.BaseAfterRenderComponent.OnAfterRenderAsync(Boolean firstRender)
   at Blazorise.BaseComponent.OnAfterRenderAsync(Boolean firstRender)
   --- End of inner exception stack trace ---

**Additional context**
I'm using version 0.9.3.8 of Blazorise.
I've tracked the error (or at least the symptom of the error) down to the Linq query in the HandleVisibilityStyles method on the Modal.razor.cs class, and will submit a PR with a proposed (temp) solution.

PR #2475, opened 2021-06-14
Solution to Bug 2474

GordonRudman:
https://github.com/Megabit/Blazorise/issues/2474

blazor.server.js:25 [2021-06-14T14:11:23.977Z] Error: System.AggregateException: One or more errors occurred. (Sequence contains no elements)
 ---> System.InvalidOperationException: Sequence contains no elements
   at System.Linq.ThrowHelper.ThrowNoElementsException()
   at System.Linq.Enumerable.First[TSource](IEnumerable`1 source)
   at Blazorise.Modal.<HandleVisibilityStyles>b__14_1()
   at Blazorise.Base.BaseAfterRenderComponent.OnAfterRenderAsync(Boolean firstRender)
   at Blazorise.BaseComponent.OnAfterRenderAsync(Boolean firstRender)
   --- End of inner exception stack trace ---

### DIFF
--- a/Source/Blazorise/Components/Modal/Modal.razor.cs
+++ b/Source/Blazorise/Components/Modal/Modal.razor.cs
@@ -211,7 +211,14 @@ namespace Blazorise
                 {
                     ExecuteAfterRender( async () =>
                     {
-                        await FocusableComponents.First().FocusAsync();
+                        //TODO: This warrants further investigation
+                        //Even with the Count>0 check above, sometimes FocusableComponents.First() fails intermittently with an InvalidOperationException: Sequence contains no elements
+                        //This null check helps prevent the application from crashing unexpectedly, until a more indepth solution can be found.
+                        var firstFocusableComponent = FocusableComponents.FirstOrDefault();
+                        if ( firstFocusableComponent != null )
+                        {
+                            await firstFocusableComponent.FocusAsync();
+                        }
                     } );
                 }
             }


David-Moreira:
//TODO: This warrants further investigation
                        //Even with the Count>0 check above, sometimes FocusableComponents.First() fails intermittently with an InvalidOperationException: Sequence contains no elements
                        //This null check helps prevent the application from crashing unexpectedly, until a more indepth solution can be found.

ExecuteAfterRender should append the function to be later processed, when on the components after render processing... So we can assume, that later when it is processed the list no longer has items. So makes sense to actually have the check you just introduced.

Now why does it not have items could be the question, if that is to be expected or not which could warrant this to be a definitive fix or not. 

When does this happen exactly? Is it by opening and closing modals?

stsrki:
@GordonRudman Seems good, although I think `if ( FocusableComponents.Count > 0 )` might not be needed anymore. We can just fire and forget with ExecuteAfterRender.

@David-Moreira The reason why it might have failed is that for some reason `HandleVisibilityStyles` was called multiple times and then on the second try there were no more items to fetch. In either case, PR should fix it.

GordonRudman:
@stsrki @David-Moreira Thanks for looking into this.

I've narrowed the error down to when the Modal is shown.
The problem also goes away if at least one of the components in the Modal has Autofocus forced to true. 
I hope that helps.

GordonRudman:
PS: You guys have done a good job on a great library

stsrki:
I have just now noticed the PR is based on `master` branch instead of `sup093`(see [CONTRIBUTING](https://github.com/Megabit/Blazorise/blob/master/CONTRIBUTING.md)).

To make it faster I will recreate it new PR and release it as part of the 0.9.3 patch.

stsrki:
Done with #2486

@GordonRudman Thank you for helping with this bug!

PR #2491, opened 2021-06-17
Pager alignment

### LINKED ISSUE #2480: Pagination Items  not aligned straight in DataGrid
image

David-Moreira:
Fixes #2480 

Moved Field inside ul outside.

Bootstrap 
image
--

Bulma
image
--

eFrolic
image
items per page does not show up on current version. Should we look for a fix? Since we'll be dropping support, right? It should be easy to fix nonetheless.
--

Material
image
--

AntDesign
image
--


David-Moreira:
What the... Where did those commits came from O.O
You can just cherry pick the commit, however if you prefer that I fix the PR just let me know.

stsrki:
Indeed, where did those commits come from??!! When you got the chance open new PR. Better be safe than sorry.

For eFrolic, just ignore it. We're dropping it anyways.

Issue #1994, opened 2021-03-09
DataGrid API


stsrki:
This will be the main ticket for the planning and work process of DataGrid API which would allow us to make changes programmatically.

## Suggestions

- [x] Add method to reset EditMode, eg `.ResetEditMode()`, `.ClearEditMode()` or `.SetEditMode(EditMode.None)`

sks-ttaylor:
I would like to see an insertable FooterTemplate:  Allowing TableFooterRow and TableFooterRowCell

tesar-tech:
- I am currently missing the option to open edit mode by clicking on the (and selecting) specific row. 
- Also, for more complicated scenarios would be helpful to have multiple detail/edit views. For example for editing Salaries in Demo table... Clicking specific button in a row would open specific detail, etc...

fori920:
I'd be helpful to trigger grid refreshes programmatically as well.

stsrki:
We've dome almost all except for

> Also, for more complicated scenarios would be helpful to have multiple detail/edit views. For example for editing Salaries in Demo table... Clicking specific button in a row would open specific detail, etc...

considering we're limited with time we will move to the next milestone.

Issue #1987, opened 2021-03-09
Support for multiple DataGridCommandColumn's


angelru:
It is not possible to do something like that, right? I'd like the edit buttons on the right and the add button on the left.
 
```
<DataGridColumns>
        <DataGridCommandColumn TItem="Item" EditCommandAllowed="false" />
        <DataGridColumn TItem="Item"  Editable="true" />
        <DataGridColumn TItem="Item" Editable="true" />
        <DataGridCommandColumn TItem="Item" NewCommandAllowed="false" EditCommandAllowed="true">
            <EditCommandTemplate>
                <Button Color="Color.Primary" Clicked="@context.Clicked">Edit</Button>
            </EditCommandTemplate>
        </DataGridCommandColumn>
    </DataGridColumns>
```

stsrki:
It is not working because internally we keep a reference to only one command column. So when you add another column it will get the options from the first.

Yuu can just create a regular `DataGridColumn` and then use `DisplayTemplate` to add the buttons.

angelru:
but then, if I want the button to show a modal, I have to do it manually and I can't take advantage of TItem, right?

stsrki:
Yes, that will only be possible once we implement the new API. I will put it on the list for next milestone.

angelru:
thanks!!! great job

angelru:
Thank you very much, but I don't want it to appear in all rows. I can also hope that this issue https://github.com/stsrki/Blazorise/issues/1986# can be set independently and thus leave only the new button and use `DataGridCommandColumn `on the left

David-Moreira:
Yes, it does work. You can have the New Button placed on the ButtonRow, which is placed on the Pager and not the table itself.
You have examples in the demo and in the docs: https://blazorise.com/docs/extensions/datagrid/#datagrid-button-row
Just let us know if you have any doubts.

angelru:
Yes, I thought of ButtonRow, but it is that the pager above is also shown as I comment in the issue, I need it to be independently (these requirements are from the client, not mine)

David-Moreira:
There are some more hacks you can do... Like applying css to the pager setting display:none to the stuff that you would like to not be showing. Just so you can satisfy your client requirements.
But of course, those are temporary hacks, hopefully we can have the suggestion of #1986 about decoupling pager and ButtonRow soon, giving some more flexibility.

angelru:
[...]

Issue #1915, opened 2021-02-25
Clearing filters with a separate button (outside the DataGrid)


kswoll:
**Is your feature request related to a problem? Please describe.**
I want to be able to clear the filters of the `DataGrid` component.  This could be straightforward, but because the `DisplayableColumns` property is protected instead of public, you must subclass `DataGrid` in order to do this.  Please see [this StackOverflow question](https://stackoverflow.com/a/66357930/189950) and answer for a realworld scenario with a concrete workaround that solves the issue.

**Describe the solution you'd like**
I suggest that the `DisplayableColumns` property be made public for user scenarios such as these.

I can make a PR for this if we agree on a solution.

kswoll:
Alternatively, since changing `protected` to public is, I believe, a breaking change, we could expose a new property that simply maps to `DisplayableColumns`.

David-Moreira:
Hello, I might be missing something here, but why is `protected `to `public `a breaking change?

We've been talking about this today. The real solution is exposing some kind of API with the internals to use the Datagrid commands, we need to find a sound solution. I have a few, but I need to investigate further first. If you have any good ideas... :D

So for the ClearFilter question, you can also just use the ClearFilter command... which generally makes more sense to call in the actual provided commands, or you might be missing on usefull internal logic by manually doing it yourself.

And actually you can check yet another temporary workaround I just tested today here: #1914

kswoll:
So I looked at the `ClearFilter` command and it appears to me that it is only for a specific column?  Please correct me if I'm wrong, as I was only introduced to this libarary while trying to help the OP on the SO question.  But the original use case was to be able to clear _all_ the filters of _all_ the columns.  This (I believe) would require iterating over all the columns individually.  That said, I agree with you that utilizing that would be preferable to setting the property to null explicitly.  I'll have to mull that over a bit to see if that means I should change my answer.

As to why it would be a breaking change, I was erroneously considering the case of _overriding_ a `protected` member.  In this case, since the method is not `virtual`, that is actually not a problem, so merely changing the access to `public` would be non-breaking.

David-Moreira:
It's the equivalent to the clear filter command on this demo page that clears every column filter:
https://bootstrapdemo.blazorise.com/tests/datagrid

(Basically does the same iteration that you did, and I think a bit more)
PS: The FirstName and City aren't being cleared because they have a bug on that demo version

stsrki:
[...]

Issue #1278, opened 2020-08-25
DataGridReadDataEventArgs does not provide order of sort columns


gkapelevich:
When using DataGrid with large data ReadData, event arguments contain a list of columns and sort direction for each column.  However there is no way to determine which column should be sorted first.  As I pass the sort arguments to the server I need to know if I should ORDER BY FirstName ASC, LastName ASC or LastName ASC, FirstName ASC for example.

1. Create DataGrid with ReadData method:

```
	<DataGrid TItem="Employee"
			Data="@employeeList"
			ReadData="@OnReadData"
			TotalItems="@totalEmployees">
	…
	</DataGrid>
```

	@code
	{
		Employee[] employeeList;
		int totalEmployees;

		async Task OnReadData( DataGridReadDataEventArgs<Employee> e )
		{
			var request = new
			{
				e.PageSize,
				e.Page,
				Sort = e.Columns
					.Where(_ => _.Direction != SortDirection.None)
					.Select(_ => new { _.Field, _.Direction })  // *** Need a property to OrderBy here ***
			};

			// calling a fictional api
		        var responseMessage = await this.httpClient.PostAsJsonAsync("some-api/employees", request);
			var response = await JsonSerializer.DeserializeAsync<PagedResult<Employee>>(await responseMessage.Content.ReadAsStreamAsync());

			employeeList = response.Data; // an actual data for the current page
			totalEmployees = response.Total;
			StateHasChanged();
		}
	}

2. Sort on one column, sort on second column

It would be very useful to be able to provide the order of columns to sort on to the server API.

stsrki:
And how do you propose the right order of sort columns? By order of clicking the column header maybe?

gkapelevich:
Yes, that would be my approach.  Looks like sortByColumns are already ordered that way.  But they are not used in ManualReadMode.

CPlusPlus17:
I think with this missing feature the data grid is completely usable with big data chunks.
Was doing a proof of concept and had the abort with this limitation.
Looking forward to hopefully see this in 0.9.5.

PR #2503, opened 2021-06-21
DataGrid public APIs

stsrki:
Work on #1994

In this PR we're going to expose as many of internal commands and made them public so that they can be consumed for easier DataGrid manipulation.

## Public DataGrid commands

- New
- Edit
- Delete
- Save
- Cancel
- ClearFilter
- Select
- Paginate
- Sort
- FilterData

## Other APIs

- UseValidation parameter
- Public AddColumn and AddAggregate methods

## Other changes
  - Add `SortIndex` to the `ReadData` event
  - `Direction` parameter made **obsolete** in favor of `SortDirection`
  - Data-annotated Validations

## Closes

- #1987 Support for multiple DataGridCommandColumn's
- #1915 Clearing filters with a separate button (outside the DataGrid)
- #2505 Consider renaming DataGridColumn.Direction to "SortDirection"
- #1278 DataGridReadDataEventArgs does not provide order of sort columns
- #1104 Data grid initial sort direction is not working
- #784 DataGrid: property to expose count of unfiltered/filtered records


stsrki:
@David-Moreira I think I'm almost done with exposing the internal APIs but I need another pair of eyes to make sure I'm not missing something. Can you think of something we might expose publicly?

PS. you can remove table resize in this PR!

David-Moreira:
I've done this on Virtualize PR, there's a few places where in my opinion, would benefit from refactoring the code blocks into a simple more concise Reload() / SetDirty() Since I'm not sure we'll be able to get Virtualize into 0.9.4, might be good to bring those changes over here.
David-Moreira@8cc752a12b3589b7d3addc7bf63b0997db5b3253
David-Moreira@f6d16403d416b5237d688cb2e05f0dc17d475cd0

image

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -383,21 +359,199 @@ protected async Task OnSaveCommand()
...
+        /// Valid <paramref name="pageName"/> values are:
+        /// 1-n:    Number of the page.
+        /// prev:   Go to first page.
+        /// next:   Go to next page.
+        /// first:  Go to first page.
+        /// last:   Go to last page.
+        /// </remarks>
+        /// <param name="pageName">Name or number(1 indexed) of the page.</param>
+        /// <returns>A task that represents the asynchronous operation.</returns>
+        public Task Paginate( string pageName )
```
David-Moreira, line 435: In addition to the already explanatory method doc, a more explanatory parameter name:
`string pageName` => `string pageNameOrNumber`  or  `string paginationCommandOrNumber`
something along those lines, in my opinion. Sometimes good parameter names are really helpful. :D

Source/Extensions/Blazorise.DataGrid/DataGridColumn.cs
```diff
@@ -17,7 +17,7 @@ public partial class DataGridColumn<TItem> : BaseDataGridColumn<TItem>
         private readonly Lazy<Func<TItem, object>> valueGetter;
         private readonly Lazy<Action<TItem, object>> valueSetter;
 
-        private Dictionary<DataGridSortMode, SortDirection> currentDirection { get; set; } = new Dictionary<DataGridSortMode, SortDirection>();
+        private Dictionary<DataGridSortMode, SortDirection> currentSortDirection { get; set; } = new Dictionary<DataGridSortMode, SortDirection>();
```
David-Moreira, line 20: Does this need to be a Dictionary to hold two items? Or could it be a fixed size array with like two value tuples?
I'm pretty sure the memory footprint is less, but we could measure it if in doubt.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -681,7 +681,7 @@ internal async Task<bool> IsSafeToProceed( EventCallback<CancellableRowChange<TI
 
         #region Filtering
 
-        protected void HandleSortColumn( DataGridColumn<TItem> column )
+        protected void HandleSortColumn( DataGridColumn<TItem> column, bool initializing )
```
David-Moreira, line 699: Maybe it could be named `changeDirection` or `switchDirection` which is the real purpose of the flag. To be able to call the method and decide if we should change or not sort direction. Which we do not want to, when initializing. Not a big deal in my opinion.

Source/Extensions/Blazorise.DataGrid/EventArguments.cs
```diff
@@ -162,6 +162,41 @@ public class DataGridReadDataEventArgs<TItem> : EventArgs
         public CancellationToken CancellationToken { get; set; }
     }
 
+    /// <summary>
+    /// Provides all the information for filtered data items.
+    /// </summary>
+    /// <typeparam name="TItem">Type of the data model.</typeparam>
+    public class DataGridFilteredDataEventArgs<TItem> : EventArgs
```
David-Moreira, line 169: Love these! Easy to extend without breaking API, in the future! 👍

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -139,7 +139,7 @@ public DataGrid()
         /// Links the child column with this datagrid.
         /// </summary>
         /// <param name="column">Column to link with this datagrid.</param>
-        internal void Hook( DataGridColumn<TItem> column )
+        public void AddColumn( DataGridColumn<TItem> column )
```
David-Moreira, line 142: Does this API actually work well, with adding columns programatically. It's different from when adding through a razor file which actually does some additional stuff behind the curtains.

For example :  `[CascadingParameter] protected DataGrid<TItem> ParentDataGrid { get; set; }` comes to mind, I think I had trouble implementing something similar to this before.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -162,7 +162,7 @@ internal void Hook( DataGridColumn<TItem> column )
         /// Links the child column with this datagrid.
         /// </summary>
         /// <param name="aggregate">Aggregate column to link with this datagrid.</param>
-        internal void Hook( DataGridAggregate<TItem> aggregate )
+        public void AddAggregate( DataGridAggregate<TItem> aggregate )
```
David-Moreira, line 165: Same as the columns above, adding these type of Razor components programatically.

David-Moreira:
For the commands, I remember I had a few ideas to present, but I didn't end up writing down, doing demos... it's a shame. 
- Bundling the api commands so they are not as scattered.
- Doing component like to bind delegates on the Razor File.
But this is a simple approach and right to the point.
I think you exposed the most important one's.

Could potentially look at:
- exposing MultiSelectAll (Select or unselect all visible rows )
- programatically opening or closing DetailRowTrigger. I know it already does it with a `Func<TItem,bool>`, but it does it in a dynamic fashion, not in a "I tell you to" way.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
stsrki: Done

Source/Extensions/Blazorise.DataGrid/DataGridColumn.cs
stsrki: I used a dictionary because it is easier to access the current sort mode, eg. `currentSortDirection[mode]`. With a list, I would need to get the item, check if it's null, and then get the value.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
stsrki: Done

Source/Extensions/Blazorise.DataGrid/EventArguments.cs
stsrki: Yeah, we should try to make it similar in other places, where it makes sense.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
stsrki: Yeah, you're right. Maybe I can make cascaded ParentDataGrid public and then let developers assign it manually?

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
stsrki: > I've done this on Virtualize PR, there's a few places where in my opinion, would benefit from refactoring the code blocks into a simple more concise Reload() / SetDirty() Since I'm not sure we'll be able to get Virtualize into 0.9.4, might be good to bring those changes over here.
> [David-Moreira/Blazorise@8cc752a](https://github.com/David-Moreira/Blazorise/commit/8cc752a12b3589b7d3addc7bf63b0997db5b3253)
> [David-Moreira/Blazorise@f6d1640](https://github.com/David-Moreira/Blazorise/commit/f6d16403d416b5237d688cb2e05f0dc17d475cd0)
> 
> image

I'm not sure what I need to do for HandleReadData. Can you explain? It is better to be in InvokeAsync because otherwise, users could get deadlock in some rare situations. I know we had it on one of the client projects. For every action that doesn't come from the Blazor event, we must wrap it in InvokeAsync.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
David-Moreira: Since it's a Parameter it should already be public. Maybe as it should be a part of DataGridColumn ctor... I don't think there's an easy way to go about this.

Source/Extensions/Blazorise.DataGrid/DataGridColumn.cs
David-Moreira: Is it not actually the same thing? Since it's for internal use, you do not need to do null checks.
You'd have the same null check problem with the Dictionary, if you did not assume it already has the item you are looking for.
Was talking about something like: `(DataGridSortMode sortMode, SortDirection sortDirection)[2]`.
(Don't know if syntax for array of ValueTuple is correcct)

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
David-Moreira: Nah, just saying the code block is easily replaced by a call to `Reload()` in several places.

Also there are several` dirtyFilter = dirtyView = true;` across the code, those would also benefit from being inside a Method `SetDirty()`

David-Moreira: Also not sure why you quoted that comment in here. >.<

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
stsrki: Well unline Parameter, a CascadedParameter is supposed to be private. At least that is the convention. But nothing is set in stone so it might be a good idea to make it public. It is the DataGrid instance that is already public so it is good in this case.

Source/Extensions/Blazorise.DataGrid/DataGridColumn.cs
stsrki: Let's just leave it as it is 😅

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
stsrki: Haha quote was done by the GitHub somehow

stsrki:
> * programatically opening or closing DetailRowTrigger. I know it already does it with a `Func<TItem,bool>`, but it does it in a dynamic fashion, not in a "I tell you to" way.

We cannot have `ToggleRow` because there is nothing we can toggle on. We don't keep any state of the item or row internally.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
David-Moreira: I was taking a look at the place where I had to handle something like that.
I personally used a **ctor**, which kinda makes it more noticeable and relevant, but I guess making the `CascadingParameter `public is fine also.
_xtabs => `CascadingParameter`
image

If someone ends up using that method for some reason, if I'm not wrong there's going to be a runtime error due to CascadingParameter being null, and the developer will notice it and set the CascadingParameter manually.

David-Moreira:
> > * programatically opening or closing DetailRowTrigger. I know it already does it with a `Func<TItem,bool>`, but it does it in a dynamic fashion, not in a "I tell you to" way.
> 
> We cannot have `ToggleRow` because there is nothing we can toggle on. We don't keep any state of the item or row internally.

I mean, I don't think it's impossible... Might not look pretty and very maintenable tough and we would need to keep more state. Which probably makes it not worth implementing.

Something like: 
```
  public Task ToggleRow( TItem item )
  {
      IsRowToggle = true;
      ToggledRow = item;
  }

@(if (IsRowToggle && ToggledRow.isEquals(item)){
    <RenderDetailRow>
})

...Manage IsRowToggle flag when there are other selection changed.
```

So yea nvm it. haha

stsrki:
> > > * programatically opening or closing DetailRowTrigger. I know it already does it with a `Func<TItem,bool>`, but it does it in a dynamic fashion, not in a "I tell you to" way.
> > 
> > 
> > We cannot have `ToggleRow` because there is nothing we can toggle on. We don't keep any state of the item or row internally.
> 
> I mean, I don't think it's impossible... Might not look pretty and very maintenable tough and we would need to keep more state. Which probably makes it not worth implementing.
> 
> Something like:
> 
> ```
>   public Task ToggleRow( TItem item )
>   {
>       IsRowToggle = true;
>       ToggledRow = item;
>   }
> 
> @(if (IsRowToggle && ToggledRow.isEquals(item)){
>     <RenderDetailRow>
> })
> 
> ...Manage IsRowToggle flag when there are other selection changed.
> ```
> 
> So yea nvm it. haha

Yeah, exactly my point. Although we might need to do it for multi-edit-row mode at one point. But that time is not now :)

stsrki:
@David-Moreira I will merge this now.

Don't make it a priority, but in case you got any time to work on the public ToggleDetailRow method do it.

CPlusPlus17:
There seems an issue with the SortDirection:

```
foreach (var pos in SortByColumns)
{
    Console.WriteLine($"Col is {pos.Field}");
    Console.WriteLine($"Col is {pos.SortDirection}");
}
```

when used on ReadData, it will output always:

> Col name is CreatedAt
> Col sort order is None

As far as I see in the code, the  public Task Sort( DataGridColumn<TItem> column ) and protected void HandleSortColumn( DataGridColumn<TItem> column, bool changeSortDirection ) should be called but I'm not sure when Sort() is triggered. Sortable && column.Sortable are met and should be processed in Sort().

Thanks
Manuel

stsrki:
`SortDirection` is used only to set the initial sort direction. After that it is handled internally whenever a user clicks on a column header.

You should be able to use `Sort` method, but as it says in the method comment, the column should be enabled as `Sortable`.

> Note that `<see cref="DataGridColumn{TItem}.Sortable"/>` must be enabled to be able to sort!

CPlusPlus17:
Ah thanks for the clarification. I know understand that I have to use the IEnumerable<DataGridColumnInfo> dataGridColumnInfos in the DataGridReadDataEventArgs event the archive my goal. Thanks.

PR #2516, opened 2021-06-24
Data annotation validation for DataGrid

stsrki:
Continuation of #2503 

This PR introduces new data-annotation validation on DataGrid. Data-annotation will be the default method. All other validation methods or handlers will override it if they are defined on the `DataGridColumn`.

To _globally_ enable or disable validation I introduce the `UseValidation` parameter on DataGrid.


Source/Extensions/Blazorise.DataGrid/_DataGridCellEdit.razor.cs
```diff
@@ -6,20 +6,47 @@
...
+
+        /// <summary>
+        /// Instance of the currently editing row item.
         /// </summary>
         [Parameter] public TItem Item { get; set; }
 
+        /// <summary>
+        /// Instance of the currently validating row item.
+        /// </summary>
+        [Parameter] public TItem ValidationItem { get; set; }
```
David-Moreira, line 48: Potentially just grab out of Cascading Datagrid? I generally prefer using the Cascading parameter for internal values, over having to set the parameters manually all the way down. What do you think?

Source/Extensions/Blazorise.DataGrid/_DataGridCell.razor.cs
```diff
@@ -34,10 +30,28 @@ protected Action<ValidatorEventArgs> Validator
         protected string ValidationPattern
             => string.IsNullOrWhiteSpace( Column.ValidationPattern ) ? null : Column.ValidationPattern;
 
+        protected Type ValidationHandlerType
```
David-Moreira, line 33: Is this clear to the user? Does the understand that the 3 existing methods of validation are mutually exclusive? 
Can there be a scenario where he has a validator or validation pattern but simply wants to use the data annotations? Or any other combination? 
Which could warrant having a `ValidationMode `enum.

And Is there a scenario where we want to run every single one of them?

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -156,6 +167,13 @@ internal void Hook( DataGridAggregate<TItem> aggregate )
             Aggregates.Add( aggregate );
         }
 
+        //public override Task SetParametersAsync( ParameterView parameters )
```
David-Moreira, line 170: commented code... boooo :P

Source/Extensions/Blazorise.DataGrid/_DataGridCellEditValidation.razor
```diff
@@ -1,8 +1,19 @@
 ﻿@typeparam TItem
 @inherits _DataGridCellEdit<TItem>
```
David-Moreira, line 2: Should this inherit `_DataGridCellEdit<TItem>` or `_DataGridCellEditValidation<TItem>`?
I don't think it quite matters though.

Source/Extensions/Blazorise.DataGrid/_DataGridCellEdit.razor.cs
```diff
@@ -6,20 +6,47 @@
...
+        /// Updated the internal cell values.
+        /// </summary>
+        /// <typeparam name="T">Type of the value being changed.</typeparam>
+        /// <param name="value">Value that is updating.</param>
+        /// <returns>A task that represents the asynchronous operation.</returns>
         protected Task OnEditValueChanged<T>( T value )
         {
             CellEditContext.CellValue = value;
 
+            Column.SetValue( ValidationItem, value );
```
David-Moreira, line 25: I'm not sure if it's a dumb question. 
Can we have various columns calling `Column.SetValue( ValidationItem, value ); `async without any worries, meaning techncally there would be various places mutating the ValidationItem at the same time?

David-Moreira:
I only took a look at code on GitHub.. I'll see if I ahve some time tomorrow, to actually test it.

Source/Extensions/Blazorise.DataGrid/_DataGridCell.razor.cs
```diff
@@ -34,10 +30,28 @@ protected Action<ValidatorEventArgs> Validator
...
+            {
+                if ( HasValidationPattern )
+                    return typeof( PatternValidationHandler );
+                else if ( HasValidator )
+                    return typeof( ValidatorValidationHandler );
+
+                // default is always data-annotations
+                return typeof( DataAnnotationValidationHandler );
+            }
+        }
```
David-Moreira, line 45: TO show up on review related to the comment above

David-Moreira: Nvm... Can't seems to handle multiple line selection properly on reviews...

David-Moreira:
Strange, shouldn't `EmailAddress `DataAnnotation validate properly? It's considering dav@dav as a valid email address. I'm not sure if they changed it? Or if it's something else... But I remember that  on .net framework MVC it validated the final part of the email address.
image

David-Moreira:
I remember you implemented async validation, could it potentially work here? 
It's pretty cool to have the feedback loop before pressing the Save button.

Source/Extensions/Blazorise.DataGrid/_DataGridCellEdit.razor.cs
stsrki: Wanted to follow the same pattern as we already have, but I agree it makes more sense to cascade. I will fix it.

Source/Extensions/Blazorise.DataGrid/_DataGridCell.razor.cs
stsrki: Most people want to use data annotation so I made it the first rule. Every other will override it just as in the regular validation. We can make it more clear in the documentation.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
stsrki: haha

Source/Extensions/Blazorise.DataGrid/_DataGridCellEditValidation.razor
stsrki: No this is good. _DataGridCellEditValidation is partial class the inherits from _DataGridCellEdit

Source/Extensions/Blazorise.DataGrid/_DataGridCellEdit.razor.cs
stsrki: No reason. Users update one field at a time. In what scenario you think it can be from multiple places?

stsrki:
> I remember you implemented async validation, could it potentially work here?
> It's pretty cool to have the feedback loop before pressing the Save button.

Enabling auto validation is just a matter of changing one parameter.

stsrki:
> Strange, shouldn't `EmailAddress `DataAnnotation validate properly? It's considering dav@dav as a valid email address. I'm not sure if they changed it? Or if it's something else... But I remember that on .net framework MVC it validated the final part of the email address.
> image

Email validation is a strange world. In original "specs" you're safe to use about any character or format as you wish. I would say that's why this validation is OK.

Source/Extensions/Blazorise.DataGrid/_DataGridCell.razor.cs
David-Moreira: Personally I could see myself using the data annotations for simple stuff,and then add a validator or validation pattern to do some more complex logic all on the same column. So I do not need to redo the simple validations according to the mode I'm using.
But, to be fair I think data annotations does have its own custom validation and pattern too.

David-Moreira:
> > I remember you implemented async validation, could it potentially work here?
> > It's pretty cool to have the feedback loop before pressing the Save button.
> 
> Enabling auto validation is just a matter of changing one parameter.

I see I'll try it later. Does it work for all validations tough?

Source/Extensions/Blazorise.DataGrid/_DataGridCellEditValidation.razor
David-Moreira: Oh I did not notice you switched to using partial class. You used to do inherit as a way of merging the razor + Cs file.

stsrki:
> > > I remember you implemented async validation, could it potentially work here?
> > > It's pretty cool to have the feedback loop before pressing the Save button.
> > 
> > 
> > Enabling auto validation is just a matter of changing one parameter.
> 
> I see I'll try it later. Does it work for all validations tough?

Yes it is the same behavior for all validation types.

Source/Extensions/Blazorise.DataGrid/_DataGridCellEditValidation.razor
stsrki: Inherit was used before Blazor supported partials. But I changed it couple of months ago.

Source/Extensions/Blazorise.DataGrid/_DataGridCellEdit.razor.cs
David-Moreira: Well when you update, you update a whole row (validation item) which will update several columns which means several oneditvaluechanged will be called potentially simultaneously I would say mutating the validation item. Again... I'm probably just over thinking this.

stsrki: Even if that is the case, we only update one field of an item for that specific column. There will not be clashes for the same field unless the user wrongly defines the field names.

Source/Extensions/Blazorise.DataGrid/_DataGridCellEdit.razor.cs
stsrki: In fact, better to leave it as a parameter. There is a feature request for editing multiple rows at once and with parameters it would be a lot easier to handle it once we start working on it.

Source/Extensions/Blazorise.DataGrid/_DataGridCell.razor.cs
stsrki: Updated documentation.

stsrki:
Going to merge now. In case we missed something we will fix it an #2503

PR #2523, opened 2021-06-25
Table Resize

### LINKED ISSUE #2313: Create column resize on Table
> I was thinking... tables don't have the resize feature like in datagrid, and in retrospective since datagrid uses table behind the scenes, maybe the feature could have been implemented in the table and be available in both, eitherway I was thinking if this is gonna mesh well with it or not, in theory it should still work exactly the same.

I just now understood this message 🤦‍♂️. It might just be a great idea to have it directly on the Table. I don't think it would be hard to move it. Good catch!

_Originally posted by @stsrki in https://github.com/stsrki/Blazorise/issues/2293#issuecomment-834517658_

David-Moreira:
Closes #2313

### DIFF
--- a/Demos/Blazorise.Demo/Pages/Tests/TablesPage.razor
+++ b/Demos/Blazorise.Demo/Pages/Tests/TablesPage.razor
@@ -508,10 +508,21 @@
                     <Column><Switch @bind-Checked="@borderless">Borderless</Switch></Column>
                     <Column><Switch @bind-Checked="@hoverable">Hoverable</Switch></Column>
                     <Column><Switch @bind-Checked="@small">Narrow rows</Switch></Column>
+                    <Column>
+                        <Switch @bind-Checked="@resizable">Resizable</Switch>
+                        @if ( resizable )
+                        { 
+                            <Select  Size="Size.Small" @bind-SelectedValue="@resizeMode">
+                                <SelectItem Value="TableResizeMode.Header">Header</SelectItem>
+                                <SelectItem Value="TableResizeMode.Columns">Columns</SelectItem>
+                            </Select>
+                        }
+                    </Column>
                 </Row>
+
             </CardBody>
             <CardBody>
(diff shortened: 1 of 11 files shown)


stsrki:
Working well except for the second header row. It doesn't detect the mouse hover effect.

stsrki:
@David-Moreira If you think it is ready I can merge it.

David-Moreira:
Let me see if I can figure out the thing with the second header row.
Also I noticed there might be a bug with having th on the tbody, hovering on first tbody columns is having incorrect behaviour.
Also what do you think of doing the Datagrid migration in this same PR or should we do on a separate one?

David-Moreira:
For general formatting so you do Ctrl + K + D or do you run an extension?
I need to figure out how to have the same settings as you on the solution. Actually I can't even do ctrl k d on the blazorise files as it completely replaces the formatting with mine, seems like editor.config only is working partially... Idk... It's gonna be a bother if you have to be reformatting every time...

stsrki:
> Let me see if I can figure out the thing with the second header row.
> Also I noticed there might be a bug with having th on the tbody, hovering on first tbody columns is having incorrect behaviour.
> Also what do you think of doing the Datagrid migration in this same PR or should we do on a separate one?

OK, try to fix it.

We can do DataGrid as part of another PR. Probably best along with my API changes to not have any merge conflict.

stsrki:
> For general formatting so you do Ctrl + K + D or do you run an extension?
> I need to figure out how to have the same settings as you on the solution. Actually I can't even do ctrl k d on the blazorise files as it completely replaces the formatting with mine, seems like editor.config only is working partially... Idk... It's gonna be a bother if you have to be reformatting every time...

Yes, I do Ctrl + K + D and it picks all the settings from editorconfig. The files that are problematic for everyone else are `.razor` because they don't work well with editorconfig.

I guess for me it works because, I also have the same settings globally on my global VS options.

David-Moreira:
@stsrki You may go ahead and merge.

I haven't figured out the 
"Working well except for the second header row. It doesn't detect the mouse hover effect." yet, I also had found out that when working with Virtualize, happens when FixedHeader + Resizable are active. Somehow the background color overrides it, even if we mess with the z-index...
Well, I'm sure I will figure it out eventually and PR a fix later.

PR #2548, opened 2021-06-30
Autocomplete improvements

### LINKED ISSUE #1485: Autocomplete: Able to type freely
I imagine a use case will be that autocomplete just augments the writing providing "suggestions". I think that it would be a nice to have, to be able to still freely type a value outside of the **AutoComplete Data** Values.

Another nice to have feature (Open another feature request if needed), is to have it display all available values on click with a scroll bar if there's alot of values. 
Then, if you type, it starts filtering the results.

Seems like a flag for both these features would be a nice to have.

### LINKED ISSUE #1890: Suggestion: Select2 like component.
I currently took in a dependency to a select2.js in a project of mine with a component wrapping the interop, but would like to remove it, if there's already something similar in a blazor component.

I was taking a look at AutoComplete extension and it already does part of it. Just that it could be augmented to have more features. Basically a merge between Select and AutoComplete.

**Examples:**
image

image

image

David-Moreira:
Closes #1485
Closes #1890 

Current Progress:
- [x] Free Typing binding support
- [x] Free Typing validation support
- [x] NotFound event callback
- [x] NotFoundTemplate
- [x] Show all items
- [x] Add support for a custom filter
- [x] Improve the selection box to have same width as text field by default

Best effort / Last Minute (if there's time) : 
- Multiple selection -- Dropped feature
- [x] Able to limit suggestions shown at a time with a scroll bar

@stsrki 
Need your opinion in regards to Free Typing + Validation. This is what I did, do you agree?

- Added a new parameter `SelectedText` just like a user had already tried. In my opinion is better and less confusing then having `SelectedValue` handling both cases.

- Flag **`AllowFreeTyping`** (I named it this, but can be something else... argh naming is hard)
  - If this is set, then, `Validator `validates the `SelectedText `instead of the `SelectedValue`
  - If this is not set. The AutoComplete will always clear the `SelectedText` if no valid value from the data source is selected, pretty much how AutoComplete already worked.

Also introduced a few features to work on, which we'd already mentioned across issues and I think are very welcome to the component, just let me know if you don't agree with any of them.


Demos/Blazorise.Demo/Pages/Tests/ComponentsPage.razor
```diff
@@ -45,13 +45,21 @@
                                       ValueField="@((item)=>item.MyTextField)"
                                       SelectedValue="@selectedSearchValue"
                                       SelectedValueChanged="@MySearchHandler"
-                                      Placeholder="Search..." Filter="AutocompleteFilter.StartsWith" />
+                                      @bind-SelectedText="selectedAutoCompleteText"
+                                      Placeholder="Search..."
+                                      Filter="AutocompleteFilter.StartsWith"
+                                      AllowFreeTyping="true">
```
stsrki, line 51: Can be named `FreeTyping`.

stsrki:
Loking good so far.

> Show all items

is already working, just need to set `MinLength` to `0`. But to make it better we need to set the scrollbar on the dropdown menu. Look at the support web for reference. I have implemented it on the issues filter.

---

I like `SelectedText`. I think it will cover most scenarios quite well. We'll see how it goes with the users.

stsrki:
PS. see if you can fix the formatting. For some reason, spacings are again broken.

David-Moreira:
For the show all items I was thinking about having an arrow just like the select has to show all items...

However I was thinking about it. I'm not sure it makes sense.

stsrki:
No, don't add an arrow. It doesn't make sense for the search field. If the search field would be inside of a dropdown menu that it would be OK, like

image

But that is a feature for some other time.

David-Moreira:
Yep that screenshot was what I was going for. But with min length I don't it it meshes well. Feature has to be better planned, for another time, yes.

Demos/Blazorise.Demo/Pages/Tests/ComponentsPage.razor
David-Moreira: Done.

Source/Extensions/Blazorise.Components/Autocomplete.razor.cs
```diff
@@ -65,20 +60,21 @@ protected async Task OnTextChangedHandler( string text )
             SelectedText = CurrentSearch;
             dirtyFilter = true;
 
-            if ( text?.Length >= MinLength && FilteredData.Any() )
```
stsrki, line 68: When I now set `MinLength="0"` it doesn't work as before. It must show a dropdown when a search is focused.

stsrki:
Once you finish all features, please update the documentation.

Source/Extensions/Blazorise.Components/Autocomplete.razor.cs
David-Moreira: My bad... I had removed this:
image

docs/_posts/2021-03-24-release-notes-094.md
```diff
@@ -125,4 +125,22 @@ Now it is possible to globally change input and button sizes. This will greatly
...
 ### Dropping eFrolic support
 
-Since eFrolic author has stopped maintaining eFrolic, I have decided to stop supporting it and move my focus elsewhere. This will allow me to make more work on new providers that are actively worked on.
\ No newline at end of file
+Since eFrolic author has stopped maintaining eFrolic, I have decided to stop supporting it and move my focus elsewhere. This will allow me to make more work on new providers that are actively worked on.
+
+### AutoComplete improvements
+
+-  `FreeTyping` binding support && `FreeTyping` validation support
+   - You are now able to use `AutoComplete` as a suggestions source to the user. Meaning you can enable this feature, and `AutoComplete` will accept any value introduced by the user and bind it on the `SelectedText` parameter.
```
stsrki, line 133: AutoComplete should be Autocomplete

Source/Extensions/Blazorise.Components/Autocomplete.razor.cs
stsrki: Now it works only the first time you focus on search. If you select a value, then clear, lose focus and enter focus it will not show the menu again.

David-Moreira: Fixed.

stsrki:
Hmm, is it OK that the menu is shown when we click on _clear_ button? 🤔

David-Moreira:
If the min len is 0, yes it seems right to me. 
The clear button clears the search text... However we are still focusing the text and "searching", so I don't see a reason why the min len 0 should not provoque the same behaviour if cleared. Should it not be like that?

stsrki:
Kinda breaks the flow. Let say a user just wants to clear some fields on a form. He clicks on autocomplete clear button and suddenly a menu is opened. He would need to click outside of the menu to close. Maybe we can handle it in a later version if people start complaining.

David-Moreira:
Oh, but is the clear button visible without focusing the auto complete? 
In that case I agree with you yes. It probably shouldn't even focus the field, if it does... 

However, If the user is already focused on the field then in my opinion it should maintain the behaviour both on clear button or escape, do you agree? 

Don't merge yet. Let me see if I can take a look at that case later today.

stsrki:
It is a native search input. It might be problematic to detect how text ia cleared.

David-Moreira:
Right. I'll just take a look. If it's unnecessarily complex. We can just leave it and listen to feedback.

David-Moreira:
Hello @stsrki 
So I didn't know input type search had an inbuilt close button haha, I tought it was Blazorise's. 
So the behaviour is that it clears the field and focus on it. So even if it shows items or not, it would always focus on it (so i'm not sure it would "break the flow" all the same). So might as well honor the configured behaviour, of `MinLen` is equal to 0 + focus, so it displays the items, like expected.

We could change it so when escape or clear is pressed it does not show the items when `MinLen` is equal to 0, I just tested this successfully, but I personally do not fully agree with that approach, if the developer set it to be opened at the `MinLen` of 0, then it should. If he wants otherwise, might as well configure it otherwise. Just let me know if this is okay and you may go ahead and merge. Otherwise, just let me know if you'd think the behaviour should instead be like mentioned. Sometimes this kind of things are kind of subjective and there isn't a real right answer in my opinion.

PR #2557, opened 2021-07-01
Datagrid set to single selection + existing MultiSelectColumn with filterable set to false fix

### LINKED ISSUE #2553: DataGridMultiSelectColumn & Filtering Misbehavior
I had my multi select column defined this way (you know, for good measure add everything to it):

`<DataGridMultiSelectColumn TItem="MyClass" Width="30px" Editable="false" Sortable="false" Filterable="false" />`

The column was followed by a simple column ("Name") that has sorting and filtering enabled, however, I got the following result:

image

The filter textboxes were shifted right by one column... once I removed all the extraneous props

`<DataGridMultiSelectColumn TItem="FileInfoShortened" Width="30px" />`

everything was back in place, ie the Name filter textbox in the Name column where it belonged.

David-Moreira:
Closes #2553


David-Moreira:
Ooops, I sent this to dev093.

PR #2565, opened 2021-07-02
DataGrid : Handle SelectionMode Changed

### LINKED ISSUE #2549: DataGrid Selection Behavior on Switching Single / Multiple
I made a little recording, for repro (using https://bootstrapdemo.blazorise.com/tests/datagrid):

* In single mode, select an item
* Switch to multiple; it seems that the selection carried over.
* Click on any other item - you now see that the selected item wasn't actually selected because now only the newly selected item is so.

The bound data is ok, but the UI doesn't seem to catch on.

Animation

David-Moreira:
Fixes #2549

### DIFF
--- a/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
+++ b/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
@@ -211,19 +211,24 @@ namespace Blazorise.DataGrid
             if ( selectionMode == DataGridSelectionMode.Multiple && SelectedRow != null )
             {
                 SelectedRows ??= new();
+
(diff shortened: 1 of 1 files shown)


David-Moreira:
Hey @stsrki 
I don't think this was picked up for dev094

David-Moreira:
@stsrki ? :P

stsrki:
Cherry picked? Might be I forgot or I failed to understand you, thinking it was already fixed in 0.9.4.

stsrki:
Can you include it in your Virtualize branch?

David-Moreira:
Done.

PR #2566, opened 2021-07-03
show snapshot testing approach

SimonCropp:
thoughts?


David-Moreira:
Hello @SimonCropp 
I have seen your presentation on "BlazorTrain" about Verify. It's pretty cool indeed, congrats. We'll take a look at your suggestion. Thanks! :D

stsrki:
Hi, @SimonCropp Thank you for taking the time to implement it in Blazorise. I hate to say it, but can I ask you to recreate PR on the **dev094** branch. We generally don't accept any new work on the `master` branch since it can lead to problems once we merge all dev work for the final release. And considering the master and dev branches are already very different it would be hard for us to maintaining it even as it is.

Once you finish we'll gladly test it. Thank you!

SimonCropp:
will do

stsrki:
I have tried tests and they're working. But there is a problem. I tried to first format HTML, eg. `ButtonComponentTest.RenderTest.01.verified.html` and then run the test. Tests fail because the strings don't match while I was expecting it to compare the HTML structure. Is this the expected behavior?

PR #2573, opened 2021-07-04
DataGrid EmptyFilterTemplate

### LINKED ISSUE #2550: Add <EmptyFilterTemplate> to datagrid
The template is used when the current filtering excludes all rows in the model. Currently, EmptyTemplate is only used when the model has no data; it isn't used when all of it is filtered out

stsrki:
resolves #2550 Add `EmptyFilterTemplate` to datagrid


Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -947,17 +947,26 @@ protected IEnumerable<DataGridColumn<TItem>> DisplayableColumns
         /// <summary>
         /// Returns true if EmptyTemplate is set and Data is null or empty.
         /// </summary>
-        protected bool IsEmptyTemplateVisible => !IsLoadingTemplateVisible && !IsNewItemInGrid && EmptyTemplate != null && ( Data == null || !Data.Any() );
+        protected bool IsEmptyTemplateVisible
+            => !IsLoadingTemplateVisible && !IsNewItemInGrid && EmptyTemplate != null && ( Data == null || !Data.Any() );
```
David-Moreira, line 951: Might be worth creating an internal extension for collections, EmptyOrNull and/or AnyNotNull

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -947,17 +947,26 @@ protected IEnumerable<DataGridColumn<TItem>> DisplayableColumns
...
         /// </summary>
-        protected bool IsEmptyTemplateVisible => !IsLoadingTemplateVisible && !IsNewItemInGrid && EmptyTemplate != null && ( Data == null || !Data.Any() );
+        protected bool IsEmptyTemplateVisible
+            => !IsLoadingTemplateVisible && !IsNewItemInGrid && EmptyTemplate != null && ( Data == null || !Data.Any() );
+
+        /// <summary>
+        /// Returns true if EmptyFilterTemplate is set and FilteredData is null or empty.
+        /// </summary>
+        protected bool IsEmptyFilterTemplateVisible
+            => !IsLoadingTemplateVisible && !IsNewItemInGrid && EmptyFilterTemplate != null && ( !( Data == null || !Data.Any() ) && ( FilteredData == null || !FilteredData.Any() ) );
```
David-Moreira, line 957: I don't have time to go test this now. But does this work ok with ReadData?
Can I not have 
**Data empty or null** + FilteredData Empty or null 
and it being a valid case for Empty Filter?

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -947,17 +947,26 @@ protected IEnumerable<DataGridColumn<TItem>> DisplayableColumns
...
 
         /// <summary>
         /// Returns true if ShowPager is true and grid is not empty or loading.
         /// </summary>
-        protected bool IsPagerVisible => ShowPager && !IsLoadingTemplateVisible && ( ( IsButtonRowVisible && ButtonRowTemplate != null ) || !IsEmptyTemplateVisible );
+        protected bool IsPagerVisible
+            => ShowPager && !IsLoadingTemplateVisible && ( ( IsButtonRowVisible && ButtonRowTemplate != null ) || !IsEmptyTemplateVisible );
```
David-Moreira, line 963: I have also been doing this approach with property lambdas, looks and reads way better!

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
stsrki: That is good idea.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
stsrki: It works with ReadData also. I specifically separated them because otherwise they could both be shown at the same time.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
stsrki: I agree, you can notice right away that is a lambda property.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
stsrki: Added IsNullOrEmpty extension

David-Moreira: Ya looks better.

PR #2581, opened 2021-07-06
Remove some redundant state machines

SimonCropp:


stsrki:
This is good PR :)

David-Moreira:
Just of note. I believe that sometimes you take the hit of explicitly setting the state machine just for improved stack trace on exceptions. Otherwise it's redundant, yes.

PR #2585, opened 2021-07-06
DataGrid UpdateEditingCell API

### LINKED ISSUE #2586: DataGrid UpdateEditingCell API
Ability to update any other cell after the current cell is changed.

stsrki:
resolves #2586 DataGrid UpdateEditingCell API

Cell can be changed in two ways:

1. through the edit context: `context.UpdateCell( nameof( Employee.FirstName ), "some value" )`
2. or through datagrid reference: `dataGridRef.UpdateCellEditValue( nameof( Employee.FirstName ), "some value" )`

### DIFF
--- a/Demos/Blazorise.Demo/Pages/Tests/DataGridPage.razor
+++ b/Demos/Blazorise.Demo/Pages/Tests/DataGridPage.razor
@@ -319,7 +319,7 @@
                                 }
                             </DisplayTemplate>
                             <EditTemplate>
-                                <Select TValue="string" SelectedValue="@((string)( context.CellValue ))" SelectedValueChanged="@(( v ) => { context.CellValue = v; context.UpdateCell( nameof( Employee.FirstName ), TitleFromGender( v ) + context.ReadCell( nameof( Employee.FirstName ) ) ); })">
+                                <Select TValue="string" SelectedValue="@((string)( context.CellValue ))" SelectedValueChanged="@(( v ) => { context.CellValue = v; context.UpdateCell( nameof( Employee.FirstName ), TitleToName( TitleFromGender( v ), context.ReadCell( nameof( Employee.FirstName ) ) ) ); })">
                                     <SelectItem TValue="string" Value="null"></SelectItem>
                                     <SelectItem TValue="string" Value="@("M")">Male</SelectItem>
                                     <SelectItem TValue="string" Value="@("F")">Female</SelectItem>
--- a/Demos/Blazorise.Demo/Pages/Tests/DataGridPage.razor
+++ b/Demos/Blazorise.Demo/Pages/Tests/DataGridPage.razor
@@ -319,7 +319,7 @@
                                 }
                             </DisplayTemplate>
                             <EditTemplate>
-                                <Select TValue="string" SelectedValue="@((string)( context.CellValue ))" SelectedValueChanged="@(( v ) => { context.CellValue = v; context.UpdateCell( nameof( Employee.FirstName ), "Mr" +(v=="F"?"s":"") + context.ReadCell(nameof(Employee.FirstName)) ); })">
+                                <Select TValue="string" SelectedValue="@((string)( context.CellValue ))" SelectedValueChanged="@(( v ) => { context.CellValue = v; context.UpdateCell( nameof( Employee.FirstName ), TitleFromGender( v ) + context.ReadCell( nameof( Employee.FirstName ) ) ); })">
                                     <SelectItem TValue="string" Value="null"></SelectItem>
                                     <SelectItem TValue="string" Value="@("M")">Male</SelectItem>
                                     <SelectItem TValue="string" Value="@("F")">Female</SelectItem>
--- a/Demos/Blazorise.Demo/Pages/Tests/DataGridPage.razor
+++ b/Demos/Blazorise.Demo/Pages/Tests/DataGridPage.razor
@@ -319,7 +319,7 @@
                                 }
                             </DisplayTemplate>
                             <EditTemplate>
-                                <Select TValue="string" SelectedValue="@((string)( context.CellValue ))" SelectedValueChanged="@(( v ) => { context.CellValue = v; context.UpdateCell( nameof( Employee.FirstName ), "Mr" +(v=="F"?"s":"") + context.Item.FirstName ); })">
+                                <Select TValue="string" SelectedValue="@((string)( context.CellValue ))" SelectedValueChanged="@(( v ) => { context.CellValue = v; context.UpdateCell( nameof( Employee.FirstName ), "Mr" +(v=="F"?"s":"") + context.ReadCell(nameof(Employee.FirstName)) ); })">
                                     <SelectItem TValue="string" Value="null"></SelectItem>
                                     <SelectItem TValue="string" Value="@("M")">Male</SelectItem>
                                     <SelectItem TValue="string" Value="@("F")">Female</SelectItem>
--- a/Demos/Blazorise.Demo/Pages/Tests/DataGridPage.razor
+++ b/Demos/Blazorise.Demo/Pages/Tests/DataGridPage.razor
@@ -319,7 +319,7 @@
                                 }
                             </DisplayTemplate>
                             <EditTemplate>
-                                <Select TValue="string" SelectedValue="@((string)( context.CellValue ))" SelectedValueChanged="@(( v ) => context.CellValue = v)">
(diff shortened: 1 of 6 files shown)


Source/Extensions/Blazorise.DataGrid/CellEditContext.cs
```diff
@@ -22,9 +23,20 @@ public class CellEditContext
     /// <typeparam name="TItem"></typeparam>
     public class CellEditContext<TItem> : CellEditContext
     {
-        public CellEditContext( TItem item )
+        /// <summary>
+        /// Method that will be called when cell is manually updated.
+        /// </summary>
+        private readonly Action<string, object> CellUpdated;
```
David-Moreira: Do you do private fields with capital?

stsrki: Only when they're methods, events or delegates

Demos/Blazorise.Demo/Pages/Tests/DataGridPage.razor
```diff
@@ -319,7 +319,7 @@
                                 }
                             </DisplayTemplate>
                             <EditTemplate>
-                                <Select TValue="string" SelectedValue="@((string)( context.CellValue ))" SelectedValueChanged="@(( v ) => context.CellValue = v)">
+                                <Select TValue="string" SelectedValue="@((string)( context.CellValue ))" SelectedValueChanged="@(( v ) => { context.CellValue = v; context.UpdateCell( nameof( Employee.FirstName ), "Mr" +(v=="F"?"s":"") + context.Item.FirstName ); })">
```
David-Moreira: Haven't tested. But isn't it missing a space between Mr/Mrs and the name?

stsrki: It's just for internal purposes. I will revert it before merge.

docs/_docs/extensions/datagrid.md
```diff
@@ -109,6 +109,26 @@ The grid can work in two different editing modes that can provide different user
...
+- or by using the `DataGrid` instance
+
+In the following example we're simply calling `context.UpdateCell` with a fieldname to change and a new value that we want it to assign: 
+
+```html
+<DataGridColumn TItem="Employee" Field="@nameof( Employee.Salary )" Caption="Salary" Editable="true">
+    <EditTemplate>
+        <NumericEdit TValue="decimal" Value="@((decimal)context.CellValue)" ValueChanged="@( v => {
+                context.CellValue = v;
+                context.UpdateCell( "Tax", v * .25m );
```
David-Moreira, line 126: Even in docs, show good practice examples if possible. `nameof` for Tax?

stsrki: Makes sense

Demos/Blazorise.Demo/Pages/Tests/DataGridPage.razor
David-Moreira: Could leave it... It's always good to have examples of the various features.

stsrki: In that case I need to make it better :)

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -494,7 +494,7 @@ public void FilterData()
         /// </summary>
         /// <param name="fieldName">Cell field name.</param>
         /// <param name="value">New cell value.</param>
-        private void UpdateCell( string fieldName, object value )
+        public void UpdateCell( string fieldName, object value )
```
David-Moreira: If this is public? Can I miss use this? This should only be used when the Grid is in EditMode right? If that is the case, maybe do nothing if grid is not in EditMode.

stsrki: It's public. Refresh page :)

David-Moreira: Well but if it's public. Can't I miss use it? Like mentioned? Or could I have a way to select the row(item) to update?

stsrki: Sorry, I was too quick to write before reading your message. I have fixed and it cannot be updated or read if not in edit state,

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -494,7 +494,7 @@ public void FilterData()
         /// </summary>
         /// <param name="fieldName">Cell field name.</param>
         /// <param name="value">New cell value.</param>
-        public void UpdateCell( string fieldName, object value )
+        public void UpdateEditCell( string fieldName, object value )
```
David-Moreira, line 497: To be more precise could be, `UpdateEditCellValue`, as it's always the value of the cell, not other cell properties. It is no big deal, I think it's still quite clear to the developer.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
David-Moreira: Is it worth having an overload for actually updating / reading a row cell's value? Any benefit to this?
We might not want to directly update/mutate the row cell values though. It won't go trough the edit pipeline... But I guess that could be up to the developer... what do you think?

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
stsrki: Hmm tough choice...

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
David-Moreira: Actually nvm, this just seems dumb.. the only way to know the row is through the item, right? If dev has the item, he can just read / mutate it directly.

stsrki: Yep. The dev can already mutate it even though it's not the right way.

Demos/Blazorise.Demo/Pages/Tests/DataGridPage.razor
```diff
@@ -319,7 +319,7 @@
                                 }
                             </DisplayTemplate>
                             <EditTemplate>
-                                <Select TValue="string" SelectedValue="@((string)( context.CellValue ))" SelectedValueChanged="@(( v ) => { context.CellValue = v; context.UpdateCell( nameof( Employee.FirstName ), "Mr" +(v=="F"?"s":"") + context.ReadCell(nameof(Employee.FirstName)) ); })">
+                                <Select TValue="string" SelectedValue="@((string)( context.CellValue ))" SelectedValueChanged="@(( v ) => { context.CellValue = v; context.UpdateCell( nameof( Employee.FirstName ), TitleFromGender( v ) + context.ReadCell( nameof( Employee.FirstName ) ) ); })">
```
David-Moreira, line 322: Is it still missing the space? haha? Meaning it will do something like: 
`Mr.Jonh` instead of `Mr. John`?

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
David-Moreira: yep, haha naming is hard! :D

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -496,6 +496,9 @@ public void FilterData()
         /// <param name="value">New cell value.</param>
         public void UpdateEditCell( string fieldName, object value )
         {
+            if ( editState == DataGridEditState.None )
```
David-Moreira, line 499: What's the behaviour if the State is New for example? Will both Update and Read work alright?

David-Moreira:
Looks good other then the tiny things. Now I understood what you meant with this feature! :D

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
stsrki: It's the same. In both states, the same dictionary is used.

Demos/Blazorise.Demo/Pages/Tests/DataGridPage.razor
stsrki: Done

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
stsrki: I will leave it as it is. No need to make it too long :)

stsrki:
> Looks good other then the tiny things. Now I understood what you meant with this feature! :D

Now users can do most of the things they needed so far. And this feature was requested many times before. It was easier that I thought :)

PR #2590, opened 2021-07-07
Remove HttpClient from demos

SimonCropp:
i dont think any of the demos do http requests?

### DIFF
--- a/Demos/Blazorise.Demo.AntDesign/Program.cs
+++ b/Demos/Blazorise.Demo.AntDesign/Program.cs
@@ -1,12 +1,9 @@
 ﻿#region Using directives
-using System;
-using System.Net.Http;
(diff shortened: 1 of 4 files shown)


stsrki:
I would leave just this. The original idea was to have some mockup rest API that would be used from the demo app as an example for users and for internal testing.

SimonCropp:
in that case the code is flawed, since it should be using the IHttpClientFactory  abstraction

stsrki:
Yep. It's an old code when Blazor was young and people still haven't figured that IHttpClientFactory is that way to go.

SimonCropp:
so given the current code is flawed, and trivial to re-add the correct implementation, my recommendation is to remove for now

stsrki:
I agree. Removing it!

PR #2592, opened 2021-07-07
Use some null-coalescing assignment operators

SimonCropp:


stsrki:
You're on a roll! 😃 And I thought the code was clean enough. Thanks for your help, I really appreciate it!

David-Moreira:
Could also consider using the new() feature then. That infers the type to create :P

SimonCropp:
@David-Moreira it is on my queue

SimonCropp:
@David-Moreira https://github.com/Megabit/Blazorise/pull/2603

David-Moreira:
Ah, haha 👍

David-Moreira:
Just wanted to add. That I like these type of features alot, specially the one's that make it more consice and easier to read.
I do think we should keep up with language improvements, whenever it makes sense.

However it can sometimes have the negative effect of making the barrier of entry to open source projects a tiny bit higher for people who have yet to catch up with the newer features and may be initially puzzled with the syntax.

Nothing that a simple google search can't help with though.

stsrki:
@David-Moreira well seeing something new will force them to learn. I know I was doing the same in the past. And it still happens from time to time when I see something new :D

David-Moreira:
Exactly, but to us, it just something more to add to already existing knowledge. I'm also learning something everyday, thank god!!

But for newer people, there are like ALOT of concepts to grasp all at the same time and these features just add on to that. However like I said, i'm all aboard to integrate them! 👍

PR #2593, opened 2021-07-07
Use catch when

SimonCropp:
to avoid the catch and re-throw when !task.IsCanceled

so we only swallow  when ( task.IsCanceled )

### DIFF
--- a/Source/Blazorise/Components/Bar/Bar.razor.cs
+++ b/Source/Blazorise/Components/Bar/Bar.razor.cs
@@ -127,12 +127,8 @@ namespace Blazorise
                     {
                         await task;
                     }
-                    catch
+                    catch when ( task.IsCanceled )
                     {
-                        if ( !task.IsCanceled )
-                        {
-                            throw;
-                        }
                     }
 
                     DisposeDotNetObjectRef( dotNetObjectRef );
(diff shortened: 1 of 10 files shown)


David-Moreira:
It is of note that this is "syntactic sugar", right? It should still be lowered and compiled to IL as a simple if.

SimonCropp:
my understanding it is not "just syntactic sugar" https://ivanderevianko.com/2015/05/c-6-0-exception-filters-try-catch-when

SimonCropp:
and another one http://www.volatileread.com/Wiki/Index?id=1087

stsrki:
I remember we had a valid reason why we re-throw an error but I can't find the original issue now.  It was something about tasks being timeout after one minute.

@David-Moreira do you remember it maybe?

David-Moreira:
Nice. I'll read those posts later. I skimmed through the first one I saw it has the compiled IL, great! 

Strski, 
I think it had to do with blazor throwing task cancelled to us on the disposal process. And we were letting it bubble up. I'll see if I can find the thread also.

Simons solution should be equivalent to what we had. It might be even better if it produces improved IL i thought it would produce the same behind the covers, apparently maybe not.

stsrki:
Based on those two articles seems to me also that this feature is a nice optimization. From what I can understand, if an error happens it will re-throw for all situations except for when the Task is canceled. Basically what we already do, but better.

David-Moreira:
Yea, pretty good feature, not just "syntactic sugar".  
And apparently it doesn't "re-throw", it's not even caught if it doesn't meet the "filter"(condition) to begin with.
Awesome! 👍

PR #2599, opened 2021-07-07
redundant verbatim prefix

SimonCropp:


Source/Blazorise/Components/Progress/PageProgress.razor.cs
```diff
@@ -120,13 +120,13 @@ public bool Visible
         [Parameter]
         public int? Value
         {
-            get => @value;
```
stsrki, line 123: This was needed before. I guess `value` is not a reserved keyword anymore.

PR #2600, opened 2021-07-07
use top level methods

SimonCropp:


stsrki:
I prefer it the _old_ way. Not a fan of new top-level methods.

SimonCropp:
fair enough :)

PR #2601, opened 2021-07-07
Simplify csproj files

SimonCropp:


Source/Directory.Build.props
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
```
stsrki, line 1: Can this be part of `Build\Blazorise.props`? Or is `Directory.Build.props` some special rule for VS to pick up when building projects?

Source/Extensions/Blazorise.Animate/Blazorise.Animate.csproj
```diff
@@ -3,20 +3,10 @@
   <Import Project="..\..\..\Build\Blazorise.props" />
 
   <PropertyGroup>
-    <PackageId>Blazorise.Animate</PackageId>
```
stsrki, line 6: What happens with PackageId? Is it going to be automatically generated based on project name?

Source/Directory.Build.props
```diff
@@ -0,0 +1,11 @@
...
+<Project>
+  <PropertyGroup>
+    <PackageLicenseFile>LICENSE.md</PackageLicenseFile>
+    <PackageIcon>Blazorise.png</PackageIcon>
+    <PackageProjectUrl>https://blazorise.com/</PackageProjectUrl>
+    <RepositoryUrl>https://github.com/Megabit/Blazorise</RepositoryUrl>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <NoWarn>1591</NoWarn>
```
stsrki, line 9: NoWarn is now enabled for all projects. I still want them for Blazorise code projects and other projects that are directly used.

Source/Directory.Build.props
SimonCropp: VS walks up the directory tree and looks for Directory.Build.props

Source/Extensions/Blazorise.Animate/Blazorise.Animate.csproj
SimonCropp: yep it defaults to the project name

stsrki:
I used existing `Blazorise.props` for shared properties instead of `Directory.Build.props`. This feature was not working in the previous version of VS. It's good they finally fixed it.

I also tested on `dotnet pack` and it seems there are no issues with new settings.

PR #2603, opened 2021-07-07
Use target typed new

SimonCropp:


stsrki:
It's hard to read what is happening in some cases with new keyword, for example:

```cs
themeCache = new( new( null, options => { } ) );
```

or

```cs
edit.Date = new( new DateTime( 2020, 4, 12 ), new( 0, 0, 0 ) );
```

Multiple new's and I have no idea what is the type that is initialized unless I go and investigate.

---

For situations like the following where I can see the actual type and the `new()` I think it is OK

```cs
private static Dictionary<IconName, string> names = new()
```

stsrki:
I've noticed some issues with formating on existing files so I've fixed it.

@David-Moreira I think even with some less-readable situations this feature is OK. Agree?

David-Moreira:
Yep in general yes. The type of cases you mentioned could potentially be the exception but I don't think they are that bad. Altough you can't explicitly see the type, you still see the arguments / parameters explicitly which should give enough context on most cases?

PR #2613, opened 2021-07-09
DataGrid Sort API

### LINKED ISSUE #2612: DataGrid Sort API
I've looked through the source code but don't see any way to retrieve the current sort column/direction for a datagrid.  Can someone point me in the right direction?

What'd I'm looking to do is fairly straightforward, display a list of items in a DataGrid that's pageable, sortable, etc. Then the user clicks an item and is navigated to a separate page to view the item details.  Then the user clicks a "Save" or "Back" button and is returned back to the same list, which has maintained the sort and page they were on.  Simple right?

I planned on embedding/storing the sort and page information in the querystring of the url that links to the item detail page.  Then in the item detail page, I'd embed that same information in the return url used to back to the list page, which would use those variables to initialize the datagrid to the appropriate state.

I see there are events which I can hook into for PageChanged, and PageSizeChanged to capture the page information, but there's no event for "SortChanged"....am I missing something?

stsrki:
resolves #2612 Capture DataGrid sort column/direction and page state

- [x] Add `Sort()` overload with the fieldName
- [x] Add `SortChanged` event
- [x] Optional direction for `Sort()` method

### DIFF
--- a/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
+++ b/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
@@ -375,13 +375,34 @@ namespace Blazorise.DataGrid
         ///
         /// Note that <see cref="DataGridColumn{TItem}.Sortable"/> must be enabled to be able to sort!
         /// </summary>
-        /// <param name="column">Column to sort.</param>
+        /// <param name="fieldName">Field name of the column to sort.</param>
+        /// <param name="sortDirection">Sort direction of the specified column, or if null it will be handled automatically.</param>
         /// <returns>A task that represents the asynchronous operation.</returns>
-        public Task Sort( DataGridColumn<TItem> column )
+        public Task Sort( string fieldName, SortDirection? sortDirection = null )
+        {
+            var column = Columns.FirstOrDefault( x => x.Field == fieldName );
+
+            if ( column != null )
+            {
+                return Sort( column, sortDirection );
+            }
+
+            return Task.CompletedTask;
+        }
+
+        /// <summary>
+        /// Sorts the Data for the specified column.
+        ///
+        /// Note that <see cref="DataGridColumn{TItem}.Sortable"/> must be enabled to be able to sort!
+        /// </summary>
+        /// <param name="column">Column to sort.</param>
+        /// <param name="sortDirection">Sort direction of the specified column, or if null it will be handled automatically.</param>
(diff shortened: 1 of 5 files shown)


Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -718,7 +739,7 @@ internal async Task<bool> IsSafeToProceed( EventCallback<CancellableRowChange<TI
 
         #region Filtering
 
-        protected void HandleSortColumn( DataGridColumn<TItem> column, bool changeSortDirection )
+        protected void HandleSortColumn( DataGridColumn<TItem> column, bool changeSortDirection, SortDirection? sortDirection = null )
```
David-Moreira, line 742: There's an async execution that's been added. Missing Task here?

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -735,14 +756,17 @@ protected void HandleSortColumn( DataGridColumn<TItem> column, bool changeSortDi
...
 
                 if ( !SortByColumns.Any( c => c.Field == column.Field ) )
                 {
                     SortByColumns.Add( column );
                 }
                 else if ( column.CurrentSortDirection == SortDirection.None )
                     SortByColumns.Remove( column );
+
+                if ( changeSortDirection )
+                    InvokeAsync( () => SortChanged.InvokeAsync( new DataGridSortChangedEventArgs( column.Field, column.CurrentSortDirection ) ) );
```
David-Moreira, line 769: I kinda forgot... Is the EventCallback safe to call without null check / delegate exists check?

David-Moreira:
Other then the questions on the review, looks good to me.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
stsrki: SortChanged is of type EventCallback (struct) soi it can never be null. It is safe.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
stsrki: Async exec is inside of `InvokeAsync` so it should be good. But I can make it better.

Issue #2500, opened 2021-06-19
DetailRowTrigger on DataGrid is called twice on each row click


GilShalit:
Using the following to close a details row if the row is clicked a second time
```
<DataGrid TItem="CourtCase"
		  @bind-SelectedRow="@selectedCourtCase"
		  DetailRowTrigger="@((item)=> item.Id == selectedCourtCase?.Id && !isRowAlreadySelected())">
...
</DataGrid>
```

with
```
bool isRowAlreadySelected()
        {
            bool isRowClicked = prevSelectdRow!=null && prevSelectdRow.Id == selectedCourtCase.Id;
            prevSelectdRow = selectedCourtCase;
            return isRowClicked;
        }
```

Does not work.  Using debug I found that isRowAlreadySelected() is called twice on each row click, so obviously the prevSelectedRow variable is already set to the row which is newly selected on the second call, and so the result is effectively always true.

I found a way around this with an additional flag, but I think this isn't how it should work.

stsrki:
I would say this is good behavior. DetailRowTrigger can be potentially called for every row in a table. That is why it has `item` as an argument. It's not just for the currently selected row.

GilShalit:
Right.  But for some reason it seems to be called twice with the same item and selectedCourtCase because the first part of the trigger condition is met twice so that isRowAlreadySelected() is visited twice for each row click.

StephenOTT:
Having similar issues: This seems to fit into the Virtualization work for the Datagrid?

David-Moreira:
The detail row should close with Ctrl+click, have you guys tried that? Or does that not satisfy your use case?

GilShalit:
Yes it works with Ctrl+Click, but this does not seem as intuitive as simply have the second click close an opened row.  How do I know it is not intuitive? EVERY new user asks me why isn't the second click closing the opened row...

stsrki:
@David-Moreira do you think we should change the toggle with a click instead of ctrl+click?

David-Moreira:
Well, @GilShalit does have a point. I would say most users, would just expect to just click back again and it would collapse the `DetailRowTrigger` .
However the reason why Ctrl+Click works, is because it will unselect the row, therefore the `DetailRowTrigger Func` will not evaluate to true anymore.

I think it's feasible.
[...]

PR #2625, opened 2021-07-11
Prototype for handling DetailRow on click, saving row state, ...

### LINKED ISSUE #2532: Datagrid : ToggleDetailRow : Public Api Feature
@David-Moreira I will merge this now.

Don't make it a priority, but in case you got any time to work on the public ToggleDetailRow method do it.

David-Moreira:
@stsrki can you take a look at these suggestions?
**Disclaimer :** Commit still not clean / not final code, since I've done some prototypying here.

I was initially going to do this with some flags and some tricks to keep track of the state... But I figured we're starting to scatter too much logic because of missing state...
So... Introduced new RowInfo Internal API as a suggestion.
I really think we are missing this, to keep state around of each Row, and it is limiting the stuff that we can do and expose to our users.
Tell me what you think? This should also be the foundation for the requested multi row edit handling, as you can finally keep track of the state of multiple rows.

Also Figured `dirtyView` was always true due to `Data` Parameter always calling set..., so everything was always recalculated, on every interaction, I did a tiny dirty fix on it, but probably should think of a different approach.

Also the DetailRow, will now behave slightly differently. Since it's triggered on rowclick, which should be more efficient, instead of always being re-evaluated, you will now be able to have several row details open, which in my opinion would be the actual correct behaviour for `DetailRowTrigger` but it is indeed different from the current behaviour, which only keeps a single detail open. (The fix is however easy, as we just need to check if the item is selected or not, to go back to the previous behaviour with this new implementation)

Also there's still a bug with changing the PageSize, as it looses track of the row info.

Closes #2532 // #2500

### DIFF
--- a/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
+++ b/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
@@ -54,7 +54,7 @@ namespace Blazorise.DataGrid
         /// <summary>
         /// Marks the grid to refresh currently visible page.
         /// </summary>
-        private bool dirtyView = true;
+        private bool dirtyView = false;
 
         /// <summary>
         /// Keeps track of whether the object has already been disposed.
@@ -215,6 +215,7 @@ namespace Blazorise.DataGrid
             {
                 await InvokeAsync( StateHasChanged );
             }
+            dirtyView = true;
         }
 
         private async void OnPageChanged( int currentPage )
@@ -232,6 +233,7 @@ namespace Blazorise.DataGrid
             {
                 await InvokeAsync( StateHasChanged );
             }
+            dirtyView = true;
         }
 
         #endregion
@@ -525,6 +527,21 @@ namespace Blazorise.DataGrid
             }
 
             return null;
+        }
+
+        /// <summary>
+        /// Toggles DetailRow while evaluating the <see cref="DetailRowTrigger"/>   
+        /// </summary>
+        /// <returns>A task that represents the asynchronous operation.</returns>
+        public Task ToggleDetailRow( TItem item )
+        {
+            if ( DetailRowTrigger != null )
+            {
+                var rowInfo = displayRowData.FirstOrDefault( x => x.Item.IsEqual( item ) );
+                if ( rowInfo != null )
+                    rowInfo.SetShowDetail( DetailRowTrigger.Invoke( item ) );
+            }
+            return Task.CompletedTask;
         }
 
         #endregion
@@ -1112,10 +1129,13 @@ namespace Blazorise.DataGrid
             get { return data; }
             set
             {
-                data = value;
+                //This was just straight up recalculating every grid interaction...
+                //Need to figure out a better way to track reloading filter/view
+                if ( value == null || ( !data?.SequenceEqual( value ) ?? true ) )
+                    // make sure everything is recalculated
+                    dirtyFilter = dirtyView = true;
 
-                // make sure everything is recalculated
-                dirtyFilter = dirtyView = true;
+                data = value;
             }
         }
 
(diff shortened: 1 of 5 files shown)


Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -1112,10 +1129,13 @@ public IEnumerable<TItem> Data
             get { return data; }
             set
             {
-                data = value;
+                //This was just straight up recalculating every grid interaction...
+                //Need to figure out a better way to track reloading filter/view
+                if ( value == null || ( !data?.SequenceEqual( value ) ?? true ) )
```
stsrki, line 1134: I didn't know about `SequenceEqual` method. Is it the same thing as we already have, a `ArrayExtensions.AreEqual`?

Source/Extensions/Blazorise.DataGrid/Models/DataGridRowInfo.cs
```diff
@@ -0,0 +1,29 @@
...
+        /// </summary>
+        /// <param name="item">Row Item</param>
+        public DataGridRowInfo( TItem item )
+        {
+            Item = item;
+        }
+
+        public TItem Item { get; }
+
+        public bool ShowDetail => showDetail;
```
stsrki, line 23: Rename to DetailRowVisible

Source/Extensions/Blazorise.DataGrid/Models/DataGridRowInfo.cs
```diff
@@ -0,0 +1,29 @@
...
+
+        public void SetShowDetail(bool toShowDetail)
```
stsrki, line 25: Rename to SetRowDetail or SetRowDetailVisibility. Also might be good to have ToggleRowDetail

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
David-Moreira: Oh I didn't even notice you had that extension, haha.
But yes it is equivalent, makes sure that for both sequences every item is compared by using the default comparer. Maybe you could actually replace it. I would think the native implementation might have slightly better performance.
Something we could compare by using, if needed:
https://benchmarkdotnet.org/articles/overview.html

stsrki:
Seems good so far. We must be careful with the DataGridRowInfo list. We're basically creating a new collection every time something changes on the grid. This is different from current, where we used only the data that the user gave us. So it can have an impact on performance if the list is too large.

David-Moreira:
> Seems good so far. We must be careful with the DataGridRowInfo list. We're basically creating a new collection every time something changes on the grid. This is different from current, where we used only the data that the user gave us. So it can have an impact on performance if the list is too large.

I agree. But we can't run from creating a reference type for each item, if we want to start tracking state of each row, enabling more flexibility for the features we can implement.
Maybe we can make it more optimized, though.

Also, at the moment the rows are tracking the display data, which means there's only a DataGridRowInfo list equivalent to the PageSize, if the user has a PageSize of thousands, he most likelly already will have a different performance problem, with rendering all that html...

Only real problem, could be Virtualize with static data, which can actually have a huge DisplayData at a time.

David-Moreira:
Maybe using this pattern for `_DataGridRow` component, since we already have to create these references.
```
OnInitialzied(){
ParentDataGrid.AddDatagridRow(this);
}
```
And making the `_DataGridRow` our state tracker? However I don't think it's as generic as having the `DataGridRowInfo` as there can be other type of rows like `_BaseDataGridRowEdit`

What would you think of trying this approach?

stsrki:
I agree. First, we can make it work as expected and later we can optimize. I like the general idea. It is on the same track that I already had in mind before.

David-Moreira:
hmm with the few days left for 0.9.4 release, and all the changes already on the Virtualize branch and ongoing internal refactoring, better leave this one for next release. More time for us to also decide/discuss/prototype the internal row state api.

David-Moreira:
@stsrki 
Might re-pick this one soon, probably going to just do a new branch, too much conflicts by now.
Any comments on something you'd like done differently or want to discuss? Otherwise I will keep this approach.

stsrki:
Yeah, better to start with a new branch. Generally, I think the approach is good so you can proceed with it.

PR #2627, opened 2021-07-11
Benchmark project | ArrayExtensions AreEqual refactor

David-Moreira:
Added Benchmark project with 
https://benchmarkdotnet.org/articles/guides/getting-started.html
so we can easily do some simple benchmarking.

Since we walked about `SequenceEquals` vs the Blazorise `AreEqual`, I compared them.
**Results (Mean is what matters most as it's the metric you can look at for performance, in nanoseconds here):**
image

After doing some unit tests for `AreEqual`, I noticed `SequenceEquals` does not handle nulls.
Refactored `AreEqual` to handle the nulls and use `SequenceEquals` internally for the comparision behind the scenes.
**Results (Mean is what matters most as it's the metric you can look at for performance, in nanoseconds here):**
image

**Legend:**
image

### DIFF
--- a/Blazorise.sln
+++ b/Blazorise.sln
@@ -84,6 +84,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Blazorise.SpinKit", "Source
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Blazorise.Markdown", "Source\Extensions\Blazorise.Markdown\Blazorise.Markdown.csproj", "{1D30D96F-71FC-49A7-AF88-5749CBAA8153}"
 EndProject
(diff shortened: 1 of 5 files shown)


stsrki:
Tested new `SequenceEquals` with validation, ran all unit tests, and E2E tests. Everything seems good. And it is a lot faster than before. Good job!

I must admit this is the first time I'm seeing a benchmark like this. Usually, I just did it old school with stopwatches 😂. This is much better.

stsrki:
You know what. This will come in handy once we start optimizing our fluent builder(ColumnSize, Margin, Flex, etc.). Once again, good job!

David-Moreira:
Yea it's pretty good. And it also can analyse the allocated memory too.

PR #2631, opened 2021-07-13
Change Tooltip content dynamically

### LINKED ISSUE #2629: Tooltip text is not updated on @bind-Text
I'm trying to dynamically update tooltip's text from a variable, but can't get it work - the text stays the same as it was on the first render.

I'm using the following code:
`<Tooltip Fade="true" 
               @bind-Text="@_tooltipText"
               Placement="TooltipPlacement.Right"
               Multiline="true"
               AlwaysActive="false">
           <Icon Name="IconName.QuestionCircle" Style="color:#969392" />
</Tooltip>`

Would be good if tooltip text would be bindable. Should we add changing event to the component like below to get it work?
`[Parameter] public EventCallback<string> TextChanged { get; set; }`

### Additional context
Environment:
VS 16.11.0 preview 2.0
.Net 6 preview 5
Blazorise 0.9.4-preview5

stsrki:
resolves #2629 Tooltip text is not updated on @bind-Text

### DIFF
--- a/Source/Blazorise/Components/Tooltip/Tooltip.razor.cs
+++ b/Source/Blazorise/Components/Tooltip/Tooltip.razor.cs
@@ -1,4 +1,5 @@
 ﻿#region Using directives
+using System.Threading.Tasks;
 using Blazorise.Utilities;
 using Microsoft.AspNetCore.Components;
 #endregion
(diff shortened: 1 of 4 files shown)


JanVargovsky:
@stsrki I just migrated from 0.9.3.5 to 0.9.5.4 and I can't get it working. Simply changing value binded to the `Text` doesn't do anything. I tried to dig into the internals and call the JS myself but it doesn't work either.

nnnhub:
It did work in 9.4.9, but seems to be broke in 9.5.4.

stsrki:
Can you open new ticket and I will look into it when I have more time?

PR #2644, opened 2021-07-16
Add Ticks Callback functionality

StefH:
## Make it possible to define JavaScript code which will be executed for Tick -> Callback

Result is that you can format the ticks on yAxis / xAxis values yourself:

image

## Usage
``` c#
new Axis
{
    //
    Ticks = new AxisTicks
    {
        CallbackJavaScript = "`${value / 1000} K`"
    }
    //
}
```


stsrki:
It really works 🎉. Good job!

Can you document the usage in the docs? https://github.com/Megabit/Blazorise/blob/dev094/docs/_docs/extensions/chart.md

---

So this is basically an implementation of https://www.chartjs.org/docs/2.8.0/axes/labelling.html#creating-custom-tick-formats, right? Do you think we can have the same thing for other callbacks? For example https://www.chartjs.org/docs/2.8.0/configuration/tooltip.html#label-callback

Looks very promising :)

StefH:
@stsrki 

Some questions/remarks
- did you change your formatting style to the default style since when?
- is the name callbackJavaScript a good name? Or should we have something else?
- I did not yet look into other callbacks yet. Maybe start with this one and implement the others in a different PR
- Can I change the docs in this same PR?

stsrki:
> @stsrki
> 
> Some questions/remarks
> 
> * did you change your formatting style to the default style since when?
> * is the name callbackJavaScript a good name? Or should we have something else?
> * I did not yet look into other callbacks yet. Maybe start with this one and implement the others in a different PR
> * Can I change the docs in this same PR?

- formating is always the same
- maybe just Callback?
- yes other callbacks can be done in another PR
- yes

stsrki:
I like the simplicity of the PR. We need to see how it would work for more complex callbacks.

Later I can make a list of other callbacks and then try and implement them.

StefH:
Callback is maybe not a good name...

I also thought about:
callbackStringFormatter or callbackValueFormatter ?

David-Moreira:
In my personal opinion, I think it's easier for a developer to understand that the callback being set by the parameter is going to be evaluated by javascript, if it says `callbackJavaScript`.

Altough I do like something like `callbackStringFormatter` best, I'm not sure it's gonna be so clear to as many developers. But they can also just straight up read the parameter documentation... :D

stsrki:
We can follow the same principle as we already have for options: `Options` > `OptionsString`. So have is as `CallbackString`. As far as I understand it is a function body that is executed by the JS. We will document it and mention that it is executed by the Java Script.

---

I have more ideas for the future coming to my mind. One idea is to have a C# Expression that would be parsed and serialized for JavaScript. For example in this case we could have C# representation of ChartJS function

Callback from https://www.chartjs.org/docs/2.8.0/axes/labelling.html#creating-custom-tick-formats

```js
callback: function(value, index, values)
```
would be

```cs
Expression<Func<TValue, int, object[], object>> Callback { get; set; }
```

or

```cs
Func<TValue, int, object[], object> Callback { get; set; }
```

and usage

```cs
Callback = ( value, index, values ) =>
{
    return $"{value / 1000} K";
};
```

We would have a special serializer that would _somehow_ get the body part `$"{value / 1000} K"`, and that body would be sent to JS for processing. Just as it is now.

StefH:
I did also think about converting a C# function to JavaScript.
But that seemed 'a bridge to far'.

Other thing I thought about was calling a c# method as Blazor interop via the .net reference from the JavaScript.

stsrki:
Calling the C# method as Blazor interop would not work. Believe me, I tried. Problem is that ChartJs expects sync functions while JSInterop needs async. So it just gets the promise as a result.

David-Moreira:
I think the accepted and common way is to call the csharp methods through the net object reference.

Also theres a library that probably does something similar to what you guys are trying to do, 
from this fellow
https://github.com/mariusmuntean/ChartJs.Blazor
Might be worth taking a look if he tackled this problem and how.

stsrki:
They made it work for WebAssembly but not for Server projects. I think it was something with **sync** JS Interop calls.

stsrki:
@StefH You can leave it as `CallbackJavaScript`. I think it is a good name. Just add usage examples in the docs and I'll merge. Hopefully just in time before the final release :)

StefH:
What is the deadline?

stsrki:
Tomorrow :)

StefH:
I will try tonight. If I can't make it. Just merge this PR. And docs will be in next release.

Btw I was investigating that Expression approach this morning. And I think I have a possible proof of concept to convert an Expression into a JavaScript string. But this needs of course an custom JSON serializer. I will make POC PR for next release.

stsrki:
Cool. Leave it for the 0.9.5. The change will be big and timely.

If you don't happen to make it by tomorrow I will try adding to the docs myself and merge the PR. Ate least it is working as expected and is good starting point for later :)

PR #2650, opened 2021-07-17
Table scroll to row feature 

### LINKED ISSUE #2646: Add Table ScrollToRow method
https://stackoverflow.com/a/7853216/833106

_Originally posted by @stsrki in https://github.com/Megabit/Blazorise/pull/1977#discussion_r671692660_

David-Moreira:
Closes #2646


Source/Blazorise/Components/Table/Table.razor.cs
```diff
@@ -140,15 +140,16 @@ private async ValueTask RecalculateResize()
...
         /// <summary>
-        /// If table has <see cref="FixedHeader"/> enabled, it will scroll position to the provided pixels.
+        /// If table has <see cref="FixedHeader"/> enabled, it will scroll to the provided value.
         /// </summary>
-        /// <param name="pixels">Offset in pixels from the top of the table.</param>
+        /// <param name="scrollTo"><see cref="TableScrollToType"/></param>
+        /// <param name="value">Value to Scroll to according to <see cref="TableScrollToType"/></param>
         /// <returns>A task that represents the asynchronous operation.</returns>
-        public ValueTask FixedHeaderScrollTableTo( int pixels )
+        public ValueTask FixedHeaderScrollTableTo( TableScrollToType scrollTo, int value )
```
stsrki, line 148: Name it ScrollTo?

Does it make sense to have an enum for operation type? Maybe just have two diferent methods?

stsrki: ScrollToRow and ScrollToOffset?

David-Moreira: I was kinda divided between both ways, haha.
My though was: If new ways come up, it's easy to add a new one, instead of making a new method for each one.

But I can change it, yes, Would you rather have the two methods, then?

David-Moreira: Refactored into two methods.

stsrki:
Looking good now! This will come in handy :)

David-Moreira (commented):
Is this coherent with the pr you accepted from Simon removing the state machine from all async calls?
Behind the covers it should still. Allocate a Task and not a ValueTask. Same result.

Issue #2287, opened 2021-05-01
missing dataset labels on PieChart


naefp:
When using a pie chart with multiple datasets, there can be set a label for each dataset which never will be displayed. Since this seems to be default behavior in Chart.Js it is not a bug in Blazorise!

But definitely it would make sense to add an option for displaying the dataset labels additionally in the tooltips. A possible solution which would perfectly fit in my use case, is to update the tooltip label callback: `return data.datasets[item.datasetIndex].label+ ": "+ data.labels[item.index]+ ": "+ data.datasets[item.datasetIndex].data[item.index];` (got this from here: https://stackoverflow.com/a/45184272)

Discussion on Chart.js: https://github.com/chartjs/Chart.js/issues/1417

CPlusPlus17:
*bump*

stsrki:
Sorry guys, lately I've been very busy with some other business-related work to make it done all as soon as possible so I tend to overlook some issues.

---

The fix seems doesn't seem that complicated. Can't promise when it will be done. I will schedule it for 0.9.4, and if anyone in the meantime could create a PR it would be released soon as part of the next preview.

stsrki:
Working on this now. Should I make modifications only for Pie chart?

naefp:
Not sure, I did not use every chart type yet. So far, I only missed the dataset label on pie chart...

stsrki:
Not sure if I'm missing something? But for me, the labels are working just fine for Pie chart

![image](https://user-images.githubusercontent.com/900302/123932980-3efaab80-d992-11eb-8bd5-133b1995ccf3.png)

naefp:
@stsrki that's correct, the labels of each entry are displayed correctly, but not the label of the dataset.

Each circle in the pie chart represents a dataset and the tooltip should be able to show the dataset label as well. Example here: https://stackoverflow.com/a/45184272

To be able to do that we need to change the tooltip callback for the label, and right now this is not possible with the Blazorise implementation. I can set the Label property in PieChartDataset, but this is never displayed and so you are not able to understand what each circle of the pie chart represents.

In my opinion the tooltip could always show that like: `<dataset.label>: <entry.label>: value`
But there may be use cases where the dataset label is not expected to be shown, I don't know. Maybe you could enable/disable that with a simple bool or you could provide a way to customize the tooltip callback directly? (which may be possible already, but I just didn't find it :D )

stsrki:
I was able to do it the same as for all other charts

[...]

PR #2653, opened 2021-07-18
show dataset label in tooltip title on pie/doughnut charts

naefp:
#2287


Source/Extensions/Blazorise.Charts/wwwroot/blazorise.charts.js
```diff
@@ -21,11 +21,10 @@ window.blazoriseCharts = {
                     tooltips: {
                         callbacks: {
                             title: function (item, data) {
-                                const label = data.labels[item[0].index];
-                                return label;
+                                return data.datasets[item[0].datasetIndex].label;
```
naefp, line 24: in the title section we need to use `item[0]` to get the datasetIndex

stsrki:
Looking good. Thanks!

PR #2659, opened 2021-07-19
Datagrid: BugFix : When data is null on init, set internal viewdata to empty list

### LINKED ISSUE #2658: DataGrid bug with 0.9.4
I just upgraded from 0.9.3 preview 5 to 0.9.4, and get a nasty null reference exception on load:

 ```
   <app>
        <component type="typeof(App)" render-mode="ServerPrerendered" />
    </app>
```

I'm trying to figure it out by stripping back the razor and code but no luck yet other than it probably is something to do with the datagrid virtualize stuff. 

Even a plain grid like:

```
<DataGrid  TItem="Animal"  >
</DataGrid>
```

causes it, so not sure where to look next...

Just letting you know and I'll keep looking tomorrow

### PR DESCRIPTION
**Pull request by David-Moreira**
Closes #2658

Tested the grid's null start with main variants.
```
<DataGrid TItem="Animal"></DataGrid>
<DataGrid TItem="Animal" Virtualize="true"></DataGrid>
<DataGrid TItem="Animal" ReadData="@((x) => null)"></DataGrid>
<DataGrid TItem="Animal" Virtualize="true" ReadData="@((x) => null)"></DataGrid>
```

### DIFF
--- a/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
+++ b/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
@@ -71,7 +71,7 @@ namespace Blazorise.DataGrid
         /// <summary>
         /// Marks the grid to refresh currently visible page.
         /// </summary>
-        private bool dirtyView;
+        private bool dirtyView = true;
 
         /// <summary>
         /// Keeps track whether the user has changed the filter for Virtualize purposes.
--- a/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
+++ b/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
@@ -1375,7 +1375,7 @@ namespace Blazorise.DataGrid
 
                 dirtyView = false;
 
-                return viewData;
+                return viewData ?? Enumerable.Empty<TItem>();
             }
         }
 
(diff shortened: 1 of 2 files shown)


stsrki:
It's strange that `dirtyView = true` would fix it. I'm trying to reproduce the error but even if I set `Data="@null"` and revert your changes it still works for me. No error whatsoever. Were you able to reproduce it?

David-Moreira:
I was, yes. I reproduced with the code in the first post on the dashboard page. It throws an object reference exception when trying to do the foreach for the the rows, due to DisplayData/viewdata not being set to an empty list through filtereddata.

David-Moreira:
You can't reproduce it? Need me to set up a repo?

stsrki:
I was able to reproduce it with the same code. But how likely is it that DataGrid will be used without any column or other params?

In any case, I have added another layer of safety for null value.

David-Moreira:
Sure. I originally was just going to do a null coalesce fix too, but I took a look at 0.9.3 and just maintained the same behaviour. Eitherway extra check shouldn't hurt.

robalexclark:
The initial example I gave with no Data= was an extreme one as the simplest example I could find - all my DataGrids do in fact have Data=@somedata and columns! But the null  issue still exists. Anyhow thanks for taking a look and Iets see if the patch works. If it doesn't I will create a project to reproduce.

PR #2680, opened 2021-07-20
Datagrid: Revert "Dirty only when data changed"

### LINKED ISSUE #2675: DataGrid fails to update when items are added/deleted
**Describe the bug**
When an item is added/deleted via code triggered by a button, it does not update in the DataGrid.
Note: you can manually trigger a DataGrid to refresh by sorting any column

**To Reproduce**
1. Clone this test repo: [https://github.com/mtbayley/BlazoriseJSExceptionTest](https://github.com/mtbayley/BlazoriseJSExceptionTest)
2. Run, go to /fetchdata
3. Click add item
4. Default table will add item. Blazorise DataGrid below does not update.
5. Change to version 0.9.3.10
6. Do same steps. DataGrid will add/ delete items.

image

David-Moreira:
Unfortunately 
image
Causes a bug... since setter value and data will pretty much always be equal. We would need to keep a copy of the old data and compare... I'm not sure it would be worth it at all... we could still take a look in the future.

Fixes #2675

### DIFF
--- a/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
+++ b/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
@@ -1314,11 +1314,7 @@ namespace Blazorise.DataGrid
             get { return data; }
             set
             {
-                if ( !value.AreEqual( data ) )
-                {
-                    SetDirty();
-                }
-
+                SetDirty();
                 data = value;
             }
         }


stsrki:
Could it be that the new `SequenceEqual` in `AreEqual` is not working properly? Can you try using the old version of `AreEqual` just to make sure?

David-Moreira:
I did debug. Both data and value are the same before this line. 
`data = value;`

But I can try the old version, no problem.

David-Moreira:
Went on branch 0.9.3 and reproduced:
image
Evaluates to true. Reference is the same.

David-Moreira:
image

stsrki:
OK makes sense. Too bad.

I will merge the change but we should definitely try to find a way to not refresh if not needed.

David-Moreira:
Yea, 
We need to benchmark if the cost of making sure to keep an "old" data in memory is worth it over refreshing every time.

stsrki:
Maybe we can just save a hash of the array?

David-Moreira:
Right, maybe an hash is enough.

PR #2682, opened 2021-07-20
DataGrid : Internally have datagrid's child component's notify of a statechanged

### LINKED ISSUE #2676: DataGrid selection mode
### Discussed in https://github.com/Megabit/Blazorise/discussions/2667

<sup>Originally posted by **AlexaBuff** July 20, 2021</sup>
After update 0.9.4 dataGrid's selection mode became multiple automatically. I can not turn on single selection mode. I noticed that multiple selection throw off after double click. My code:

```cshtml
 <DataGrid TItem="Employee" @ref="dataGrid"
              Data="@employeeList"
              Editable="true"
              CommandMode="DataGridCommandMode.ButtonRow"
              ShowPageSizes="true"
              EditMode="DataGridEditMode.Popup"
              Bordered="true"
              Hoverable="true"
              Responsive="true"
              ShowPager="true"
              RowInserted="@CreateEmployee"
              PopupSize="ModalSize.Large"
              RowDoubleClicked="@ShowModal"
              UseInternalEditing="true"
              PagerPosition="DataGridPagerPosition.Top"
              Resizable="true"           
              FixedHeader="true"
              FixedHeaderDataGridHeight="80%"
              ResizeMode="TableResizeMode.Columns"
              SelectionMode="DataGridSelectionMode.Single" >
```</div>

### PR DESCRIPTION
**Pull request by David-Moreira**
Fixes #2676

Since Datagrid on 0.9.4 does not refresh every single time due to alot of internal parameter passage between components being removed. 

There should be a way for a child to notify the parent that there have been changes that may affect other children. 
(Like, in the case of this issue, recalculating the style of other rows, when a row is clicked/selected)

### DIFF
--- a/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
+++ b/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
@@ -831,11 +831,12 @@ namespace Blazorise.DataGrid
         }
 
         /// <summary>
-        /// Notifies the <see cref="DataGrid{TItem}"/> that state has changed.
+        /// Tells the <see cref="DataGrid{TItem}"/> to refresh.
+        /// <para>Internally calls StateHasChanged.</para>
         /// </summary>
         /// <returns></returns>
-        protected internal new void  StateHasChanged()
-            =>  base.StateHasChanged();
+        public async Task Refresh()
+            => await InvokeAsync( StateHasChanged );
 
         protected async Task HandleReadData( CancellationToken cancellationToken )
         {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs [hunk 1]
```diff
@@ -830,6 +830,13 @@ public async Task Reload( CancellationToken cancellationToken = default )
             }
         }
 
+        /// <summary>
+        /// Notifies the <see cref="DataGrid{TItem}"/> that state has changed.
+        /// </summary>
+        /// <returns></returns>
+        protected internal new void  StateHasChanged()
```
stsrki, line 837: Instead of overriding StateHasChanged and making it accessible from outside I would have a `Task Refresh()` that would call `InvokeAsync( StateHasChanged );`

David-Moreira: Good idea. Will change.

stsrki:
Looking good now!

PR #2692, opened 2021-07-21
Show a real name of the icon inside tooltip

Luk164:
Just added a tooltip that shows a real name of the icon if hovered over.


stsrki:
Looking good. Thank you!

PR #2707, opened 2021-07-22
chartjs: Add support for Formattable String Expression

StefH:


StefH:
@stsrki

I've started a proof of concept solution for using an C# expression for the callback function.

A callback can now be defined like:
``` c#
Callback = ( value, index, values ) => $"{value / 1000} K"
```

See the example code, and also review this PR to check if the unit tests are clear enough.

Source/Blazorise/Utilities/ExpressionConverter.cs
```diff
@@ -0,0 +1,160 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq.Expressions;
+
+namespace Blazorise.Utilities
+{
+    /// <summary>
+    /// Helper methods to convert a <see cref="LambdaExpression"/> to valid JavaScript.
+    /// </summary>
+    public static class ExpressionConverter
```
stsrki, line 10: I think it's better to have it on the `Blazorise.Charts` project as it is only going to be used there. In case we have a reason we can later move to the core project.

Source/Extensions/Blazorise.Charts/ChartOptions.cs
```diff
@@ -330,6 +332,22 @@ public class AxisTicks
         /// </summary>
         [DataMember( EmitDefaultValue = false )]
         public string CallbackJavaScript { get; set; }
```
stsrki, line 334: Do we need to have this now with the Callback? Can the Callback be serialized automatically with `Converters.ToDictionary`?

stsrki:
Already looking great. It reminds of SQL function in [linq2db](https://github.com/linq2db/linq2db) in a way that they are executed on the server. While in our case they are executed by JavaScript.

As long as we document it well and people understand what it represents and how it works I think it will a great feature.

Source/Blazorise/Utilities/ExpressionConverter.cs
StefH: If this ExpressionConverter is moved to the `Blazorise.Charts` project, this means that it cannot be used in the `Converters.ToDictionary`. So in that case this converter should stay in the common location. Or do I miss something?

Source/Extensions/Blazorise.Charts/ChartOptions.cs
StefH: 1. See my other comment.

2. And maybe it's good to keep the CallbackJavaScript, to give the user total freedom to write JavaScript code.

Source/Blazorise/Utilities/ExpressionConverter.cs
stsrki: You're right, I didn't think it that good 😅

OK it can stay in the core project.

Source/Extensions/Blazorise.Charts/ChartOptions.cs
stsrki: I don't think Callback will prevent them from writing JavaScript code. As far as I can see there is no difference between what `Callback` and `CallbackJavaScript` return. It's just that with `Callback` you also have access to the params of the function, which is good.

StefH: OK. In that case, I can remove the CallbackJavaScript and call the converter in the Converters.ToDictionary.

(However this means that you cannot use the callback when using **OptionsAsJson**.)

StefH:
Hello @stsrki,

I did remove the `callbackJavaScript` property and updated the Converter to handle the `Callback` LambdaExpression.

StefH:
@stsrki 
Can you review this PR again and check if there are still things to do ?

stsrki:
@StefH I have tested your work so far. It works and looks great. Did you plan to work more on this or I can already merge it?

StefH:
For now I think it finished, so you can merge it.

stsrki (approved):

stsrki:
Thank you @StefH

PR #2723, opened 2021-07-24
Added intervals configuration for NotificationAlert component.

maisak:
NotificationAlert component lacks an ability to set custom interval for a toast. So, added a parameter to the component and a property to NotificationOptions (to allow individual settings for notifications). Also updated documentation.


stsrki:
Looking good, thanks! I have realized IntervalBeforeClose should be better null so I made some changes.

Issue #2524, opened 2021-06-26
Minify support + bundling 


David-Moreira:
Related #1910

Implement minifying support, meaning there should be resources both available in the normal version and in the minified version. *.min.css | *.min.js

Separate js from one big file to a more modular approach and just bundle them all back into blazorise.js / blazorise.min.js in the end.

David-Moreira:
@stsrki agree with the feature?
Is it okay that I grab this one eventually? I think we could do it for 0.9.4.

stsrki:
I have assigned it for you. Do it only after we clear all other issues in 0.9.4 and if we have enough time left.

PR #2728, opened 2021-07-25
WIP: Bundling & compilation of resources files into the .net's compilation process

David-Moreira:
Hello,
Here's the suggestion we've talked about before.

Moved Bundling & compilation of resources files into the .net's compilation process by adding two new assemblies.
image
Introduced bundling concept and splitted the big js file into minor component like js files, so it's easier to develop.
I still haven't splitted all the "components", it's an example/suggestion:
image
**In the end, it will still all be bundled into blazorise.js**
image
image

Ideally we would also provide our users with the minified version of the blazorise js. But for some reason, it's not working, I think there are some invalid javascript which are not supported by the minification process. So for now we've turned it off.


David-Moreira:
**PS:** And apparently it broke the build, haha.
Gotta check if it's an easy fix.

Source/Blazorise/Blazorise.csproj
```diff
@@ -23,4 +24,44 @@
...
+  <ItemGroup>
+    <None Remove="Scripts\blazorise.dependencies.js" />
+    <None Remove="Scripts\blazorise.functions.js" />
+    <None Remove="Scripts\blazorise.main.js" />
+    <None Remove="Scripts\blazorise.table.js" />
+    <None Remove="Scripts\blazorise.tooltip.js" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Content Include="Scripts\blazorise.table.js">
```
stsrki, line 40: Instead of naming every file you can just use a wildcard.

Source/Blazorise/bundleconfig.json
```diff
@@ -0,0 +1,15 @@
+[
+  {
+    "outputFileName": "wwwroot/blazorise.js",
+    "inputFiles": [
+      "Scripts/blazorise.dependencies.js",
```
stsrki, line 5: No need for `blazorise.` prefix for source files

Source/Blazorise/Blazorise.csproj
```diff
@@ -23,4 +24,44 @@
...
+      <CopyToPublishDirectory>PreserveNewest</CopyToPublishDirectory>
+    </Content>
+    <Content Include="Scripts\blazorise.dependencies.js">
+      <ExcludeFromSingleFile>true</ExcludeFromSingleFile>
+      <CopyToPublishDirectory>PreserveNewest</CopyToPublishDirectory>
+    </Content>
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="BuildBundlerMinifier" Version="3.2.449" />
```
stsrki, line 63: What happens with nuget packages? Are these assemblies included or excluded on dotnet pack? You can check by running dotenet-pack.cmd in NuGet folder

Source/Blazorise/Blazorise.csproj
David-Moreira: You're right. It's just automatic from vs ide.

Source/Blazorise/bundleconfig.json
David-Moreira: Think this config also supports wildcard. So something like Scripts/*.js

Source/Blazorise/Blazorise.csproj
David-Moreira: Oh that's a good question, you're worried it's smthg that goes embbeded on the final Product, is that it?

David-Moreira: Hmm I don' see anything wrong with the packages. I will have to setup a local nuget server later to test... But i'm pretty sure, that be default is the standard assembly behaviour. It will list those new references as Blazorise dependencies.

stsrki: You only need to run the script and open .nuget file in any zip viewer to see the result.

David-Moreira: I did that. I don't see anything wrong with it. 
But it did pick up the scripts... that should be setting Copy to Ouput to false.
image

image

stsrki: Yeah, scripts should not be there

David-Moreira:
Hello,
Take a look at web compiler. If you like it, we set it up on the remaining projects.

Demos/Blazorise.Demo.Bootstrap.RC/Pages/_Host.cshtml
```diff
@@ -18,7 +18,7 @@
     <link rel="stylesheet" href="https://use.fontawesome.com/releases/v5.12.0/css/all.css">
     <link rel="stylesheet" href="https://unpkg.com/easymde/dist/easymde.min.css">
 
-    <link href="_content/Blazorise/blazorise.css?v=0.9.4.1" rel="stylesheet" />
+    <link href="_content/Blazorise/blazorise.min.css.gz?v=0.9.4.1" rel="stylesheet" />
```
stsrki, line 21: Don't use minified files in demo apps. It becomes harder to debug any JS code.

David-Moreira: You're right. It was kinda curiosity getting the best of me. I was experimenting with gzip compressed resources 🤣🤣 I hadn't used them before.

David-Moreira: I'll fix and finish the remaining projects later today then.

stsrki: No problem :)

David-Moreira:
@stsrki any idea why these paths do not work well in CI?
CICD

stsrki:
> not

Maybe this

image

David-Moreira:
I think i figured it... Computers are fun....
replace ..\..\ with ../../

stsrki:
I get these errors in VS

image

stsrki:
🤯🤔🤔 I submitted the previous message, haven't touch anything in VS and now it builds without any error. Strange...

David-Moreira:
Closes ##2524

David-Moreira:
Maybe we should still minfy the js files if we aren't.
It's stil useful for automatic scss on build. and also css minify

David-Moreira:
Should we close this and #2524, since it doesn't work 100%? We're doing fine with the `webcompiler` extension for the time being.

stsrki:
Yep, closing it now.

PR #2736, opened 2021-07-27
Dynamically calculate HorizontalBarHeight if theme settings are undefined

### LINKED ISSUE #2695: <LayoutHeader Fixed="true"> causes overlap of LayoutContent
It seems like in 0.9.4.1 LayoutHeader Fixed="true" causes overlap of LayoutContent

```
<Layout>
    <LayoutHeader Fixed="true">
        <MainMenu></MainMenu>
    </LayoutHeader>
        <LayoutContent>
            @Body
        </LayoutContent>
    <LayoutFooter Fixed="true">
        <MainFooter></MainFooter>
    </LayoutFooter>
</Layout>
```

stsrki:
fixes #2695

### DIFF
--- a/Source/Blazorise/Components/Bar/Bar.razor.cs
+++ b/Source/Blazorise/Components/Bar/Bar.razor.cs
@@ -72,6 +72,18 @@ namespace Blazorise
                     await Toggle();
                 }
             }
+            else
+            {
+                if ( string.IsNullOrEmpty( Theme?.BarOptions?.HorizontalHeight ) )
+                {
+                    var rect = await JSRunner.GetElementInfo( ElementRef, ElementId );
+
+                    if ( rect.BoundingClientRect.Height > 0 )
+                    {
+                        await JSRunner.AddThemeVariable( ThemeVariables.HorizontalBarHeight, $"{rect.BoundingClientRect.Height}px" );
+                    }
+                }
+            }
 
             await base.OnFirstAfterRenderAsync();
         }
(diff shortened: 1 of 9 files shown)


Source/Blazorise/Components/Bar/Bar.razor.cs
```diff
@@ -72,6 +72,18 @@ protected override async Task OnFirstAfterRenderAsync()
...
             }
+            else
+            {
+                if ( string.IsNullOrEmpty( Theme?.BarOptions?.HorizontalHeight ) )
+                {
+                    var rect = await JSRunner.GetElementInfo( ElementRef, ElementId );
+
+                    if ( rect.BoundingClientRect.Height > 0 )
+                    {
+                        await JSRunner.AddThemeVariable( ThemeVariables.HorizontalBarHeight, $"{rect.BoundingClientRect.Height}px" );
```
David-Moreira, line 83: Enough to be on firstRender? Does it ever need any recalc, if for some reason the users are able to change?

David-Moreira (commented):
Other then the single comment. Looks good to me.

Source/Blazorise/Components/Bar/Bar.razor.cs
stsrki: I was thinking about it. But then again how many times do you think topbar height will be changed. I think very rarely. In case it does get reported I can just move it to recalculate in OnAfterRender.

PR #2783, opened 2021-08-05
Remove instruction to include DataGrid CSS

juan-db:
The DataGrid CSS file was removed in commit a0363dcbb7c6cf61a9417c6e9a01ed4255df68a5. Including it in version 0.9.3.10 or higher will result in a 404 response.


stsrki:
Since documentation is hosted from the `rel-0.9.4` branch I have recreated your PR on that branch. Rebasing from master into release branch is not possible.

In any case, thank you for the fix.

PR #2793, opened 2021-08-09
DataGrid initializer fix

### LINKED ISSUE #2785: datagrid begin edit
if entity has an complex property with no default constructor, this exception will thrown

`Default constructor not found for type [complex type]`
stack trace is 
```

at System.RuntimeType.CreateInstanceMono(Boolean nonPublic, Boolean wrapExceptions)
   at System.RuntimeType.CreateInstanceSlow(Boolean publicOnly, Boolean wrapExceptions, Boolean skipCheckThis, Boolean fillCache)
   at System.RuntimeType.CreateInstanceDefaultCtor(Boolean publicOnly, Boolean skipCheckThis, Boolean fillCache, Boolean wrapExceptions)
   at System.Activator.CreateInstance(Type type, Boolean nonPublic, Boolean wrapExceptions)
   at System.Activator.CreateInstance(Type type, Boolean nonPublic)
   at System.Activator.CreateInstance(Type type)
   at Blazorise.DataGrid.Utils.RecursiveObjectActivator.CreateInstanceRecursive(Object currObjInstance, PropertyInfo[] properties)
   at Blazorise.DataGrid.Utils.RecursiveObjectActivator.CreateInstanceRecursive(Object currObjInstance, PropertyInfo[] properties)
   at Blazorise.DataGrid.Utils.RecursiveObjectActivator.CreateInstance[Tbl_AST_EstateOwnership]()
   at Blazorise.DataGrid.DataGrid`1[[AppModel.Web.Model.Tbl_AST_EstateOwnership, Sayeh.AppModel, Version=1.0.0.2, Culture=neutral, PublicKeyToken=null]].InitEditItem(Tbl_AST_EstateOwnership item)
```
blazorise version : 9.4.2

stsrki:
fixes #2785 

In this PR I have used `FormatterServices.GetUninitializedObject` instead of `Activator.CreateInstance`. The reason is that FormatterServices creates an empty object without the need for any parameterless constructor. And an empty object is what we want to have.

Source: https://stackoverflow.com/questions/390578/creating-instance-of-type-without-default-constructor-in-c-sharp-using-reflectio

Note: Test data in the demo should be removed once we make it work!

@David-Moreira Now the problem is in `FunctionCompiler.CreateValidationExpressionGetter`. When an access field is null it cannot generate member access. Can you also have a look?

### DIFF
--- a/Demos/Blazorise.Demo/Pages/Tests/DataGridPage.razor
+++ b/Demos/Blazorise.Demo/Pages/Tests/DataGridPage.razor
@@ -226,7 +226,8 @@
                           PageChanged="(e) => currentPage = e.Page"
                           FilteredDataChanged="@OnFilteredDataChanged"
                           UseValidation="true"
-                          SortChanged="@OnSortChanged">
+                          SortChanged="@OnSortChanged"
+                          NewItemCreator="@OnEmployeeNewItemCreator">
                     <EmptyTemplate>
                         <Span>
                             <Icon Name="IconName.InfoCircle" TextColor="TextColor.Info" Margin="Margin.Is2.FromRight" />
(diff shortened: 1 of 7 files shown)


stsrki:
I think I know what is the problem.

Currently, we have this method to generate an accessor for data-annotation validation.

```cs
public static Expression<Func<TValue>> CreateValidationExpressionGetter<TItem, TValue>( TItem item, string fieldName )
{
    var parameter = Expression.Parameter( typeof( TItem ), "item" );
    var property = GetPropertyOrField( parameter, fieldName );

    var convertExpression = Expression.MakeMemberAccess( Expression.Constant( item ), property.Member );

    return Expression.Lambda<Func<TValue>>( convertExpression );
}
```

The problem is in our test demo we have a nested field `Test.Name` inside of `Employee` class, while `CreateValidationExpressionGetter` and `Expression.MakeMemberAccess` are expecting the field property of the `TItem`, or in this of `Employee` class. `Test.Name` is a nested field and obviously, it will not work. 

Looking at this now: https://stackoverflow.com/questions/7663501/dataannotations-recursively-validating-an-entire-object-graph/8090614

David-Moreira:
Like we talked the `MakeMemberAccess` wasn't accessing the correct instance. Initially we thought that the expression was missing the item traversal, and that was only partially correct. Actually It was always accessing the **item** instance. So that's why nested objects weren't working.

Changing the code to create a getter for the correct instance fixed this enabling the data validation on nested objects. 
image

image

I left your testing class, so you can go ahead and do your testing before merging.

PR #2817, opened 2021-08-17
Update start.md - Fixed typo

koenbeuk:
I assumed you meant 3.1 here, not 3.2 which is not a thing. Otherwise this should be 5.0 :)


stsrki:
Thank you!

PR #2819, opened 2021-08-17
Removes DataGridColumn from ParentDataGrid if disposed.

### LINKED ISSUE #2809: Blazorise DataGrid dynamic column is not removed on datagrid reload
I have created a datagrid using bazorise. The issue I am facing is that when the user adds query to the text box and press the Start query button the data is populated in the gridview. However if the user press the same button again to get the updated data from db, the new columns and values get appended to the datagrid, meanwhile the old columns and value doesn't get disposed.

Note: If you are wondering how clear query works, we are just reloading the whole page. which removes the datagrid alltogether.

gridissue

datagridresult.razor:

`<Card>
            <CardBody>
                <Fields>
                    <Field ColumnSize="ColumnSize.Is12.OnDesktop">
                        <TextEdit TextChanged="@((string value) => QueryService.SetQueryText(value))" />
                    </Field>
                    <Divider />
                    <Field ColumnSize="ColumnSize.IsAuto.OnDesktop">
                        <Button Color="Color.Primary" Clicked="@(() => dataGridRef.Reload())">Start Query</Button>
                    </Field>
                </Fields>
                <DataGrid @ref="dataGridRef"
                          TItem="DataRowView"
                          Data="@queryDataView"
                          ReadData="@OnReadData"
                          ShowPager="true" PageSize="@RowCount"
                          RowStyling="@DataGrid_OnRowStyling"
                          Filterable="false">
                    <NextPageButtonTemplate>
                        Next
                    </NextPageButtonTemplate>
                    <LoadingTemplate>
                        <div class="box">
                            <progress class="progress is-small is-primary" max="100" />
                        </div>
                    </LoadingTemplate>
                    <DataGridColumns>
                        @foreach (DataColumn column in QueryTable.Columns)
                        {
                            <DataGridColumn TItem="DataRowView" Caption="@column.ColumnName" Field="@nameof(DataRowView.RowVersion)" Context="item" Filterable="false">
                                <DisplayTemplate>
                                    @($"{item[column.ColumnName]}")
                                </DisplayTemplate>
                            </DataGridColumn>
                        }
                    </DataGridColumns>
                </DataGrid>
            </CardBody>
        </Card>`

datagridresult.razor.cs:

`async Task OnReadData(DataGridReadDataEventArgs<DataRowView> e)
    {
        if (QueryText != "")
        {
            //dataGridRef = new DataGrid<DataRowView>();
            await Task.Run(() => QueryService.StartTable());
            queryDataView = QueryTable.DefaultView.Cast<DataRowView>();
            displayRefreshField = Display.Always;
            StateHasChanged();
        }
    }`

I am not sure why the code is not showing in a formatted way so attaching screenshot of the same
datagridresult.razor:
image

datagridresult.razor.cs:
image

David-Moreira:
Fixes #2809 

Removes `DataGridColumn` from ParentDataGrid if disposed.

It is of note that the `DataGridColumn` component was being disposed asynchronously, meaning that when invoking the regular Dispose (which was invoked from base component), the flag disposing was set to false, and not running the correct disposal of the component.
I personally have no idea as to the criteria of .NET invoking async or sync disposal, but we should keep this current quirk of our disposal proccess in mind going forward.

**Code used to reproduce & test:**
```
<DataGrid ElementId="ColoniesGrid-DataGrid-0" TItem="Employee" Data="@colonies" Striped="true" Bordered="true" Editable="true" ShowPager="true" Narrow="true" Resizable="true" ResizeMode="TableResizeMode.Columns" RowSelectable="@(v => false)">
    <DataGridColumns>
        @foreach (ColumnData column in columns)
                {
        <DataGridColumn TItem="Employee" Field="FirstName" Caption="@column.Name" >
            <DisplayTemplate>
                @context.FirstName
            </DisplayTemplate>    
        </DataGridColumn>
        }
    </DataGridColumns>
</DataGrid>
<NumericEdit @bind-Value="colQuantity">Quantity</NumericEdit>
<Button Clicked="CreateNewColumns">Create Columns</Button>
    @code {
    public class Employee
    {
        public int Id { get; set; }

    }

    private int colQuantity;

    public class ColumnData
    {
    }

    private List<ColumnData> columns = new();
    private void CreateNewColumns()
    {
        Random rand = new();
        columns = new(){};
        for ( var i = 0; i < colQuantity; i++)
        {
            columns.Add( new() { Name = rand.Next( 1, 10 ).ToString() } );
        }
        StateHasChanged();
    }

    private IEnumerable<Employee> colonies = new List<Employee>();
    protected override Task OnInitializedAsync()
    {
        var result = new List<Employee>();
        for (var i = 1; i<=10; i++ )
        {
            result.Add( new() { Id = i, FirstName = $"Test_{i}" } );
        }
        colonies = result;
        return Task.CompletedTask;
    }
}
```

### DIFF
--- a/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
+++ b/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
@@ -188,6 +188,16 @@ namespace Blazorise.DataGrid
         }
 
         /// <summary>
+        /// Removes an existing link of a child column with this datagrid.
+        /// <para>Returns:
+        ///     true if item is successfully removed; otherwise, false. 
+        /// </para>
+        /// </summary>
+        /// <param name="column">Column to link with this datagrid.</param>
+        public bool RemoveColumn( DataGridColumn<TItem> column )
+            => Columns.Remove( column );
+
+        /// <summary>
         /// Links the child column with this datagrid.
         /// </summary>
         /// <param name="aggregate">Aggregate column to link with this datagrid.</param>
--- a/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
+++ b/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
@@ -194,8 +194,8 @@ namespace Blazorise.DataGrid
         /// </para>
         /// </summary>
         /// <param name="column">Column to link with this datagrid.</param>
-        public bool RemoveColumn( DataGridColumn<TItem> column )
-            => Columns.Remove( column );
+        internal bool RemoveColumn( DataGridColumn<TItem> column )
+            => Columns.Remove( column ); // TODO: mark as public in v0.9.5
 
         /// <summary>
         /// Links the child column with this datagrid.
--- a/Source/Extensions/Blazorise.DataGrid/DataGridColumn.cs
+++ b/Source/Extensions/Blazorise.DataGrid/DataGridColumn.cs
@@ -2,6 +2,7 @@
 using System;
 using System.Collections.Generic;
 using System.Text;
+using System.Threading.Tasks;
 using Blazorise.DataGrid.Utils;
 using Microsoft.AspNetCore.Components;
 #endregion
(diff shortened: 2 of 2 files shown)


Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -187,6 +187,16 @@ public void AddColumn( DataGridColumn<TItem> column )
...
         }
 
+        /// <summary>
+        /// Removes an existing link of a child column with this datagrid.
+        /// <para>Returns:
+        ///     true if item is successfully removed; otherwise, false. 
+        /// </para>
+        /// </summary>
+        /// <param name="column">Column to link with this datagrid.</param>
+        public bool RemoveColumn( DataGridColumn<TItem> column )
```
stsrki, line 197: Should we add the `public` method? While not breaking-change it still changes the API.

David-Moreira: I forgot. Let's make it internal. And introduce it public on 9.5

stsrki: Done!

PR #2829, opened 2021-08-23
DataGrid: edit item creator

puschie286:
Add a callback for creating a new instance on editing to ensure nested object lifecycle


stsrki:
Looking good to me. @David-Moreira any comment?

David-Moreira:
It's a nice idea for one of datagrid's limitations. 
Might provoke unexpected behaviours if we like provide an entirelly new object for example? Other than that, looks good.

stsrki:
Yeah, as long as the item reference stays the same I think it should be good. We need to document it so that people are aware of it.

stsrki:
Sorry, it took me too long to respond. Thank you for the PR!

PR #2838, opened 2021-08-27
Fixed PreviouslySelectedSlide being null

### LINKED ISSUE #2832: Carousel with only one slide: Slide change (incl. Autoplay) not working after upgrade to 0.9.4.3
This doesn't work:
`<Carousel SelectedSlide="0" Class="border-0">                
        <CarouselSlide Name="0">
            <h1>This is announcement number 1</h1>
        </CarouselSlide>
</Carousel>`

This works:
`<Carousel SelectedSlide="0" Class="border-0">                
        <CarouselSlide Name="0">
            <h1>This is announcement number 1</h1>
        </CarouselSlide>
        <CarouselSlide Name="1">
            <h1>This is announcement number 2</h1>
        </CarouselSlide>
</Carousel>`

In your carousel.cs carousel.PreviouslySelectedSlideIndex is set to -1 which doesn't go great:

image

David-Moreira:
Fixes #2832

Used code provided in the issue to test. Also tested with SelectedSlide parameter not being set, which would also cause the bug/error.

### DIFF
--- a/Source/Blazorise/Components/Carousel/Carousel.razor.cs
+++ b/Source/Blazorise/Components/Carousel/Carousel.razor.cs
@@ -200,7 +200,10 @@ namespace Blazorise
 
             if ( carouselSlides.Count == 1 && string.IsNullOrEmpty( SelectedSlide ) )
             {
-                SelectedSlide = carouselSlides.Single().Name;
+                state = state with
+                {
+                    SelectedSlide = carouselSlides.Single().Name
+                };
             }
         }
 
@@ -336,9 +339,16 @@ namespace Blazorise
             {
                 AnimationRunning = false;
 
-                carouselSlides[PreviouslySelectedSlideIndex].Clean();
-                carouselSlides[SelectedSlideIndex].Clean();
-                carouselSlides[SelectedSlideIndex].Active = true;
+                if ( PreviouslySelectedSlideIndex >= 0 && PreviouslySelectedSlideIndex < NumberOfSlides )
+                {
+                    carouselSlides[PreviouslySelectedSlideIndex].Clean();
+                }
+
+                if ( SelectedSlideIndex >= 0 && SelectedSlideIndex < NumberOfSlides )
+                {
+                    carouselSlides[SelectedSlideIndex].Clean();
+                    carouselSlides[SelectedSlideIndex].Active = true;
+                }
 
                 await InvokeAsync( StateHasChanged );
 
@@ -382,10 +392,18 @@ namespace Blazorise
 
             //Trigger Animation
             carouselSlides[SelectedSlideIndex].Left = Direction == CarouselDirection.Previous;
-            carouselSlides[PreviouslySelectedSlideIndex].Left = Direction == CarouselDirection.Previous;
+
+            if ( PreviouslySelectedSlideIndex >= 0 && PreviouslySelectedSlideIndex < NumberOfSlides )
+            {
+                carouselSlides[PreviouslySelectedSlideIndex].Left = Direction == CarouselDirection.Previous;
+            }
 
             carouselSlides[SelectedSlideIndex].Right = Direction == CarouselDirection.Next;
-            carouselSlides[PreviouslySelectedSlideIndex].Right = Direction == CarouselDirection.Next;
+
+            if ( PreviouslySelectedSlideIndex >= 0 && PreviouslySelectedSlideIndex < NumberOfSlides )
+            {
+                carouselSlides[PreviouslySelectedSlideIndex].Right = Direction == CarouselDirection.Next;
+            }
 
             await InvokeAsync( StateHasChanged );
 
@@ -605,8 +623,8 @@ namespace Blazorise
 
                 state = state with
                 {
-                    SelectedSlide = value,
-                    PreviouslySelectedSlide = state.PreviouslySelectedSlide is null ? value : SelectedSlide
+                    PreviouslySelectedSlide = state.SelectedSlide,
+                    SelectedSlide = value
                 };
 
                 InvokeAsync( () => SelectedSlideChanged.InvokeAsync( state.SelectedSlide ) );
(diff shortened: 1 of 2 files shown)


Source/Blazorise/Components/Carousel/Carousel.razor.cs
```diff
@@ -200,10 +200,7 @@ internal void AddSlide( CarouselSlide slide )
 
             if ( carouselSlides.Count == 1 && string.IsNullOrEmpty( SelectedSlide ) )
             {
-                state = state with
-                {
-                    SelectedSlide = carouselSlides.Single().Name
```
stsrki, line 205: I called SelectedSlide on the record because I didn't want to trigger unnecesary event callback in SelectedSlide property. I remember it was giving me problems before.

David-Moreira: Well... The value did change... So in my opinion you should inform the user.. That the property has been changed...
Can you remember the problem...? 😅Maybe its in the commit history I'll take a look when I'm at pc.

David-Moreira: Don't see anything wrong with it.
**First introduced in :** 
a578b04a Select first slide of SelectedSlide is not set on initialization (#2576)

On #2572 you say its unnecessary: 
"Also, I made slight optimization so that SelectedSlideChanged doesn't get called unnecessary on initialization."

If this is the reason, I would have to disagree.
What do you think?

Also using the previous version and using `@bind-SelectedSlide` would cause an exception if the bound property was null or empty.

Code used to test working fine with the newest changes:
```
<Carousel @bind-SelectedSlide="SelectedSlide" Class="border-0">
    <CarouselSlide Name="0">
        <h1>This is announcement number 1</h1>
    </CarouselSlide>
</Carousel>


<Carousel @bind-SelectedSlide="@SelectedSlideRegularCarousel" Class="border-0"> <CarouselSlide Name="0"> <h1>This is announcement number 1</h1> </CarouselSlide> <CarouselSlide Name="1"> <h1>This is announcement number 2</h1> </CarouselSlide> </Carousel>

Selected Sldie is : @SelectedSlide

Selected Sldie 2 is : @SelectedSlideRegularCarousel

@code {
    public string SelectedSlide { get; set; } 

    public string SelectedSlideRegularCarousel { get; set; }
}
```

stsrki: Well, once the component is initialized and you provide it a value it means it is initialized with that same value. It didn't change as there was nothing before. So there is no reason to call a change event at the initialization.

David-Moreira: There is, cause you have mutated it?
The component basically says: if you don't provide me with a starting slide, I will set it myself.
If that should only be internal then the getter shouldn't grab the internal value... Well at least in my opinion... 
Either way if you don't agree, I will change it later... It will need more code as it causes some off cases bugs.

stsrki: So, I did the quick test in WinForms(I know, I know 😅)

```
public Form1()
        {
            InitializeComponent();

            textEdit2.Text = "1";
        }

        private void Form1_Load( object sender, EventArgs e )
        {
            textEdit2.Text = "2";
        }

        private void simpleButton1_Click( object sender, EventArgs e )
        {
            textEdit2.Text = "3";
        }

        private void textEdit2_EditValueChanged( object sender, EventArgs e )
        {
            Debug.WriteLine( textEdit2.Text );
        }
```

console writes only 2 and 3.

David-Moreira: I don't think its even comparable 😂😂
It's a constructor vs two events. And the event form load which is part of the life cycle does mutate it and notifies accordingly...
Also Windows forms has its own opinated framework, while blazor has its own... Don't think it makes sense to compare... we have more ways and freedom of doing thing nowadays...specially on the web.... 

Either way I see you're dead set in not invoking the changed there. This means that the user will not receive that first change event if he's listening to it, possibly missing a slide? Either way... I gotcha... We'll change it later when I get back on pc.

David-Moreira: So I was reverting and retesting... with this:

```
internal void AddSlide( CarouselSlide slide )
{
    carouselSlides.Add( slide );

    if ( carouselSlides.Count == 1 && string.IsNullOrEmpty( SelectedSlide ) )
    {
        var selectSlide = carouselSlides.Single().Name;
        state = state with
        {
            SelectedSlide = selectSlide,
            PreviouslySelectedSlide = state.SelectedSlide is null ? selectSlide : SelectedSlide
        };
    }
}
```

 there's a very very weird behaviour... with this code:
```
<Carousel @bind-SelectedSlide="SelectedSlide" Class="border-0">
    <CarouselSlide Name="carousel1">
        <h1>This is announcement number 1</h1>
    </CarouselSlide>
</Carousel>

<Carousel @bind-SelectedSlide="SelectedSlideRegularCarousel" Class="border-0">
    <CarouselSlide Name="Test1"> <h1>This is announcement number 1</h1> </CarouselSlide>
    <CarouselSlide Name="Test2"> <h1>This is announcement number 2</h1> </CarouselSlide>
</Carousel>

Selected Sldie is : @SelectedSlide

Selected Sldie 2 is : @SelectedSlideRegularCarousel

@code {
    public string SelectedSlide { get; set; }

    public string SelectedSlideRegularCarousel { get; set; }
}
```
Both this examples above, work just well, if alone on the page. Altough the first example won't update the Selected Slide from the AddSlide... But when together like above... causes this...

Both images from same call stack. First one at top, where error was thrown, second one further down with a call from the other carousel? Does this call stack make any sense to you?

stsrki:
I have made some changes and it seems now to work properly.  @David-Moreira once you confirm I will merge it.

David-Moreira:
So like we've talked. We've decided to slightly refactor the carousel slide getters and some logic making it more readable and maintanable and making it safer by just straight up null checking.

Re-tested on both WASM & Server, the original issue, the carousel page & the other bug I had found.

stsrki (approved):

stsrki:
It works well now. I like how you structured the code. Thanks!

PR #2864, opened 2021-09-02
Disable Bar animation during component initialization

### LINKED ISSUE #2863: Documentation: The menu shouldn't be opened for 1 second at page load on a smartphone
**To Reproduce**
1. Go to https://preview.blazorise.com/docs on a smartphone (I tested it in Google Chrome 92 on Samsung Galaxy S7).
2. The menu will open, and then hide after 1 second

I reproduced it also on Windows 10 after selecting Galaxy S5 device (360x640 resolution) in Developer Tools in Google Chrome 92. So I assume it is also reproductible on other smartphones.

**Expected behavior**
Opening menu for 1 second looks strange, I suppose it is a bug. No other component library documentation does this.

**Screenshots**
This menu is displayed for 1 second at page load on my smartphone:

Screenshot_20210902-114821_Chrome

stsrki:
fixes #2863 Documentation: The menu shouldn't be opened for 1 second at page load on a smartphone

We had a problem when the page is opened on mobile, the vertical Bar would show for a moment, and then it would minimize. It looks strange.

In this PR I have added a flag that tells us the `initial` state of the Bar and if it is `true` it would disable any animation and transition. It does so only for vertical mode.

@David-Moreira can you review and test? The change could potentially be breaking. We need to test demo apps and docs apps. Resize, responsive, collapse, etc. should work.

PS. i just realized I named branch `dev-v2-bar-initial` instead of `docs-v2-bar-initial` 🤦‍♂️


David-Moreira:
Well you kept the changes to a minimum, so there isn't much room to break something. :P Seems like you did a good job, don't see anything wrong, everything's working.

PR #2868, opened 2021-09-03
Add Polish language resources

cagrin:
* Add Polish language resources


stsrki:
Looking good. Thank you!

PR #2904, opened 2021-09-10
Fixes and improvements on AsyncValidator

### LINKED ISSUE #2750: Async Validation for DataGrid
Could you also add Async Validation for DataGrid? Thanks.

stsrki:
resolves #2750 Async Validation for DataGrid

With this PR I have:

- removed ValidatingAll event from Validations
- added AsyncValidator on DataGridColumn

The reason for the removal of `ValidatingAll` is that it didn't play well with the async code. Since it uses regular event subscribers they cannot be awaited and canceled afterward. So instead I have added two new APIs on the `IValidation` and run validation directly. It works quite well.

Since this is possibly a breaking change we will need to document it properly.


David-Moreira:
PR looks good. What's the behaviour for having both an Async and regular Validator set?

stsrki:
When `AsyncValidator` is defined it will be awaited when the user clicks on the save button in Datagrid edit mode. That part is working just fine.

Problem is that `.ValidateAll()` method on Validations is not working properly when mixed with async code. Since it is synchronous it will not wait for any `AsyncValidator`. I think we should remove it and replace it with the new `.ValidateAllAsync()`. But it is a breaking change 🤔

David-Moreira:
Oh right I think I had one of our demos form where that happened. Form would validate every normal validation but miss the async one. I agree it should wait every single validation, yea.

stsrki:
I have removed ValidateAll in favor of ValidateAllAsync. It's better to do it now than later. And we should probably think more about other APIs that could potentially better be async. Like `modal.ShowAsync`, or `dropdownToggler.ToggleAsync`. Basically anything that currently has async calls internally(like JSInterop) or it could have had at one time in it future.

PR #2911, opened 2021-09-13
Add DisabledDates to DatePicker

### LINKED ISSUE #2825: Disable range of dates in DatePicker
Example on https://flatpickr.js.org/examples/ under section **Disabling dates**.

gcardinale:
resolves #2825 Disable range of dates in DatePicker


stsrki:
@gcardinale, Thank you for the PR. The approach looks good to me. I only didn't like the use of strings for the disabled dates so I have used TValue instead. And I have used a built-on array compared to make sure dates are changed.

Before I merge please test one more time on your side to make sure it works as expected. Thank you!

Source/Blazorise.Bootstrap/Blazorise.Bootstrap.csproj
```diff
@@ -4,6 +4,7 @@
 
   <PropertyGroup>
     <PackageTags>blazorise blazor components bootstrap</PackageTags>
+    <TargetFrameworks>net5.0</TargetFrameworks>
```
stsrki, line 7: Can you please remove this from `.csproj` files? The framework is defined only through the shared `.props` file.

Source/Blazorise/Components/DatePicker/DatePicker.razor.cs
```diff
@@ -311,10 +312,43 @@ public override async Task SelectAsync( bool focus = true )
         [Parameter] public bool TimeAs24hr { get; set; }
 
         /// <summary>
-        /// List of disabled dates.
+        /// List of disabled Dates. Format string "yyyy-MM-dd".
         /// </summary>
-        [Parameter] public IEnumerable<TValue> DisabledDates { get; set; }
+        [Parameter] public string[] DisabledDates { get; set; } = new string[0];
```
stsrki, line 317: Why did you convert back to string values?

Issue #2810, opened 2021-08-13
Modal Closing is an Action instead of an EventCallback


njannink:
Model Closing event is of type `Action<ModalClosingEventArgs>` so when I want to do something async in this OnClosing it will have to be something like this `async void OnClosing(ModalClosingEventArgs args) { ....  }` which is undesirable. The type of this event should be an EventCallback so async operations are supported.

PR #2923, opened 2021-09-16
Async refactoring's

### LINKED ISSUE #1359: Modal - Show event
Hello,

On you [Model documentation](https://blazorise.com/docs/components/modal/#closing) you have a closing event, you could also have a `Showing` and/or `Shown`.  
This is a requirement for what I want to use the models for, currently in the app I am developing I use `RenderFragment` for my dialogs, so every time a dialog shows then the component does it's `Init`, inside the init method I can then setup the form and the object the form is editing.

I am looking to replace my `RenderFagments` with models in the hop to improve performance.

What do you think?

stsrki:
resolves #1359 Modal - Show event

When we change the component state imperatively instead of declaratively, eg. `modal.Show()`, in most of the cases it can also call JS interop or some other async call behind the scenes. Problem is that currently we have `void Show()` and similar methods. Void should generally be avoided for any async code.

With this PR I have started converting all current void methods into Task-able methods.

`void Show()` into `Task Show()`

This will be best in the long run, and it is the best time before 1.0.

---

| Component | Before          | After           |
|-----------|-----------------|-----------------|
| Modal     | `void Show()`   | `Task Show()`   |
|           | `void Hide()`   | `Task Hide()`   |
| Alert     | `void Show()`   | `Task Show()`   |
|           | `void Hide()`   | `Task Hide()`   |
|           | `void Toggle()` | `Task Toggle()` |
| Dropdown     | `void Show()`   | `Task Show()`   |
|           | `void Hide()`   | `Task Hide()`   |
|           | `void Toggle()` | `Task Toggle()` |
| BarDropdown     | `void Show()`   | `Task Show()`   |
|           | `void Hide()`   | `Task Hide()`   |
|           | `void Toggle()` | `Task Toggle()` |
| Button     | `void Focus()`   | `Task Focus()`   |
| Collapse   | `void Toggle()`   | `Task Toggle()`   |
| DropdownList     | `void Focus()`   | `Task Focus()`   |
| DropdownToggle     | `void Focus()`   | `Task Focus()`   |
| ListGroup     | `void SelectItem()`   | `Task SelectItem()`   |
| ProgressBar     | `void Animate()`   | `Task Animate()`   |
| Steps     | `void SelectStep()`   | `Task SelectStep()`   |
| StepsContent     | `void SelectPanel()`   | `Task SelectPanel()`   |
| Tabs     | `void SelectTab()`   | `Task SelectTab()`   |
| TabsContent     | `void SelectPanel()`   | `Task SelectPanel()`   |
| Validations     | `void ValidateAll()`   | `Task ValidateAll()`   |
|     | `void ClearAll()`   | `Task ClearAll()`   |
| SelectList  | `void Focus()`   | `Task Focus()`   |
| Autocomplete  | `void Focus()`   | `Task Focus()`   |
|  ALL input components   | `void Focus()`   | `Task Focus()`   |
|        | `void FocusAsync()`   | **removed**   |

---

The reason why I didn't add the `Async` suffix is that there will be only an asynchronous method.


njannink:
With relation to this Async work could you also look at this one: https://github.com/Megabit/Blazorise/issues/2810

Maybe the Action<ModelClosingEventArgs> can be changed to a Func<ModelClosingEventArgs,Task> but best would be an EventCallback. Already had a quick look at it, but what happens there is a bit tricky

stsrki:
> With relation to this Async work could you also look at this one: #2810
> 
> Maybe the Action can be changed to a Func<ModelClosingEventArgs,Task> but best would be an EventCallback. Already had a quick look at it, but what happens there is a bit tricky

Yes, that is also one of the reasons for this refactorings. I think it will `Func<ModelClosingEventArgs,Task>` because EventCallback can trigger re-render and we dont want that on closing event.

Source/Blazorise/Components/Modal/Modal.razor.cs
```diff
@@ -182,7 +182,7 @@ internal protected void Hide( CloseReason closeReason )
                 // finally reset close reason so it doesn't interfere with internal closing by Visible property
                 this.closeReason = CloseReason.None;
 
-                InvokeAsync( StateHasChanged );
+                await InvokeAsync( StateHasChanged );
```
David-Moreira, line 188: Can return

Source/Blazorise/Components/MessageAlert/MessageAlert.razor.cs
```diff
@@ -86,7 +86,7 @@ protected Task OnCancelClicked()
         {
             return InvokeAsync( async () =>
```
David-Moreira, line 87: redudancy? Can extract inner delegate?

Source/Blazorise/Components/Alert/Alert.razor.cs
```diff
@@ -109,7 +113,7 @@ private void HandleVisibilityState( bool visible )
         /// <param name="visible">True if <see cref="Alert"/> is visible.</param>
         private void RaiseEvents( bool visible )
```
David-Moreira, line 117: Possible and worth it converting?

Source/Blazorise/Components/Collapse/CollapseHeader.razor.cs
```diff
@@ -29,11 +29,12 @@ protected override void BuildClasses( ClassBuilder builder )
...
         /// <returns>A task that represents the asynchronous operation.</returns>
-        protected Task ClickHandler()
+        protected async Task ClickHandler()
         {
-            ParentCollapse.Toggle();
+            if ( ParentCollapse != null )
+                await ParentCollapse.Toggle();
 
-            return Clicked.InvokeAsync( null );
+            await Clicked.InvokeAsync( null );
```
David-Moreira: I think you've removed the explicit null argument on previous examples... 
`await Clicked.InvokeAsync( null );`

Source/Blazorise/Base/BaseInputComponent.razor.cs
```diff
@@ -62,7 +62,7 @@ public override async Task SetParametersAsync( ParameterView parameters )
                     }
                     else
                     {
-                        ExecuteAfterRender( () => FocusAsync() );
+                        ExecuteAfterRender( () => Focus() );
```
David-Moreira, line 65: Does this ExecuteAfterRender support async?

Source/Blazorise/Components/ColorEdit/ColorEdit.razor.cs
```diff
@@ -70,13 +70,7 @@ protected override Task<ParseValue<string>> ParseValueFromStringAsync( string va
...
-        public virtual void Select( bool focus = true )
-        {
-            InvokeAsync( () => SelectAsync( focus ) );
-        }
-
-        /// <inheritdoc/>
-        public virtual async Task SelectAsync( bool focus = true )
+        public virtual async Task Select( bool focus = true )
         {
             await JSRunner.Select( ElementRef, ElementId, focus );
```
David-Moreira, line 75: Might choose to not set up the state machine in the methods from this commit, like in the code base.

David-Moreira (changes requested):

Source/Blazorise/Components/Alert/Alert.razor.cs
stsrki: This one is problematic because it can be called from Visible setter. It is private so it doesn't affect users. We can leave it.

Source/Blazorise/Components/Collapse/CollapseHeader.razor.cs
stsrki: Yes. In olded blazor versions there was no parameterless InvokeAsync so null was needed. But now we can safely remove it.

Source/Blazorise/Base/BaseInputComponent.razor.cs
stsrki: It does.

Source/Blazorise/Components/ColorEdit/ColorEdit.razor.cs
stsrki: We can return without await, and add AsTask.

Source/Blazorise/Components/MessageAlert/MessageAlert.razor.cs
stsrki: This is OK. We have multiple operations that needs to be awaited.

PR #2936, opened 2021-09-19
Carousel: Fixes disabling of AutoPlay not working

### LINKED ISSUE #2910: Carousel with only 1 item has a UI flaw
jumpy

When the Carousel component only has one item then it makes an unfortunate UI flaw. I have enclosed a gif showing the flaw. 

I have tried playing around with the AutoPlay property but it makes no difference. 

When there is more than 1 item then there is no issue. 

I am creating my carousel like this

```
<Carousel SelectedSlide="0" Class="border-0" Autoplay="false">
                @{
                    var i = 0;
                    foreach (var item in DataComponentModel.Announcements)
                    {
                        if (item.Announcement is null)
                        {
                            continue;
                        }
                        var name = i.ToString();
                        var id = item.CommunityGroupAnnouncementId ?? 0;
                        <CarouselSlide Name="@name">
                            <div class="community-group-announcement-carousel-item-container-overlay" @onclick="@(() => ShowPopup(id))"></div>
                            <div class="community-group-announcement-carousel-item-container overflow-hidden p-3">
                                @((MarkupString) item.Announcement)
                            </div>
                        </CarouselSlide>
                        i++;
                    }
                }
            </Carousel>
```

I am using version 0.9.4.4

David-Moreira:
Closes #2910

User noticed that disabling AutoPlay on Carousel was not working. This PR fixes it.

Used Blazorise's Demo CarouselPage to test. Changing the `AutoPlay` and `Interval` Parameters to test.

### DIFF
--- a/Demos/Blazorise.Demo/Pages/Tests/CarouselPage.razor
+++ b/Demos/Blazorise.Demo/Pages/Tests/CarouselPage.razor
@@ -16,7 +16,7 @@
                 </Fields>
             </CardBody>
             <CardBody>
-                <Carousel @bind-SelectedSlide="@selectedSlide" ShowIndicators="@showIndicators" ShowControls="@showControls">
+                <Carousel @bind-SelectedSlide="@selectedSlide" ShowIndicators="@showIndicators" ShowControls="@showControls" Autoplay="false" Interval="100">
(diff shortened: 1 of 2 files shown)


stsrki (approved):
Looking good!

PR #2954, opened 2021-09-22
Nested dropdowns and refactor of closeable components

### LINKED ISSUE #1559: Nested Dropdowns malfunctioning
**Describe the bug**
I am trying to nest BarDropdown according to the example at [https://blazorise.com/docs/components/bar/#nested-dropdowns](https://blazorise.com/docs/components/bar/#nested-dropdowns)
When clicking "Nested toggler" the entire dropdown collapses.

**To Reproduce**
1. Add the code from the example
2. Click on the nested menu item "Nested toggler"
3. The entire dropdown disappears and no sub-items are displayed.

**Expected behavior**
The expected behavior is for the nested submenuitem "Nested item" to be displayed.

### LINKED ISSUE #2138: Escape key in Autocomplete component closes Modal
**Describe the bug**
While typing in the Autocomplete text box that is present in an open Modal, the escape key propagates up to the Modal. The result is that the Autocomplete clears its value, but the Modal also closes.

**To Reproduce**
1. Embed an Autocomplete with some options in a Modal 
2. Open the Modal
3. Start typing in the Autocomplete and have some autocomplete options appear
4. Press the Escape key on your keyboard

**Expected behavior**
I expect that the Autocomplete will clear its value, and that the Modal will stay open. The Autocomplete does clear its value, but the event propagates to the Modal and the Modal closes.

### LINKED ISSUE #2551: Refactor closable components
Affected components

- Modal
- Dropdown
- BarDropdown
- Select (AntDesign)

Currently, all our closable components are tracked using the JS once they are in an `open` state. The component will close if it loses focus or the user press `Escape` key. What is problematic is how we detect lost focus. We do it by checking elementId, from the opened component and the one we clicked on. With this, it is impossible to do any nesting of the closable components.

The only way to fix is to refactor the entire code and implement some new way of opening and closing states.

Once done it should fix

- #2138 Escape key in Autocomplete component closes Modal
- #1559 Nested Dropdowns malfunctioning

David-Moreira:
closes #2551 Refactor closable components
closes #1559 Nested Dropdowns malfunctioning
closes #2138 Escape key in Autocomplete component closes Modal


stsrki:
Looking good so far. Bootstrap is working nice. AntDesign is broken and the dropdown is not closing when you click outside of the menu. You will probably need to add mouse events to handle the `ShouldClose` boolean. Also `ShouldClose` naming feels strange but I cannot think of a better alternative 🤔

The same approach would have to be done for the `BarDropdown` component, in Horizontal mode. You can do it once we settle on the final version for regular Dropdown.

David-Moreira:
CanClose or MayClose? :O

Source/Extensions/Blazorise.Components/Autocomplete.razor
```diff
@@ -34,6 +34,8 @@
 
         @if ( DropdownVisible )
         {
+            JSRunner.RegisterClosableComponent( dotNetObjectRef, ElementRef );
```
stsrki, line 37: Just noticed this. Isn't it _dangerous_ to call JS async code in here? And it could potentially be called every time a component is rendered. We need to check for a boolean value jsRegistered to make sure it is registered only once. And possibly make use of `ExecuteAfterRender`.

PR #2957, opened 2021-09-22
Refactor Modal Visible parameter and state management

### LINKED ISSUE #1359: Modal - Show event

### LINKED ISSUE #2810: Modal Closing is an Action instead of an EventCallback

stsrki:
closes #2810 Modal Closing is an Action instead of an EventCallback
closes #1359 Modal - Show event

I have refactored the `Visible` parameter and `Closing` event. Instead of doing all the logic in the `Visible` setter, I have moved it to the `SetParametersAsync` lifecycle.

- The benefit of this is that the `Closing` event can now be properly awaited.
- The disadvantage is that `Visible` is used only as a hint for us that the parameter was received by the Blazot lifecycle. It should never be used directly as that would leave it in a bad state, eg. `modal.Visible = true`.

This will also come in handy for #1359

### DIFF
--- a/Source/Blazorise.Bulma/wwwroot/blazorise.bulma.js
+++ b/Source/Blazorise.Bulma/wwwroot/blazorise.bulma.js
@@ -24,7 +24,11 @@ window.blazoriseBulma = {
     modal: {
         open: (element, scrollToTop) => {
             if (scrollToTop) {
-                element.querySelector('.modal-card-body').scrollTop = 0;
+                const modalBody = element.querySelector('.modal-card-body');
+
+                if (modalBody) {
+                    modalBody.scrollTop = 0;
+                }
             }
         },
         close: (element) => {
--- a/Source/Blazorise/Components/Modal/Modal.razor.cs
+++ b/Source/Blazorise/Components/Modal/Modal.razor.cs
@@ -68,17 +68,11 @@ namespace Blazorise
             {
                 if ( visibleResult == true && await IsSafeToOpen() )
                 {
-                    state = state with { Visible = true };
-
-                    HandleVisibilityStyles( true );
-                    RaiseEvents( true );
+                    SetVisibleState( true );
                 }
                 else if ( visibleResult == false && await IsSafeToClose() )
                 {
-                    state = state with { Visible = false };
-
-                    HandleVisibilityStyles( false );
-                    RaiseEvents( false );
+                    SetVisibleState( false );
                 }
                 else
                 {
@@ -180,10 +174,7 @@ namespace Blazorise
 
             if ( await IsSafeToOpen() )
             {
-                state = state with { Visible = true };
-
-                HandleVisibilityStyles( true );
-                RaiseEvents( true );
+                SetVisibleState( true );
 
                 await InvokeAsync( StateHasChanged );
             }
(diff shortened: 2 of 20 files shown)


Source/Blazorise.Bulma/wwwroot/blazorise.bulma.js
```diff
@@ -24,7 +24,11 @@ window.blazoriseBulma = {
     modal: {
         open: (element, scrollToTop) => {
             if (scrollToTop) {
-                element.querySelector('.modal-card-body').scrollTop = 0;
+                const modalBody = element.querySelector('.modal-card-body');
```
David-Moreira, line 27: Forgot material.js?

Source/Blazorise/Components/Modal/Modal.razor.cs
```diff
@@ -61,6 +61,32 @@ public partial class Modal : BaseComponent, ICloseActivator
...
+                    HandleVisibilityStyles( true );
+                    RaiseEvents( true );
+                }
+                else if ( visibleResult == false && await IsSafeToClose() )
+                {
+                    state = state with { Visible = false };
+
+                    HandleVisibilityStyles( false );
+                    RaiseEvents( false );
+                }
```
David-Moreira: If these are always done in 'bulk' maybe you can extract these into a method with a bool argument?

David-Moreira (changes requested):
Tested on bootstrap, looking good.

Update Doc Attributes, `Closing` is now wrong? Also missing alot of attributes.
image

Two async calls, setup Async?
image

Could `Opening` and `Closing` be EventCallbacks? Or you think it's not worth it?
Also if there's a `Closed` EventCallback, maybe we should introduce an `Opened` EventCallback?
image

Source/Blazorise.Bulma/wwwroot/blazorise.bulma.js
stsrki: Done

njannink:
Since setting the Visible paramater isn't really a good idea, because it will leave the component in a bad state. Wouldnt it be better to make the setter private or protected?

stsrki:
> Since setting the Visible parameter isn't really a good idea, because it will leave the component in a bad state. Wouldnt it be better to make the setter private or protected?

Every parameter must be made public and that is Blazor constraint. Otherwise, it will not compile. With the changes in this PR the use of the `Visible` parameter is made in the right way. Meaning it can only be defined declaratively. And if you need to define it imperatively, you still have the option to use `Show()` and `Hide()`.

PR #2961, opened 2021-09-23
Diferent structure for dropdown submenus

stsrki:
This PR adds several improvements

- Places sub-Dropdown on the same level as sibling DropdownItem
- Custom CSS to adjust submenus next to the toggle button
- No need for `Direction="Direction.Right"`

I probably need to remove `.dropdown-submenu` but I will leave it until we settle for the final version.

Example

```cshtml
<Dropdown>
    <DropdownToggle Color="Color.Primary">Level 1</DropdownToggle>
    <DropdownMenu>
        <DropdownItem>Item 1.1</DropdownItem>
        <Dropdown Direction="Direction.Right">
            <DropdownToggle>Level 2</DropdownToggle>
            <DropdownMenu>
                <DropdownItem>Item 2.1</DropdownItem>
                <DropdownItem>Item 2.2</DropdownItem>
                <Dropdown Direction="Direction.Right">
                    <DropdownToggle>Level 3</DropdownToggle>
                    <DropdownMenu>
                        <DropdownItem>Item 3.1</DropdownItem>
                        <DropdownItem>Item 3.2</DropdownItem>
                        <Dropdown Direction="Direction.Right">
                            <DropdownToggle>Level 4</DropdownToggle>
                            <DropdownMenu>
                                <DropdownItem>Item 4.1</DropdownItem>
                                <DropdownItem>Item 4.2</DropdownItem>
                            </DropdownMenu>
                        </Dropdown>
                    </DropdownMenu>
                </Dropdown>
            </DropdownMenu>
        </Dropdown>
    </DropdownMenu>
</Dropdown>
```

### DIFF
--- a/Demos/Blazorise.Demo/Pages/Tests/ButtonsPage.razor
+++ b/Demos/Blazorise.Demo/Pages/Tests/ButtonsPage.razor
@@ -230,33 +230,27 @@
                     <DropdownToggle Color="Color.Primary">Level 1</DropdownToggle>
                     <DropdownMenu>
                         <DropdownItem>Item 1.1</DropdownItem>
-                        <DropdownItem>
-                            <Dropdown Direction="Direction.Right">
-                                <DropdownToggle Color="Color.Secondary">Level 2</DropdownToggle>
-                                <DropdownMenu>
-                                    <DropdownItem>Item 2.1</DropdownItem>
-                                    <DropdownItem>Item 2.2</DropdownItem>
-                                    <DropdownItem>
-                                        <Dropdown Direction="Direction.Right">
-                                            <DropdownToggle Color="Color.Secondary">Level 3</DropdownToggle>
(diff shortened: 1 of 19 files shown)


David-Moreira:
Is dropdown-submenu part of the providers styling or is it something you came up with? 
I wouldn't be suprised to just apply regular dropdown class, it's a regular dropdown, whether it is nested or not. Also I don't know if this is a dumb question, why does <DropdownMenu> even exist? is it possible to two or more menus inside a dropdown is that it?

stsrki:
I saw `.dropdown-submenu` was actually part of blazorstrap, so I tried to c/p it. But it isn't needed actually. With custom CSS it can also work and that's what I did. But, even though we don't need it for Bootstrap we will need it for the AntDesign, and possibly for Bulma. They have dedicated class names for submenus.

PR #2978, opened 2021-10-01
WIP: ColorPicker

### LINKED ISSUE #2656: Add ColorPicker component
based on https://github.com/Simonwep/pickr

stsrki:
resolves #2656 

Code to test:

```
<Row>
    <Column>
        <ColorPicker @bind-Color="@color1" />
    </Column>
    <Column>
        <TextEdit Text="@color1" />
        <Button Clicked="@(()=>color1 = "#00ff00")">SET COLOR</Button>
    </Column>
</Row>

<ColorPicker Swatches="@(new string[]{ "rgba(244, 67, 54, 1)", "rgba(233, 30, 99, 0.95)" })" />

@code {
    string color1 = "#ff0000";
}
```


David-Moreira:
With your example, 
Cancel behaviour is weird in my opinion. I'd expect it to just go back to the original bound color.
Clear does clear the color completly, but that's expected. I'm not sure, about straight up closing or not the window.

Can I choose to output the RGBA value instead of the hexadecimal one?

stsrki:
I know, the cancellation is really weird and I will need to investigate more. Their API is really weird to work with.

As for RGBA, it is not a bad idea to have.

David-Moreira:
Not sure what swatches is personally, but consider being able to change them on the fly...? Like for example binding from a property.

stsrki:
They are buttons with default colors, and yes they can be changed with the parameter.

image

David-Moreira:
They can be changed while running?

stsrki:
At the moment no, but I plan to add it.

David-Moreira:
On the example you sent.
On the second color pick, there's only two swatches defined... but I get all of these, should I not get just two?
image

On first color picker, set swatches to null, they still all show up?
image

stsrki:
I still need to decide if I don't define swatches should I just not show them or fall back to the default set of swatches.

David-Moreira:
Can we have the best of both worlds somehow? :D
If not, I guess having the default set is the best, in my opinion. 

Maybe a kind of Builder pattern would be good? So having this kind of fluent api for swatches:
`Swatches.None()` or `Swatches.Default()` or `Swatches.AddRGB(blabla).AddRGB(BLABLA).AddHEX(blabla) `

stsrki:
It is not a bad idea. But in this case, I think it is easier to just define them as a string array. The builder would be too much.

David-Moreira:
If you think it's overkill, we have two options I guess:

- null = not show
- empty = default

- null = default 
- empty still displays default and there's no way to hide swatches.
- or just additionally provide a bool DisplaySwatches....

David-Moreira:
The first option would work with some documentation on the parameter... 
However I like the 2nd option with the flag... I guess that when people want to disable something, they look for flags.

PR #2984, opened 2021-10-02
WIP : Autocomplete tag

### LINKED ISSUE #2528: Tag autocomplete
I created a simple tag autocomplete, with a few changes (read: ugly AF) it should be possible to add this as another component.

### Additional context
Here is the current code. AllTags is the list of all tags to be shown, Selected tags is the list to which the tags will be added upon user selection. Tag object will have to be replaced by a generic and we will also need an accessor for the string, but I wanted to hear some feedback before I do it and create a pull request.

```
<div>
    <div class="tagAutocomplete">
        @foreach (var tag in SelectedTags)
        {
            <Badge Color="Color.Primary" CloseClicked="() => SelectedTags.Remove(tag)">@tag.Name</Badge>
        }
    </div>

    <Autocomplete TItem="Tag"
                  Data="AllTags"
                  SelectedValue="Value"
                  TextField="@(item => item.Name)"
                  ValueField="@(item => item)"
                  SelectedValueChanged="MySearchHandler"
                  Placeholder="Vyberte označenie"/>
</div>

@code{
    private string Value { get; set; }

    [Parameter]
    public IList<Tag> SelectedTags { get; set; } = null!;

    [Parameter]
    public IList<Tag> AllTags { get; set; } = null!;

    //TODO This will be broken by next version of blazorise
    void MySearchHandler(object newValue)
    {
        if (newValue is Tag tag)
        {
            SelectedTags.Add(tag);
            Value = string.Empty;
        }
    }
}

```

David-Moreira:
**WIP...** Closes #2528 
I looked up at the js libraries you suggested, I honestly think it's not worth to interop js for this feature, as I saw nothing special js did.

So basically the example you had given in the thread is pretty much the way to go and what I seen every library for this feature doing...
A bordered div wrapper with the tag html elements + input inside.
Only thing special is getting the css right for every provider. For now it's only done for bootstrap and not 100% right.

For now we've introduced these new parameters:
image

Demo updated with new feature.
image

Need your opinion on `FreeTyping`...
So the new `SelectedValues` `Parameter `allows `List<TValue>` support, for the `TValue` provided....
I'm pretty sure people will eventually want to turn on `FreeTyping` and be able to bind a `List<string>` with the values that the user types freely... Do you agree in introducing a `SelectedTexts` `Parameter`? Just seem like too many `Parameters` to keep track of selected stuff, but I pretty much think it's needed!

### DIFF
--- a/Demos/Blazorise.Demo/Pages/Tests/ComponentsPage.razor
+++ b/Demos/Blazorise.Demo/Pages/Tests/ComponentsPage.razor
@@ -30,7 +30,7 @@
         </Card>
     </Column>
     <Column ColumnSize="ColumnSize.IsFull.OnMobile.IsHalf.OnTablet">
-        <Card Margin="Margin.Is4.FromBottom">
+        <Card Margin="Margin.Is4.OnY">
             <CardHeader>
                 <CardTitle>Dropdown List</CardTitle>
             </CardHeader>
@@ -60,7 +60,7 @@
     </Column>
 </Row>
 <Row>
-            <Column ColumnSize="ColumnSize.IsFull.OnMobile.IsHalf.OnTablet">
+    <Column ColumnSize="ColumnSize.IsFull.OnMobile.IsHalf.OnTablet">
         <Card Margin="Margin.Is4.OnY">
             <CardHeader>
                 <CardTitle>Autocomplete</CardTitle>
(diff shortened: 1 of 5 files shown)


stsrki:
I still need to try it, but from what I can see in the code it seems good. The only thing I would change is to rename `MultipleSelection` into `Multiple` so that it is the same as on the `Select` component.

And I agree `SelectedTexts` is also needed.

stsrki:
All these APIs should be in documentation with explanation on why and how it works. Especially with SelectedText.

David-Moreira:
I'll update the auto complete docs when I finish.

PR #2985, opened 2021-10-02
Numeric Min and Max of TValue

### LINKED ISSUE #2876: Numeric value does not handle values too low or too high for type TValue
**Describe the bug**
When the value of a numeric input is higher or lower than the `TValue` can handle, it does not throw an error. Instead, it keeps the last valid value it had.

**To Reproduce**
1. Clone my repository https://github.com/InzeNL/Blazorise-NumericEdit-Bug-Demo
2. Launch the application
3. Enter 0 as a value
4. The div above the input should now show "0"
5. Remove the value
6. The div above the input should now show nothing
7. Enter 1 as a value
8. The div above the input should now show "1"
9. Add a "-" in front of the 1
10. The div above the input still shows "1"
11. Enter 1111111111 as a value
12. The div above the input should now show "1111111111"
13. Add another number to the end of this value
14. The div above the input still shows "1111111111"

**Expected behavior**
I am not sure what the best behaviour would be, but I think the most reasonable would be as follows:
If the value (e.g. -1, or 11111111112) does not fit in the value type (e.g. uint), set the value to "null"

This makes it possible to verify that the values are valid.
Right now, if I enter a valid value, and then enter a value too big, it uses the last valid value.
The user does not see any feedback, and thinks "11111111112" is a valid value. The verification thinks the value is actually "1111111111", and uses that.

**Screenshots**
image
image
image
image
image
image

stsrki:
resolves #2876 Numeric value does not handle values too low or too high for type TValue

This PR tries to resolve the problem when we type a numeric value that is larger or smaller than the TValue can hold, eg. for `int` max value is 2147483647. The problem is that we handle numeric validations in JS and even if it passes, it will fail on the C# part once we try to parse it. Thing is, we shouldn't be able to write anything above max or min.

My approach is to handle the min and max of the actual type in JS. And if it fails it will prevent the input and replace the entered value with the min or max, depending on what failed. I'm still not sure if this is the right way but I don't have any other idea at the moment.

It is still ugly as it is just a concept. All unit tests are passing so in theory it shouldn't break people's code. 🤞

```cshtml
<CardBody>
    <div>
        byte: @ByteValue
    </div>
    <NumericEdit TValue="byte?" @bind-Value="ByteValue" Min="10" Max="20" />

    @code {
    byte? ByteValue;
    }
</CardBody>
<CardBody>
    <div>
        int? @NumericValue
    </div>
    <NumericEdit TValue="int?" @bind-Value="NumericValue" />

    @code
    {
    int? NumericValue;
    }
</CardBody>
<CardBody>
    <div>
        long? @LongValue
    </div>
    <NumericEdit TValue="long?" @bind-Value="LongValue" />

    @code
    {
    long? LongValue;
    }
</CardBody>
<CardBody>
    <div>
        decimal? @DecimalValue
    </div>
    <NumericEdit TValue="decimal?" @bind-Value="DecimalValue" Min="10" Max="20" />

    @code
    {
    decimal? DecimalValue;
    }
</CardBody>
```

### DIFF
--- a/Source/Blazorise/Components/NumericEdit/NumericEdit.razor.cs
+++ b/Source/Blazorise/Components/NumericEdit/NumericEdit.razor.cs
@@ -92,7 +92,9 @@ namespace Blazorise
                 Separator = DecimalsSeparator,
                 Step,
                 Min = Min.IsEqual( default ) ? minFromType : Min,
-                Max = Max.IsEqual( default ) ? maxFromType : Max
+                Max = Max.IsEqual( default ) ? maxFromType : Max,
+                TypeMin = minFromType,
+                TypeMax = maxFromType
             } );
 
             await base.OnFirstAfterRenderAsync();
--- a/Source/Blazorise/Components/NumericEdit/NumericEdit.razor.cs
+++ b/Source/Blazorise/Components/NumericEdit/NumericEdit.razor.cs
@@ -238,7 +238,9 @@ namespace Blazorise
             // make sure that null values also starts from zero
             value ??= Converters.ChangeType<TValue>( 0 );
 
-            return MathUtils<TValue>.Add( value, Converters.ChangeType<TValue>( Step.GetValueOrDefault( 1 ) * sign ) );
+            return sign > 0
+                ? MathUtils<TValue>.Add( value, Converters.ChangeType<TValue>( Step.GetValueOrDefault( 1 ) ) )
+                : MathUtils<TValue>.Subtract( value, Converters.ChangeType<TValue>( Step.GetValueOrDefault( 1 ) ) );
         }
 
         /// <summary>
--- a/Source/Blazorise/Components/NumericEdit/NumericEdit.razor.cs
+++ b/Source/Blazorise/Components/NumericEdit/NumericEdit.razor.cs
@@ -91,8 +91,8 @@ namespace Blazorise
                 Decimals,
                 Separator = DecimalsSeparator,
                 Step,
-                Min = Min.IsEqual( default ) ? minFromType : Min,
-                Max = Max.IsEqual( default ) ? maxFromType : Max,
+                Min = MinDefined ? (object)Min : null,
+                Max = MaxDefined ? (object)Max : null,
                 TypeMin = minFromType,
                 TypeMax = maxFromType
             } );
--- a/Source/Blazorise/wwwroot/blazorise.js
+++ b/Source/Blazorise/wwwroot/blazorise.js
@@ -732,44 +732,33 @@ window.blazorise = {
             let value = this.element.value,
                 selection = this.carret();
 
-            function toFixed(x) {
-                var result = '';
-                var xStr = x.toString(10);
-                var digitCount = xStr.indexOf('e') === -1 ? xStr.length : (parseInt(xStr.substr(xStr.indexOf('e') + 1)) + 1);
-
-                for (var i = 1; i <= digitCount; i++) {
-                    var mod = (x % Math.pow(10, i)).toString(10);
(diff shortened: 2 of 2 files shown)


stsrki:
Code cleaned a little. All tests are passing and I think it is safe to merge. @David-Moreira If you can please test on your side to make sure I didn't miss something before I merge it. This is the last ticket for the 0.9.5.5.

David-Moreira:
0.9.4.5? :P
I think I can still take a look today, in a few hours.

stsrki:
> I think I can still take a look today, in a few hours.

Yes, for the _patch_ update.

David-Moreira (commented):
What if it's dynamic? Should we check OnParametersSet?

Source/Blazorise/Components/NumericEdit/NumericEdit.razor.cs
```diff
@@ -92,7 +92,9 @@ protected override async Task OnFirstAfterRenderAsync()
                 Separator = DecimalsSeparator,
                 Step,
                 Min = Min.IsEqual( default ) ? minFromType : Min,
-                Max = Max.IsEqual( default ) ? maxFromType : Max
+                Max = Max.IsEqual( default ) ? maxFromType : Max,
```
David-Moreira: What if user provides a max that is higher then the type max? Do we care about that?

David-Moreira:
Looks good to me.

David-Moreira:
image
If you increase above the limit, it goes around to the negative value limit. 

Is that expected?

David-Moreira:
Byte step up and step down doesn't work, is that expected?

David-Moreira:
If I step up on long max, I get a different behaviour then int max
image

David-Moreira:
Is this correct for long min?
image

David-Moreira:
On int if I type -999999999999 I get 0.
Overall I don't feel like everything has the same behaviour, can you take a look?

Source/Blazorise/Components/NumericEdit/NumericEdit.razor.cs
stsrki: Didn't I already fix this?

David-Moreira: Not sure. What I mean is, using numeric edit like this <NumericEdit TValue="int" Max="value higher then int max here" />

stsrki: There is no reason to support a value higher than int max because that same value cannot be stored in an `int`. You would store 300 into byte that has a max of 255. Same thing for the `int`.

David-Moreira: Yep I was just asking if we handle that use case correctly.

stsrki: Seems to me it is handled by the framework :)

image

Since TValue is recognized it will give an error if we want to assign a value larger than it can handle. I also tried to trick it by casting but then it breaks at runtime. Which makes sense.

stsrki:
> Byte step up and step down doesn't work, is that expected?

The problem with `byte` is that it doesn't have the operators for adding and subtracting. So the only way is to have a special rule for byte. We can ignore this for now.

Source/Blazorise/Components/NumericEdit/NumericEdit.razor.cs
David-Moreira: Ah great, haha

stsrki:
@David-Moreira I have fixed most of the problems that you noticed. Almost everything works except for a very large number where exponential will convert to zeroes. I'm still not sure if this is the limitation of JS numerics or something else. But for 99.99% of situations, the new behavior will be just fine. So considering the original problem is fixed I think it is safe to merge.

PR #2995, opened 2021-10-05
Autocomplete selected text bug fix

### LINKED ISSUE #2879: Autocomplete - updating bound SelectedValue from codebehind won't update the SelectedText
**Describe the bug**
I believe this has been broken by the newest improvements to *Autocomplete* in #2548. The whole `SelectedText` parameter introduction is quite a breaking change which was **NOT** documented very well unfortunately.

Currently if you bind the `SelectedValue` to a codebehind property, updating it will not update the newly introduced `SelectedText` as well. This is noticable especially in *edit*-like forms where you need to initialize *Autocomplete* with initial value but also when you have a presentation logic causing the `SelectedValue` to change (e.g. to `null`).

Changing the `SelectedText` parameter will update the `SelectedValue` properly - thanks to the `OnTextChangedHandler` implementation.

**Expected behavior**
I expected the `SelectedValue` to work as it was working before the improvements introduced in 0.9.4, if I am not intrested in managing `SelectedText` by myself. Now I have to manage state of two properties even though I have already told the component where it can get the text to display.

There seems to be already something prepared (which is actually currently only wasting performance) in the setter implementation of `SelectedValue` (lines 423-425).

https://github.com/Megabit/Blazorise/blob/8c5deda7ab61279036865e54a085e786c05063d4/Source/Extensions/Blazorise.Components/Autocomplete.razor.cs#L412-L427

I understand that updating the `SelectedText` from the setter of `SelectedValue` can possibly cause infinite loop of event callbacks. I am not really a fan of the introduced `SelectedText` functionality because it caused a lot of confusion in my opinion. There is still the `CurrentSearch` functionality as well and the respective events `SelectedTextChanged` and `SearchChanged` are almost completely duplicate. 

I am not sure what is the correct way out of this. I have seen somewhere that Blazorise team is going to completely remove the value selection behavior in *Autocomplete* and instead introduce autocomplete-like filtering in *Select* components in the future. But for now, this is really a hybrid that is not working very well.

### LINKED ISSUE #2991: Autocomplete cannot be set with initial value
Something is broken with the Autocomplete and if you define the SelectedValue it will show the item text in the search field.

I suspect it might be something with the removal of

```
SelectedText = item != null
    ? TextField?.Invoke( item )
    : string.Empty;
```

in one of the previous commits https://github.com/Megabit/Blazorise/pull/2548/files#diff-a895443c3b823714561da792f1cf43839cf68919cb8056adc1581d49426199eeL362

@David-Moreira can you look into this? It is a priority.

### LINKED ISSUE #2993: AutoComplete - SelectedValue updated through Parameter doesn't update text or validation status
**Describe the bug**
In AutoComplete, when using parameter SelectedValue to set a value, SelectedText or validation status are not updated.

**To Reproduce**
1. Pass a value through parameter SelectedValue. AutoComplete will be rendered "empty".  
2. After selecting an option using the dropdown, change the SelectedValue through the parameter and AutoComplete will keep showing the previous selected text.

**Expected behavior**
AutoComplete should update accordingly when parameter SelectedValue is set with a value.

**Additional context**
I have a repository with demo setting initial value and added a button to set the value through the parameter.
https://github.com/paulosdfigueiredo/Blazorise/tree/rel-0.9.4-autocomplete-selectedvalue-codebehind

Maybe a duplicate of #2991 ?

David-Moreira:
Closes #2991
Closes #2879
Closes #2993

I had infact removed 
```
SelectedText = item != null
    ? TextField?.Invoke( item )
    : string.Empty;
```
incorrectly like reported in #2991 

However, by reintroducing it, there would still be a minor bug, that when both `SelectedValue` and `SelectedText` are bound to `Autocomplete`. The `SelectedText` setter would run after the `SelectedValue` setter and set `SelectedText` to null (or whatever is set in the parameter initially) causing incoherent behaviour. So I moved the logic to `SetParametersAsync`.

Also introduced tests to support the PR, and will most likelly do so going forward so we start getting better coverage and can hopefully diminish the bugs caused by introducing new features.

### DIFF
--- a/Demos/Blazorise.Demo/Pages/Tests/ComponentsPage.razor
+++ b/Demos/Blazorise.Demo/Pages/Tests/ComponentsPage.razor
@@ -42,13 +42,12 @@
                                       TValue="string"
                                       Data="@myDdlData"
                                       TextField="@(( item ) => item.MyTextField)"
-                                      ValueField="@(( item ) => item.MyTextField)"
-                                      SelectedValue="@selectedSearchValue"
-                                      SelectedValueChanged="@MySearchHandler"
+                                      ValueField="@(( item ) => item.MyValueField.ToString())"
+                                      @bind-SelectedValue="selectedSearchValue"
                                       @bind-SelectedText="selectedAutoCompleteText"
                                       Placeholder="Search..."
                                       Filter="AutocompleteFilter.StartsWith"
-                                      FreeTyping="true"
+                                      FreeTyping
                                       CustomFilter="@(( item, searchValue ) => item.MyTextField.IndexOf( searchValue, 0, StringComparison.CurrentCultureIgnoreCase ) >= 0 )">
                             <NotFoundContent> Sorry... @context was not found! :( </NotFoundContent>
                         </Autocomplete>
@@ -163,11 +162,6 @@
     void MyDropValueChangedHandler( int newValue )
     {
         selectedDropValue = newValue;
-    }
-
-    void MySearchHandler( string newValue )
-    {
-        selectedSearchValue = newValue;
     }
 
     private void ShuffleList()
--- a/Source/Extensions/Blazorise.Components/Autocomplete.razor.cs
+++ b/Source/Extensions/Blazorise.Components/Autocomplete.razor.cs
@@ -50,6 +50,26 @@ namespace Blazorise.Components
 
         #region Methods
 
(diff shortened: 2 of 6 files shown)


stsrki:
Possibly it will. I think it makes sense to have it call any events after the rendering is done. It doesn't hurt since parameters were already processed.

PR #3017, opened 2021-10-13
MemoEdit(textarea) tab indentation

### LINKED ISSUE #2135: Enable input of tabs in MemoEdit

I would like to enable the input of the tab character `/t` in `<MemoEdit />`

An option to prevent the default tab functionality (change focus to next element) and instead add a tab to selected text.

Also support:
- adding a tab in the middle of text (not just at the end)
- indenting multiple selected lines
- shift + tab (un-indenting)

### Additional context

https://stackoverflow.com/questions/6637341/use-tab-to-indent-in-textarea
https://stackoverflow.com/questions/6140632/how-to-handle-tab-in-textarea
https://github.com/alanhogan/Tabby
https://github.com/fregante/indent-textarea
https://github.com/wjbryant/taboverride

stsrki:
closes #2135 Enable input of tabs in Memo

This PR brings the ability to use the Tab key in MemoEdit(textarea) component. First I tried to make it with the code from [css-tricks](https://css-tricks.com/snippets/javascript/support-tabs-in-textareas/) but I realized if I would to make it behave as expected I would need to write a lot more code. And I really don't want to reinvent the wheel.

So after I searched a bit, I found an excellent library that had all the features I needed. It is named Behave(https://github.com/jakiestfu/Behave.js). And while it is coming at a cost of increasing JS file by 8KB I think it is justified. 

Currently, we support:

- ReplaceTab: to enabled or disable new tab replacing
- SoftTabs: if true, uses space characters instead of `\t`
- TabSize: number of spaces when `SoftTabs` is enabled

@David-Moreira as usual, see if I missed something or you have any idea to improve it.

**Test code**

```cshtml
<MemoEdit ReplaceTab="@replaceTab" TabSize="@tabSize" SoftTabs="@softTabs" />
<Button Color="Color.Primary" Clicked="@(()=>replaceTab = !replaceTab)">TOGGLE TABS</Button>
<Button Color="Color.Primary" Clicked="@(()=>tabSize = 2)">TAB SIZE 2</Button>
<Button Color="Color.Primary" Clicked="@(()=>tabSize = 4)">TAB SIZE 4</Button>
<Button Color="Color.Primary" Clicked="@(()=>softTabs = !softTabs)">TOGGLE SOFT</Button>
@code {
    bool replaceTab = true;
    int tabSize = 4;
    bool softTabs = true;
}
```


David-Moreira:
Haven't tested yet. But from looking at PR looks good.

PR #3018, opened 2021-10-14
JS modules

### LINKED ISSUE #2898: Import JavaScript on-demand with IJSObjectReference
Sources:

- https://docs.microsoft.com/en-us/aspnet/core/blazor/javascript-interoperability/call-javascript-from-dotnet?view=aspnetcore-5.0#javascript-isolation-in-javascript-modules-1
- https://blog.elmah.io/using-js-object-references-in-blazor-wasm-to-wrap-js-libraries/

---

I don't know how I missed this but it seems it is possible to import JS on runtime without needing to manually add `_content` files into the `index.html` or `_Host.cshtml`.

@David-Moreira With this we could split our big JS file into multiple smaller, for modal, date inputs, numerics, etc., and then import them once the component is initialized. We need to see if there are any performance issues. And also how to handle different operations for providers, For example, modal can have `open()` method in Bootstrap, Bulma, and AntDesign.

### LINKED ISSUE #3015: Fix chartjs-plugin-streaming CDN
Should be https://cdn.jsdelivr.net/npm/chartjs-plugin-streaming@1.9.0/dist/chartjs-plugin-streaming.min.js instead of latest version.

stsrki:
resolves #2898 Import JavaScript on-demand with IJSObjectReference
resolves #3015 Fix chartjs-plugin-streaming CDN

This is the first try at the JS modules. I have moved the button into a separate JS file and imported it with `IJSObjectReference`. The modules are abstracted similar to IJSRuntime, but instead of having everything in one giant file, the plan is to separate everything into **modules**.

So far it works well. JSButtonModule is loaded only once by the browser so there should not be any performance cost. But as always I have a couple of concerns.

1. How to properly dispose of JSModule? Since it is registered as scoped it should be removed once the app is closed. Is this a good way of handling it? What about hard closing the browser window?

2. Not all modules are the same. For example, an **IJSModalModule** (not yet created) should be loaded from each provider separately. That means each provider has its own JS implementation. That's why I started with interfacing every module. That way we can register each implementation. The only thing left is to properly handle the `_content` path for each provider.

### DIFF
--- a/Documentation/Blazorise.Docs/Layouts/MainLayout.razor.cs
+++ b/Documentation/Blazorise.Docs/Layouts/MainLayout.razor.cs
@@ -9,8 +9,6 @@ namespace Blazorise.Docs.Layouts
     {
         #region Members
 
-        private Bar topBar;
-
         private bool topbarVisible = false;
 
         #endregion
--- a/Documentation/Blazorise.Docs/Layouts/NewsLayout.razor.cs
+++ b/Documentation/Blazorise.Docs/Layouts/NewsLayout.razor.cs
@@ -12,8 +12,6 @@ namespace Blazorise.Docs.Layouts
     {
         #region Members
 
-        private Bar topBar;
-
         private bool topBarVisible;
 
         private bool disposed;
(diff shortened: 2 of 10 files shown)


stsrki:
Everything in the Blazorise project is now moved into JS modules. As a result, there is no need for `blazorise.js` and `JSRunner` anymore so I have removed them. Docs are also updated. I will continue with extensions.

@David-Moreira you will need to updates tests. I have no strength for them 😅

David-Moreira:
Hello,
Fixed the tests.
But latest Bunit version starts breaking some of the Datagrid's tests, I am not sure why... only happens starting on BUnit 1.2.49. 
1.2.36 preview below seems to work fine.

I'm gonna update the branch to the breaking version, and I'll open an issue over on @egil 's side... See if he can help with it.

PR #3076, opened 2021-10-31
Fix unreleased resources

David-Moreira:
This is to investigate and fix unreleased resources... as per DisposeAsync on base component making it so regular Dispose in Parent component is not called, while trying to keep internal changes to a minimum...

I still can't figure out what's keeping Validations/Validation in memory it's driving my crazy....
image

### DIFF
--- a/Source/Blazorise/Base/BaseValidationResult.cs
+++ b/Source/Blazorise/Base/BaseValidationResult.cs
@@ -61,7 +61,6 @@ namespace Blazorise
             }
         }
 
-
         /// <inheritdoc/>
         protected override void OnParametersSet()
         {
--- a/Source/Blazorise/Base/BaseValidationResult.cs
+++ b/Source/Blazorise/Base/BaseValidationResult.cs
@@ -15,8 +15,6 @@ namespace Blazorise
 
         private Validation previousParentValidation;
 
-        private readonly EventHandler<ValidationStatusChangedEventArgs> validationStatusChangedHandler;
-
         #endregion
 
         #region Constructors
@@ -26,11 +24,6 @@ namespace Blazorise
         /// </summary>
         public BaseValidationResult()
         {
-            validationStatusChangedHandler += async ( sender, eventArgs ) =>
-            {
-                OnValidationStatusChanged( sender, eventArgs );
-                await InvokeAsync( StateHasChanged );
-            };
         }
 
         #endregion
@@ -64,7 +57,7 @@ namespace Blazorise
             DetachValidationStatusChangedListener();
             if ( ParentValidation is not null )
             {
-                ParentValidation.ValidationStatusChanged -= validationStatusChangedHandler;
+                ParentValidation.ValidationStatusChanged -= OnValidationStatusChanged;
             }
         }
 
@@ -75,7 +68,7 @@ namespace Blazorise
             if ( ParentValidation != previousParentValidation )
             {
                 DetachValidationStatusChangedListener();
-                ParentValidation.ValidationStatusChanged += validationStatusChangedHandler;
+                ParentValidation.ValidationStatusChanged += OnValidationStatusChanged;
                 previousParentValidation = ParentValidation;
             }
         }
@@ -84,13 +77,14 @@ namespace Blazorise
         {
             if ( previousParentValidation != null )
             {
-                previousParentValidation.ValidationStatusChanged -= validationStatusChangedHandler;
+                previousParentValidation.ValidationStatusChanged -= OnValidationStatusChanged;
             }
         }
 
         /// <inheritdoc/>
-        protected virtual void OnValidationStatusChanged( object sender, ValidationStatusChangedEventArgs eventArgs )
+        protected virtual async void OnValidationStatusChanged( object sender, ValidationStatusChangedEventArgs eventArgs )
         {
+            await InvokeAsync( StateHasChanged );
         }
 
         #endregion
--- a/Source/Blazorise/Base/BaseValidationResult.cs
+++ b/Source/Blazorise/Base/BaseValidationResult.cs
@@ -35,7 +35,7 @@ namespace Blazorise
         {
             if ( disposing )
             {
-                ReleaseResources();
+                DisposeResources();
             }
 
             base.Dispose( disposing );
@@ -46,13 +46,13 @@ namespace Blazorise
         {
             if ( disposing )
             {
-                ReleaseResources();
+                DisposeResources();
             }
 
             return base.DisposeAsync( disposing );
         }
 
-        private void ReleaseResources()
+        private void DisposeResources()
         {
             DetachValidationStatusChangedListener();
 
--- a/Source/Blazorise/Base/BaseValidationResult.cs
+++ b/Source/Blazorise/Base/BaseValidationResult.cs
@@ -1,5 +1,6 @@
 ﻿#region Using directives
 using System;
+using System.Threading.Tasks;
 using Microsoft.AspNetCore.Components;
 #endregion
 
@@ -41,11 +42,32 @@ namespace Blazorise
         {
             if ( disposing )
             {
-                DetachValidationStatusChangedListener();
+                DisposeResources();
             }
 
             base.Dispose( disposing );
         }
+
+        /// <inheritdoc/>
+        protected override ValueTask DisposeAsync( bool disposing )
+        {
+            if ( disposing )
+            {
+                DisposeResources();
+            }
+
+            return base.DisposeAsync( disposing );
+        }
+
+        private void DisposeResources()
+        {
+            DetachValidationStatusChangedListener();
+            if ( ParentValidation is not null )
+            {
+                ParentValidation.ValidationStatusChanged -= validationStatusChangedHandler;
+            }
+        }
+
 
         /// <inheritdoc/>
         protected override void OnParametersSet()
--- a/Source/Blazorise/Base/BaseValidationResult.cs
+++ b/Source/Blazorise/Base/BaseValidationResult.cs
@@ -31,34 +31,19 @@ namespace Blazorise
         #region Methods
 
         /// <inheritdoc/>
-        protected override void Dispose( bool disposing )
-        {
-            if ( disposing )
-            {
-                DisposeResources();
-            }
-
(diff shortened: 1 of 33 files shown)


David-Moreira:
Simple dumb test of doing through every page with the .net mem diagnoser generates slightly better end result.

**rel0.9.4**
ClickAllPagesRel094

**this branch**
ClickAllPagesFix

Source/Blazorise/Base/BaseValidationResult.cs
```diff
@@ -41,19 +35,40 @@ protected override void Dispose( bool disposing )
...
             {
-                DetachValidationStatusChangedListener();
+                DisposeResources();
             }
 
             base.Dispose( disposing );
         }
 
+        /// <inheritdoc/>
+        protected override ValueTask DisposeAsync( bool disposing )
```
stsrki, line 45: Wait, so we need to have BOTH `Dispose` and `DisposeAsync` on every component that needs to be disposed of?

David-Moreira: It inherits a component with both interfaces... Due to blazor async nature... I have found out that it pretty much always calls DisposeAsync. 

So either that component deals with it... By calling the synchronous version if disposed async...(Which in my opinion makes no sense... Either you dispose or async or sync...) Or the parent deals with it like in this PR. 

That's why I'm advocating on removing async from base component...

> "A good implementation pattern of the IAsyncDisposable interface is to be prepared for either synchronous or asynchronous dispose."

stsrki: What about having just DisposeAsync? In 095?

David-Moreira: It's an idea... I don't know if there's ever a gotcha or any implications. Gotta do some research.

For read later : https://github.com/dotnet/aspnetcore/issues/9960

stsrki:
@David-Moreira test again and if you think this is OK I will merge and then proceed to 095. In there we can do the proper refactor and optimizations if needed.

stsrki:
I cannot agree with myself on this one either 🙄

I just dont want to redo everything on 095. Should we then close and leave it as it is? And do proper dispose on 095?

David-Moreira:
In my opinion, I wouldn't close the PR and leave 094 like this either.
I'd prefer the approach I suggested even if it does not look pretty, at least it's somewhat according to the guidelines... 

If we can't reach a concensus I'd rather then, that we take your proposal of switching to async disposal, at least that is an improvement over what we have now that for the grand majority of cases does not seem to be releasing resources at all.

For 095 we have to fix it definitely even if it's alot of work independently of what we do here... I don't think we should sacrifice the 094 PR in any way because of merging or having more work on 095. 

And we should open a new thread / discuss the changes we should make. Because again, we might have different takes on it. You seem to be advocating to have async on all cases while in my opinion it should be on a per case basis. I honestly just think introducing IAsyncDisposable on the BaseAfterRender is a bad design choice, even the BaseAfterRender itself has no async disposal work to do...

stsrki:
@David-Moreira 

What do you think of this pattern?

```
public class OurBaseAfterRenderComponent : ComponentBase
{
    public void Dispose()
    {
        Dispose( true );
    }

    protected virtual void Dispose( bool disposing )
    {
        if ( !Disposed )
        {
            Disposed = true;

            if ( disposing )
            {
            }
        }
    }

    public async ValueTask DisposeAsync()
    {
        await DisposeAsync( true );

        Dispose( false );
    }

    protected virtual ValueTask DisposeAsync( bool disposing )
    {
        try
        {
            if ( !AsyncDisposed )
            {
                AsyncDisposed = true;

                if ( disposing )
                {
                }
            }

            return default;
        }
        catch ( Exception exc )
        {
            return new( Task.FromException( exc ) );
        }
    }

    protected bool Disposed { get; private set; }

    protected bool AsyncDisposed { get; private set; }
}

public class SomeComponentA : OurBaseAfterRenderComponent, IDisposable
{

}

public class SomeComponentB : OurBaseAfterRenderComponent, IAsyncDisposable
{

}

public class SomeComponentC : OurBaseAfterRenderComponent, IDisposable, IAsyncDisposable
{

}
```

The `OurBaseAfterRenderComponent` is not disposable by itself, but internally it has the dispose code. Then if any component that inherits from it needs to add dispose of pattern we just add `IDisposable`, or `IAsyncDisposable` interface.

We still have the same code and all the benefits but we get the freedom of choosing when to opt-in for disposal.

David-Moreira:
Looks good, but taking into account `IDiposable` is the standardized way of disposing resources, I think our `BaseComponent` should at least implement `IDisposable`, because it does have stuff to release synchronously, meaning that every component that inherits it will successfully dispose of these.

So I would do these changes:
```diff
+public class OurBaseAfterRenderComponent : ComponentBase, IDisposable
{
 public void Dispose()
    {
        Dispose( true );
    }

    protected virtual void Dispose( bool disposing )
    {
        if ( !Disposed )
        {
            Disposed = true;

            if ( disposing )
            {
+            RLSRESOURCESSYNC();
            }
        }
    }

    public async ValueTask DisposeAsync()
    {
        await DisposeAsync( true );

        Dispose( false );
    }

    protected virtual ValueTask DisposeAsync( bool disposing )
    {
        try
        {
            if ( !AsyncDisposed )
            {
                AsyncDisposed = true;

                if ( disposing )
                {
+            RLSRESOURCESSYNC();
                }
            }

            return default;
        }
        catch ( Exception exc )
        {
            return new( Task.FromException( exc ) );
        }
    }

    protected bool Disposed { get; private set; }

    protected bool AsyncDisposed { get; private set; }
}

public class SomeComponentA : OurBaseAfterRenderComponent, IDisposable
{
    protected override void Dispose( bool disposing )
    {
        base.Dispose( disposing );
    }
}

+**//THIS ONE is now equivalent to IDisposable + IAsyncDisposable, which is a guideline from Microsoft when you provide 
+IAsyncDisposable, meaning that you should override the Dispose here providing
+ also a way to release resources sync. Why? Because you can't expect every situation, DI container, framework,
+library, whatever to always be able to dispose this asynchronously, it may only be prepared to dispose of this
+synchronously.**
+But we do expect, that under normal circumstances, Blazor will prioritize IAsyncDisposable like we've researched.
public class SomeComponentB : OurBaseAfterRenderComponent, IAsyncDisposable
{

    protected override void Dispose( bool disposing )
    {
+We Expect this to be async disposed, but still provide a way to dispose resources sync, in case the "one" disposing of 
+this does not support async dispose, or calls sync dispose for some reason.
+DisposeMYResources();
        base.Dispose( disposing );
    }
    
    protected override async ValueTask DisposeAsync( bool disposing )
    {
        await DisposeMYResourcesASync();
        return base.DisposeAsync( disposing );
    }
}

public class SomeComponentC : OurBaseAfterRenderComponent, IDisposable, IAsyncDisposable
{
    protected override void Dispose( bool disposing )
    {
        base.Dispose( disposing );
    }

    protected override ValueTask DisposeAsync( bool disposing )
    {
        return base.DisposeAsync( disposing );
    }
}
```

stsrki:
> Looks good, but taking into account `IDiposable` is the standardized way of disposing resources, I think our `BaseComponent` should at least implement `IDisposable`, because it does have stuff to release synchronously, meaning that every component that inherits it will successfully dispose of these.
> 
> So I would do these changes:
> 
> ```diff
> +public class OurBaseAfterRenderComponent : ComponentBase, IDisposable
> {
>  public void Dispose()
>     {
>         Dispose( true );
>     }
> 
>     protected virtual void Dispose( bool disposing )
>     {
>         if ( !Disposed )
>         {
>             Disposed = true;
> 
>             if ( disposing )
>             {
> +            RLSRESOURCESSYNC();
>             }
>         }
>     }
> 
>     public async ValueTask DisposeAsync()
>     {
>         await DisposeAsync( true );
> 
>         Dispose( false );
>     }
> 
>     protected virtual ValueTask DisposeAsync( bool disposing )
>     {
>         try
>         {
>             if ( !AsyncDisposed )
>             {
>                 AsyncDisposed = true;
> 
>                 if ( disposing )
>                 {
> +            RLSRESOURCESSYNC();
>                 }
>             }
> 
>             return default;
>         }
>         catch ( Exception exc )
>         {
>             return new( Task.FromException( exc ) );
>         }
>     }
> 
>     protected bool Disposed { get; private set; }
> 
>     protected bool AsyncDisposed { get; private set; }
> }
> 
> public class SomeComponentA : OurBaseAfterRenderComponent, IDisposable
> {
>     protected override void Dispose( bool disposing )
>     {
>         base.Dispose( disposing );
>     }
> }
> 
> +**//THIS ONE is now equivalent to IDisposable + IAsyncDisposable, which is a guideline from Microsoft when you provide 
> +IAsyncDisposable, meaning that you should override the Dispose here providing
> + also a way to release resources sync. Why? Because you can't expect every situation, DI container, framework,
> +library, whatever to always be able to dispose this asynchronously, it may only be prepared to dispose of this
> +synchronously.**
> +But we do expect, that under normal circumstances, Blazor will prioritize IAsyncDisposable like we've researched.
> public class SomeComponentB : OurBaseAfterRenderComponent, IAsyncDisposable
> {
> 
>     protected override void Dispose( bool disposing )
>     {
> +We Expect this to be async disposed, but still provide a way to dispose resources sync, in case the "one" disposing of 
> +this does not support async dispose, or calls sync dispose for some reason.
> +DisposeMYResources();
>         base.Dispose( disposing );
>     }
>     
>     protected override async ValueTask DisposeAsync( bool disposing )
>     {
>         await DisposeMYResourcesASync();
>         return base.DisposeAsync( disposing );
>     }
> }
> 
> public class SomeComponentC : OurBaseAfterRenderComponent, IDisposable, IAsyncDisposable
> {
>     protected override void Dispose( bool disposing )
>     {
>         base.Dispose( disposing );
>     }
> 
>     protected override ValueTask DisposeAsync( bool disposing )
>     {
>         return base.DisposeAsync( disposing );
>     }
> }
> ```

Yes, the `RLSRESOURCESSYNC` will be there. I just didn't include it in this example. But obviously, we need it. This new approach I plan to do on 095, while for this PR I will just merge what we already have. If you agree?

PR #3081, opened 2021-10-31
Tabs Lazy Load Content

### LINKED ISSUE #2284: Lazy load of tab content
When loading tabs, would be great to have a lazy load option so tab content is not always executed if the tab is not active.  Common use case is when queries/HTTP/DB queries are executed within a tab.

similar arguments to #1110

David-Moreira:
Closes #2284


Source/Blazorise/Components/Tabs/TabPanel.razor
```diff
@@ -1,5 +1,15 @@
...
-</div>
\ No newline at end of file
+
+@if ( Mode == TabsMode.Default )
+{
+    <div id="@ElementId" class="@ClassNames" style="@StyleNames" @attributes="@Attributes">
+        @ChildContent
+    </div>
+}
+else if ( ( Mode == TabsMode.LazyLoad && Active ) )
```
stsrki, line 10: Can be written as `Mode == TabsMode.Default || ( Mode == TabsMode.LazyLoad && Active )` ?

stsrki: Also, I think checking for `Active` is not right. This means it will render content every time we click on a tab. The idea was to postpone rendering until we click on a tab for the first time. After that, it behaves as usual.

Source/Blazorise/Components/Tabs/Tabs.razor.cs
```diff
@@ -221,6 +221,19 @@ public TabPosition TabPosition
...
+        /// Gets or sets the TabsMode.
+        /// </summary>
+        [Parameter]
+        public TabsMode Mode
+        {
+            get => state.Mode;
+            set
+            {
+                state = state with { Mode = value };
+            }
```
stsrki, line 234: missing `DirtyClasses();`

Source/Blazorise/Components/Tabs/TabPanel.razor
David-Moreira: Oh, we have different take on this then. I do believe it's a good feature to re-evaluate each time. Should we do a 3rd mode?

Source/Blazorise/Components/Tabs/Tabs.razor.cs
David-Moreira: Does it need to Dirty classes here??

Source/Blazorise/Components/Tabs/TabPanel.razor
David-Moreira: > Can be written as Mode == TabsMode.Default || ( Mode == TabsMode.LazyLoad && Active )

😂😂 Yes

David-Moreira: https://github.com/Megabit/Blazorise/issues/2284#issuecomment-835825568
3rd mode for this?

Don't know what to call it yet. So we could call this version, TabsMode.Reload?
And for LazyLoad implement what you said :
> The idea was to postpone rendering until we click on a tab for the first time. After that, it behaves as usual.

Source/Blazorise/Components/Tabs/Tabs.razor.cs
stsrki: You never know. Maybe some provider might need it so it is good to have it. Especially when you update the `state` object.

Source/Blazorise/Components/Tabs/TabPanel.razor
stsrki: LazyReload?

Source/Blazorise/Components/Tabs/TabPanel.razor
```diff
@@ -1,5 +1,15 @@
...
+{
+    <div id="@ElementId" class="@ClassNames" style="@StyleNames" @attributes="@Attributes">
+        @ChildContent
```
stsrki, line 13: I would wrap `@ChildContent` with an `if` statement. That way a `div` will still render(without content), but it is a good approach in case someone wished to have a CSS selector targeting a specific child. Or maybe they want to have a tab with a certain size occupying the page.

Source/Blazorise/Components/Tabs/Tabs.razor.cs
David-Moreira: I see, OK.

Source/Blazorise/Components/Tabs/TabPanel.razor
David-Moreira: Nice idea.

Source/Blazorise/Components/Tabs/TabPanel.razor.cs
```diff
@@ -46,6 +52,14 @@ protected override void OnInitialized()
             base.OnInitialized();
         }
 
+        /// <inheritdoc/>
+        protected override Task OnParametersSetAsync()
+        {
+            if ( Active )
+                lazyLoaded = ( Mode == TabsMode.LazyLoad );
```
stsrki, line 59: Since this is placed in `OnParametersSetAsync` isn't it's going to be executed every time a page has refreshed?

David-Moreira: Due to cascaded state should be pretty much every interaction with tabs. At least from my testing.
Why? This is negligible work.

stsrki: Yeah, it will work. It just felt weird at first.

PR #3082, opened 2021-11-01
DataGrid non mandatory field parameter

### LINKED ISSUE #3079: DataGrid : Permit DataGridColumn usage without specifying Field Parameter
DataGrid : Permit DataGridColumn usage without specifying Field Parameter

Will allow to have columns that do not need to have a mandatory associated field/property on the Model.

David-Moreira:
Still WIP...
Closes #3079


David-Moreira:
Working fine from my tests. 
Still need to do docs.  Need your opinion on something.
Currently if user does not provide Field it still works quite well even with Validator, and the user will get the values like this:
Since there's no Field, the ElementId is now the column key for these cases... What do you think??
image
Also the user will have to handle the save somehow... With Field & UseInternalEditing set to true, we would save the value to the TItem correspondent Property/Field, but without Field....Should we save the value internally? Or up to the user to deal with it completly? What do you think here?

stsrki:
Can we just not assign anything to a column if `Field` is not unspecified? And skip any internal saving if there is no `Field`? The sole purpose of Field'less column is for it to be on its own. The users can do whatever they want basically.

I think even with ElementId as an identifier it would be difficult for users to distinguish them, being they are autogenerated.

Source/Blazorise/Components/Validation/EditContextValidator.cs
```diff
@@ -150,44 +150,48 @@ public virtual void ValidateField( EditContext editContext, ValidationMessageSto
...
-
-                // This is used only for custom localization. We assume that unformatted ErrorMessage will be
-                // used as an localization key, so we need to replace it in case it is undefined with
-                // the internal ErrorMessageString that has unformatted message. eg. "The field {0} is invalid."
-                var validationAttributes = ValidationAttributeHelper.GetValidationAttributes( propertyInfo );
-                var formattedValidationAttributes = ValidationAttributeHelper.GetValidationAttributes( propertyInfo );
+                var cacheKey = (ModelType: fieldIdentifier.Model.GetType(), fieldIdentifier.FieldName);
 
-                if ( forLocalization )
+                if ( !propertyInfoCache.TryGetValue( cacheKey, out validationPropertyInfo ) )
```
stsrki, line 158: Hmm, I must test this. Caching might be tricky. There is a risk of getting the field reference to an older model instance.

stsrki: Tested and seems to be working fine.

David-Moreira:
> Can we just not assign anything to a column if `Field` is not unspecified? And skip any internal saving if there is no `Field`? The sole purpose of Field'less column is for it to be on its own. The users can do whatever they want basically.
> 
> I think even with ElementId as an identifier it would be difficult for users to distinguish them, being they are autogenerated.

At the moment it does not do any internal saving yes. And we could remove it all together from the key/values info we provide on update/add to the user. Think we should? Users could still find this usefull I guess? 
ElementId can be set manually right? So user could generate his ids and handle appropriately.
Yes, it's still most up to the user if he so chooses to ignore Field as Datagrid can't do much else than this.

stsrki:
Well, yes, they can define ElementIt on a column. But even so, it would be good not to include those columns in the saving process. Let's say a user just sends the edited key-value pairs to their backend. In this scenario, they would need to filter the named columns, while if they are not provided or left empty they can easily exclude them.

David-Moreira:
It's ready. Can you take a look. :)

stsrki:
Looking good to me.

I took time to clean the projects a little as it becomes too crowded lately. So I moved some objects into a separate folder/files.

PR #3094, opened 2021-11-04
Reduce memory foot print from components in memory

David-Moreira:
Like we've talked, let's at least try to reduce the memory footprint caused by components that are kept in memory.
**Before**
Builder
**After**
BuilderAfter


Source/Blazorise/Base/BaseComponent.cs
```diff
@@ -142,6 +143,22 @@ protected override async Task OnAfterRenderAsync( bool firstRender )
         protected virtual Task OnFirstAfterRenderAsync()
             => Task.CompletedTask;
 
+        /// <inheritdoc/>
+        protected override void Dispose( bool disposing )
+        {
+            ClassBuilder = null;
```
stsrki, line 149: Are there really any benefits in setting the member instances to null? According to this https://blog.stephencleary.com/2010/02/q-should-i-set-variables-to-null-to.html, and other sources I'm reading, it should be avoided.

Source/Blazorise/Base/BaseInputComponent.razor.cs
```diff
@@ -89,21 +89,37 @@ protected override void Dispose( bool disposing )
...
+            if ( disposing )
+            {
+                ReleaseResources();
             }
 
-            base.Dispose( disposing );
+            return base.DisposeAsync( disposing );
+        }
+
+        protected void ReleaseResources()
```
stsrki, line 109: The `BaseInputComponent` is marked only with `IDisposable`. So I don't think we need to overload `DisposeAsync`.

stsrki: Hmmm, actually I understand what you wanted to. If any of the sub-components has an `IAsyncDisposable` then it would pass through both methods. Yeah, makes sense.

Source/Blazorise/Utilities/ClassBuilder.cs
```diff
@@ -94,7 +94,7 @@ public string Class
             {
                 if ( dirty )
                 {
-                    builder = new();
+                    builder.Clear();
```
stsrki, line 97: Good optimizations. According to the online benchmarks, it seems this is a little faster than `new().` To be honest I didn't know about it.

Source/Blazorise/Base/BaseComponent.cs
David-Moreira: I haven't read the blog yet. But generally you don't need to do it, yes.
But in thsi case. Since we have the problem that the components are being kept in memory it won't release these At least from my tests... See prints...

Source/Blazorise/Base/BaseInputComponent.razor.cs
David-Moreira: Exactly... 🤭

Source/Blazorise/Base/BaseComponent.cs
stsrki: Just wanted to confirm if you're aware of it. I will merge.

PR #3101, opened 2021-11-06
DataGrid RowSelectable EventArgs

### LINKED ISSUE #3092: DataGrid RowSelectable EventArgs
Hello, 
We have an application where we are utilizing datagrid with a checkbox (DataGridMultiSelectColumn). We intentionally wanted row selection to be explicit via checkbox by setting :
RowSelectable="@(_=>false)"
When upgrading to 0.9.4.7 from 0.9.3.8 we noticed this breaks the ability to use the checkboxes to select the rows, and therefore nothing is selectable.

David-Moreira:
Closes #3092


stsrki (commented):
Looks good to me. Other than missing comments I cannot find any issue. Tomorrow I will run it and test it.

Source/Blazorise/Components/Link/Link.razor
```diff
@@ -1,5 +1,6 @@
 ﻿@namespace Blazorise
 @inherits BaseComponent
+@implements IDisposable
```
stsrki, line 3: We already inherit IDisposable in code behind so this is not needed.

PR #3102, opened 2021-11-07
Fix Autocomplete inside modal loosing focus on dropmenu scroll click

### LINKED ISSUE #3069: Autocomplete dropdown in modal dialog closes when scroll bar used (repro available)
When an autocomplete is used in a modal dialog and the scroll bar of the autocomplete dropdown is clicked on, the dropdown closes before any selection can be made.

this seems to only be when a dialog is used, otherwise this does not occur.

scroll wheel does work but if trying to use the mouse pointer to scroll down there is a problem.

A repro is available at:
https://github.com/robalexclark/BlazoriseAutocompleteInModalBug

Instructions:
1) Click on Setup-> Colonies
2) Click on Open Picker
3) In the dropdown start searching using  AAA-  to get the dropdown to open.
4) click on the scroll bar with the mouse pointer, the dropdown will close prematurely.

David-Moreira:
Closes #3069

Removing it from the DIV does fix the problem. Don't see any problem with removing it. Tabbing keeps working normally. 
Also applying -1 seems pretty close to not having the attribute set at all, both ways make it so it's not tab targetable.
https://developer.mozilla.org/pt-BR/docs/Web/HTML/Global_attributes/tabindex

### DIFF
--- a/Source/Blazorise.Bootstrap/Modal.razor
+++ b/Source/Blazorise.Bootstrap/Modal.razor
@@ -1,7 +1,7 @@
 ﻿@inherits Blazorise.Modal
 <CascadingValue Value="@this" IsFixed="true">
     <CascadingValue Value="@State">
-        <div @ref="@ElementRef" id="@ElementId" class="@ClassNames" style="@StyleNames" tabindex="-1" role="dialog" @attributes="@Attributes">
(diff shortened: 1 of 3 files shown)


stsrki (approved):
As I mentioned before, it's hard to tell if there are going to be any side effects. In case there is, we will figure it later. So I will merge it in good faith :)

PR #3125, opened 2021-11-17
JS relative paths

### LINKED ISSUE #3122: JSInterop Exception using Tooltip in v0.9.5
**Describe the bug**
When using the Tooltip component on v0.9.5, an error is thrown in the console and the Blazor Server application crashes:
```
Loading module from “https://localhost:5001/_content/Blazorise/tooltip.js” was blocked because of a disallowed MIME type (“”).
Error: Microsoft.JSInterop.JSException: error loading dynamically imported module
   at async ValueTask<TValue> Microsoft.JSInterop.JSRuntime.InvokeAsync<TValue>(long targetInstanceId, string identifier, object[] args)
   at async ValueTask Blazorise.Modules.JSTooltipModule.Initialize(ElementReference elementRef, string elementId, object options)
   at async void Blazorise.Tooltip.OnInitialized()+(?) => { }
   at async Task Blazorise.BaseAfterRenderComponent.OnAfterRenderAsync(bool firstRender)
   at async Task Blazorise.BaseComponent.OnAfterRenderAsync(bool firstRender)
   at async Task Microsoft.AspNetCore.Components.RenderTree.Renderer.GetErrorHandledTask(Task taskToHandle)
```
I have tried in current versions of both Firefox and Chrome, with both Blazorise.Bootstrap and Blazorise.Material. All combinations work under v0.9.4.8 (and earlier), and throw an error under v0.9.5.

**To Reproduce**
Steps to reproduce the behaviour:
1. Have a project using the `<Tooltip>` component working correctly under v0.9.4.8 (possibly not in the root path, see below).
2. Upgrade the appropriate Blazorise package (e.g. Blazorise.Bootstrap) to v0.9.5.
3. Start the project and navigate to a page showing the tooltip.
4. See the error in the console.
5. Downgrade to v0.9.4.8, it starts working again.

**Expected behaviour**
The tooltip should display in the browser without throwing any error or severing the websocket connection to Blazor Server.

**Additional context**
I don't know whether this is relevant, but my Blazor Server application resides in a subdirectory. There is an ASP.NET API served on `https://localhost:5001` and handled by controllers, while Blazor is running on `https://localhost:5001/admin`.

This setup has been working perfectly with previous versions of Blazorise, but I mention it because the invalid MIME type error displayed immediately before the exception does not include that custom path. That is, it fails to load `https://localhost:5001/_content/Blazorise/tooltip.js`, which makes sense, because the file is at `https://localhost:5001/admin/_content/Blazorise/tooltip.js`. Although I'm not sure why the error is that the module has an invalid MIME type, instead of that it doesn't exist...

stsrki:
closes #3122 JSInterop Exception using Tooltip in v0.9.5

### DIFF
--- a/Source/Blazorise.AntDesign/wwwroot/tooltip.js
+++ b/Source/Blazorise.AntDesign/wwwroot/tooltip.js
@@ -1,4 +1,4 @@
-import { initialize as baseInitialize, destroy, updateContent } from "/_content/Blazorise/tooltip.js";
+import { initialize as baseInitialize, destroy, updateContent } from "../Blazorise/tooltip.js";
 
 export function initialize(element, elementId, options) {
     baseInitialize(element, elementId, options);
--- a/Source/Blazorise.Bootstrap/wwwroot/modal.js
+++ b/Source/Blazorise.Bootstrap/wwwroot/modal.js
@@ -1,4 +1,4 @@
-﻿import { addClassToBody, removeClassFromBody } from "/_content/Blazorise/utilities.js";
+﻿import { addClassToBody, removeClassFromBody } from "../Blazorise/utilities.js";
 
 export function open(element, scrollToTop) {
     // adjust modal and page padding BEFORE modal is shown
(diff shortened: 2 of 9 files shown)


David-Moreira (approved):
yep lgtm

PR #3140, opened 2021-11-18
Extract hex digits cache regex

### LINKED ISSUE #3133: ExtractHexDigits high CPU Usage + Time
Using Theme Provider slows down the whole application. This is caused by ExtractHexDigits being called multiple 1000 times.
Taking up to 3500ms for 2500 calls.

Use the method:
https://github.com/Megabit/Blazorise/blob/e0f1f9a641114705d127ddb6b4d60dac143e60f1/Source/Blazorise/Themes/ThemeGenerator.cs#L1317

Fill it with any hex color string and run it 2500 times in a loop.
This will take arround 2600ms.

2500 runs taking only couple of ms.

Simple fix to reduce the time and usage drasticly:
```
private static Regex isHexDigit = new( "[abcdefABCDEF\\d]+", RegexOptions.Compiled );

    protected static string ExtractHexDigits( string input )
    {
        // remove any characters that are not digits (like #)
        string newnum = "";
        foreach ( char c in input )
        {
            if ( isHexDigit.IsMatch( c.ToString() ) )
                newnum += c.ToString();
        }
        return newnum;
    }
```
Declaring the same regex each time the function is called slows it down.

David-Moreira:
Closes #3133 

Didn't have time to finish yet.
Need to test Regex.Matches vs our version as I am not 100% it's 100% logically equivalent. If it is, then it's way faster.

image

Also need to do some reading up on https://docs.microsoft.com/en-us/dotnet/standard/base-types/best-practices as static method `Regex.IsMatch` might also be worth considering...
**Edit:** I'm not sure if additional steps need to be taken to actually take advantage of `RegexOptions.Compiled` ?

Also consider using Span in the future to do internal optimizations in string manipulation... might be worth it for 1.0.0
A good resource: 
https://www.youtube.com/watch?v=FM5dpxJMULY&ab_channel=NickChapsas

### DIFF
--- a/Source/Blazorise/Themes/ThemeGenerator.cs
+++ b/Source/Blazorise/Themes/ThemeGenerator.cs
@@ -1313,7 +1313,7 @@ namespace Blazorise
         /// <summary>
         /// Checks for characters that are Hexadecimal
         /// </summary>
-        protected static Regex isHexDigit = new( "[abcdefABCDEF\\d]+", RegexOptions.Compiled );
+        protected static Regex IsHexDigit = new( "[abcdefABCDEF\\d]+", RegexOptions.Compiled );
 
         /// <summary>
         /// Extract only the hex digits from a string.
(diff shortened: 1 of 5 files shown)


stsrki (approved):
LGTM

When changing the theme it feels faster now. Maybe it is subjective I don't know.

Regarding the use of Span. I thought so too already, but I don't have a clear use-case for it. If you want to make use of it just create a new ticket for later,

David-Moreira:
I felt like it was faster too haha but I didn't bother to go compare both versions side by side haha. 
We'll eventually figure out some ways to benchmark the full rendering of the pages so we can have tangible numbers. 

I think every internal string manipulation can use it. You have some substringing in the theme generator for example. It should be a bit faster and without allocations, we'll do it once we clear up out the big stuff in the release.

PR #3147, opened 2021-11-20
Modal retain css animation

### LINKED ISSUE #2893: Hiding Modal with animation
**Describe the bug**
I am unable to hide the modal with an animation. 

**To Reproduce**
1. Add css
```
@keyframes slide-in-right {
    0% { right: -100%; }
    100% { right: 0; }
}

@keyframes slide-out-right {
    0% { right: 0; }
    100% { right: -100%; }
}

.modal .modal-dialog {
     animation: slide-out-right 0.5s;
     right: -100%;
}

.modal.show .modal-dialog {
    animation: slide-in-right 0.5s;
    right: 0;
}
```
2. Open modal
2. Close modal

**Expected behavior**
The modal opens and closes with an animation. Manually adding and removing the `show` class on the modal will show the expected behaviour.

**Additional context**
I am using the bootstrap provider and it looks like they use a special check for the [`fade` class and defers the closing until the animation is complete](https://github.com/twbs/bootstrap/blob/v4.6.0/js/src/modal.js#L168), I guess something similar needs to happen for the css animation to complete before the `Visibility` and/or `display: block` is removed. 

Applogies if this is not clear, I am a little unsure of css animations in general

David-Moreira:
Closes #2893


stsrki:
Here are some things I noticed, compared to the native behavior.

- BS5 animation works on open and on close. This is OK.
- BS4 animation works only on open. Close is not animated. Not OK.
- Bulma, I didn't notice anything different. This is OK.
- Material animation works on open. On close, it doesn't work but since it is not working on their own examples I would say this is OK.
- AntDesign animation doesn't work on open or on close. Not OK.

So I think you only need to fix BS4 closing animation and AntDesign open and close.

Works really really good I must admit, so far. Good job!

Source/Blazorise/Components/Modal/Modal.razor.cs
```diff
@@ -313,26 +313,53 @@ protected virtual void HandleVisibilityStyles( bool visible )
...
+        /// <summary>
+        /// Handles the modal styling in a way that still triggers css animations.
+        /// </summary>
+        /// <param name="visible"></param>
+        /// <returns></returns>
+        private async Task ModalDirtyStylingAnimation( bool visible )
+        {
+            var hasFadeAnimation = !string.IsNullOrWhiteSpace( ClassProvider.ModalFade() );
+
+            if ( visible )
```
stsrki, line 328: Shouldn't both DirtyStyles and DirtyClasses be called? I don't we can assume that all providers would control animation with styles and classes in the same order.

Source/Blazorise/Components/Modal/Modal.razor.cs
```diff
@@ -313,26 +313,53 @@ protected virtual void HandleVisibilityStyles( bool visible )
...
+                DirtyStyles();
+            else
+                DirtyClasses();
+
+            if ( hasFadeAnimation )
+            {
+                await InvokeAsync( StateHasChanged );
+                await Task.Delay( 150 );
```
stsrki, line 336: Maybe we can add animation duration as a parameter?

David-Moreira:
Bs4 can you clear cache and try again, I removed css. 
Bulma didn't change as their native modal does not have animation. I left our current behaviour. 
I think material clear cache also? 
Ant design I don't think I made any change at all. Should be working the same. At least it was when I tested.

Source/Blazorise/Components/Modal/Modal.razor.cs
David-Moreira: The order is changed according to visibility because of style block or none that would stop the animation.

David-Moreira: Srsly doubt it will affect. Otherwise we can just make the logic focus on the display style.

Source/Blazorise/Components/Modal/Modal.razor.cs
David-Moreira: Sure. Not sure how much boiler plate needed for that yet.

stsrki:
Hmm BS4 seems to be working now. Probably it was something cached as you said. That leaves only AD not working.

David-Moreira:
What do you think? Just the interfaces, for the sake of having common api/named methods?

Can't think of an abstraction to do on top... 
Maybe an adapter that could detect if component is `ICloseable` or `IAnimated` and act accordingly. But then `IAnimated` would need to have more methods like `BeforeAnimation` `AfterAnimation`. 
So you could have the adapter call into these if `IsAnimated` or just call regular `Open` / `Close` if not animated.

But honestly I'm not sure it's worth all the boilerplate / trouble? What do you think?
Something similar to this...? O.o 
```
public async Task Open()
            { 
                if (this.component is IAnimatedComponent animated)
                {
                    if (animated.IsAnimated)
                    {
                        await animated.BeforeAnimation();
                    
                        await Task.Delay( this.AnimationDuration );
                        await animated.Open();

                        await animated.AfterAnimation();
                    }
                     else
                    {
                        await this.component.Open();
                    }
                }
                 else
                 {
                        await this.component.Open();
                  }
            }
```

stsrki:
Yeah, I think having `ICloseable` or `IAnimated` and separated would make it best. `IAnimated` could hold the reference to `ICloseable` and then inside of it have something like `BeforeAnimation` and `AfterAnimation` that would open and close it. Or vice versa, not sure at the moment until we(you) try it.

We can then reuse it for another closable component.

David-Moreira:
If you don't want just the interfaces we can test then.

David-Moreira:
Ok, so what do you think of the approach?

David-Moreira:
Material and AntDesign both don't use the libraries original animations it seems.
However seems complicated as they use js. At the moment they have a custom animation only for Fading in.
Do you think it's worth it extracting the same animations as in the lbiraries? 
Or just add a fadeout animation? Or similar to bs?

David-Moreira:
Changes look good.

Keeping this gives more flexibility, in doing a specific Show or Hide operation? Altough you're right, it can just be done on the Begin / End Animations according to the visible flag also.
image

- Still missing a way to configure the Animation Duration as a Parameter and propagate it to the css, also might be worth the IsAnimated to also be a Parameter and the user can Enable/Disable?

Also, you didn't answer what you think about Material and AntDesign. :P

stsrki:
> Also, you didn't answer what you think about Material and AntDesign

Well, the Material is working. And we also need to implement AntDesign animations. Or if it is too hard just get back the CSS fade animations.

David-Moreira:
It's working as it was before?

We can try to copy.

stsrki:
> It's working as it was before?

The material worked last time I tried. Not the same as before but the same as it is working on djibe documentation. So, this is OK.

---

AntDesign doesn't work. What I meant, we need to revert to CSS keyframe animations if their native animation cannot be done.

David-Moreira:
I'm pretty sure I left both working as before... At least that's what I intended. I'll take a look.

David-Moreira:
I just checked the live demos.
https://antdesigndemo.blazorise.com/tests/modals
https://materialdemo.blazorise.com/tests/modals
Behaviour remains the same.

I'll check if js can be replicated.

David-Moreira:
@stsrki It's working as before. There's an open animation and not a close animation.
What if we just add the close animation similar to how the current open fade animation was already added?

stsrki:
It's working well. If you don't have anything else to add I will merge it?

PR #3152, opened 2021-11-22
Support types without object equal operator ( extend test for DateTime, TimeSpan )

puschie286:
Implement compare for value types.

Add test for DateTime and TimeSpan


David-Moreira:
So basically this is to access value types within structs? Is that it? 
Can you add more tests for other value types within the struct just to make sure everything works fine?

@stsrki are we doing net6 only on 1.0.0? Maybe we can think about adding the new Date Types in .NET6 to the tests also.

@puschie286 why `Activator` over `Expression.Default`?  Would the later not make more sense and be more performant in these cases? Could you please test?

puschie286:
@David-Moreira its for comparing struct that doesnt have a equal operator which accept objects.
Express.Constant( null ) is a object so it fail for some struct and Expression.Default / Activator always create a valid instance. 
But im not 100% sure if this is how it works internaly - we use a lot of TimeSpan objects, which caused the expression to fail, so this is what i came up with^^

you are right with Expression.Default, thanks for pointing that out :)

replaced the reflection approach with your suggested expression approach

David-Moreira:
Way simpler change, makes more sense. If all tests are passing, LGTM.
(Do take note, that you're merging this in for 1.0.0, this change would only be live in a few months.)

@stsrki  wanna consider this for 0.9.5? Seems worthy honestly.

stsrki:
@David-Moreira Yeah, it is not breaking any API so it can be done as part of 095. LGTM

PR #3155, opened 2021-11-23
Fix Table THead Query Selector to target only first THead

### LINKED ISSUE #3151: TableHeaderCell in DetailRowTemplate of virtualized DataGrid is shifted
**Describe the bug**
When a table is added as the DetailRowTemplate in a DataGrid, the header row is shifted down by an X amount of px

**To Reproduce**
1. Add a virtualized DataGrid with a table in row details.
2. When the row is selected, show the details row
3. Row will be shifted down

**Example**
https://github.com/mtbayley/BlazoriseTableBug

**Screenshots**

image

image

image

**Additional context**

As a workaround, remove TableRow from table header. This does not seem to cause any issues except that it does not follow the documentation. OR, do not use a virtualized DataGrid

David-Moreira:
Closes #3151

### DIFF
--- a/Source/Blazorise/wwwroot/table.js
+++ b/Source/Blazorise/wwwroot/table.js
@@ -11,7 +11,8 @@
     }
 
     function resizeHandler(element) {
-        const tableRows = element.querySelectorAll("thead tr");
+        const thead = element.querySelector("thead:first-child");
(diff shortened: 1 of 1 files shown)


stsrki (approved):
LGTM

PR #3159, opened 2021-11-24
Lazy load JS modules

### LINKED ISSUE #3150: Button Issue Migrating 0.9.4.4 on RC 1 to 0.9.5 on .NET 6.0 RTM
I used https://blazorise.com/news/release-notes/095 to do the migration. I see the following error on navigating to a page that has Blazorise components on it:

```
blazor.webassembly.js:1 
        
       crit: Microsoft.AspNetCore.Components.WebAssembly.Rendering.WebAssemblyRenderer[100]
      Unhandled exception rendering component: Cannot read properties of null (reading 'type')
      TypeError: Cannot read properties of null (reading 'type')
          at Module.initialize (https://localhost:5001/_content/Blazorise/button.js?v=0.9.5.0:6:17)
          at https://localhost:5001/_framework/blazor.webassembly.js:1:3332
          at new Promise (<anonymous>)
          at Object.beginInvokeJSFromDotNet (https://localhost:5001/_framework/blazor.webassembly.js:1:3306)
          at Object.Rt [as invokeJSFromDotNet] (https://localhost:5001/_framework/blazor.webassembly.js:1:59738)
          at _mono_wasm_invoke_js_blazor (https://localhost:5001/_framework/dotnet.6.0.0.tj42mwroj7.js:1:193780)
          at wasm://wasm/00945abe:wasm-function[2588]:0x8065b
          at wasm://wasm/00945abe:wasm-function[2307]:0x689a7
          at wasm://wasm/00945abe:wasm-function[7015]:0x17bac4
          at wasm://wasm/00945abe:wasm-function[6830]:0x172bc5
Microsoft.JSInterop.JSException: Cannot read properties of null (reading 'type')
TypeError: Cannot read properties of null (reading 'type')
    at Module.initialize (https://localhost:5001/_content/Blazorise/button.js?v=0.9.5.0:6:17)
    at https://localhost:5001/_framework/blazor.webassembly.js:1:3332
    at new Promise (<anonymous>)
    at Object.beginInvokeJSFromDotNet (https://localhost:5001/_framework/blazor.webassembly.js:1:3306)
    at Object.Rt [as invokeJSFromDotNet] (https://localhost:5001/_framework/blazor.webassembly.js:1:59738)
    at _mono_wasm_invoke_js_blazor (https://localhost:5001/_framework/dotnet.6.0.0.tj42mwroj7.js:1:193780)
    at wasm://wasm/00945abe:wasm-function[2588]:0x8065b
    at wasm://wasm/00945abe:wasm-function[2307]:0x689a7
    at wasm://wasm/00945abe:wasm-function[7015]:0x17bac4
    at wasm://wasm/00945abe:wasm-function[6830]:0x172bc5
   at Microsoft.JSInterop.JSRuntime.<InvokeAsync>d__16`1[[Microsoft.JSInterop.Infrastructure.IJSVoidResult, Microsoft.JSInterop, Version=6.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60]].MoveNext()
   at Microsoft.JSInterop.JSObjectReferenceExtensions.InvokeVoidAsync(IJSObjectReference jsObjectReference, String identifier, Object[] args)
   at Blazorise.Modules.JSButtonModule.Initialize(ElementReference elementRef, String elementId, Object options)
   at Blazorise.Button.<OnInitialized>b__12_0()
   at Blazorise.BaseAfterRenderComponent.OnAfterRenderAsync(Boolean firstRender)
   at Blazorise.BaseComponent.OnAfterRenderAsync(Boolean firstRender)
   at Microsoft.AspNetCore.Components.RenderTree.Renderer.GetErrorHandledTask(Task taskToHandle, ComponentState owningComponentState)
```

Funny thing is, the buttons do work, and on navigating to another page & then back this error doesn't come up again. Sadly this message doesn't give (at least) me a clue where to start looking - do you have any pointers for me what might be going on?

The app started on 5.0 and I migrated to 6 during the previews once the upload limits (JS interop limits) were lifted. All the time I also upgraded Blazorise along the way.

stsrki:
closes #3150

Adds the lazy load support for JS modules. It should work better when running along with the async code.


David-Moreira:
I don't have time to test this today. But this seems equivalent to the previous version? Even if you had the Lazy, it should be working the same? Am I missing smthg?

stsrki:
> I don't have time to test this today. But this seems equivalent to the previous version? Even if you had the Lazy, it should be working the same? Am I missing smthg?

It was a guess since I cannot reproduce it.

stsrki:
I think my last comment where I check the button `element` is not null has fixed the issue. But I still think we should keep the Lazy loading changes. @David-Moreira Do you agree?

David-Moreira:
I think it's okay to deploy it with either of the ways. But Let me do some testing later today. Unless you want to go with a release right away.

stsrki:
I will rather wait. If all is good then I will release it tomorrow.

David-Moreira:
We've gone back to have race conditions on the js calls.
If you switch very quickly between pages you'll eventually get it. (I'm guessing it might be more noticeable on worst connections on server or slower processing devices on wasm)

I believe this might be because we ended up fixing Dispose? Since it's now calling dispose correctly... now these race conditions are back.
To solve this:
- Maybe null checking every dotnet object reference in js?
- Delay dotnet object reference disposal by a few ms?
- Is this such a low happening that we could swallow the exception?
- Any "Cancellable work" patterns (similar to async Cancel token) out there that we can do in js?
- Would not explicitly setting `dotNetObjectRef = null;` be helpful? Since the GC should eventually get rid of it anyway, null or not null. (At least under normal circumstances, if it thinks no component has a hold of it.)

If you figure it out in the meantime great! Otherwise I can help you on the weekend.

image
image
image

stsrki:
I tried to fast switch between pages but maybe my PC is faster.

---

- _Maybe null checking every dotnet object reference in js?_ Yes, we can do that.
- _Delay dotnet object reference disposal by a few ms?_ NO, I don't want this to slow down the app in any way if possible.
- _Is this such a low happening that we could swallow the exception?_ What do you mean? `try catch` block?
- _Any "Cancellable work" patterns (similar to async Cancel token) out there that we can do in js?_ Really don't know 🤔
- `Would not explicitly setting dotNetObjectRef = null; be helpful? Since the GC should eventually get rid of it anyway, null or not null. (At least under normal circumstances, if it thinks no component has a hold of it.)` We can set them to null if we didn't already.

David-Moreira:
Some pages are easier. Try between table / forms pages. I tried with blazor.server so there shouldn't be much diff in processing between our pcs? It's not like my pc is a stone! 😅

stsrki:
I'm still unable to reproduce. Clicking as fast as I can between pages. Even enabled Slow 3G.

image

David-Moreira:
- Maybe null checking every dotnet object reference in js? Yes, we can do that.
R: OK.
Delay dotnet object reference disposal by a few ms? NO, I don't want this to slow down the app in any way if possible.
Is this such a low happening that we could swallow the exception? What do you mean? try catch block?
R: YEA
Any "Cancellable work" patterns (similar to async Cancel token) out there that we can do in js? Really don't know 🤔
R: haha
Would not explicitly setting dotNetObjectRef = null; be helpful? Since the GC should eventually get rid of it anyway, null or not null. (At least under normal circumstances, if it thinks no component has a hold of it.) We can set them to null if we didn't already.
R: Oh, I meant removng the dotNetObjectRef = null instead.

stsrki:
@David-Moreira So what did we agree on in the end. This is OK to merge?

PR #3173, opened 2021-11-25
Chart v3

### LINKED ISSUE #2229: Update Chart.JS to v3
https://www.chartjs.org/docs/latest/getting-started/v3-migration.html

stsrki:
closes #2229

Done:

- Adjusted DataSets and Options for new structure(breaking change)
- All Data and Options are made nullable
- Every field is marked with `System.Text.Json` attributes
- Removed custom serialized in favor of native `System.Text.Json`
- Support for callback serialization of JS function and compiling then back on client-side

New:

- Bubble chart
- Scatter chart


Demos/Blazorise.Demo/Pages/Tests/ChartsPage.razor.cs
```diff
@@ -19,14 +19,11 @@ public partial class ChartsPage
...
+                Y = new()
                 {
-                    new()
+                    Ticks = new AxisTicks
                     {
-                        Ticks = new AxisTicks
-                        {
-                            Callback = ( value, index, values ) => $"{value / 1000} K"
-                        }
+                        Callback = ( value, index, values ) => value / 1000 + "K"
```
David-Moreira, line 26: Actually the previous way explicits tells .NET it's a string in there, right?
This way calls in an operator with the int value boxed, probably more costly? These things are still neglible eitherway.. :P
image

Source/Extensions/Blazorise.Charts/LambdaConverter.cs
```diff
@@ -0,0 +1,139 @@
...
+            sb.Append( ") {" );
+
+            sb.Append( "return " );
+            sb.Append( jsBody );
+            sb.Append( "; }" );
+
+            return sb.ToString();
+        }
+
+        static string BuildFunctionParameters( Expression<T> value )
```
David-Moreira, line 43: Should this just take in the existing SB instead of making a new one?

Source/Extensions/Blazorise.Charts.Streaming/wwwroot/charts.streaming.js
```diff
@@ -1,13 +1,22 @@
...
+
 export function initialize(dotNetAdapter, canvas, canvasId, vertical, streamOptions) {
     const chart = getChart(canvasId);
 
     if (chart) {
-        const scalesOptions = getStreamingOptions(dotNetAdapter, vertical, chart.options, streamOptions);
+        // In Chart v3 options now containt all kind of objects and functions. Thats why we need to deep
+        // clone options without getting the reference to any running object or function. Otherwise we
+        // will run into recursion errors.
+        const options = JSON.parse(JSON.stringify(chart.options));
```
David-Moreira, line 14: Might be worth as an internal js util named DeepClone.

David-Moreira:
https://localhost:5001/tests/charts
Click Change Type

image

David-Moreira:
Horizontal bar does not seem to work with more than a data set.
Vertical bar seems weird, is this the expected behaviour?
image

David-Moreira:
Is this still true?
image

David-Moreira:
Empty chart 
image

David-Moreira:
Do these remain the same? Should we update the versions to the current latest?
image

Demos/Blazorise.Demo/Pages/Tests/ChartsPage.razor.cs
stsrki: This will be compiler and run in JS, never in C#. That's why an expression needs to be as simple a possible.

Source/Extensions/Blazorise.Charts/LambdaConverter.cs
stsrki: Yeah it can.

PR #3180, opened 2021-11-28
Data grid filter post form fix

### LINKED ISSUE #3178: Data grid filter causes page to reload when enter is pressed and only one column is filterable
**Describe the bug**
Data grid filter causes page to reload when enter is pressed and only one column is filterable. If you have more than 1 column filter, it works fine. If there is only one column filter, the page will reload to /?.

**To Reproduce**
Clone this repo and run...
https://github.com/mtbayley/BlazoriseTableBug/tree/FilterReload

**Expected behavior**
There should not be a page reload when the enter key is pressed

**Screenshots**
Animation

David-Moreira:
Closes #3178 

Was thinking in introducing a PreventDefaultOnEnter Parameter to TextEdit, but since we don't make api changes, went with this solution to fix it. We could however do that for next release, if you'd like.

The commit that introduced "SubmitFormOnEnter" on Datagrid: a80e91c9793c31961162bae27a9101f6da054


stsrki:
I don't like this approach that much. It feels like a hack to me. And it will not work for TextEdit that are placed inside of a template unless the user adds the same attribute.

The main problem here is that everything is placed inside of a Form within the DataGrid. And the default behavior for the `enter` key is to submit the form. I think we need to detect if the input is inside of a datagrid, within form, within header. SImilar to the current implementation but without the data attribute.

David-Moreira:
> I don't like this approach that much. It feels like a hack to me.
Programming is basically hacking haha

> and it will not work for TextEdit that are placed inside of a template unless the user adds the same attribute. 
You are right. I forgot about this. Only way is to do some general js that applies it then. 

In this case I'll probably do it to any inputs in the thead and make them not submittable like suggested.

stsrki (approved):
Looking good now. I've added a null check. And realized that I left the same filename as before do I renamed it.

PR #3183, opened 2021-11-29
Revert lazy module

stsrki:
While testing before the release I have noticed that with Lazy the JS would get called even though disposal happened already.


David-Moreira:
Ah, ya, you forgot to allocate it on a field. But yea, there shouldn't be any gains in using Lazy here eitherway.

PR #3185, opened 2021-11-30
DataGrid : Makes sure to close the modal with explicit CloseReason.UserClosing, if Save ran to success

### LINKED ISSUE #3149: Problem in PopupClosing Event of Datagrid
When you want to prevent closing the Edit Popup you can use PopupClosing Event of the Datagrid. When you set the Cancel property on FocusLostClosing the CloseReason property stays with FocusLostClosing and the Popup never closes again.

I have attached an example project.

[BlazoriseTest.zip](https://github.com/Megabit/Blazorise/files/7576376/BlazoriseTest.zip)

David-Moreira:
Closes #3149

### DIFF
--- a/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
+++ b/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
@@ -197,7 +197,7 @@ namespace Blazorise.DataGrid
         /// <summary>
         /// Links the child row with this datagrid.
         /// </summary>
-        /// <param name="column">Column to link with this datagrid.</param>
+        /// <param name="row">Row to add.</param>
         public void AddRow( DataGridRowInfo<TItem> row )
         {
             Rows.Add( row );
@@ -216,7 +216,7 @@ namespace Blazorise.DataGrid
         /// <summary>
         /// Links the child row with this datagrid.
         /// </summary>
-        /// <param name="column">Column to link with this datagrid.</param>
+        /// <param name="row">Row to remove.</param>
         public bool RemoveRow( DataGridRowInfo<TItem> row )
             => Rows.Remove( row );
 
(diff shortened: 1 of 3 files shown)


stsrki:
Still not good. It now calls the event two times with the following result

image

Code:

`PopupClosing="@((e)=>{Console.WriteLine(e.CloseReason); return Task.CompletedTask; } )"`

David-Moreira:
Nice catch. I'll take a look.

David-Moreira:
Apparently same happens with `Cancel`. Thing is we have two concurrent ways of `Closing` the modal, by API or by mutating the `Visible` Parameter workign at the same time. 
Theorically it still should work, but it goes like this:
- `Hide` call --> Reason: UserClosing
- Concurrent `SetParametersAsync` still thinks visible is true, and runs visible set to true pipeline...
- In the meantime Datagrid tells the correct visible state, false, and now runs the false pipeline... => Reason: None

Most likely the fix we're gonna work on, is to only use one of these ways in the DataGrid. Might be confusing to try handling the "concurrency" instead.

stsrki:
I believe that's why I used the parameter Visible to control the Modal. If I remember correctly. If you want to control it properly, you will need to choose one way. Possibly with `Show` `Hide` would be the best.

stsrki:
It is behaving well now. The only problem was that `OpenModal and ``validations.ClearAll` were called from sync code so I fixed it. You can check one more time to confirm and then I will merge it.

David-Moreira:
OpenModal should have still be async? It was on the AfterRender Task queue remember? :D 
Validations.ClearAll was already wrong, ddin't notice, assumed it was a sync call.

stsrki:
> OpenModal should have still be async? It was on the AfterRender Task queue remember?

`modalRef.Open` was on AfterRender Task but `validations.ClearAll` was not. This is just a simple optimization to prevent any problem in the future.

PR #3222, opened 2021-12-09
DataGrid : MacOs Meta key as Ctrl Click

### LINKED ISSUE #1705: Datagrid - Unselect child data grid
Currently when you select a datagrid row that has a child grid, the child grid expands when you click on the parent row but if you click on the parent row again, you can not deselect the row and collapse all the child grid rows. The only way to collapse the child grid row is to click on another row.

Can you add a feature that collapses the Child Grid within a selected parent row when you click on the parent row?

David-Moreira:
Closes #1705


David-Moreira:
Everything good here?

stsrki (approved):
Yes, everything is good. Didn't code much last few days :)

PR #3245, opened 2021-12-19
Syncronize SelectedRows set to empty or null vs SelectedRow

### LINKED ISSUE #3233: Datagrid SelectedRows not Resetting/Clearing as expected.
**Describe the bug**
I Have a DataGrid  with `SelectedRows="@myList"`  and `SelectedRowsChanged="@AddOrRemoveMyList"`

When i select a row AddOrRemoveMyList gets hit
```
private void AddOrRemoveMyList (List<MyType> someList)
{
        myList= someList;
        +adding stuff to another list based on the Selection made
}
```
The Problem is: If i try to reset the selection like `myList= new List<MyType>();`
`AddOrRemoveMyList ` is hitting afterwards but not with an empty list but with a list where only the last selected one is in it.

The Selected Ui Items are also not changed. Leading to some weird selections.

**To Reproduce**
1. Create razorPage with code example below.
2. Start and open The Page
3. Select Some rows.
4. Click Reset button

**Expected behavior**
Selection should be Empty, AddOrRemoveMyList shouldn't be hit, and if, then with an empty List of items.

**Sample Code**
```
<DataGrid TItem="string" Data="@dataList" Narrow="true" FixedHeader="true" Virtualize="true"
          VirtualizeOptions="@(new VirtualizeOptions() { DataGridHeight="calc(-8px + 70vh) !important", DataGridMaxHeight="calc(-8px + 70vh) !important" })"
          Striped="true" Bordered="true" Filterable="true" Sortable="false" SelectionMode="DataGridSelectionMode.Multiple"
          Responsive="true"
          SelectedRows="@myList"
          SelectedRowsChanged="@AddOrRemoveMyList">
    <DataGridColumn TItem="string" Caption="No">
        <DisplayTemplate>
            @context
        </DisplayTemplate>
    </DataGridColumn>
</DataGrid>
<Button Color="Color.Danger" Clicked="@ResetSelection">Reset</Button>
@code {
    private List<string> dataList = new List<string>() { "a", "b", "c", "d", "e", "f" };
    private List<string> myList = new List<string>();

    private void AddOrRemoveMyList(List<string> someList)
    {
        myList = someList;
        StateHasChanged();
    }

    private void ResetSelection()
    {
        myList = new List<string>();
        // also tried myList = null;
        // myList.Clear();
        StateHasChanged();
    }
}
```
**Screenshots**

**Additional context**
Vs 2019
Latest Version 0.9.5.2
net5.0 (5.0.12)
Blazor Serverside on Windows

David-Moreira:
Closes #3233

Currently this fixes, user setting the `SelectedRows` Parameter with items in it, to empty and `DataGrid` setting `SelectedRows` to contain the last `SelectedRow` as per `HandleSelectionModeChanged` which syncronizes both.

There is yet another problem, which is, the internal rows aren't updating themselves as per `StateHasChanged`, they think nothing has changed so they still show up as selected visually.

I guess if we go back to having Parameters going through the `DataGrid` directly to the component, it will solve it. But I kinda don't like it... I.E:
`<_DataGridRow @key="@item" TItem="TItem" Item="@item" />`
to:
`<_DataGridRow @key="@item" TItem="TItem" Item="@item" SelectedRows="SelectedRows" SelectedRow="SelectedRow" />`

There's another way which I was gonna suggest for v1.0, since it's a technically a wider change. But we might consider taking it here? What do you think?
Setting the DataGrid `IsFixed="true"` to `IsFixed="false"`  Theorically IsFixed should be set to false, as `DataGrid` is a `CascadingParameter` that keeps changing and should inform its children that it did so. 
(I think this would solve 100% alot of the problems we had related to `DataGrid` not knowing it should update internal components)

Code used to test:
```
<DataGrid 
@ref="dataGridRef"
            TItem="string" Data="@dataList" Narrow="true" FixedHeader="true" Virtualize="false"
          VirtualizeOptions="@(new Blazorise.DataGrid.Configuration.VirtualizeOptions() { DataGridHeight="calc(-8px + 70vh) !important", DataGridMaxHeight="calc(-8px + 70vh) !important" })"
          Striped="true" Bordered="true" Filterable="true" Sortable="false" SelectionMode="DataGridSelectionMode.Multiple"
          Responsive="true"
          SelectedRows="@myList"
          SelectedRowsChanged="@AddOrRemoveMyList">
    <DataGridColumn TItem="string" Caption="No">
        <DisplayTemplate>
            @context
        </DisplayTemplate>
    </DataGridColumn>
</DataGrid>
<Button Color="Color.Danger" Clicked="@ResetSelection">Reset</Button>
@String.Join(',' , myList)
@code {
    private DataGrid<string> dataGridRef;
    private List<string> dataList = new List<string>() { "a", "b", "c", "d", "e", "f" };
    private List<string> myList = new List<string>();

    private void AddOrRemoveMyList(List<string> someList)
    {
        myList = someList;
        StateHasChanged();
    }

    private void ResetSelection()
    {
        myList = new List<string>();
        //myList.Clear();
        StateHasChanged();
    }
}
```


stsrki:
Setting it to `IsFixed="false"` would lead to other problems, and DataGrid would be refreshed multiple times. Trust me we don't want to go that way. I had the same problem with the validation component that was eluding me for a long time until I figured the problem was in `IsFixed="false"`. And DataGrid is a much bigger component. If you really want to cascade states then it is better to use a `record` type. Just like we have for `Modal`, `Bar`, etc. It works perfectly and it will not refresh unless it is truly changed.

stsrki (approved):
LGTM

PR #3267, opened 2021-12-28
First draft of ShakaPlayer.js for Blazorise

sturlath:
I just thought I´d go as far as I could with this for fun today.

I just copied what was done with the TimePicker and hope I didn´t miss anything but have not tested this in any shape or form **or** implemented enough to make it work.

If you want to poke into this a bit I can do some more work just to help this along. I´m ready to do the grunt work.


Source/Blazorise/wwwroot/video.js
```diff
@@ -0,0 +1,66 @@
...
+const _instancesInfos = [];
+
+export function initialize(dotnetAdapter, element, elementId, options) {
+    element = utilities.getRequiredElement(element, elementId);
+
+    if (!element)
+        return;
+
+    // Install built-in polyfills to patch browser incompatibilities.
+    shaka.polyfill.installAll();
```
stsrki, line 13: I would probably move this up, before the `initialize` method so that it doesn't get called every time a new video is created.

Source/Blazorise/wwwroot/video.js
```diff
@@ -0,0 +1,66 @@
...
+
+    // Check to see if the browser supports the basic APIs Shaka needs.
+    if (shaka.Player.isBrowserSupported()) {
+        // Everything looks good!
+        initPlayer();
+    } else {
+        // This browser does not have the minimum set of APIs we need.
+        console.error('Browser not supported!');
```
stsrki, line 21: Mentioned that it is for the video component.

stsrki:
Looking good for the first version. Thanks. I will try to test it as soon as possible.

Source/Blazorise/wwwroot/video.js
sturlath: I lean towards the first and will update as such. But if you don´t like it I also considered these here.

- "Browser not supported for Video player"
- "Browser not supported for Blazorise Video"
- "Browser not supported for Shaka video player"

stsrki: Yes, the first version is OK.

sturlath:
Just one thing about the ShakePlayer.js I took the content from the Player.js from the Shake repository. See [this doc](https://shaka-player-demo.appspot.com/docs/api/tutorial-welcome.html). I think that is the correct one but thought to let you know.

stsrki:
Move work to #3298

PR #3302, opened 2022-01-10
Autocomplete CloseOnSelection feature

### LINKED ISSUE #3285: Autocomplete multiselection
When the menu is opened in multiselection mode it should not close when you click on items. The menu should stay open until the escape key is pressed on focus has been lost by going out of bounds.

David-Moreira:
Closes #3285


stsrki:
Tested and it is almost there. The menu should IMO stay open only in multi-select mode if `CloseOnSelection` is enabled. In single mode, it should behave as usual.

David-Moreira:
Yea there are still some bugs.
I guess it couldn't hurt to allow both. Up to the user if he wants or not. But we can do it only multi selection then. Better change parameter to `CloseOnMultiSelection`? Or at leçast mention on the parameter docs.

stsrki:
Leave it the same name, but mention it in the docs.

stsrki (approved):
lgtm

PR #3315, opened 2022-01-13
Enum docs update + Visibility vs Display Note

### LINKED ISSUE #3273: Docs : Add missing enums / Note about Visibility vs Display
### Discussed in https://github.com/Megabit/Blazorise/discussions/3272

>P.s
When I was searching for `Visibility` in the docs I didn´t find it and see that there are [loads of enums missing](https://github.com/Megabit/Blazorise/tree/67da3e1fb3256f83f25eb378026eda10768322b5/Source/Blazorise/Enums). I would help with adding them to the documentation if I wasn't super pressed for time atm.</div>

Check if there are missing enums and add. Add Note about recurring `Visibility` vs `Display` question.

David-Moreira:
Closes #3273


stsrki (approved):

stsrki:
So what does this do? Generates the `EnumsOutput.txt` file with all enums found in Blazorise project? Then we need to manually copy-paste it into the documentation?

David-Moreira:
Yea. For now that's it. Maybe we can even segregate into separate files if needed.
But this is better then manually copy pasting in my opinion.

David-Moreira:
We can automate in the future when we have more time?

stsrki:
Currently, we have enums separated into different pages. How will this new generator work in that page structure, considering you generate one large file?

image

David-Moreira:
I did it manually in this commit:
Separate enums into corresponding pages from generator & Add DisplayT… 

But if we really want to, we can do smthg simple like a dicionatary with the page and corresponding enums and generate a page for each... I just didn't bother as we don't update them that often...

stsrki:
It would be good to have something that autogenerates links from the Attributes when the enum is defined. But we can do that after 1.0

David-Moreira:
Yea we should automate more stuff on docs. Just that there's already so much to do haha. Better after 1.0

Documentation/Blazorise.Docs.Generator/Program.cs
```diff
@@ -0,0 +1,82 @@
...
+//        </UnorderedList>
+//    </DocsPageSectionHeader>
+//</DocsPageSection>
+
+var sb = new StringBuilder();
+
+foreach ( var enumGroup in enums )
+{
+    sb.AppendLine( "<DocsPageSection>" );
+    sb.AppendLine( $"\t<DocsPageSectionHeader Title=\"{enumGroup.Key}\">" );
```
stsrki, line 58: Can you use spaces instead of tabs? We have 4 spaces as tab configured in `.editorconfig` and they should match.

David-Moreira: image
That's how its setup

PR #3317, opened 2022-01-13
New SelectIndexedList Component

lolochristen:
The components Select and SelectList are not able to directly return and bind to an item from the assigned list. I created a new component with this capability as the existing SelectList would have most likely experienced a breaking change. the new component uses the index of a IList for the SelectItem and synchronizes with SelectedItem property.
Please consider this capability as I at was a crucial missing feature in my projects. Feel free to integrate directly or replace SelectList. 

Example:
``` cs 

<SelectIndexedList TItem="Item" Data="items" @bind-SelectedItem="@selectedItem" TextField="@((item)=>item.Name)"  /> 
@code {

    private IList<Item> items = new List<Item>()
    {
        new Item() { Id = 1, Name = "Value 1" },
        new Item() { Id = 2, Name = "Value 2" }, 
        new Item() { Id = 3, Name = "Value 3" }
    };

    private Item? selectedItem;

    internal class Item
    {
        public int Id { get; set; }
    }
}
```

### DIFF
--- a/Source/Extensions/Blazorise.Components/SelectIndexedList.razor
+++ b/Source/Extensions/Blazorise.Components/SelectIndexedList.razor
@@ -1,0 +1,21 @@
+﻿@typeparam TItem
+<Select @ref="@selectRef" ElementId="@ElementId" TValue="int" SelectedValue="@SelectedIndex" SelectedValueChanged="@HandleSelectedValueChanged" Class="@Class" Style="@Style" Size="@Size" MaxVisibleItems="@MaxVisibleItems" TabIndex="@TabIndex" Disabled="@Disabled" Attributes="@Attributes">
+    <ChildContent>
+        @if ( DefaultItemText != null )
+        {
+            <SelectItem @key="default" TValue="int" Value="-1" Disabled="@DefaultItemDisabled" Hidden="@DefaultItemHidden">@DefaultItemText</SelectItem>
+        }
+        @if ( Data != null )
+        {
+            @for (int i = 0; i < Data.Count; i++)
+            {
+                var text = TextField != null ? TextField.Invoke(Data[i]) : Data[i].ToString();
+                <SelectItem @key="i" TValue="int" Value="@i">@text</SelectItem>
+            }
+        }
+        @ChildContent
+    </ChildContent>
+    <Feedback>
+        @Feedback
+    </Feedback>
+</Select>
--- a/Source/Extensions/Blazorise.Components/SelectIndexedList.razor.cs
+++ b/Source/Extensions/Blazorise.Components/SelectIndexedList.razor.cs
@@ -1,0 +1,174 @@
+﻿#region Using directives
+using System;
+using System.Collections.Generic;
+using System.Linq.Expressions;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Components;
+#endregion
+
+namespace Blazorise.Components
+{
+    /// <summary>
+    /// Dynamically builds select component and it's items based in the supplied data.
+    /// </summary>
+    /// <typeparam name="TItem">Data item type.</typeparam>
+    public partial class SelectIndexedList<TItem> : ComponentBase
+    {
+        #region Members
+
+        /// <summary>
+        /// Reference to the <see cref="Select{TValue}"/> component.
+        /// </summary>
+        private Select<int> selectRef;
+
+        #endregion
+
+        #region Methods
+
+        protected async Task HandleSelectedValueChanged( int index )
+        {
+            if ( SelectedIndex != index )
+            {
+                SelectedIndex = index;
+                await SelectedIndexChanged.InvokeAsync( index );
+                await SelectedItemChanged.InvokeAsync( SelectedItem );
+            }
+        }
+
+        protected override async Task OnParametersSetAsync()
+        {
+            if ( DefaultItemText == null && SelectedIndex == -1 ) // when no default text is defined, selected item should be first item in list
+                await HandleSelectedValueChanged( 0 );
+        }
+
+        /// <summary>
+        /// Sets focus on the input element, if it can be focused.
+        /// </summary>
+        /// <param name="scrollToElement">If true the browser should scroll the document to bring the newly-focused element into view.</param>
+        /// <returns>A task that represents the asynchronous operation.</returns>
+        public Task Focus( bool scrollToElement = true )
+        {
+            return selectRef.Focus( scrollToElement );
+        }
+
+        #endregion
+
+        #region Properties
+
+        /// <summary>
+        /// Gets or sets the select element id.
+        /// </summary>
+        [Parameter] public string ElementId { get; set; }
+
+        /// <summary>
+        /// Gets or sets the select data-source.
+        /// </summary>
+        [Parameter] public IList<TItem> Data { get; set; }
+
+        /// <summary>
+        /// Method used to get the display field from the supplied data source.
+        /// </summary>
+        [Parameter] public Func<TItem, string> TextField { get; set; }
+
+        /// <summary>
+        /// Currently selected item value index.
+        /// </summary>
+        [Parameter] public int SelectedIndex { get; set; } = -1;
+
+        /// <summary>
+        /// Occurs after the selected value has changed.
+        /// </summary>
+        [Parameter] public EventCallback<int> SelectedIndexChanged { get; set; }
+
+        /// <summary>
+        /// Currently selected item value.
+        /// </summary>
+        [Parameter]
+        public TItem SelectedItem
+        {
+            get
+            {
+                if (SelectedIndex >= 0 && Data != null)
+                    return Data[SelectedIndex];
+
+                return default(TItem);
+            }
+            set
+            {
+                int index = -1;
+                if ( Data != null )
+                    index = Data.IndexOf( value );
+
+                HandleSelectedValueChanged( index ).Wait();
+            }
+        }
+
+        /// <summary>
+        /// Occurs after the selected value has changed.
+        /// </summary>
+        [Parameter] public EventCallback<TItem> SelectedItemChanged { get; set; }
+
+        /// <summary>
+        /// Display text of the default select item.
+        /// </summary>
+        [Parameter] public string DefaultItemText { get; set; }
+
+        /// <summary>
+        /// If true, disables the default item.
+        /// </summary>
+        [Parameter] public bool DefaultItemDisabled { get; set; } = false;
+
+        /// <summary>
+        /// If true, disables the default item.
+        /// </summary>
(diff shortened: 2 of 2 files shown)


stsrki:
Seems good, although I would like it better if it was part of an existing component. I will try to make work on SelectList in the next few days, my schedule is full at the moment so I cannot properly test it. Thanks for the PR, btw 💪

stsrki:
The problem with `SelectIndexedList` is that it would lead to several problems.

1. Data-Annotation validation would not work because it doesn't bind to any field on an `TItem`
2. `@key` parameter would potentially be invalid in certain cases if `Data` is changed
3. Use of `.Wait();` in synchronous context

---

This is my first try at adding the SelectedIndex to the SelectList. Seems to be working fine but more tests are welcome on your side.

Test

```cshtml
<Button Color="Color.Primary" Clicked="@(()=>{ selectedItem = items[1]; return Task.CompletedTask; })">SELECT ITEM</Button>
<Button Color="Color.Danger" Clicked="@(()=>{ items.RemoveAt(1); return Task.CompletedTask; })">REMOVE ONE</Button>

<SelectList TItem="Item" TValue="int" Data="items" @bind-SelectedItem="@selectedItem" ValueField="@((item)=>item.Id)" TextField="@((item)=>item.Name)" />

<Div>
    Selected item: @selectedItem?.Name
</Div>
@code {
    private IList<Item> items = new List<Item>()
    {
        new Item() { Id = 1, Name = "Value 1" },
        new Item() { Id = 2, Name = "Value 2" },
        new Item() { Id = 3, Name = "Value 3" },
        new Item() { Id = 4, Name = "Value 4" },
        new Item() { Id = 5, Name = "Value 5" },
    };

    private Item? selectedItem;

    internal class Item
    {
        public int Id { get; set; }
    }
}
```

Source/Extensions/Blazorise.Components/SelectList.razor.cs
```diff
@@ -69,7 +103,23 @@ public Task Focus( bool scrollToElement = true )
...
+        [Parameter]
+        public TValue SelectedValue
+        {
+            get => selectedValue;
+            set
+            {
+                if ( !selectedValue.Equals( value ) )
+                {
+                    selectedValue = value;
+                    SelectedValueChanged.InvokeAsync( selectedValue ).ConfigureAwait( false );
```
stsrki, line 115: You must not call InvokeAsync in synchronous code.

Source/Extensions/Blazorise.Components/SelectList.razor.cs
```diff
@@ -22,14 +23,16 @@ public partial class SelectList<TItem, TValue> : ComponentBase
...
+        private TValue selectedValue;
+
         #endregion
 
         #region Methods
 
         protected Task HandleSelectedValueChanged( TValue value )
         {
             SelectedValue = value;
-            return SelectedValueChanged.InvokeAsync( value );
```
stsrki, line 32: Any reason why this is removed?

Source/Extensions/Blazorise.Components/SelectList.razor.cs
```diff
@@ -147,6 +235,13 @@ public Task Focus( bool scrollToElement = true )
         /// </summary>
         [Parameter] public RenderFragment ChildContent { get; set; }
 
+
+        /// <summary>
+        /// Returns true if SelectedItem property is supported.
+        /// </summary>
+        public bool SelectedItemSupported =>
+            ValueField == null && Data is IList && typeof(TValue) == typeof(int);
```
stsrki, line 243: Why support only `typeof(int)`? I think practically any `TValue` type should be indexed.

lolochristen: In a IList, the index is always int. A standard collection supporting other types as index would be a dictionary which is not assignable from IEnumerable<TItem>. We would need to move to object for Data (btw as WPF/UWP does with ItemSource)

Source/Extensions/Blazorise.Components/SelectList.razor.cs
lolochristen: As changes also can be trigger from SelectedItem change, the raising of the events shall when the property changes.

Source/Extensions/Blazorise.Components/SelectList.razor.cs
lolochristen: I had a look at asp.net core components and the pattern they use is to trigger changed events use within a internal CurrentValueProperty. I propose to the same as Property can be changed from different sources. And there is no other way then raising InvokeAsync in synch code. At least I use no _ to indicate that code does not care of the task output.

stsrki:
@David-Moreira can you try looking at this if you have time?

David-Moreira:
I'll take a look over these days. 
To start off. Could you guys clarify what does this component solve?
I kinda don't get this phrase:
>"The components Select and SelectList are not able to directly return and bind to an item from the assigned list. "

Does `SelectList` not already support this?
The shown example also seems to be supported by `SelectList`... but I'm sure I am missing something elementar.... again could you guys enlighten me, please :D

David-Moreira:
Ah great! Yea no point in doing a new component. We should also update the other one's if they do not support it.

David-Moreira:
Hello @lolochristen ,
So I was starting to do some testing based on your examples, can you take a look at these:

1- If I mutate `selectedItem` `SelectList` is not updating the dropdown value, it remains value1.
image
**Test code:**
```
<Paragraph>
    <Button Color="Color.Primary" Clicked="@(()=>{ selectedItem = items[1]; return InvokeAsync(StateHasChanged); })">SELECT ITEM</Button>
    <Button Color="Color.Danger" Clicked="@(()=>{ items.RemoveAt(1); return Task.CompletedTask; })">REMOVE ONE</Button>

    <SelectList TItem="Item" TValue="int" Data="items" @bind-SelectedItem="@selectedItem" ValueField="@((item)=>item.Id)" TextField="@((item)=>item.Name)" />
    <Paragraph>Selected item: @selectedItem?.Name</Paragraph>
</Paragraph>
```





2- If I have two `SelectList` and I try to mutate `selectedItem` it will throw this error:
image

```
<Paragraph>
    <Button Color="Color.Primary" Clicked="@(()=>{ selectedItem = items[1]; return InvokeAsync(StateHasChanged); })">SELECT ITEM</Button>
    <Button Color="Color.Danger" Clicked="@(()=>{ items.RemoveAt(1); return Task.CompletedTask; })">REMOVE ONE</Button>

    <SelectList TItem="Item" TValue="int" Data="items" @bind-SelectedItem="@selectedItem" ValueField="@((item)=>item.Id)" TextField="@((item)=>item.Name)" />
    <Paragraph>Selected item: @selectedItem?.Name</Paragraph>
</Paragraph>

<Paragraph>
    <SelectList TItem="Item" TValue="int" Data="items" @bind-SelectedItem="@selectedItem2" TextField="@((item) => item.Name)" />
    <Paragraph>Selected item 2: @selectedItem2?.Name</Paragraph>
</Paragraph>
```


Finally here's the full code that I was going to start testing that's based on your examples, before I found these issues:
```
<Paragraph>
    <Button Color="Color.Primary" Clicked="@(()=>{ selectedItem = items[1]; return InvokeAsync(StateHasChanged); })">SELECT ITEM</Button>
    <Button Color="Color.Danger" Clicked="@(()=>{ items.RemoveAt(1); return Task.CompletedTask; })">REMOVE ONE</Button>

    <SelectList TItem="Item" TValue="int" Data="items" @bind-SelectedItem="@selectedItem" ValueField="@((item)=>item.Id)" TextField="@((item)=>item.Name)" />
    <Paragraph>Selected item: @selectedItem?.Name</Paragraph>
</Paragraph>

<Paragraph>
    <SelectList TItem="Item" TValue="int" Data="items" @bind-SelectedItem="@selectedItem2" TextField="@((item) => item.Name)" />
    <Paragraph>Selected item 2: @selectedItem2?.Name</Paragraph>
</Paragraph>



<Paragraph>
    <SelectList TItem="Item" TValue="int" Data="items" @bind-SelectedItem="@selectedItem3" TextField="@((item)=>item.Name)" DefaultItemText="make your choice" DefaultItemValue="-1" />
    <Paragraph>Selected item 3: @selectedItem3?.Name</Paragraph>
</Paragraph>


<Paragraph>

    <SelectList TItem="Item" TValue="int" Data="items" @bind-SelectedValue="@selectedValue" TextField="@((item)=>item.Name)" />
    <Paragraph>Selected value: @selectedValue</Paragraph>

</Paragraph>


<Paragraph>
    <SelectList TItem="Item" TValue="int" Data="items" @bind-SelectedValue="@selectedValue2" TextField="@((item)=>item.Name)" ValueField="@((item)=>item.Id)" />
    <Paragraph>Selected value 2: @selectedValue2</Paragraph>
</Paragraph>

@code {
    private IList<Item> items = new List<Item>()
    {
        new Item() { Id = 1, Name = "Value 1" },
        new Item() { Id = 2, Name = "Value 2" },
        new Item() { Id = 3, Name = "Value 3" },
        new Item() { Id = 4, Name = "Value 4" },
        new Item() { Id = 5, Name = "Value 5" },
    };

    private Item? selectedItem;
    private Item? selectedItem2;
    private Item? selectedItem3;

    private int selectedValue;
    private int selectedValue2;

    internal class Item
    {
        public int Id { get; set; }
    }
}
```

stsrki:
I would say the reason for the error is that `GetKey` should return an actual `item` and set that as a `@key` instead of the hash code.

David-Moreira:
@lolochristen just a friendly reminder. :) Understand if you're busy, of course.

David-Moreira:
@stsrki Let's resume this for **1.3**?

Also, some time has passed, I'm having a hard time remembering how this component is different then a regular `SelectList`?

stsrki:
> @stsrki Let's resume this for **1.3**?
> 
> Also, some time has passed, I'm having a hard time remembering how this component is different then a regular `SelectList`?

As far as I remember, this PR introduces an internal List that holds the Items, and then we can access them by the Index. Then again, I was thinking we should probably go forward with making something similar on the regular `Select` component. Have an internal List or Dictionary to hold data items. That way we could, in theory, solve problems with storing object types as values.

David-Moreira:
> > @stsrki Let's resume this for **1.3**?
> > Also, some time has passed, I'm having a hard time remembering how this component is different then a regular `SelectList`?
> 
> As far as I remember, this PR introduces an internal List that holds the Items, and then we can access them by the Index. Then again, I was thinking we should probably go forward with making something similar on the regular `Select` component. Have an internal List or Dictionary to hold data items. That way we could, in theory, solve problems with storing object types as values.

I see. Let's investigate and we might abandon this one.

stsrki:
We have decided that we will not pursue this feature. Closing the PR.

@lolochristen Thank you for the work, and I am sorry for not being able to go through it.

PR #3330, opened 2022-01-17
DetailRowTrigger API DetailRowTriggerEventArgs

### LINKED ISSUE #3226: DetailRowTrigger Bug
After opening a DetailRow it is possible to close the DetailRow by clicking on the parent cell of the DetailRow.
Normally this behavior is expected, but our DetailRow is openend by a button inside of the parent row with a additional method.

Create a grid and open a DetailRow then click on its parent DetailRow. 

`DetailRowTrigger=@((item) => selectedPatches.Contains(item.ID))`
This is used to know which row is opened. Our button simply adds the of a row to the List which then opens the DetailRow.

David-Moreira:
Closes #3226


David-Moreira:
Ready, take a look when you have time.

Source/Extensions/Blazorise.DataGrid/Contexts/DetailRowTriggerContext.cs
```diff
@@ -0,0 +1,34 @@
...
+
+        /// <summary>
+        /// Gets the model.
+        /// </summary>
+        public TItem Item { get; private set; }
+
+        /// <summary>
+        /// If DetailRow is showing, and if the trigger is re-evaluated the DetailRow will be set to false. Defaults to <c>true</c>.
+        /// </summary>
+        public bool Toggleable { get; set; } = true;
```
stsrki, line 22: Should this be changeable from the outside? Can we make it private or internal?

David-Moreira: THe point is for the user to mutate yes, according to their needs. If they want they set it to false therefore having a different behaviour.
I kinda debated with myself if this should be an OptionsAPI instead...

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -703,11 +703,10 @@ public Task ToggleDetailRow( TItem item, bool forceDetailRow = false )
         protected internal Task ToggleDetailRow( TItem item, DetailRowTriggerType detailRowTriggerType, bool forceDetailRow = false, bool skipDetailRowTriggerType = false )
         {
             var rowInfo = GetRowInfo( item );
+
             if ( rowInfo is not null )
             {
-                if ( forceDetailRow )
-                    rowInfo.ToggleDetailRow();
```
David-Moreira, line 706: Did you remove this?

stsrki: Yes. It seems to me that it was not needed. ` rowInfo.ToggleDetailRow();` was called in here and in `else` when `if ( DetailRowTrigger is not null )` was not satisfied. So I just left it in an `else` block.

Source/Extensions/Blazorise.DataGrid/Contexts/DetailRowTriggerContext.cs
stsrki: OK if mutability was wanted then it is fine like this.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
David-Moreira: Maybe you missunderstood that parameter purpose, the ForceDetailRow argument is to, toggle it either way... even if DetailTrigger would evaluate to false?

stsrki: Hmm I think you're right. Going to reverse it.

Source/Extensions/Blazorise.DataGrid/Contexts/DetailRowTriggerContext.cs
```diff
@@ -0,0 +1,34 @@
+﻿namespace Blazorise.DataGrid
+{
+    /// <summary>
+    /// Holds the DetailRowTrigger context configuration.
+    /// </summary>
+    /// <typeparam name="TItem"></typeparam>
+    public class DetailRowTriggerContext<TItem>
```
stsrki, line 7: Maybe this would be better named `DetailRowTriggerEventArgs`? The Context suffix is mostly on parameters that are event callbacks.

David-Moreira: Well Action/Func as Parameters, pretty much do work like callbacks in blazor if you think about it, just that they do not have the `StateHasChanged` implicit call.
Eitherway I don't really mind either I guess. I'll make the change.

stsrki: We might actually consider a standard way of naming stuff. For example, `Context` can be reserved for render fragments(templates), and EventArgs can be for everything else, `Action`, `Func`, or `EventCallback`.

David-Moreira: Might be worth having a technical doc with this kind of stuff... our conventions / blazorise best practices etc... Which would also be in the how to contribute.

PR #3356, opened 2022-01-25
Markdown: add missing actions

WolfgangKluge:
EasyMDE has three actions not described in the [readme](https://github.com/Ionaru/easy-markdown-editor#toolbar-icons), yet.
They are in the code:
https://github.com/Ionaru/easy-markdown-editor/blob/master/src/js/easymde.js#L1614-L1626 (undo/redo)
https://github.com/Ionaru/easy-markdown-editor/blob/master/src/js/easymde.js#L1553-L1558 (upload-image)

Undo/Redo is already available through shortcuts.

(Another approach to support this would be to allow custom actions that are passed through to EasyMDE - currently that's restricted by the internal [MarkdownActionProvider](https://github.com/Megabit/Blazorise/blob/master/Source/Extensions/Blazorise.Markdown/Providers/MarkdownActionProvider.cs)).

### DIFF
--- a/Source/Extensions/Blazorise.Markdown/Enums.cs
+++ b/Source/Extensions/Blazorise.Markdown/Enums.cs
@@ -3,7 +3,8 @@
     /// <summary>
     /// Markdown toolbar actions.
     /// </summary>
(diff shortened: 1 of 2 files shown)


stsrki:
LGTM. Thank you for making the necessary changes.

PR #3364, opened 2022-01-27
DatePicker range and multiselect

### LINKED ISSUE #3425: DatePicker range and multiple select

soenneker:
closes #3425

Adds range capability to the flatpickr port (https://flatpickr.js.org/examples/#range-calendar)


stsrki:
The range would mean we need to have two `Date` values. But we only have only one on `DatePicker`.

soenneker:
Thanks for reviewing. Your comment makes sense. I figured I didn't have everything covered. I'll make that change. Is there anything else you feel is needed before merging?

stsrki:
There's probably a lot more work needed. You need to cover:

- two-way binding. Meaning, values must be received from outside through the parameter, eg `TValue[] Dates`, and they also must be returned through `DatesChanged`.
- if multi-mode is on, the validation mechanism should also work

In case you're lost feel free to ask. I can also try looking at it over the weekend if I have much time.

soenneker:
Thanks @stsrki! This is going to be great to use.

PR #3373, opened 2022-01-29
Add EasyMDE options

### LINKED ISSUE #3297: Markdown shortcuts
Add an API to change keyboard shortcuts https://github.com/Ionaru/easy-markdown-editor#keyboard-shortcuts

### LINKED ISSUE #3361: Markdown: Full control over EasyMDE
I try to configure the MDE instance but I don't have full control over the options you provide to the EasyMDE constructor.

Especially I miss `spellChecker` (bit like #1316), but I'm interested in `status` and `renderingConfig`, too.

Every (or at least most) options EasyMDE provides, should be configurable by the Markdown editor.
https://github.com/Ionaru/easy-markdown-editor#options-list

WolfgangKluge:
closes #3297
closes #3361

I've added more options from https://github.com/Ionaru/easy-markdown-editor/#options-list

As discussed in #3361, it's a simple extension to the current implementation of the options.
So the new options work more or less as the current ones.

- `autofocus` seems not to work in general (even the most easiest example)
  What I can see is, that the option is properly (and I think correctly) forwarded to the codemirror instance
  I've' opend a ticket for that: https://github.com/Ionaru/easy-markdown-editor/issues/399
- `autoRefresh`: I have no idea how to test it, but what I can see is, that the option is properly (and I think correctly, too) forwarded to the codemirror instance
- `parsingConfig`: The option `underscoresBreakWords` is not supported in CodeMirror.
  https://github.com/Ionaru/easy-markdown-editor/issues/401
- Fixes the option `UploadImage`, so this option has an effect at all (always `true` before)
- I considered to implement `status` like the `toolbar`, but you want to have access to the codemirror instance in the functions there - and that's a bigger problem (would be very useful in the `toolbar` action, too)

Since it's a slightly larger PR, I've splitted the commits on every option..


stsrki:
From what I can see in the PR I think you have covered most of the missing stuff and you did an excellent job. The thing that you mention is not working we can leave them for later if needed. The API will not change so once EasyMDE fixes those we can include it in Blazorise.

Do you think that is OK? If you don't have any more work I think we can merge the PR.

WolfgangKluge:
Yes, sure it's ok. I don't have further work to do on that.

netz0:
It seems this also causes an issue with forms that have the required="" attribute with Chromium based browsers returning an error that the form control is invalid and not focusable.

stsrki:
> It seems this also causes an issue with forms that have the required="" attribute with Chromium based browsers returning an error that the form control is invalid and not focusable.

Can you open a new issue with code to reproduce it?

netz0:
> > It seems this also causes an issue with forms that have the required="" attribute with Chromium based browsers returning an error that the form control is invalid and not focusable.
> 
> Can you open a new issue with code to reproduce it?

Since the auto focus does not work, if your text area has the required attribute Chromium based browsers like Chrome, Edge, Vivaldi, etc. will fail because they can't focus on the field that has the required field. I suspect this is related to the original issue that autofocus is not working.

Just add a require field to the text area:
` <textarea id="easyMDE" name="easyMDE" required>{comment}</textarea>`
 
Can't submit the form and browser fails with:
**An invalid form control with name='easyMDE' is not focusable.**

PR #3381, opened 2022-01-31
Cascading type parameter

### LINKED ISSUE #3300: CascadingTypeParameter support
Now that we support .NET 6 only, we might add `CascadingTypeParameter` support.

https://docs.microsoft.com/en-us/aspnet/core/blazor/components/templated-components?view=aspnetcore-6.0#infer-generic-types-based-on-ancestor-components

David-Moreira:
closes #3300

Heads up, from my initial tests **null** does not seem to mesh well with the inference, so you still have to explicitly define the generic in those cases otherwise It will throw a compiler error.
My guess is that it should work, but maybe there's a technical aspect to it? Think we should ask over at the aspnetcore/blazor github?

image


stsrki:
Null was always difficult. Even in regular C# code if you try to assign null to a `generic` type it will complain. I think this is OK. We just need to add a note so that people are aware of it. And they can always just continue using TValue.

stsrki:
PS. Just write it as `Value="@((string)null)"`

stsrki:
Looks good to me so far. Have you tried the same on DataGrid?

stsrki:
DataGrid templates are problematic. I'm trying to do it

image

David-Moreira:
> DataGrid templates are problematic. I'm trying to do it
> 
> image

I didn't get to Datagrid, no. Was doing normal components first. 
What do you mean? When you have multiple children using the same tvalue? Maybe you have to add the attribute to the next child as it goes down the chain?

stsrki:
I'm not sure really. Seems to me that TItem is only propagated to the Blazor components but for regular C# code it doesn't get applied.

David-Moreira:
You didn't commit did you end up doing smthg?

stsrki:
No, I've just stashed it. There were too many changes and no need to break the PR.

David-Moreira:
Just tried it. Yea, but like only the templates that do not have TItem bug... pretty strange. Might be a bug. Have to investigate further.

stsrki:
> Just tried it. Yea, but like only the templates that do not have TItem throw that error... pretty strange. Might be a bug. Have to investigate further. Best thing is to do a simpler component in a demo that replicates this. If it still happens we report over on aspnetcore/blazor git.

I agree. It might be a scenario they didn't think of.

WolfgangKluge:
I ran into the same problem some days ago..
I had to set `[CascadingTypeParameter(nameof(TItem))]` not only on the root, but on the Context-classes as well.. Didn't get into this any further

David-Moreira:
I did try to cascade the attribute down to the children. Maybe I did not put it into the right place, or you're not having the exact same problem.

(Problem is that it actually cascades down successfully the TItem, but RenderFragment's that do not have TItem as an Argument, for some reason stop working.)

WolfgangKluge:
Yeah, seems like it's another problem - more like https://github.com/dotnet/aspnetcore/issues/37482.
As a workaround, you could make NewCommandContext generic, too (and never use the non-generic CommandContext directly)

David-Moreira:
yea the repro here:
https://github.com/jirikanda/Blazor-CascadingTypeParameterIssue/tree/main/ClassLibrary1
Seems to replicate the same exact issue. 
I'm not sure about making stuff generic on purpose, but at least I guess that does not have an impact on our usrs and we can make use of the functionality.

Since It seems like the actual fix is only planned for .NET7, right? :(

David-Moreira:
So with this in mind and the suggestion of applying the `TItem` to the `RenderFragment` arguments that do not have it...
I tried with just the Datagrid Aggregates on the Demo DataGridPage.

By turning 
`[Parameter] public RenderFragment<AggregateContext> DisplayTemplate { get; set; }`
into 
`[Parameter] public RenderFragment<AggregateContext<TItem>> DisplayTemplate { get; set; }`
the error with the RenderFragment was gone.

But then we had a new one with this parameter 
`[Parameter] public Func<IEnumerable<TItem>, DataGridColumn<TItem>, object> AggregationFunction { get; set; }`
now it suddently has trouble accepting an `implicit delegate` (or however we call this)
So this line:
`AggregationFunction=@(DataGridAggregate<Employee>.TrueCount)`
Only works like this:
`AggregationFunction=@((x,y) => DataGridAggregate<Employee>.TrueCount(x,y))`
or with the old version by having TItem explicitly defined.

So I guess they still have some ironing to do with this new feature...

So @stsrki what do you think, do we still wanna take this feature on the DataGrid taking into account this caveats?

1- We have to change any Datagrid child Renderfragments arguments to be generic of TItem. This does not seem to have any negative impact on our users usage of the api. 
2- There's this caveat with "implicit delegates" based on `TItem` which also does not seem to mesh well. (still with workaround, explicit TItem or anonymous method call)

There might even be more stuff that we figure out with time... haha but I'm hoping not.

Personally I'd take the leap and do the changes. My guess is that for most of the scenarios will work just fine giving our users the opportunity of enjoying this feature... Only some more specific stuff like that one with the `AggregationFunction` will need to be explicitly defined.
And with time maybe we can find some other workaround...

stsrki:
Will `AggregationFunction=@(DataGridAggregate<Employee>.TrueCount)` still work if `TItem` is explicitly defined on DataGridColumns? If that is true then I guess we could proceed with the feature. I don't want that suddenly people's code stops working because of this feature.

David-Moreira:
Yes it still works with `TItem` explicitly defined. In fact we can leave it like that on our demo example. And add a note to the docs.
We'll proceed then.

stsrki (approved):
That's a lot of changes :)

I have no complaint on this one. You did a really good job.

PR #3385, opened 2022-02-01
Is list or collection fix

### LINKED ISSUE #3318: DataGrid "Cannot create an abstract class" when UseValidation=true
**Describe the bug**
If I try to add or edit a row with a specific type (contains an `ICollection<>`-Property), I get the above error message. The stack trace is
```
System.MemberAccessException: Cannot create an abstract class.
   at System.Runtime.CompilerServices.RuntimeHelpers.GetUninitializedObjectInternal(Type type)
   at System.Runtime.CompilerServices.RuntimeHelpers.GetUninitializedObject(Type type)
   at Blazorise.DataGrid.Utils.RecursiveObjectActivator.CreateInstanceRecursive(Object currObjInstance, PropertyInfo[] properties, ValueTuple`2 typeTracker, Int32 depthLevel, Int32 circularReferenceLevel)
   at Blazorise.DataGrid.Utils.RecursiveObjectActivator.CreateInstanceRecursive(Object currObjInstance, PropertyInfo[] properties, ValueTuple`2 typeTracker, Int32 depthLevel, Int32 circularReferenceLevel)
   at Blazorise.DataGrid.Utils.RecursiveObjectActivator.CreateInstance[TItem]()
   at Blazorise.DataGrid.DataGrid`1.InitEditItem(TItem item)
   at Blazorise.DataGrid.DataGrid`1.Edit(TItem item)
   at Blazorise.DataGrid._BaseDataGridRowCommand`1.<get_Edit>b__6_0()
```

**To Reproduce**
Here's an example
```c#
@code{
    class Test {
        public int Num { get; set; }

        //public List<string>? Values { get; set; } // works
        public IList<string>? Values { get; set; } // should work
        //public IList<string> Values { get; set; } = new List<string>(); // should work
    }

    List<Test> Data = new List<Test> {
        new() { Num = 1 },
        new() { Num = 2 },
        new() { Num = 3, Values = new List<string>() }
    };
}

<DataGrid
    TItem="Test"
    Data="@Data"
    Editable
    UseValidation
>
    <DataGridColumns>
        <DataGridCommandColumn TItem="Test" />
        <DataGridColumn TItem="Test" Field="Num" Editable />
        <DataGridColumn TItem="Test">
            <DisplayTemplate>@(context.Values == null)</DisplayTemplate>
        </DataGridColumn>
    </DataGridColumns>
</DataGrid>
```

1. Click `New` or `Edit` on any entry and see the error
2.
   a) Even on an object where `Values` is already set (use of property initializer or `Num = 3`) no editing is available
   b) If you use `List<>` instead of `IList<>` it works.

**Expected behavior**
Ideally, there's no difference between `List<>` and `IList<>`-properties.
Since that will mean that you have to guess what implementation of an interface one will have to use, I think at least the property should be empty (and initializers should not be overwritten).
So in that cases empty (without error message) would/should be fine.

**Additional context**
While investigating, I've tested `IsListOrCollection`, too (don't know if that's part of the problem/soluton). And that behaves a bit strange to me (but I'm unsure)
```c#
typeof(List<string>).IsListOrCollection() // true
typeof(string[]).IsListOrCollection() // true
typeof(IList<string>).IsListOrCollection() // false
typeof(IEnumerable<string>).IsListOrCollection() // false
typeof(ICollection<string>).IsListOrCollection() // false
```

The current code is https://github.com/Megabit/Blazorise/blob/8990c56620c54480b5fe531039287c16606b7c18/Source/Extensions/Blazorise.DataGrid/ExtensionMethods.cs#L53-L56

Since `IList` inherits `ICollection`, the first line is superflous. But the striking part is the last line. I've never seen open generics used in that way and I think (and have tested) that this returns (nearly) always false.
```c#
typeof(IEnumerable<>).IsAssignableFrom(typeof(List<string>)) // false
typeof(IEnumerable<>).IsAssignableFrom(typeof(string[])) // false
typeof(IEnumerable<>).IsAssignableFrom(typeof(IList<string>)) // false
typeof(IEnumerable<>).IsAssignableFrom(typeof(IEnumerable<string>)) // false
typeof(IEnumerable<>).IsAssignableFrom(typeof(ICollection<string>)) // false
typeof(IEnumerable<>).IsAssignableFrom(typeof(ICollection<>)); // false
typeof(IEnumerable<>).IsAssignableFrom(typeof(IEnumerable<>)); // true
```

If you want to detect all enumerables, just use `typeof(IEnumerable).IsAssignableFrom(type)` (`IEnumerable<>` inherits `IEnumerable`). But by doing so, you will have all the collection types returning true, thus e.g. `typeof(string)` (as it's a collection of `char`), too.. Don't know if that's the desired behavior.
From the name `IsListOrCollection` I assume you want to detect collections only, so this might be a solution

```c#
public static bool IsListOrCollection( this Type type )
    => typeof(ICollection).IsAssignableFrom( type )
    || type.IsGenericCollection()
    || Array.Find( type.GetInterfaces(), IsGenericCollection ) != null;

private static bool IsGenericCollection( this Type type )
    => type.IsGenericType
    && type.GetGenericTypeDefinition() == typeof( ICollection<> );
```
(should be renamed to `IsCollection`..)

David-Moreira:
Closes #3318


stsrki:
Not sure if this is a mistake. Should it go into the master or 095 branch?

image

David-Moreira:
mistake. :D

David-Moreira:
Hmm I just switched branch. I think I based off master by mistake, I'll have to rebase later.

stsrki:
I think it would be best to close this PR and open new.

David-Moreira:
Ahhh It was based of the correct branch, you merged in master I just noticed haha. I'll just redo the branch/pr.

PR #3399, opened 2022-02-04
Add Bootstrap Icons

### LINKED ISSUE #3390: Add Bootstrap Icons
[Bootstrap icons](https://icons.getbootstrap.com) are free, MIT licensed, ever expanding collection of high quality, open source icons with almost 1,700 icons and growing. Can be used with or without Bootstrap in any project. And I think it would be great if they were included as part of Blazorise.

image

I can open PR with all 1683 icons auto-parsed as fields and keep them updated, but I'm not quite sure how to tackle `IconName` mappings. I understand it was meant to create common ground for various icon providers, but I'm not convinced that's ever going to work as intended. It covers barely couple % of icons and many if not most of them don't click across various providers. It's a nightmare to find matching icons (what one calls "adjust" another calls "circle-half"), others aren't there at all and trying to switch between providers will be poor experience either way. Not to mention it's going to be nightmare to maintain and even then it will be barely usable to find specific icons you might want, it's highly opinionated ... Arrow down? There could be 6 versions of it.

Bootstrap icons don't have "styles", everything is a different icon. Some icons have filled variants some don't, some have circle/square variants etc. I don't know how to deal with this.

```razor
<Icon Name="BootstrapIcons.ArrowDown"></Icon>
<Icon Name="BootstrapIcons.ArrowDownShort"></Icon>
<Icon Name="BootstrapIcons.ArrowDownCircle"></Icon>
<Icon Name="BootstrapIcons.ArrowDownSquare"></Icon>
<Icon Name="BootstrapIcons.ArrowDownSquareFill"></Icon>
```

Xeevis:
Fixes #3390 

`BootstrapIcons.cs` was auto-parsed so shouldn't contain any typos (code in appendix for future use). 

I've matched all `IconName`s I possibly could with my best effort and opinion. I've used field references instead of duplicating magic strings, I think it's more maintainable, less error-prone and intellisense is of great help. I've commented out entries I couldnt' match, I think it's more maintainable than removing them, but it's your call.

This should be all the heavy lifting. Only thing left should be docs and manifests. Perhaps this can make it into 1.0 release? Let me know if I need to do anything else. Thanks.

# Appendix (Icon parser)
```csharp
using System.Globalization;
using System.Text;
using System.Text.Json;

Stream? json = await new HttpClient().GetStreamAsync("https://raw.githubusercontent.com/twbs/icons/v1.8.0/font/bootstrap-icons.json");
JsonDocument? jsonDoc = await JsonDocument.ParseAsync(json);

StringBuilder? sb = new StringBuilder()
    .AppendLine("namespace Blazorise.Icons.Bootstrap;")
    .AppendLine()
    .AppendLine("/// <summary>")
    .AppendLine("/// Strongly-typed list of bootstrap icon names.")
    .AppendLine("/// </summary>")
    .AppendLine("public static class BootstrapIcons")
    .AppendLine("{");

int iconCount = 0;
foreach (JsonProperty item in jsonDoc.RootElement.EnumerateObject())
{
    iconCount++;
    string? normalizedName = CultureInfo.InvariantCulture.TextInfo.ToTitleCase(item.Name).Replace("-", string.Empty);
    if (char.IsDigit(normalizedName[0]))
    {
        normalizedName = "_" + normalizedName;

    }
    _ = sb.AppendLine(CultureInfo.InvariantCulture, $"    public static readonly string {normalizedName} = \"bi-{item.Name}\";");
}

sb.AppendLine("}");

await File.WriteAllTextAsync("BootstrapIcons.cs.txt", sb.ToString());
Console.WriteLine($"Parsed {iconCount} icons.");
```


stsrki:
It's looking really good. I think it might go into v1.0. Just need to clear those unmapped icons, but even so, it might be good enough as long as the most internal icons are covered.

I might even enable it as a default icon for bootstrap providers.

stsrki:
It's merged now! I have tried to map missing icons but not all could be included. I have reported some of them on https://github.com/twbs/icons. We'll see how it goes.

Thank you for making this work!

Issue #3410, opened 2022-02-07
Feature - QR-Code extension


njannink:
Component for generating QR codes based on something like

https://github.com/codebude/QRCoder

    <Blazorise.QRCode Text="https://blazorise.com" ECCLevel="L" EciMode="Default" DarkColor="Black" LightColor="White" />

or
   
    <Blazorise.QRCode ECCLevel="L" EciMode="Default" DarkColor="Black" LightColor="White">
        https://blazorise.com
    </Blazorise.QRCode>

stsrki:
There might be some problems with this approach https://github.com/codebude/QRCoder/issues/361#issuecomment-992152570

njannink:
My project is still on net5 but we plan to move to net6 in some months. I will have a look if it works or not. I believe system.drawing.common just works in Blazor

njannink:
generate a base64 image and use that as source on the image

njannink:
I will create a PR when I have it working for you to have a look and then you can decide to incorporate it or not

PR #3411, opened 2022-02-07
#3410 Add QRCode extensions

njannink:
#3410 Initial PR for Blazorise QRCode component


stsrki:
Looking good and it is almost ready to merge but I have noticed the PR was created on the wrong branch so I have moved it to a new PR #3417 as it was faster to do so than deal with merge conflicts.

BTW really good job.

PR #3428, opened 2022-02-11
DateOnly and TimeOnly support

### LINKED ISSUE #3427: DateOnly and TimeOnly support
Add to DateEdit, TimeEdit, DatePicker and TimePicker parsers.

stsrki:
closes #3427


Source/Blazorise/Utilities/Parsers.cs
```diff
@@ -153,6 +153,12 @@ public static bool TryParseDate<TValue>( string value, DateInputMode inputMode,
                 return true;
             }
 
+            if ( type == typeof( DateOnly ) && DateOnly.TryParse( value, out var parsedDto3 ) )
+            {
+                result = (TValue)(object)parsedDto3;
```
David-Moreira, line 158: These seem hella weird. Would this be better or the same?? 
It will still be a boxing operation I guess. Just don't know if more efficient?
Box

David-Moreira (approved):
LGTM

Source/Blazorise/Utilities/Parsers.cs
stsrki: We could use boxing but since we have at this point already parsed the value it is safe to convert it to a TValue.

David-Moreira: I didn't understand. So you're saying you prefer the solution I posted?

stsrki: Aaah I was focused on the screenshot and failed to see that we already use boxing 😂, Yeah, your solution seems better.

stsrki: Or maybe not

image

David-Moreira: What do you mean? In your screenshot I think you are missing the where constraint, that's why it's complaining.

PR #3453, opened 2022-02-15
API: Enum defaults

### LINKED ISSUE #2479: API changes for the enum values
Currently for a lot of enum types like `Size`, `Color`, `TextAlignment` and many others, we have a default enum value of `None`. The value `None` means the CSS class name or some other element style will not be set. For the majority of components `None` is OK and doesn't lead to problems. But in some rare cases, `None` has a different meaning. For example `Display.None` doesn't mean style will not be applied. It means an element will be hidden. This inconsistency I would like to solve and I think it is the right time now to do it before we reach v1.0.

---

What I propose is to convert all enum parameters to nullable, eg.

```cs
public Size? Size { get; set; }
```

This will have the benefit where `null` will mean the parameters are not set, and its class names or styles shouldn't need to be applied. With this approach, `None` can be safely removed from the majority of enum types, and we can leave it only on those that actually have a usable meaning.

---

All of these changes are breaking changes and a lot of files will be affected. But I don't think it will be much of a problem since even now when users set the parameter they don't use `Size.None` or `Color.None`. It is already a default value for all parameters. So converting it to null would not be much of a problem. We would need to document all changes and mention it on the migration guide along with v0.9.4 release notes.

stsrki:
closes #2479


Documentation/Blazorise.Docs/Pages/News/2022-02-01-release-notes-100.razor
```diff
@@ -45,6 +45,9 @@
     <UnorderedListItem>
         <Code>RightAligned</Code> parameter moved from <Code Tag>BarDropdownMenu</Code> to <Code Tag>BarDropdown</Code> component.
     </UnorderedListItem>
+    <UnorderedListItem>
+        Change enum value from <Code>None</Code> to <Code>Default</Code> for the floowing enum types: <Code>Color</Code>, <Code>Background</Code>, <Code>TextColor</Code>, <Code>Alignment</Code>, <Code>BorderRadius</Code>, <Code>BorderSize</Code>, <Code>Direction</Code>, <Code>DisplayDirection</Code>, <Code>FigureSize</Code>, <Code>IconSize</Code>, <Code>JustifyContent</Code>, <Code>OverflowType</Code>, <Code>Target</Code>, <Code>TextAlignment</Code>, <Code>TextOverflow</Code>, <Code>TextTransform</Code>, <Code>TextWeight</Code>, <Code>VerticalAlignment</Code>, <Code>Visibility</Code>, <Code>Size</Code>, and <Code>SnackbarLocation</Code>.
```
David-Moreira, line 49: > floowing enum types?

:D

Demos/Blazorise.Demo/Pages/Tests/ListGroupPage.razor
```diff
@@ -75,7 +75,7 @@
             </CardBody>
             <CardBody>
                 <ListGroup Flush="flush">
-                    <ListGroupItem Color="Color.None">None</ListGroupItem>
+                    <ListGroupItem Color="Color.Default">Default</ListGroupItem>
```
David-Moreira, line 78: Just an unrelated idea... could probably just go every available enum dynamically? Easier to stay up to date.

Documentation/Blazorise.Docs/Pages/News/2022-02-01-release-notes-100.razor
stsrki: 😂

Demos/Blazorise.Demo/Pages/Tests/ListGroupPage.razor
stsrki: Maybe later.

PR #3472, opened 2022-02-17
API: Immediate and Debounce

### LINKED ISSUE #3471: Rename ChangeTextOnKeyPress and DelayTextOnKeyPress
| Current name         | New name  |
|----------------------|-----------|
| ChangeTextOnKeyPress | Immediate |
| DelayTextOnKeyPress  | Debounce |

DebounceInterval instead of DelayTextOnKeyPressInterval

stsrki:
closes #3471


David-Moreira:
LGTM. What was the idea for the name change? 
I don't dislike the previous one haha. My guess is that I need to think... what is `Immediate`, immediate what?

David-Moreira (approved):
lgtm

stsrki:
> What was the idea for the name change?

I was looking at other libraries and they also used the same naming, which I found easier to understand. 

Immediate might not be obvious at first but I think people would get used to it. Better than ChangeTextOnKeyPress :). Unless you come up with a better name I think this is good.

PR #3478, opened 2022-02-17
Dropdown Nested Directions

### LINKED ISSUE #3170: Nested Dropdown Direction
### Discussed in https://github.com/Megabit/Blazorise/discussions/3166

<sup>Originally posted by **itsLollo1000** November 25, 2021</sup>
Hey,

so yet another question of mine.

As I have seen nested dropdowns have been added to Blazorise.
After using them for a bit, what I have noticed that the orientation of the nested ones can not be set, is that true?

Because unfortunately they are certainly limited, which leeds to them being drawn behind the boudaries of the browser window. (overflow isnt set correctly but this would not change a lot)

image

Kind Regards
Sebastian</div>

David-Moreira:
Closes #3170


stsrki (commented):
The last submenu is broken

image

David-Moreira:
I know haha no idea why yet. Maybe zindex or something not sure. Didn't have time yesterday to finish PR. But looks promising, to be able to set direction to every nested menu.

And not sure how we should handle going out of bounds, that'd need js checking I guess... MB just leave it up to the user to make sure he renders with the appropriate directions.

stsrki:
All menus are opening below.

Bulma

image

---

Bootstrap 4

image

---

Bootstrap 5

image

David-Moreira:
Cache...? xD

David-Moreira:
Can you take a look at Bulma? There doesn't really exist a left/right direction, it's more of an aligment... Should we leave as it is?

stsrki:
Will you be able to do the same for the BarDropdown?

Demos/Blazorise.Demo/Pages/Tests/ButtonsPage.razor
```diff
@@ -226,21 +226,21 @@
                 <CardTitle>Dropdown Nesting</CardTitle>
             </CardHeader>
             <CardBody>
-                <Dropdown>
+                <Dropdown Direction="Direction.End" Display="Display.InlineBlock">
                     <DropdownToggle Color="Color.Primary">Level 1</DropdownToggle>
                     <DropdownMenu>
                         <DropdownItem>Item 1.1</DropdownItem>
-                        <Dropdown>
+                        <Dropdown Direction="Direction.End">
```
stsrki, line 233: I think submenus to the right(End) should be the default option.

David-Moreira: How so? If nested menus default is right, but if main menu default is down?

David-Moreira:
> Icons are not aligned as before
> 
> image
> 
> image

AH goddamn css... :X

Demos/Blazorise.Demo/Pages/Tests/ButtonsPage.razor
stsrki: I think we had diferent CSS selectors for submenus.

David-Moreira: What do you mean? Since we only did a single direction before for nested menus we had custom css to just make everything go right.

But we can make the default right, the solution will be to apply the dropright classes when direction is default. This is preferable as classless drop down now corresponds to down direction basically, so that should stay as it is now.

stsrki:
That was my idea. When the direction is undefined it needs to behave the same as before, to the right. And then if users want to change it they need to set the direction.

David-Moreira:
> > Icons are not aligned as before
> > image
> > image
> 
> AH goddamn css... :X

I can't reproduce, which provider?

David-Moreira:
We can try `BarDropdown` for 1.0.0 but I'm not sure I have much time this weekend.

stsrki:
Now, these are broken 😂

image

YOu can ignore BarDropdown for this release. Focus on making regular Dropdown work. Or we can leave it for v1.1.

stsrki (approved):
Looking good now.

PR #3482, opened 2022-02-20
PWA && Offline Support Docs

### LINKED ISSUE #3379: Js modules vs PWA and offline capabilities
> We are also aware that there might need to be extra setup when dealing with PWA and offline capabilities if you want your app to remain responsive. We are working on providing you with clear steps for these use cases.

David-Moreira:
Closes #3379


stsrki (commented):
Looks good.

I would also make a PWA guide on a new page and then add a URL pointing to it from various places.

Example

image

PR #3493, opened 2022-02-21
DataGrid : Handle Delete Edge Case : SelectedRow being edited is deleted

### LINKED ISSUE #3480: Datagrid selection stops working
**Describe the bug**
Datagrid selection stops working

**To Reproduce**
1. Go to https://bootstrapdemo.blazorise.com/tests/datagrid
2. Change  Command Mode  to Button Row
3. Select a row
4. Click Edit button
5. Click Delete button
6. Try to select a line. =====> This wont work <=====
7. Click New button
8. Click Cancel button
9. Try to select a line. This works again

I believe selection gets turned off when in edit mode, and will only come back after the user explicitly clicks Cancel button. If he deletes the line without canceling the Edit, selection stops working.

**Expected behavior**
When the edit mode is finished by any action other than by clicking the Cancel button, the selection option should work again.

David-Moreira:
Closes #3480

### DIFF
--- a/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
+++ b/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
@@ -441,6 +441,9 @@ namespace Blazorise.DataGrid
                     {
                         if ( data.Contains( item ) )
                             data.Remove( item );
+
+                        if ( editState == DataGridEditState.Edit && SelectedRow.IsEqual( item ) )
+                            editState = DataGridEditState.None;
                     }
 
                     await RowRemoved.InvokeAsync( item );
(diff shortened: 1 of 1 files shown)


stsrki:
> I was thinking... included also:
> 
> * SelectedRow being edited is deleted sets edit to none even if UseInternalEditing is set to false...
>   
>   * I don't think it makes sense to be kept in edit mode here, so we are also covering that, Agree?
> * If UseInternalEditing && If Item being deleted is selected.
>   
>   * It now de-selects the item as it does not make sense to be kept selected if removed && if user is using Datagrid's internal management.

Makes sense to do so.

stsrki (approved):

PR #3495, opened 2022-02-22
Allow Sort Template when There is No Sort Direction, and Display a Default Sort Icon on Sortable Columns

### LINKED ISSUE #3455: Eliminate Check in DataGrid that Prevents a `SortDirectionTemplate` from Displaying when `column.CurrentDirection == SortDirection.None`
## Problem I'm Trying to Solve
Remove a check for `SortDirection.None` in `Source/Extensions/Blazorise.DataGrid/DataGrid.razor`. The `if` statement won't allow the `SortDirectionTemplate` to render when the sort direction is `SortDirection.None`. This prevents implementers from specifying a custom icon for when there is no sort (but sorting is available). 

## Possible Solutions
Change Line 53 (Source/Extensions/Blazorise.DataGrid/DataGrid.razor) in the following manner:
```blazor
@if ( Sortable && column.Sortable && column.CurrentDirection != SortDirection.None )

@* TO *@

@if ( Sortable && column.Sortable )
```

## Additional Context
This doesn't appear to be a breaking change, since the only checks are against the SortDirection context. This would just allow users to check the context against `SortDirection.None` to display an icon when unsorted.

P.S. Happy to make the change and do the PR! LMK!

gkochera:
Edited Datagrid component to allow a SortTemplate to render when `column.CurrentDirection == SortDirection.Default`. Additionally, columns that are sortable will now display the FontAwesome Sort icon `IconName.Sort`.

Passed all unit tests, visually inspected performance in the Bootstrap 5 Demo Client. Works as intended.

Resolves #3455

### DIFF
--- a/Source/Extensions/Blazorise.DataGrid/DataGrid.razor
+++ b/Source/Extensions/Blazorise.DataGrid/DataGrid.razor
@@ -53,7 +53,7 @@
                                             @column.Caption
                                         }
                                     }
-                                    @if ( Sortable && column.Sortable && column.CurrentSortDirection != SortDirection.Default )
+                                    @if ( Sortable && column.Sortable )
                                     {
                                         @if ( column.SortDirectionTemplate != null )
                                         {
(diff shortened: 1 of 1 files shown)


stsrki:
Tested and looking good. Thank you for taking the time to help with this feature!

gkochera:
You're welcome! Excited for the future of this project.

PR #3502, opened 2022-02-22
MemoEdit pattern

### LINKED ISSUE #3501: Pattern Validation for MemoEdit
My TextEdits and MemoEdit do not use @text-Bind; rather, it uses Text and therefore instead of utilizing data annotations for my pattern validations, I have to use Manual Validation and using the Pattern attribute for the fields. This works great for everything BUT the MemoEdit. 

Request that MemoEdit utilizes the Pattern attribute for Manual Validation. 

This is only a problem because we are using a data grid where you add/edit items using a modal. When making changes I. The modal while @text-Bind is being used, the changes reflect in the data grid and do not clear those changes out on cancellation of the modal.

stsrki:
closes #3501


David-Moreira (approved):
Same implementation as was working on TextEdit so... **LGTM.**

PR #3506, opened 2022-02-23
Drag & Drop component

### LINKED ISSUE #2633: Drag and drop functionality
Hi,

It would be nice to have a drag and drop functionality. So developers can easily build interactive functionality for websites.

stsrki:
closes #2633


David-Moreira:
If I drag something to an invalid drop zone, it stays invalid even after I stop dragging. Also I can't drag anything out of that invalid zone.

image

David-Moreira:
Have we testes with actual components? Like Texts and other inputs? Theoretically everything should work just fine. But I guess we should test just in case.

stsrki:
> If I drag to a valid zone, but drop it on the background instead. I can't drag anything on that valid zone until I go interact with the other zones.

Can't reproduce this.

David-Moreira:
> > If I drag to a valid zone, but drop it on the background instead. I can't drag anything on that valid zone until I go interact with the other zones.
> 
> Can't reproduce this.

Drag from middle group and hover left group then drop it on the background.
Then go to the left group and try dragging an item, it will not be draggable.

stsrki:
Still cannot reproduce. Might be fixed with the last commit.

PR #3511, opened 2022-02-25
Update start guides according to .NET 6 templates

### LINKED ISSUE #3485: Adjust start guide and usage documentation
The newest Blazor templates come with top-level statements and because of that our code samples are not very usable. We need to adjust them according to the new MS guidelines.

stsrki:
closes #3485


David-Moreira:
Worth leaving both versions or too confusing?
Update demos? 😇

stsrki:
Hmm well we might leave old version but considering we now only have NET6 maybe itnis not worth it. And who ever uses old versions they already have it set up.

That being said. Don't touch the demos 👊😂

David-Moreira (approved):

PR #3562, opened 2022-03-10
FocusTrap component

### LINKED ISSUE #3550: Add trap focus to Modal
https://uxdesign.cc/how-to-trap-focus-inside-modal-to-make-it-ada-compliant-6a50f9a70700

stsrki:
closes #3550

This PR introduces a new `FocusTrap` component. It is inspired by the work from https://github.com/Blazored/Modal with some modifications.

`FocusTrap` is also made an internal part of the modal component. Since it basically wrapping the content of the modal with multiple `divs` I'm still not sure if this is a good approach or should we make it with https://github.com/focus-trap/focus-trap


Source/Blazorise/Components/FocusTrap/FocusTrap.razor.cs
```diff
@@ -0,0 +1,84 @@
...
+            {
+                await endFirstRef.FocusAsync();
+            }
+        }
+
+        private void HandleKeyPresses( KeyboardEventArgs args )
+        {
+            if ( args.Key == "Tab" )
+            {
+                _shiftPressed = args.ShiftKey;
```
David-Moreira, line 64: No biggy, but maybe rename to _shiftTabPressed? Which is what it represents? I was reading the code and I was like, wth is this _shiftPressed thing until I got to this method.

Source/Blazorise/Components/Modal/ModalContent.razor.cs
```diff
@@ -75,6 +108,21 @@ internal void NotifyHasModalFooter()
         /// </summary>
         protected virtual bool AsDialog => false;
 
+        /// <summary>
+        /// True if modal should keep the input focus.
+        /// </summary>
+        protected bool ActivateFocusTrap => Options?.ActivateModalFocusTrap == true;
```
David-Moreira, line 114: I'm sure someone will eventually want to configure this per modal, can we have it as a Parameter also?

David-Moreira:
Well seems to work well, I'd much prefer to have only c# if possible haha Thing is, we might be missing stuff that the library solves and we just haven't noted yet.

I guess we can leave it implemented with c# and if something really complex comes up, we'll implement focus-trap js if really neeeded.

I guess modals are the most common, but what other use cases are there? Maybe focus trap on a formulary div? Have we tried/tested that?

Source/Blazorise/Components/FocusTrap/FocusTrap.razor.cs
stsrki: Haha yeah you're right. I thought the same when I seen it but then forgot about it.

Source/Blazorise/Components/Modal/ModalContent.razor.cs
stsrki: Yep

stsrki:
> Well seems to work well, I'd much prefer to have only c# if possible haha Thing is, we might be missing stuff that the library solves and we just haven't noted yet.
> 
> I guess we can leave it implemented with c# and if something really complex comes up, we'll implement focus-trap js if really neeeded.
> 
> I guess modals are the most common, but what other use cases are there? Maybe focus trap on a formulary div? Have we tried/tested that?

Might we worth on tabs maybe. But we can leave that for users to decide.

What is formulary div?

David-Moreira:
Well. Just a formulary? Maybe someone could have some formulary on the page that given a certain criteria it would focus trap on it until the user ends submitting for formulary? Does that make sense?

stsrki:
Yeah, kinda understand it now. Makes sense.

Source/Blazorise/Components/Modal/ModalContent.razor.cs
stsrki: Added

stsrki:
> I guess modals are the most common, but what other use cases are there? Maybe focus trap on a formulary div? Have we tried/tested that?

I have made the public method `SetFocus()` so if anyone wishes to implement `FocusTrap` they would be able to do it, same as we have done it on Modal.

Source/Blazorise/Components/FocusTrap/FocusTrap.razor.cs
```diff
@@ -38,26 +38,45 @@ protected override async Task OnAfterRenderAsync( bool firstRender )
...
             }
         }
 
-        private void HandleKeyPresses( KeyboardEventArgs args )
+        /// <summary>
+        /// Handles the focus start event.
+        /// </summary>
+        /// <param name="args">Supplies information about a keyboard event that is being raised.</param>
+        /// <returns>A task that represents the asynchronous operation.</returns>
+        protected virtual void OnKeyPressesHandler( KeyboardEventArgs args )
```
David-Moreira, line 79: Is the method documentation right? I don't know if it's github change that's displaying incorrectly. Seems like it has the Focus documentation instead.
Should it be `KeyPressesHandler` or `KeyPressesHandler`?

Source/Blazorise/Components/FocusTrap/FocusTrap.razor.cs
```diff
@@ -38,26 +38,45 @@ protected override async Task OnAfterRenderAsync( bool firstRender )
             }
         }
 
-        internal async Task SetFocus()
+        /// <summary>
+        /// Sets the focus trap.
+        /// </summary>
+        /// <returns>A task that represents the asynchronous operation.</returns>
+        public async Task SetFocus()
```
David-Moreira, line 45: So there's a `SetFocus` so we're trapper inside a container for tabbing/focus right? What if I want to get out of the "trap"?

Source/Blazorise/Components/FocusTrap/FocusTrap.razor.cs
stsrki: Yep, it's a typo.

Source/Blazorise/Components/FocusTrap/FocusTrap.razor.cs
stsrki: Then `Active` would be set to `false`.

David-Moreira: So `Active` + `SetFocus` does nothing
`Active` + `SetFocus` sets the trap, and if I want to disable after, I just set `Active` to false, that's it, right?

Source/Blazorise/Components/FocusTrap/FocusTrap.razor.cs
David-Moreira: Wait did I just made a type too when asking? haha I meant 
Should it be `KeyPressesHandler` or `KeyPressedHandler`?

Source/Blazorise/Components/FocusTrap/FocusTrap.razor.cs
stsrki: `Active` will basically enable tabindex in the wrapper components. But it will not do anything to actually focus the input inside of its content. With `SetFocus` we force it to focus on the wrapper which then keeps track of everything else, tab press, leave event, etc.

Source/Blazorise/Components/FocusTrap/FocusTrap.razor.cs
stsrki: You did :D

Source/Blazorise/Components/FocusTrap/FocusTrap.razor
```diff
@@ -1,6 +1,6 @@
 ﻿@namespace Blazorise
 @inherits BaseComponent
-<div @ref="@containerRef" @onkeydown="@OnKeyPressesHandler" @onkeyup="@OnKeyPressesHandler">
+<div @ref="@containerRef" @onkeydown="@OnKeyPressedHandler" @onkeyup="@OnKeyPressedHandler">
```
David-Moreira, line 3: Why both keydown and keyup, will this not double trigger the Handler?

stsrki: Yes is it true that it will raise two events for the same key; But, I assume it is OK as Chris hopefully knew better at the time 😅

stsrki:
@David-Moreira can you try it once you have time? FocusTrap is now responsible for focusing on nested input and it works by setting the `Active` parameter. The `SetFocus` still stays for scenarios where we need to manually focus.

David-Moreira:
Yea I'll let you know once I test it.

David-Moreira:
I know this line is kinda generic whenever you want to check if a parameter has changed
image
but for boolean is probably simpler and easier to read by having `Active == false`?
Does not make a difference if it stays at it is eitherway.

David-Moreira:
New focus trap demo page is exactly what I saying. Works simple & good. :)

David-Moreira (approved):

stsrki:
> 



> 

I want to be consistent across code. That's why it is the same for booleans.

David-Moreira:
> > 
> 
> > 
> 
> I want to be consistent across code. That's why it is the same for booleans.

In that case maybe consider using a method or extension? HasParameterChanged or smthg.

PR #3570, opened 2022-03-11
Fix Markdown Image Upload: Allow Images with Sizes > 0 bytes

Knuhl:
#3213 broke Image-Upload in the `Markdown`-Control: It added the new parameter `long maxFileSize` to both `RemoteFileEntryStreamReader` and `RemoteFileEntryStream`, but the parameter is always set to `0` from the Markdown-Control. 

This Check in both `RemoteFileEntryStreamReader` and `RemoteFileEntryStream` now prevents any Image-Upload through the Markdown-Control:
``` csharp
if ( maxFileSize < fileEntry.Size )
{
    await fileEntryNotifier.UpdateFileEndedAsync( fileEntry, false, FileInvalidReason.MaxLengthExceeded );
    return;
}
```

This fix now uses the `ImageMaxSize` Property of the Markdown-Control for `maxFileSize`.

**To Reproduce**
Any Markdown Image-Upload using `file.WriteToStreamAsync` or `file.OpenReadStream` should fail. The Image-Upload-Example at https://blazorise.com/docs/extensions/markdown also currently fails.

Maybe the intention behind `maxFileSize: 0` was "unlimited"?


David-Moreira:
Strange seems like I missed it.
Yea seems to work just fine with your fix. 

JS error handling doesn't seem quite good though, throws on easymde javascript
@stsrki any idea?

image

stsrki:
> JS error handling doesn't seem quite good though, throws on easymde javascript

I cannot reproduce this. How did you do it?

David-Moreira:
By attempting upload of a file bigger then the set max size.

stsrki:
Probably something internal to `easymde`. I cannot catch it in `errorCallback`. We can ignore it for now I think.

PR #3574, opened 2022-03-11
DataGrid: show readonly Check when DataGridCheckColumn

### LINKED ISSUE #2897: DataGridCheckColumn - render it as a readonly checkbox by default
Perhaps a logic of:

if no string formatter is set, show a checkbox

OR

if a bool RenderAsCheck is true, show a box, otherwise show a text representation of the value, given by the formatter

stsrki:
closes #2897


David-Moreira:
Yea I guess that'll look better by default even if internals aren't as nice. :P

David-Moreira (approved):

PR #3576, opened 2022-03-12
Carousel : Periodic Timer : First Attempt

David-Moreira:
Can you elaborate on `public PeriodicTimer TransitionTimer { get; set; }`?

When replacing the `TransitionTimer` `Timer` with `PeriodicTimer` it kinda didn't work well, but when looking at the actual code, it seems to be that the `OnTransitionTimerEvent` can just be done as part of the standard `Timer` event, no?

The change below that adds the same code as in the `OnTransitionTimerEvent ` to the main `RunAnimations` ... seems to have an equivalent result, but without the use of the `TransitionTimer`, but I'm not sure. Sometimes stuff are there for a reason... do you know?
Sem Título


stsrki:
> 

TBH I really don't remember.

Source/Blazorise/Components/Carousel/Carousel.razor.cs
```diff
@@ -269,79 +257,71 @@ private async void OnLocalizationChanged( object sender, EventArgs eventArgs )
...
         }
 
-        private void SetTimer()
+        private Task InitializeTransitionTimer()
         {
-            TimerEnabled = ( Interval > 0 );
+            TransitionTimer = new( TimeSpan.FromMilliseconds( 2000 ) );
 
-            if ( Timer == null && TimerEnabled )
+            Task.Run( async () =>
```
stsrki, line 279: Do we still need `Task.Run`?

Source/Blazorise/Components/Carousel/Carousel.razor.cs
```diff
@@ -429,7 +408,17 @@ protected virtual async Task RunAnimations()
 
             await InvokeAsync( StateHasChanged );
 
-            ResetTransitionTimer();
+            await ResetTransitionTimer();
+
+            await Task.Delay( TimeSpan.FromMilliseconds( 2000 ) );
```
stsrki, line 413: Is this delay `await Task.Delay( TimeSpan.FromMilliseconds( 2000 ) );` for debugging only?

Source/Blazorise/Components/Carousel/Carousel.razor.cs
```diff
@@ -269,79 +257,71 @@ private async void OnLocalizationChanged( object sender, EventArgs eventArgs )
...
-                InitializeTimer();
-            }
-
-            if ( AutoPlayEnabled )
-            {
-                Timer.Start();
-            }
+                while ( await TransitionTimer.WaitForNextTickAsync() )
+                {
+                    //await OnTransitionTimerEvent( this, null );
```
stsrki, line 283: This is commented on purpose?

Source/Blazorise/Components/Carousel/Carousel.razor.cs
David-Moreira: Well since instead of callbacks, it setups a while.loop it needs to be running on a background thread. Suggestions?

Source/Blazorise/Components/Carousel/Carousel.razor.cs
David-Moreira: Yea on purpose I replaced the transition timer calbsck with the new call on run animations... So basically. It seems we don't need transition timer at all but I'm. Not sure...

Source/Blazorise/Components/Carousel/Carousel.razor.cs
David-Moreira: Transition timer was running in a 2s  interval, so while not exactly. The same it's pretty close to what it was.... I guess I must further investigate why transition timer exists and is called on a 2s interval...

Source/Blazorise/Components/Carousel/Carousel.razor.cs
stsrki: I think just having `while ( await Timer.WaitForNextTickAsync() )` will work.

David-Moreira: Am I missing something? The while is a sync operation it'll lock execution of the method until completed...

David-Moreira: Exactly my point. the Outter method (`OnInitializedAsync`) will not return until the timer is disposed of. Now I don't know if that has negative consequences or not.

stsrki: Are you sure it will not return? I need to confirm it....

David-Moreira: Pretty sure... just debug it. Why would it, again while is a sync operation.

Source/Blazorise/Components/Carousel/Carousel.razor.cs
stsrki: I have tested and you're right. It doesn't pass on with initializing cycle. And I found another even worse problem. if you navigate away from the carousel page the timer is still called every time, even though it is disposed of.

stsrki: God damn it https://docs.microsoft.com/en-us/dotnet/api/system.threading.periodictimer.waitfornexttickasync?view=net-6.0#parameters

David-Moreira: So what about just returning to the previous background implementation?

stsrki: I think it is best to just not use PeriodicTimer and just leave it as it is now.

David-Moreira: Right, that also went through my mind.
Altough the implementation was starting to look a lot cleaner.

PR #3584, opened 2022-03-14
Moved padding for buttons to default message options 

wexman:
... so you can override these via the button class properties (which isn't possible currently)


wexman:
I just realized my solution was kind of bootstrap-only, didn't take into account the other ui libraries. Then I guess we should have DefaultPadding properties for the message dialog buttons ...

PR #3587, opened 2022-03-15
2nd try: making message dialog button padding customizable

wexman:
See title

### DIFF
--- a/Source/Blazorise/Components/MessageAlert/MessageAlert.razor.cs
+++ b/Source/Blazorise/Components/MessageAlert/MessageAlert.razor.cs
@@ -230,6 +230,22 @@ namespace Blazorise
             => Options?.CancelButtonClass;
 
         /// <summary>
+        /// Gets the OK button padding.
+        /// </summary>
+        protected virtual IFluentSpacing OkButtonPadding
+            => Options?.OkButtonPadding ?? Blazorise.Padding.Is2.OnX;
+        /// <summary>
+        /// Gets the cancel button padding.
+        /// </summary>
+        protected virtual IFluentSpacing CancelButtonPadding
+            => Options?.CancelButtonPadding ?? Blazorise.Padding.Is2.OnX;
+        /// <summary>
+        /// Gets the confirm button padding.
+        /// </summary>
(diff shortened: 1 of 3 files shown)


Source/Blazorise/Components/MessageAlert/MessageAlert.razor.cs
```diff
@@ -193,6 +195,18 @@ protected virtual Color OkButtonColor
         protected virtual string OkButtonClass
             => Options?.OkButtonClass;
 
+        /// <summary>
+        /// Gets the css class for the title.
+        /// </summary>
+        protected virtual string TitleClass
+            => Options?.TitleClass ?? "";
```
stsrki, line 202: Let it just return null instead of empty string. Eg. `Options?.TitleClass`.

Source/Blazorise/Components/MessageAlert/MessageAlert.razor.cs
```diff
@@ -229,57 +243,85 @@ protected virtual Color CancelButtonColor
...
+        /// Gets the confirm button padding.
+        /// </summary>
+        protected virtual IFluentSpacing ConfirmButtonPadding
+            => Options?.ConfirmButtonPadding ?? Blazorise.Padding.Is2.OnX;
+
         /// <summary>
         /// Gets or sets the <see cref="IMessageService"/> to which this dialog is responding.
         /// </summary>
-        [Inject] protected IMessageService MessageService { get; set; }
+        [Inject]
```
stsrki, line 267: If get set are on the same line we prefer to also have attributes on the same line.

stsrki (approved):
Fixed some formatting but overall it is a good PR. Thanks for your help!

stsrki:
It is already merged as part of this PR.

stsrki:
Yes, That's why you need to create separate branches for any features. See our guidelines https://github.com/Megabit/Blazorise/blob/master/CONTRIBUTING.md#branch-organization

Now it was OK since it is a small change but usually, we don't allow merging from the master.

wexman:
I see - sorry then ;-)

I will know for the next time.

PR #3590, opened 2022-03-15
Route Fix

David-Moreira:


stsrki:
Why the error did not say the route was duplicate? Is this changed in .NET6??? 🤦‍♂️

Thanks for fixing it.

David-Moreira:
> Why the error did not say the route was duplicate? Is this changed in .NET6??? 🤦‍♂️
> 
> Thanks for fixing it.

It does say it, but on `Server` rendermode. Certaintly something technical with `ServerPreRendered` I guess...

PR #3591, opened 2022-03-16
Trendline bugfix: don't call chart.update() after you add the trendlines

ledpup:
…. Doing so causes the line data to be rendered in the wrong part of the grid.


stsrki:
Closed in favor of #3644 because this PR was done on the wrong branch. 

PS. thanks for fixing it.

PR #3609, opened 2022-03-18
Add strict version to JS imports

### LINKED ISSUE #3608: The requested module './utilities.js' does not provide an export named 'firstNonNull'
When I upgrade our Blazor Server application from Blazorise 0.9.5.5 to 1.0.1 I get the following error when using the new NumericPicker component: "The requested module './utilities.js' does not provide an export named 'firstNonNull'"

This seems to be a caching issue when using JavaScript import statement. As the browser already have a cached version of utilities.js (0.9.5.5) it will not get the latest version of utilities.js (1.0.1) and will therefore give an error when trying to import the new "firstNonNull"-function. 

The workaround is to start a new browser session or make sure "Disable cache" under Developer Tools is checked, but I don't want to force the users of our application to do that. I think this is generic issue for all JavaScript import statements that do not include the version number, as a cached version could be used.

image

This import does not have a ?v=1.0.1.0 parameter and will therefore will use the cached version and will give an error when trying to use the new "firstNonNull"-function.

image

stsrki:
closes #3608

Trying to come up with caching problem if JS modules imported from another JS module or file. This approach manually adds a version number through a query param. It works but the downside is that we would have to manually update them with each release.

Any suggestion is welcome.


WillianGruber:
> closes #3608
> 
> Trying to come up with caching problem if JS modules imported from another JS module or file. This approach manually adds a version number through a query param. It works but the downside is that we would have to manually update them with each release.
> 
> Any suggestion is welcome.

You can try to run pre-compilation steps to automatically increase the number

PR #3614, opened 2022-03-19
Modal Provider

### LINKED ISSUE #2986: way to open modals programmatically
something that doesn't need to explicitly put the modal on the page and thus refer only to the class and its parameters.

example:
https://www.matblazor.com/DialogService

David-Moreira:
Closes #2986 
Here's a simple POC for the new ModalProvider. This should basically be the core of how it works.

I guess we can go from here, agree?


Source/Blazorise/Components/ModalProvider/ModalProvider.razor.cs
```diff
@@ -0,0 +1,131 @@
...
+    /// <summary>
+    /// A modal provider to be set at the root of your app, providing a programatic way to invoke modals with custom content by using ModalService.
+    /// </summary>
+    public partial class ModalProvider
+    {
+        #region Members
+
+        private Modal modalRef;
+        private RenderFragment childContent;
+        private bool useModalDefinition = true; //TOOD : Choose a better name
```
stsrki, line 24: This will probably be set through the options, right?

stsrki:
By the look of it seems good to me. I like that we will have the freedom of choosing a standard header/body/footer modal and then rendering custom content. It's a good start. We also need to have many other options to set modal title, buttons colors, classes, styles, etc. Similar to what we already have on MessageAlert.

Source/Blazorise/Components/ModalProvider/ModalProvider.razor.cs
David-Moreira: Ye. Global options and then overriden options when calling `Show`

Source/Blazorise/Components/ModalProvider/ModalProvider.razor.cs
```diff
@@ -34,37 +35,110 @@ public partial class ModalProvider
...
+        /// Gets or Sets the modal's title.
+        /// </summary>
+        public string Title { get; set; }
+
+    }
+
+    /// <summary>
+    /// Provide the parameters to be passed onto your custom content to be rendered under the modal.
+    /// </summary>
+    public class ModalComponentParameter
```
stsrki, line 87: Name if ModalProviderParameter? Also, make it a struct instead.

stsrki (commented):
So, the purpose of parameters is to populate any custom parameter type typeof(SomeComponent) can have? Why not just use dictionary `<string, object>`?

Source/Blazorise/Components/ModalProvider/ModalProvider.razor.cs
David-Moreira: Maybe ModalProviderComponentParameter or content? The idea is that it represents the user custom component parameters. I guess if we don't need additional behaviour or helper methods we can make the dictionary.

The advantage is that if in the future we need to add behaviour for some reason we don't need to change the api.

Structure change seems fine. I always forget nowadays haha

David-Moreira:
I think it might be close, what do you think of the provided APIs?

I'm gonna start doing demos next to test it.

Demos/Blazorise.Demo/App.razor
```diff
@@ -14,7 +14,7 @@
     <MessageAlert />
     <NotificationAlert />
     <PageProgressAlert />
-    <ModalProvider />
+    <ModalProvider UseModalStructure />
```
stsrki, line 17: I think `UseModalStructure` should be passed with the options. That way users can have better control over individual uses cases.

David-Moreira: That's the global options. Isn't it passed on the options?  If it isn't I meant to.

David-Moreira: Yea it seems like it is.
`protected virtual bool GetUseModalStructure => ModalProviderOptions?.UseModalStructure ?? UseModalStructure;`

stsrki: I missed it, sorry. It's all good.

stsrki: I missed it, sorry. It's all good.

Source/Blazorise/Components/ModalProvider/ModalProvider.razor.cs
```diff
@@ -0,0 +1,182 @@
...
+
+        /// <summary>
+        /// Gets or sets whether the component has any animations.
+        /// </summary>
+        [Parameter] public bool Animated { get; set; } = true;
+
+        /// <summary>
+        /// Gets or sets whether the component has any animations.
+        /// </summary>
+        protected virtual bool GetAnimated => ModalProviderOptions?.Animated ?? Animated;
```
stsrki, line 147: It's strange to have a `Get` prefix on properties. Maybe better convert to function in that case. Or change the naming convention.

David-Moreira: Properties with lambda are pretty much functions xD . 
Suggestions for naming convention? I don't think it's that bad...

David-Moreira: But I just noticed the doc is wrong "Gets or sets" :D

stsrki: I know that every property is actually a function internally, but it is disguised as a field.
 
You can leave it with `Get` prefix but make a real function: `protected virtual bool GetAnimated() => ModalProviderOptions?.Animated ?? Animated;`

David-Moreira:
SO I was going through @drma-dev request/suggestion 

And I noticed we missed something important, the provider should handle multiple modals. Next thing to do.

>I already do these suggestions you gave me. but let me give you two examples that can justify this type of functionality:

>1 - a page can have several modals (maybe even dozens). this makes the page layout a mess with so much. plus having to load all this code on every page load. having a service like this I would just open what the user requested.

>2 - I have a page that has a modal A, and this modal A opens a modal B. If from modal B I want to open the A again, I can't, because the blazor understands this as a cyclic reference. so if I open this modal by code, the blazor will have no reason to complain.

David-Moreira:
Any suggestions for examples using this feature?

stsrki:
> Any suggestions for examples using this feature?

Just create a simple component with a few input fields. Might also have a couple of parameters to receive.

David-Moreira:
Done the multiple modals handling, how do you think we should handle the passage of most BaseComponent attributes? Should they all just be on the ModalProviderOptions? And then grab from ModalProvider or from ModalProviderOptions, accordingly?

Source/Blazorise/Components/ModalProvider/ModalProvider.razor.cs
```diff
@@ -0,0 +1,149 @@
...
+{
+
+    /// <summary>
+    /// A modal provider to be set at the root of your app, providing a programmatic way to invoke modals with custom content by using ModalService.
+    /// </summary>
+    public partial class ModalProvider : BaseComponent
+    {
+        #region Members
+
+        private List<ModalInstance> modalInstances;
```
stsrki, line 23: Might be better to use `Stack<>`?

Source/Blazorise/Components/ModalProvider/ModalProvider.razor
```diff
@@ -0,0 +1,39 @@
+﻿@namespace Blazorise
+@inherits BaseComponent
+@using Blazorise.Extensions
+
+<CascadingValue Value="@this" IsFixed>
+    @if (!modalInstances.IsNullOrEmpty())
+    {
+        @foreach (var modal in modalInstances)
+        {
+            <Modal @ref="modal.ModalRef" @bind-Visible="modal.Visible" Attributes="modal.Attributes" Animated="modal.Animated()" AnimationDuration="modal.AnimationDuration()" Background="modal.Background" Border="modal.Border" Casing="modal.Casing" Class="modal.Class"
```
stsrki, line 10: Add a `@key` for optimization.

Source/Blazorise/Components/ModalProvider/ModalInstance.cs
```diff
@@ -0,0 +1,120 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Components;
+#endregion
+
+namespace Blazorise
+{
+    /// <summary>
+    /// Generates a new instance of a Modal
+    /// </summary>
+    public class ModalInstance : BaseComponent
```
stsrki, line 12: ModalInstance is only used to hold information about the modal, right? If true, then it doesn't need to inherit from BaseComponent and we don't need to have all of the BaseComponent parameters.

Source/Blazorise/Components/ModalProvider/ModalProvider.razor.cs
David-Moreira: Actually I don't recall if I setup the API, but the Show returns the ModalInstance, so the user could use the service to close it, so there should be a lookup to close a certain modal instance. In that case the List would make more sense in my opinion. I'll look it up, also see if it even works properly (Closing other instances)

Source/Blazorise/Components/ModalProvider/ModalProvider.razor
David-Moreira: Right... damn memory... there should be some kinda of intelisense for that once you use a foreach haha

Source/Blazorise/Components/ModalProvider/ModalInstance.cs
David-Moreira: Right, It was kinda of a "shortcut" to having all the base component attribute gibberish inherited. That way you don't have to manually update it every time. But since we'll have to add them manually to ModalProviderOptions... I guess we might as well do the same here.

Source/Blazorise/Components/ModalProvider/ModalProvider.razor.cs
David-Moreira: Yep, added an Hide(modalInstance), my guess is that it won't have much use, but it's there if some use case comes up.
So I guess we should keep the List?

Source/Blazorise/Components/ModalProvider/ModalInstance.cs
David-Moreira: So added a ModalProviderBaseAttributes with all the attributes, can you take a look and remove the one's that might not make sense / do anything at all on Modal? You certainly know this better/faster then me.

Source/Blazorise/Components/ModalProvider/ModalProvider.razor.cs
stsrki: I don't see a scenario where the user could close the previous modal. of always goes from the front visible to the back.

Source/Blazorise/Components/ModalProvider/ModalInstance.cs
stsrki: Will do.

Source/Blazorise/Components/ModalProvider/ModalProvider.razor
stsrki: Well, ask for the feature on the DotNetCore repo. Maybe they will listen 😆

Source/Blazorise/Components/ModalProvider/ModalProvider.razor.cs
David-Moreira: UHm there's always someone ingenious haha...
maybe some warning modal that generates another modal and then it can make sense to remove the warning modal sooner...
Think we should just remove that API, go with the stack?

stsrki: Yeah, I guess the warning modal would mess things up if it shows behind the currently open modal. OK in that case we'll stick with the List.

stsrki:
Apart from my failed optimizations(that aren't able to be done until some extra API changes in v2), it seems to be working quite nicely. I don't think there is much left to add.

---

There is one thing, though, that is problematic with the regular Modal also. When multiple modals are opened, then the backdrop stays at the same level as the first modal backdrop. It doesn't cover the previously opened modal. I think we should add a dynamic detection of the previously opened modal (also its backdrop) `z-index` and then assign it to the currently opened modal, but with + 1. What do you think?

David-Moreira:
Right we can try that, leave it open for the time being.

Know of any css trick? Or most likelly gotta use js, right?

stsrki:
I think only the JS will work. You can do it on another PR since it is more of a general modal problem.

David-Moreira:
Kk fine better make sure we open an issue then. I think we are still missing docs here I'll get on it.

stsrki:
@David-Moreira, I think it is ready for merge. Unless you have more work to do?

David-Moreira:
Think so too. Works pretty good. New features will come from user requests now certainly. :)

PR #3631, opened 2022-03-25
Fix dropdown inside responsive table

### LINKED ISSUE #3605: Dropdown positioned below datagrid rows
**Describe the bug**
When using responsive `DataGrid` with `Dropdown` component inside, the menu is positioned below datagrid rows. The second problem I see here is that it shouldn't stay behind datagrid container. I mean that there is the same problem as was in #1675.

**Expected behavior**
1. Dropdown is positioned next to the Actions button.
2. Dropdown doesn't stay behind datagrid container.

**To Reproduce**
```
<Container Fluid Style="border: solid red;">
    <DataGrid TItem="int" Data="new List<int>{1,2,3}" Responsive="true" Bordered="true" Style="border: solid blue;">
        <ChildContent>
            <DataGridColumn Caption="Id">
                <DisplayTemplate>@context</DisplayTemplate>
            </DataGridColumn>
            <DataGridColumn>
                <DisplayTemplate>
                    <Dropdown>
                        <DropdownToggle Color="Color.Primary">Actions</DropdownToggle>
                        <DropdownMenu>
                            <DropdownItem>Do something 1</DropdownItem>
                            <DropdownItem>Do something 2</DropdownItem>
                            <DropdownItem>Do something 3</DropdownItem>
                        </DropdownMenu>
                    </Dropdown>
                </DisplayTemplate>
            </DataGridColumn>
        </ChildContent>
    </DataGrid>
</Container>
```

**Screenshots**
image

David-Moreira:
Closes #3605

Reported issues:
> 1. Dropdown is positioned next to the Actions button.
> 2. Dropdown doesn't stay behind datagrid container.

This reverts #1712 pretty much. Maybe it was working at the time, but it is no longer. It is of note that with time we removed most providers js dependencies, so that probably had an effect.

For issue 1. it fixes it by placing it on the correct place. Tested in every provider. With and without table set to responsive. 

For issue 2.
- Without Responsive works fine
- With Responsive I'd argue it still works fine, it looks like it is behind something but that's just the effect of the table being responsive, and therefore scrollable, so if you actually scroll down you'll see the entire dropdown content. 

ResponsiveTable

  - If the user wants different behaviour then he should either not use `Responsive`, or style accordingly, by changing the overflow, or working with the dropdown position, for example.
  - However Bulma and AntDesign do not have proper behaviour neither before revert or after revert. I'll see if I can possibly fix it.


stsrki:
It works. But what I'm concerned is, should the menu be visible at the top of everything. Without showing the scrollbar. Seems like it is expected behavior.

David-Moreira:
> It works. But what I'm concerned is, should the menu be visible at the top of everything.

You mean when Responsive is set and the scroll bar is visible? I'm not even sure that's possible with how overflow is handled? Maybe if you play with the position fixed attribute. Think that should be the behaviour instead? Want to try?

stsrki:
> > It works. But what I'm concerned is, should the menu be visible at the top of everything.
> 
> You mean when Responsive is set and the scroll bar is visible? I'm not even sure that's possible with how overflow is handled? Maybe if you play with the position fixed attribute. Think that should be the behaviour instead? Want to try?

It might be worth trying. But I think it will be possible only with JS. Maybe see what they got in bootstrap?

David-Moreira:
> > > It works. But what I'm concerned is, should the menu be visible at the top of everything.
> > 
> > 
> > You mean when Responsive is set and the scroll bar is visible? I'm not even sure that's possible with how overflow is handled? Maybe if you play with the position fixed attribute. Think that should be the behaviour instead? Want to try?
> 
> It might be worth trying. But I think it will be possible only with JS. Maybe see what they got in bootstrap?

What if we do this version? I think it might be easier to get it right, but I am not sure, then we have to handle the directions also. Anyway I'll look it up
image

David-Moreira:
So I went to bs docs, and I just edited the table and added the dropdown, this is what it looks like:

If you click the 2nd:
image
If you click the 3rd:
image

stsrki:
The menu now starts from the middle of the parent element, see the image:

image

Demos/Blazorise.Demo/Pages/Dashboard.razor
```diff
@@ -1,30 +1,99 @@
 ﻿@page "/"
 @inject IVersionProvider VersionProvider
-<Heading Size="HeadingSize.Is1" Margin="Margin.Is3.FromBottom">Blazorise</Heading>
+<DataGrid TItem="int" Data="new List<int>{1,2,3}" Responsive="true" Bordered="true" Style="border: solid blue;">
```
stsrki, line 3: Don't forget to revert demo code.

Source/Blazorise/Components/Dropdown/Dropdown.razor.cs
```diff
@@ -92,10 +93,9 @@ protected override void BuildClasses( ClassBuilder builder )
         {
             builder.Append( ClassProvider.Dropdown( IsDropdownSubmenu ) );
             builder.Append( ClassProvider.DropdownGroup(), IsGroup );
-            builder.Append( ClassProvider.DropdownShow(), Visible );
+            //builder.Append( ClassProvider.DropdownShow(), Visible );
```
stsrki, line 96: I think `DropdownShow` should still be here because we're doing this for BS only. Other providers should still work the same, right?

Source/Blazorise/Components/Dropdown/Dropdown.razor.cs
```diff
@@ -121,15 +121,15 @@ protected override void Dispose( bool disposing )
...
-            // used to prevent toggle event call if Open() is called multiple times
             if ( Visible )
-                return Task.CompletedTask;
+                return;
 
+            
             Visible = true;
-
-            return InvokeAsync( StateHasChanged );
+            await JSModule.Initialize(this.ElementRef, this.ElementId, null);
```
stsrki, line 131: Add `if ( Rendered ) { }` check?

Source/Blazorise/wwwroot/dropdown.js
```diff
@@ -0,0 +1,43 @@
...
...r, options: i }) } return c(l).forEach((function (t) { var e = t.destroy, a = t.enable, s = t.disable; t.destroy = function (t) { void 0 === t && (t = !0), t && r.forEach((function (t) { t.destroy() })), r = [], n.forEach((function (t) { var e = t.node, n = t.eventType, r = t.handler, i = t.options; e.removeEventListener(n, r, i) })), n = [], e() }, t.enable = function () { a(), r.forEach((function (t) { return t.enable() })), o = !1 }, t.disable = function () { s(), r.forEach((function (t) { return t.disable() })), o = !0 }, function (t) { var e = t.reference; v(e, "touchstart", d, i), v(e, "mouseover", d), v(e, "focusin", d), v(e, "click", d) }(t) })), l }, _.hideAll = function (t) { var e = void 0 === t ? {} : t, n = e.exclude, r = e.duration; N.forEach((function (t) { var e = !1; if (n && (e = h(n) ? t.reference === n : t.popper === n.popper), !e) { var i = t.props.duration; t.setProps({ duration: r }), t.hide(), t.state.isDestroyed || t.setProps({ duration: i }) } })) }, _.roundArrow = '<svg width="16" height="6" xmlns="http://www.w3.org/2000/svg"><path d="M0 6s1.796-.013 4.67-3.615C5.851.9 6.93.006 8 0c1.07-.006 2.148.887 3.343 2.385C14.233 6.005 16 6 16 6H0z"></svg>', _ }));
```
stsrki, line 4: This might be good to have as a module so that we can import it here and on `tooltip.js`. See if it can be done.

Source/Blazorise/wwwroot/dropdown.js
```diff
@@ -0,0 +1,43 @@
...
+    element = getRequiredElement(element, elementId);
+
+    if (!element)
+        return;
+
+    const btnToggle = element.querySelector(".dropdown-toggle");
+    const menu = element.querySelector(".dropdown-menu");
+
+    //TODO : Set options / direction
+    const instance = Popper.createPopper(btnToggle, menu, {
```
stsrki, line 18: So, `Popper.createPopper` is called every time we need to show the menu? Can it be created once when the component is first created? Then let it handle itself. We do this pattern for all other components. Otherwise rename initialise/destroy to show/hide.

stsrki:
It's looking good so far but it needs more work to make it production-ready. Once done it would be great I think.

Source/Blazorise/Components/Dropdown/Dropdown.razor.cs
David-Moreira: Actually I'd like to do this for every provider. BS/BS5/Material are already pretty similar.
Bulma and ant-design aren't working right with responsive tables and hopefully this will help.

Source/Blazorise/wwwroot/dropdown.js
David-Moreira: Yea probably a Popper Module, where we have our popper utils and how to create it, options etc.

Source/Blazorise/wwwroot/dropdown.js
David-Moreira: Yea was kinda just trying it out. I already told you I commit stuff most of the times when I don't have more time haha Popper module

Source/Blazorise/Components/Dropdown/Dropdown.razor.cs
stsrki: OK, that makes sense.

Source/Blazorise/wwwroot/popper.js
```diff
@@ -0,0 +1,37 @@
...
+    if (direction == DIRECTION_UP)
+        return "top-start";
+    if (direction == DIRECTION_END)
+        return "right-start";
+    if (direction == DIRECTION_START)
+        return "left-start";
+    return "bottom-start";
+}
+
+export function createPopper(elementOffset, menu, options) {
```
stsrki, line 23: Are we gonna be able to import a popper module like this for the `tooltip.js`?

Source/Blazorise/wwwroot/dropdown.js
```diff
@@ -0,0 +1,34 @@
...
+
+    if (!element)
+        return;
+
+    const btnToggle = element.querySelector(".dropdown-toggle");
+    const menu = element.querySelector(".dropdown-menu");
+
+    const instance = createPopper(btnToggle, menu, options);
+
+    element.classList.add(SHOW_CLASS);
```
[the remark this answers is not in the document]
stsrki: I've done this and it seems to be working fine.

Source/Blazorise/wwwroot/popper.js
David-Moreira: I actually didn't see how tooltip. Js is done but I don't see why not

stsrki:
Found a bug with split dropdown

image

Should be:

image

Source/Blazorise/wwwroot/popper.js
David-Moreira: So for tooltip.js I see a tippy call 
I can't tell, aren't those two distinct libraries?

Source/Blazorise/Components/Dropdown/Dropdown.razor.cs
```diff
@@ -77,7 +78,7 @@ protected override void OnAfterRender( bool firstRender )
         {
             if ( firstRender )
             {
-                JSModule.Initialize( ElementRef, ElementId, new
+                JSModule.Initialize( ElementRef, ElementId, buttonList?.FirstOrDefault()?.ElementId, new
```
David-Moreira: Is it always the first occurrence?

Source/Blazorise/wwwroot/popper.js
stsrki: I suspect that tippy is expecting a popper not as a module but as a regular JS script. I think the only way to make it work is to change the tippy code to initialize popper using our own exported `createPopper` method, somehow.

Source/Blazorise/Components/Dropdown/Dropdown.razor.cs
stsrki: What do you mean by the first occurrence? The idea is to initialize it so that is already ready by the time we click on a toggle button. This way it will be faster to show and will not do the weird `pop-out`.

stsrki (approved):

PR #3653, opened 2022-03-30
DataGrid : Static data filter data only once on first load

### LINKED ISSUE #3637: DataGrid fires CustomFilter 3 times x N rows at first time loading
**Describe the bug**
Using the example provided in the doc ```https://blazorise.com/docs/extensions/datagrid/features``` to enable a CustomFilter ```CustomFilter="@OnCustomFilter"``` I realize that CustomFilter function fires (3 x N rows) times, so in the example we have 5 rows so CustomFilter fires 15 times.

**To Reproduce**
1. Copy definition and code provided in 'Custom Filtering' section at 'https://blazorise.com/docs/extensions/datagrid/features'
2. Add a private class Employee to make it compile and work:
```
private class Employee
   {
       public string FirstName { get; set; } = String.Empty;
   }
```
3. Set a breakpoint inside ```private bool OnCustomFilter(Employee model)```
4. Debug & see how ```OnCustomFilter``` is called 15 times (3 x 5 rows)

**Expected behavior**
At first run, when DataGrid is loading, CustomFilter callback must not be called because the data is loading and the user is not able to input a filter at that moment.

David-Moreira:
Closes #3637

Somehwere in time `Reload`  substitued the explicit ReadData for code reuse. Well it does not make sense to `Reload` in the cases of static data, it will set the dirty flag again for apparently no reason.

### DIFF
--- a/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
+++ b/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
@@ -254,7 +254,9 @@ namespace Blazorise.DataGrid
                 paginationContext.SubscribeOnPageSizeChanged( OnPageSizeChanged );
                 paginationContext.SubscribeOnPageChanged( OnPageChanged );
 
-                await Reload();
+                if(ManualReadMode)
+                    await Reload();
+
                 return;
             }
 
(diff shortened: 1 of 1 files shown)


stsrki (approved):
Tested and it's working.

PR #3673, opened 2022-04-03
NumericPicker: only allow decimal if the TValue is not integer

### LINKED ISSUE #3650: NumericPicker: detect TValue and remove decimals if needed
If TValue is an integer there is no need for decimals, and they should be ignored.

stsrki:
closes #3650

**Usage**

```
<NumericPicker TValue="int" />
<NumericPicker TValue="decimal" />
```

### DIFF
--- a/Source/Blazorise/Components/NumericPicker/NumericPicker.razor.cs
+++ b/Source/Blazorise/Components/NumericPicker/NumericPicker.razor.cs
@@ -46,6 +46,23 @@ namespace Blazorise
         /// </summary>
         private bool hasValueToChangeOnBlur;
 
+        /// <summary>
+        /// True if the TValue is an integer type.
+        /// </summary>
+        private bool isIntegerType;
+
+        #endregion
+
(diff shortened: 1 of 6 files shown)


Source/Blazorise/Components/NumericPicker/NumericPicker.razor.cs
```diff
@@ -86,7 +103,7 @@ public override async Task SetParametersAsync( ParameterView parameters )
                 {
                     ExecuteAfterRender( async () => await JSModule.UpdateOptions( ElementRef, ElementId, new
```
David-Moreira, line 104: Is decimalsChanged == false enough when `Integer`? I see other Decimal related configuration. Or would those not make sense to check for `integer`?

stsrki: They don't make sense if the type is Integer. That's why I ignore it.

PR #3677, opened 2022-04-04
DataGrid: RowContextMenu

### LINKED ISSUE #3660: Feature Request: add RowContextMenu event in DataGrid
It's currently not trivial to implement context menu for data grid row (JS has to be involved). I wish DataGrid can expose the RowContextMenu event.

stsrki:
closes #3660

### DIFF
--- a/Source/Blazorise/Base/BaseDraggableComponent.cs
+++ b/Source/Blazorise/Base/BaseDraggableComponent.cs
@@ -1,5 +1,6 @@
 ﻿#region Using directives
 using System.Threading.Tasks;
+using Blazorise.Utilities;
 using Microsoft.AspNetCore.Components;
 using Microsoft.AspNetCore.Components.Web;
 #endregion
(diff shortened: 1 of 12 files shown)


David-Moreira:
LGTM.
This is pretty much the same as just tracking right-click, no? There's nothing really special you could do with the "context menu"?

Also should we have an example? Right click would trigger row edit for example? Or cancel row edit?

I tried this on our data grid demo and works great:
```
RowContextMenuPreventDefault
RowContextMenu="@((e) => Console.WriteLine(e.Item.FirstName))"
```

Any chance we can keep re-filtering and other operations to a minimum? Or to be expected?
image

stsrki:
The context menu is doing more than just triggering on right-click. It can replace the browser context menu and then you can use it to draw your own menu.

---

> Any chance we can keep re-filtering and other operations to a minimum? Or to be expected?

We should look into it. I think we would have to check old and new filter values to stop it from triggering re-filter.

David-Moreira:
> The context menu is doing more than just triggering on right-click. It can replace the browser context menu and then you can use it to draw your own menu.
> 
> > Any chance we can keep re-filtering and other operations to a minimum? Or to be expected?
> 
> We should look into it. I think we would have to check old and new filter values to stop it from triggering re-filter.

Curious how that works, do we have enough power on C# to do this? I think people would greatly appreciate an example with this then.

stsrki:
Yeah, I will add an example. Need to finish 1.0 issues first.

PR #3693, opened 2022-04-07
Fix typo in PageButtonContext.PageNumer

WolfgangKluge:
At least I think it's a typo ;)


stsrki:
It is a typo. But the problem now is, is this a breaking change? We're changing the API so by the definition it should go into v2.

@David-Moreira

WolfgangKluge:
That's definitely a breaking change.

(I think I don't get the version number for the branch name right)

stsrki:
Branching is done like:

- All hotfixes are done from and onto the `rel-1.0.0` branch
- All new development is done onto the `master`

---

I think it is best to open a new issue and schedule this change for v2.0. I hate it but we don't have much choice or we'll break semver rules.

stsrki:
@WolfgangKluge Thanks for the work but it needs to be done 😔

Closing in favor of #3694

PR #3706, opened 2022-04-11
Chart: Remove default value from options and datasets

### LINKED ISSUE #3703: Chart: Remove default values
Continuation from #3664

We need to remove all default values from options and datasets and let the charts handle it.

stsrki:
resolves #3703

This is the continuation from #3701. Generally, we want to only serialize exactly what is defined by the user. Everything else will be handled by the ChartJS and by removing default values this will be possible.


David-Moreira (approved):
lgtm. So this just makes sure it takes in the correct library js defaults always?

stsrki:
Yes, that is the idea. Everything that is null will not be serialized and so it will be handled by the chart library.

PR #3712, opened 2022-04-12
NumericPicker: control the behavior of mouse wheel

### LINKED ISSUE #3691: NumericPicker: add an option to enable or disable mouse wheel
Use modifyValueOnWheel from https://github.com/autoNumeric/autoNumeric/

stsrki:
Closes #3691


David-Moreira (approved):
Implementation LGTM, have we tested starting enabled and disabling, and vice versa?

stsrki:
> have we tested starting enabled and disabling, and vice versa?

Let's leave it in good faith it will work 🙄😂

PR #3726, opened 2022-04-18
DataGrid : ToggleDetailRow : Fix stale/disposed row info being used on ToggleDetailRow logic

### LINKED ISSUE #3704: Datagrid details open logic
The administration of opened details of the datagrid is not consistent

**To Reproduce**
1. Go to https://bootstrap5demo.blazorise.com/tests/datagrid
2. Open details of "Irvin" and of "Gerald"
3. Got to page 2, and back to the first page of the datagrid

The same details should be opened, but only Gerald is opened. Why?

When after this, you collapse Gerald again, and go to page 2 and back to page 1, now Gerald is expanded again....

In my application I have a related problem:
I have a custom filter on the grid, that contains details. When this custom filter is changed all detail-grids are hidden.
I use the DetailRowTrigger mechanism to control the visibility of the details.

David-Moreira:
Closes #3704

`GetRowInfo` was returning a `RowInfo` that is in the process of being disposed. 
When `RowInfo` is first created, use it straight up as a parameter for `DataGrid.ToggleDetailRow`, no point in doing another lookup.
Change `GetRowInfo` to grab the last row instead, making it safer to grab an actual non stale row if something similar could happen.

Also this does happen because the row is being re-rendered (Disposed of && re-Initialized), in this case it was just a filter so it probably should have kept the same row instance, instead of rendering a new one... going to open an issue so we can take a look if there's some optimization we can do here, we do have `@key="@item"` but it might not be working in certain cases, due to conditional renderings. Not sure if we can do something about it.

### DIFF
--- a/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
+++ b/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
@@ -714,9 +714,10 @@ namespace Blazorise.DataGrid
             => ToggleDetailRow( item, DetailRowTriggerType.Manual, forceDetailRow, true );
 
         protected internal Task ToggleDetailRow( TItem item, DetailRowTriggerType detailRowTriggerType, bool forceDetailRow = false, bool skipDetailRowTriggerType = false )
-        {
-            var rowInfo = GetRowInfo( item );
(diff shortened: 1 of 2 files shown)


stsrki (approved):
LGTM

PR #3735, opened 2022-04-20
NumericPicker: change inputmode based on TValue

Kreisverkehr:
This pr changes the behavior of NumericEdit and NumericPicker on devices that choose a different input method based on the field type.

I had the issue that users on iOS devices were unable to input decimal numbers in euther NumericEdit or NumericPicker despite the type being decimal. This is due to the fact that iOS only presents an integer keyboard when inputmode is set to "numeric". Only when set to "decimal" the user is able to enter a decimal seperator.
For more information on this refer to: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode

> decimal
> Fractional numeric input keyboard containing the digits and decimal separator for the user's locale (typically . or ,). Devices may or may not show a minus key (-).
> 
> numeric
> Numeric input keyboard, but only requires the digits 0–9. Devices may or may not show a minus key.

I've chosen to evaluate the inputmode to be used based on TValue as I found this the least inversive and most intuitive way of setting this attribute.


stsrki:
@Kreisverkehr Sorry it took so long to review. The PR looks good; I just made some small modifications. Can you please sign the CLA before I merge it? Thanks

Kreisverkehr:
@stsrki no problem. Your changes look good to me. The fix should still work as intended. I just missed the other components.

Kreisverkehr:

PR #3743, opened 2022-04-24
CLA Workflow: Revert back to Blazorise Master Branch

David-Moreira:


stsrki:
Seems it's still failing. Or does it need to be merged to master first?

David-Moreira:
Not sure actually, because I think workflows are always based on latest master workflow configuration, is it not? That's how our PR in rel-1.0 for example already running new cla workflow even tough configuration does not exist there. Eithereway, Can't hurt merging haha.

stsrki:
Merged. We'll see what happens

Issue #3775, opened 2022-05-08
 Documentation enhancement: Horizontally center Check component in Horizontal Field


oaldrian:
**Is your feature request related to a problem? Please describe.**
### This is a Documentation enhancement request, but I can only create a Issue type Feature Request

When using a horizontal Field with a Check component as FieldBody the checkbox is not centered horizontally but slightly upwards in comparison to the FieldLabel.

**To Reproduce**

Use following code on a page 

 ```
<Field Horizontal>
          <FieldLabel ColumnSize="ColumnSize.Is5">Abgeschlossen</FieldLabel>
          <FieldBody ColumnSize="ColumnSize.Is7">
            <Check TValue="bool" />
          </FieldBody>
        </Field>
```

**Describe the solution you'd like**
Either under Check or under Field in the documentation it should be mentioned that this behavior can be mitigated with putting Margin="Margin.IsAuto" on the FieldBody.

Like this

```
 <Field Horizontal>
          <FieldLabel ColumnSize="ColumnSize.Is5">Abgeschlossen</FieldLabel>
          <FieldBody Margin="Margin.IsAuto" ColumnSize="ColumnSize.Is7">
            <Check TValue="bool" />
          </FieldBody>
        </Field>
```

### Additional context
This only happens to the Check component to my knowledge, and as the other components behaved "normal" aka horizontally centered, I think it is worth mentioning it in the docu.

oaldrian:
If this "feature" gets a go from you, I will "implement" it into the doc.

stsrki:
It's because the check doesn't have the same height by default as other inputs do. Yeah, you can add it to the docs. Just make sure you're working on/from the right branch. You can make it on `rel-1.0.0`.

PR #3786, opened 2022-05-11
Docs: adds missing comments

oaldrian:
- adds description and example for horizontal Field with Check component (#3775)
- adds missing PageSizeChanged EventHandler in DataGrid attributes list


stsrki:
Made some adjustments, but overall, it looks good. Thank you for helping us improve the documentation!

stsrki (approved):

PR #3827, opened 2022-05-22
DataGrid: pager options improvements

### LINKED ISSUE #3528: Question about DataGrid pager button sizes
How do you change the button sizes of the pager buttons?

Is there a setting or enum to change the built-in button sizes?

### LINKED ISSUE #3823: Question: Is it possible to change the Pager buttons to either side instead of it being pushed to the middle when ButtonRowTemplate are set?
Currently I have some tables which contain some buttons at the bottom, which are always aligned to the left.
The thing is, if they are not set, the pager buttons are on the left.
This way it is very inconsistent for the user where the pager buttons are.

Is it possible to change the location of these elements? I haven't found one yet.
If there is no way yet, and it is something which can be added, a good implementation could be to add a location to the templates (e.g. left, middle, right)?

image
ps: I am using Blazorise version 0.9.5.7 because I am stuck on vs 2019 due to company policy :S

David-Moreira:
Closes #3823 
Closes #3528


Source/Extensions/Blazorise.DataGrid/Configuration/PagerOptions.cs
```diff
@@ -0,0 +1,26 @@
...
+{
+    /// <summary>
+    /// Defines the options for <see cref="DataGrid{TItem}"/> pager.
+    /// </summary>
+    public class PagerOptions
+    {
+        /// <summary>
+        /// Configures the pager buttons size.
+        /// </summary>
+        public Size Size { get; set; }
```
stsrki, line 11: Might it be better renamed to ButtonsSize? Or ButtonSize?

Source/Extensions/Blazorise.DataGrid/Configuration/PagerOptions.cs
```diff
@@ -0,0 +1,26 @@
+﻿namespace Blazorise.DataGrid.Configuration
```
stsrki, line 1: Do it under the `Blazorise.DataGrid.Configuration` namespace? I think Configuration is redundant and it can be removed completely.

David-Moreira: I mean... maybe the word redudant isn't the best, it's well organized under that namespace in my opinion, you could argue 
 instead that you'd want to have it have a better discovery by not putting it under a specific namespace instead?

Nothing against that. I'll change it.

Source/Extensions/Blazorise.DataGrid/Configuration/PagerOptions.cs
David-Moreira: Right, I actually debated that myself, but I guess I just though people could easily infer that it would do.
We can change. `ButtonSize` seems okay.

Source/Extensions/Blazorise.DataGrid/Configuration/PagerOptions.cs
stsrki: Yeah, maybe redundant isn't the best. I think only three classes are under the `Blazorise.DataGrid.Configuration` namespace, and all of them have a `DataGrid` prefix, so I think it is better to have them all under the `Blazorise.DataGrid` namespace. At least they will be easier to discover.

David-Moreira: Right sure I'll remove the namespace. Should we  rename to `DataGridPagerOptions` then?

stsrki: DataGridPagerOptions sounds good

stsrki (approved):

PR #3797, opened 2022-05-13
A couple fixes for bugs related to rapid double-clicking of MessageAlert and Modal buttons

Bobbar:
MessageAlert will throw an exception if any of the submit or close buttons are clicked twice in rapid succession. This results in client crashes/disconnects.

> Microsoft.AspNetCore.Components.Server.Circuits.RemoteRenderer[100]
> Unhandled exception rendering component: An attempt was made to transition a task to a final state when it had already completed.
> System.InvalidOperationException: An attempt was made to transition a task to a final state when it had already completed.
> at System.Threading.Tasks.TaskCompletionSource`1.SetResult(TResult result)
> at Blazorise.MessageAlert.<OnConfirmClicked\>b__4_1()
> at Microsoft.AspNetCore.Components.Rendering.RendererSynchronizationContextDispatcher.InvokeAsync(Action workItem)
> at Microsoft.AspNetCore.Components.ComponentBase.InvokeAsync(Action workItem)
> at Blazorise.MessageAlert.<OnConfirmClicked\>b__4_0()
> at Microsoft.AspNetCore.Components.ComponentBase.CallStateHasChangedOnAsyncCompletion(Task task)
> at Blazorise.Button.ClickHandler()
> at Microsoft.AspNetCore.Components.ComponentBase.CallStateHasChangedOnAsyncCompletion(Task task)
> at Microsoft.AspNetCore.Components.RenderTree.Renderer.GetErrorHandledTask(Task taskToHandle, ComponentState owningComponentState)

The Modal component will enter a (potentially) infinite loop if Hide is called immediately following a Show call. Easily reproducible on the demo site by quickly double-clicking any of the show modal buttons. This results in a flashing unresponsive page while also spamming the circuit with events.

PR #3838, opened 2022-05-25
MessageAlert: fix double-clicking invocation

stsrki:
Recreation of #3797 because it was done on the wrong branch, and rebase would bring dev changes onto the stable release.

@Bobbar

### DIFF
--- a/Source/Blazorise/Adapters/CloseableAdapter.cs
+++ b/Source/Blazorise/Adapters/CloseableAdapter.cs
@@ -33,12 +33,6 @@ namespace Blazorise
 
                 await Task.Delay( animatedComponent.AnimationDuration );
 
-                // NOT SURE IF WE NEED THIS?
-                if ( visible )
-                    await animatedComponent.Show();
-                else
-                    await animatedComponent.Hide();
-
                 await animatedComponent.EndAnimation( visible );
             }
             else
--- a/Source/Blazorise/Components/MessageAlert/MessageAlert.razor.cs
+++ b/Source/Blazorise/Components/MessageAlert/MessageAlert.razor.cs
@@ -70,7 +70,7 @@ namespace Blazorise
             {
                 await ModalRef.Hide();
 
-                if ( IsConfirmation && Callback != null )
+                if ( IsConfirmation && Callback != null && !Callback.Task.IsCompleted )
                 {
                     await InvokeAsync( () => Callback.SetResult( true ) );
                 }
@@ -89,7 +89,7 @@ namespace Blazorise
             {
                 await ModalRef.Hide();
 
-                if ( IsConfirmation && Callback != null )
+                if ( IsConfirmation && Callback != null && !Callback.Task.IsCompleted )
                 {
                     await InvokeAsync( () => Callback.SetResult( false ) );
                 }


Source/Blazorise/Adapters/CloseableAdapter.cs
```diff
@@ -33,12 +33,6 @@ public async Task Run( bool visible )
 
                 await Task.Delay( animatedComponent.AnimationDuration );
 
-                // NOT SURE IF WE NEED THIS?
```
stsrki, line 36: @David-Moreira do you think we still need this?

David-Moreira: The comment? xD Or the `Task.Delay`? I think the `Task.Delay` was to give it the time to animate before proceeding with the next operation. But I can review that.

stsrki: No, the Show and Hide method calls.

stsrki: This part

```cs
if ( visible )
    await animatedComponent.Show();
else
    await animatedComponent.Hide();
```

David-Moreira: I no longer remember this by heart. I'll take a look.

David-Moreira: So, why do you think we don't need that part? That part is how it works without animation.
image

Bobbar: That same show/hide block was included above for the animated component path after the Task. Delay.

stsrki: > So, why do you think we don't need that part? That part is how it works without animation. image

Well, the comment said we

```
// NOT SURE IF WE NEED THIS?
```

so just wanted to make sure. Just going to remove the comment so that it doesn't confuse us again 🤣

David-Moreira: > That same show/hide block was included above for the animated component path after the Task. Delay.

AHHH I didn't get that, I just re-tested yea, you're right seems to work just the same.
But let's leave it as original for now, I think it's better if we still run it if animated. 

From an API perspective, doesn't matter if animated or not, seems like Show/Hide should be expected to run I'd guess... We might want to run some additional or centralized logic there. 
But It's true, I guess it's just the way that animation works, ends up being pretty close to `Show`, because most times you'll have to actually start showing the component to animate it... and ends up being a bit redudant... But we might get different components that need it like this in the future...

For example, `Hide` seems to have a centralized closeReason reset there.
image

@Bobbar 
Now I was taking a look at the original branch commit: 
> "an infinite loop within the Modal component if a Hide is called before a Show call finishes." 
Is this easily reproducable? Can you help re-create this?

David-Moreira:
As for the fix itself LGTM I guess.

What happens if I manage to click both `Cancel` and `Confirm`? (Because an operation is slow or something)
For example: Click `Confirm`... long operation... click `Cancel`?

stsrki:
Probably an error. But unlikely to happen I guess.

Source/Blazorise/Adapters/CloseableAdapter.cs
Bobbar: It's easily reproducible on the boostrap demo site by quickly double-clicking the basic modal button.

stsrki: I cannot reproduce it myself.

Bobbar: I can reproduce it easily on the bootstrap demo site by double-clicking the standard modal button.

David-Moreira: @Bobbar you're right.
Looks like disco night :)

So you solved this by removing that `show/hide` block, right?

Bobbar: Right. I don't recall any other components implementing the animation. And the modal calls the show/hide internally. IIRC.

PR #3848, opened 2022-05-27
DropZone: reorder items

### LINKED ISSUE #3733: Drag and drop reordering inside a list
Drag and drop component is great!
But, we can't reorder the items inside a DropZone and it'd be great if we could have a list where we could reorder the items within a list and not just between DroZones, something like this => https://codepad.co/snippet/sortable-list-with-drag-and-drop
In order to distinguish it from similar components it should emit a position inside a list on draggable event.

stsrki:
Closes #3733

### DIFF
--- a/Source/Blazorise/Components/Droppable/DropZone.razor
+++ b/Source/Blazorise/Components/Droppable/DropZone.razor
@@ -5,18 +5,51 @@
     <div @ref="@ElementRef" id="@ElementId" class="@ClassNames" style="@StyleNames" @ondragenter="@OnDragEnterHandler" @ondragleave="@OnDragLeaveHandler" @ondrop="@OnDropHandler" @attributes="@Attributes">
         @ChildContent
 
-        @foreach ( var item in GetItems() )
+        @if ( OnlyZone != true )
         {
-            <_Draggable @key="@item" TItem="TItem" Item="@item" ZoneName="@Name" DraggingClass="@GetItemDraggingClass()" Disabled="@GetItemDisabled(item)" DisabledClass="@GetItemDisabledClass()" DragStarted="@OnDragStarted" DragEnded="@OnDragEnded">
-                @{
-                    var renderer = ItemTemplate ?? ParentContainer?.ItemTemplate;
+            int index = 0;
+            var transactionIndex = ParentContainer?.GetTransactionIndex() ?? -1;
+            var items = GetItems();
+
+            @if ( AllowReorder )
+            {
+                @if ( items.Any() == false )
+                {
+                    <div class="@PlaceholderClassBuilder.Class"></div>
+                }
+                else
+                {
+                    if ( transactionIndex == -1 )
+                    {
+                        <div class="@PlaceholderClassBuilder.Class"></div>
+                    }
+
+                    <_Draggable @key="@ElementId" TItem="TItem" Item="default" ZoneName="@Name" Disabled="true" HideContent="false" Class="draggable-preview-start" />
+                }
+            }
+
+            @foreach ( var item in items )
+            {
+                var indexCopy = index;
+
+                <_Draggable @key="@(item.GetHashCode())" TItem="TItem" Item="@item" ZoneName="@Name" DraggingClass="@GetItemDraggingClass()" Disabled="@GetItemDisabled(item)" DisabledClass="@GetItemDisabledClass()" DragStarted="@OnDragStarted" DragEnded="@OnDragEnded" Index="@indexCopy">
+                    @{
+                        var renderer = ItemTemplate ?? ParentContainer?.ItemTemplate;
+                    }
+
+                    @if ( renderer != null )
+                    {
+                        @renderer(item)
+                    }
+                </_Draggable>
+
+                if ( transactionIndex == indexCopy && IsOrigin( indexCopy ) == false )
+                {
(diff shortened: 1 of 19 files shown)


Source/Blazorise/Components/Droppable/DropZone.razor
```diff
@@ -5,18 +5,51 @@
     <div @ref="@ElementRef" id="@ElementId" class="@ClassNames" style="@StyleNames" @ondragenter="@OnDragEnterHandler" @ondragleave="@OnDragLeaveHandler" @ondrop="@OnDropHandler" @attributes="@Attributes">
         @ChildContent
 
-        @foreach ( var item in GetItems() )
+        @if ( OnlyZone != true )
```
David-Moreira, line 8: !OnlyZone 
Matter of personal preference and Blazorise prefered code convention I guess.

Source/Blazorise/Components/Droppable/DropZone.razor
```diff
@@ -5,18 +5,51 @@
...
-            <_Draggable @key="@item" TItem="TItem" Item="@item" ZoneName="@Name" DraggingClass="@GetItemDraggingClass()" Disabled="@GetItemDisabled(item)" DisabledClass="@GetItemDisabledClass()" DragStarted="@OnDragStarted" DragEnded="@OnDragEnded">
-                @{
-                    var renderer = ItemTemplate ?? ParentContainer?.ItemTemplate;
+            int index = 0;
+            var transactionIndex = ParentContainer?.GetTransactionIndex() ?? -1;
+            var items = GetItems();
+
+            @if ( AllowReorder )
+            {
+                @if ( items.Any() == false )
```
David-Moreira, line 16: Same thing as before do we prefer this `items.Any() == false` over `!items.Any()` ?

Source/Blazorise/Components/Droppable/DropZone.razor.cs
```diff
@@ -107,53 +140,86 @@ private void OnContainerTransactionStarted( object sender, DraggableTransaction<
...
+            dragCounter = 0;
 
             if ( GetApplyDropClassesOnDragStarted() )
             {
                 dropAllowed = false;
             }
 
+            if ( e.Success )
+            {
+                if ( e.OriginDropZoneName == Name && e.DestinationDropZoneName != e.OriginDropZoneName )
```
David-Moreira, line 154: Could probably evaluate only once like:

var isOriginDropZone = blabla
var isDestinationDropZone = blabla

And then just use in each if clause, still method is so short and logic so simple... only a matter of preference

Source/Blazorise/Components/Droppable/DropZone.razor.cs
```diff
@@ -107,53 +140,86 @@ private void OnContainerTransactionStarted( object sender, DraggableTransaction<
...
+                {
+                    indices.Remove( e.Item );
+                }
+
+                if ( e.OriginDropZoneName == Name || e.DestinationDropZoneName == Name )
+                {
+                    int index = 0;
+
+                    foreach ( var item in indices.OrderBy( x => x.Value ).ToArray() )
```
David-Moreira, line 163: Decison about Dictionary<TItem, int> instead of Dictionary<int, TItem> should the int be the actual key and the value, the TItem?

Wouldn't the index always be unique?
Would we ever have two 'equal' TItem? Or is that a dumb assumption?

Would that make the implementation harder? To reorder the items?

Source/Blazorise/Components/Droppable/DropZone.razor.cs
```diff
@@ -212,14 +323,14 @@ private async Task OnDropHandler()
...
-            Func<TItem, bool> predicate = ( item ) => ParentContainer.ItemsFilter( item, Name );
+            Func<TItem, bool> predicate = ( item ) => ParentContainer.ItemsFilter( item, Name ?? string.Empty );
 
             if ( ItemsFilter != null )
             {
                 predicate = ItemsFilter;
             }
 
-            return ( ParentContainer?.Items ?? Enumerable.Empty<TItem>() ).Where( predicate ).ToArray();
+            return ( ParentContainer?.Items ?? Enumerable.Empty<TItem>() ).Where( predicate ).OrderBy( x => GetItemIndex( x ) ).ToArray();
```
David-Moreira, line 333: Are we still using the index functionality if `AllowReorder` is false? Does this make it work differently?

David-Moreira (approved):
Tested & functionality works good couldn't find an issue. LGTM

Source/Blazorise/Components/Droppable/DropZone.razor.cs
stsrki: It will work the same I think.

Source/Blazorise/Components/Droppable/DropZone.razor.cs
stsrki: TItem is the key because the TItem is the one that can have a different index when reordered. Kinda make more sense to me.

Source/Blazorise/Components/Droppable/DropZone.razor.cs
stsrki: I will leave it as it is now as there is reusage in the current implementation.

PR #3849, opened 2022-05-28
Mark *Alert APIs as absolete and introduce new *Provider APIs

### LINKED ISSUE #3840: Replace *Alert components with *Provider
In v2, the components:

- MessageAlert
- NotificationAlert
- PageProgressAlert

will be removed. But until then we can introduce the new aliases

- MessageProvider
- NotificationProvider
- PageProgressProvider

Until v2, we can mark *Alert components obsolete.

---

Also need to update the documentation with the new names.

stsrki:
Closes #3840


David-Moreira (approved):
LGTM. Like this naming better too.

PR #3854, opened 2022-05-29
DatePicker: compare array values to prevent duplicate change events

### LINKED ISSUE #3798: DatePicker's DateChanged occurs when date is changed in parent and passed through parameter
**Describe the bug**
DatePicker's DateChanged occurs when date is changed in parent and passed through parameter. It occurs 2 times if the Date has already been changed before.

**To Reproduce**
Using [demo-bugs-page](https://github.com/paulosdfigueiredo/Blazorise/tree/demo-bugs-page)

0. Run Demo
1. Go to 'tests/bugs'
2. Click on 'Set Today'
4. See a notification alert with date changed. If date has already been changed, you see 2 alerts.

NumericPicker is there as a Control showing the correct behavior.

**Expected behavior**
DateChanged should only occur when the date is changed inside the component.

stsrki:
Closes #3798

DatePicker's internal value is an array. So when `value.IsEqual( InternalValue )` was used in the base class, it would always fail. I have introduced a new API in the base input class, to be able to override the compare function. Currently, only DatePicker uses it. I think a `Select` should also benefit from it because it too uses an array internally.

### DIFF
--- a/Source/Blazorise/Base/BaseInputComponent.razor.cs
+++ b/Source/Blazorise/Base/BaseInputComponent.razor.cs
@@ -201,6 +201,13 @@ namespace Blazorise
             => value;
 
         /// <summary>
+        /// Check if the internal value is same as the new value.
+        /// </summary>
+        /// <param name="value">Value to check against the internal value.</param>
+        /// <returns>True if the internal value matched the supplied value.</returns>
+        protected virtual bool IsSameAsInternalValue( TValue value ) => value.IsEqual( InternalValue );
+
+        /// <summary>
         /// Raises and event that handles the edit value of Text, Date, Numeric etc.
         /// </summary>
         /// <param name="value">New edit value.</param>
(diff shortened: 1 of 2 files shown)


David-Moreira:
Another approach would be for the internal implementation to be to handle that without overriding case by case? 

I feel like the name should be more along the lines of "`InternalValueEqualityComparer`" or something similar. Somehow reword to use the word "Equal" which is the wording generally used on the comparision apis I guess.

stsrki:
> Another approach would be for the internal implementation to be to handle that without overriding case by case?

Can you explain this?

David-Moreira:
> > > Another approach would be for the internal implementation to be to handle that without overriding case by case?
> > 
> > 
> > Can you explain this?
> 
> Maybe I didn't understand the question... But the problem is that TValue is generic and we can't know what type it can be.

Basically change the implementation of `IsEqual` to be smarter about it? Isn't that possible? It may be generic, but the types are still there. Or am I thinking wrongly about this?
image

stsrki:
Not possible as far as I know.

We have `class BaseInputComponent<TValue>`, and it can be inherited like:

- `class BaseTextInput<TValue> : BaseInputComponent<TValue>`
- `DatePicker<TValue> : BaseTextInput<IReadOnlyList<TValue>>`

And only in the DatePicker class do we know that `TValue` is actually `IReadOnlyList<TValue>`. In the BaseInputComponent we don't know it unless we try to figure it out with a reflection.

Issue #3424, opened 2022-02-10
Document Blazorise conventions and best practices


stsrki:
Add:

- naming guidelines
- formating
- code conventions
- etc.

David-Moreira:
Notes: 
- Prefer prefixless booleans, example: `Loading` over `IsLoading`

stsrki:
> Notes:
> 
> * Prefer prefixless booleans, example: `Loading` over `IsLoading`
> * Follow Regions template Members/Properties/etc... Provide the template.

`IsLoading `is allowed for property getters in case when we need to determine something from multiple sources, eg

`IsLoading => Options?.Loading ?? Loading ?? false;`

PR #3856, opened 2022-05-30
AutoComplete : ReadData

### LINKED ISSUE #3450: Autocomplete : Load data on demand (Similar to ReadData on DataGrid) : Feature
Autocomplete : Load data on demand (Similar to ReadData on DataGrid) : Feature

Requested on #410

David-Moreira:
Closes #3450

### DIFF
--- a/Source/Extensions/Blazorise.Components/Autocomplete.razor.cs
+++ b/Source/Extensions/Blazorise.Components/Autocomplete.razor.cs
@@ -26,11 +26,6 @@ namespace Blazorise.Components
         #region Members
 
         /// <summary>
-        /// Returns true if ReadData will be invoked.
-        /// </summary>
-        protected bool Loading { get; set; }
-
-        /// <summary>
         /// Tells us that modal is tracked by the JS interop.
         /// </summary>
         private bool jsRegistered;
@@ -208,6 +203,7 @@ namespace Blazorise.Components
             if ( firstRender )
             {
                 dotNetObjectRef ??= DotNetObjectReference.Create( new CloseActivatorAdapter( this ) );
+
                 if ( ManualReadMode )
                     await Reload();
             }
@@ -232,7 +228,8 @@ namespace Blazorise.Components
             await SearchChanged.InvokeAsync( CurrentSearch );
             await SelectedTextChanged.InvokeAsync( SelectedText );
 
-            await HandleReadData();
+            if ( ManualReadMode )
+                await HandleReadData();
 
             if ( FilteredData?.Count == 0 && NotFound.HasDelegate )
                 await NotFound.InvokeAsync( CurrentSearch );
@@ -633,6 +630,11 @@ namespace Blazorise.Components
         #endregion
 
         #region Properties
+
+        /// <summary>
+        /// Returns true if ReadData will be invoked.
+        /// </summary>
+        protected bool Loading { get; set; }
 
         /// <summary>
         /// Event handler used to load data manually based on the current search value.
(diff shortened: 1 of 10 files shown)


David-Moreira:
(Ongoing)

Source/Extensions/Blazorise.Components/Autocomplete.razor.cs
```diff
@@ -181,6 +232,8 @@ protected async Task OnTextChangedHandler( string text )
             await SearchChanged.InvokeAsync( CurrentSearch );
             await SelectedTextChanged.InvokeAsync( SelectedText );
 
+            await HandleReadData();
```
stsrki, line 235: Should this be behind the `if ( ManualReadMode )`?

Source/Extensions/Blazorise.Components/Autocomplete.razor.cs
```diff
@@ -65,10 +70,54 @@ public partial class Autocomplete<TItem, TValue> : BaseAfterRenderComponent, ICl
...
+        }
+
+        /// <summary>
+        /// Triggers the reload of the <see cref="Autocomplete{TItem, TValue}"/> data.
+        /// Makes sure not to reload if the <see cref="Autocomplete{TItem, TValue}"/> is in a loading state.
+        /// </summary>
+        /// <returns>Returns the awaitable task.</returns>
+        public async Task Reload( CancellationToken cancellationToken = default )
+        {
+            if ( IsLoading )
```
stsrki, line 106: Usually, we don't have an Is prefix.

stsrki, line 106: Usually, we don't have an Is prefix.

David-Moreira: Based on DataGrid :) We can fix both then.

Source/Extensions/Blazorise.Components/Autocomplete.razor.cs
```diff
@@ -25,6 +25,11 @@ public partial class Autocomplete<TItem, TValue> : BaseAfterRenderComponent, ICl
     {
         #region Members
 
+        /// <summary>
+        /// Returns true if ReadData will be invoked.
+        /// </summary>
+        protected bool IsLoading { get; set; }
```
stsrki, line 31: Move down in file under Properties regien

Source/Extensions/Blazorise.Components/Autocomplete.razor.cs
David-Moreira: Well I was just starting a PR to fix DataGrid... And there's a bunch of them there prefix 'Is'.  Most of them are protected, so the impact wouldn't be big, but we might have public one's....
Leave it for 2.0? Can I open an issue with this kind of cleanup?
image

I do have an habbit of using the prefix `Is`...So I guess the majority might have been introduced by me....
But even utilities we may have that pattern... And I think you're the author of these?
image

So finish the document for Blazorise conventions first, and leave it for 2.0 (on 2.0 we try to make sure we comply to our own defined conventions)?

David-Moreira:
Closes #3865

Source/Extensions/Blazorise.Components/Autocomplete.razor.cs
stsrki: Property getter are allowed to have prefix, see https://github.com/Megabit/Blazorise/issues/3424#issuecomment-1146800048

David-Moreira: I see.

stsrki:
Is this ready for final review?

David-Moreira:
yes

stsrki (approved):
LGTM

PR #3862, opened 2022-06-01
Fix for Modal double-click infinite loop.

Bobbar:
I tested with all providers with backdrop visible & hidden and compared VisibleChanged events and CloseReason states and the fix does not appear to effect the behavior.

### DIFF
--- a/Source/Blazorise/Adapters/CloseableAdapter.cs
+++ b/Source/Blazorise/Adapters/CloseableAdapter.cs
@@ -33,11 +33,6 @@ namespace Blazorise
 
                 await Task.Delay( animatedComponent.AnimationDuration );
 
-                if ( visible )
-                    await animatedComponent.Show();
-                else
-                    await animatedComponent.Hide();
-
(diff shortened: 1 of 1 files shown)


David-Moreira (approved):
LGTM. Thanks for taking the time.

We'll see on the next animated components that we implement if we actually end up needing that.

Bobbar:
Yeah. If they implement the same pattern of BeginAnimation/EndAnimation and Hide/Show being called internally, they should be fine. 

Although that Begin - Delay - End with the classes and styles being manually injected is a bit of a hack, and doesn't actually render the animations correctly.  The animation duration ends up delaying the start of the animation, as well as the animation itself. This is easy to see if you crank the duration way up.

I've looked at ways to make it work more like the native frameworks, but doesn't look feasible without importing more JS and/or JQuery.

stsrki:
Yeah there are some quirks with long animations. But overall it works OK. There's always going to be some edge cases. For now it is good :)

Btw, thanks for helping with this issue!

PR #3869, opened 2022-06-03
DatePicker: Improve SetParametersAsync and prevent incorrect DateChanged or DatesChanged

### LINKED ISSUE #3798: DatePicker's DateChanged occurs when date is changed in parent and passed through parameter

paulosdfigueiredo:
Refactored SetParametersAsync based on NumericPicker. Check changes should only occur if picker is rendered and CurrentValueHandler should not be called to prevent unwanted event.

Closes #3798

### DIFF
--- a/Source/Blazorise/Components/DatePicker/DatePicker.razor.cs
+++ b/Source/Blazorise/Components/DatePicker/DatePicker.razor.cs
@@ -25,54 +25,53 @@ namespace Blazorise
         /// <inheritdoc/>
         public override async Task SetParametersAsync( ParameterView parameters )
         {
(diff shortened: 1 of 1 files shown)


stsrki (approved):
LGTM

PR #3870, opened 2022-06-05
AutoComplete multiple checkbox

### LINKED ISSUE #3401: Autocomplete: Multiple selection with checkboxes
### Discussed in https://github.com/Megabit/Blazorise/discussions/3397

<sup>Originally posted by **Lasoty** February  4, 2022</sup>
A very useful component would be DropDown with a checkbox list.

image
</div>

### LINKED ISSUE #3828: AutoComplete: selected element not visible on keyboard navigation
**Describe the bug**
When autocomplete shows a somewhat larger list (= there's a scrollbar), you can select those items with the arrow keys.
But after selecting an element that is out of view this way, the selection is not brought into view.

**To Reproduce**
1. Go to https://blazorise.com/docs/extensions/autocomplete
2. Type "a" in the first example
3. Select an element out of view (e.g. Netherlands Antilles) with the arrow keys
4. The value has changed, the element is selected - but the selected item is not visible

**Expected behavior**
The selected item should always be visible (scrollIntoView)

### LINKED ISSUE #3867: Autocomplete Component: several enhancements
Hey, 
im Silas from DextraData, we have some problems relating the Autocomplete component. 
As descriped in Issue #3770 and #64, previous selected Items and then deselected items aren't shown in the Dropdown anymore.
I.e. if you deselect a item by accident and then want to select it again, you aren't able to select it again.
You already recognized it as a Bug but its still there, on Version 1.0.4.
We would also suggest some changes/enhancement for the Component:

> - There should also be a search bar in the dropdown, before the first Item in the List.
>- In front of the items should be checkboxes, which obviously show which item is selected and which not.
>- There should be an Arrow/Chevron on the right side of the mainbar, so its more Intuitive for the User, that you can open a dropdown.

It would be nice if you could tell me how long it would take you to implement it, respectively when you plan to take on that Problem.

Thank you for your transparency.
Please tell me if you need more informations.
Kind Regards,
Silas

### LINKED ISSUE #3980: Autocomplete better keyboard support
In the Autocomplete demo, it is not possible to remove the selected badges with a backspace and not possible to tab to a specific badge to delete it.
image

Add keyboard navigation support

### Additional context
Thanks for your work

David-Moreira:
Closes #3867
Closes #3401
Closes #3828
Closes #3980


Source/Extensions/Blazorise.Components/Autocomplete.razor.cs
```diff
@@ -905,6 +915,17 @@ public TValue SelectedValue
         /// </summary>
         [Parameter] public bool CloseOnSelection { get; set; } = true;
 
+        /// <summary>
+        /// Shows selected option(s) when presenting the searched values.
+        /// </summary>
+        [Parameter] public bool ShowSelected { get; set; }
```
stsrki, line 921: What exactly ShowSelected is doing. It's hard to tell from the naming.

Source/Extensions/Blazorise.Components/Autocomplete.razor
```diff
@@ -37,41 +37,53 @@
...
-        @if ( NotFoundVisible )
+        @if (NotFoundVisible)
         {
             <DropdownItem Disabled>
                 @NotFoundContent( SelectedText )
             </DropdownItem>
         }
 
-        @if ( DropdownVisible )
+        @if (DropdownVisible)
```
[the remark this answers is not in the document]
David-Moreira: Vs is annoying... :X

Source/Extensions/Blazorise.Components/Autocomplete.razor.cs
David-Moreira: So, if you select a value autocomplete will no longer show it on the drop down options. This will still show it on the list....

stsrki: Ah, that makes sense. Yeah, it is a good feature.

David-Moreira: Right, the feature is basically what you're seeing on the print I posted on the first comment.
Just that is conflicts with the arrow/up selection, as the arrow/up selection has the same look.

David-Moreira: Maybe `SuggestSelected` I still think that's pretty similar to `Show` prefix eitherway. haha

Source/Extensions/Blazorise.Components/Autocomplete.razor.cs
```diff
@@ -671,6 +677,12 @@ protected bool IsTextSearchable
         protected string DropdownClassNames
             => $"{Class} b-is-autocomplete {( Multiple ? "b-is-autocomplete-multipleselection" : string.Empty )} {( TextFocused ? "focus" : string.Empty )}";
 
+        /// <summary>
+        /// Gets the custom class-names for dropdown element.
+        /// </summary>
+        protected string DropdownItemClassNames(int index)
+            => $"b-auto-complete-suggestion {(ActiveItemIndex == index ? "focus" : string.Empty)}";
```
stsrki, line 684: Naming should be consistent

image

Source/Extensions/Blazorise.Components/Autocomplete.razor.cs
```diff
@@ -916,16 +928,15 @@ public TValue SelectedValue
...
+        /// Suggests already selected option(s) when presenting the options.
         /// </summary>
-        [Parameter] public bool ShowSelected { get; set; }
+        [Parameter] public bool SuggestSelectedItems { get; set; }
 
         /// <summary>
-        /// Shows selected option(s) when presenting the searched values with checkboxes to ease selection when <see cref="Autocomplete{TItem, TValue}"/> is set to Multiple.
+        /// Suggests already selected option(s) when presenting the options with checkboxes to ease selection when <see cref="Autocomplete{TItem, TValue}"/> is set to Multiple.
         /// </summary>
         [Parameter] public bool ShowMultipleCheckbox { get; set; }
```
stsrki, line 938: Should this also be renamed to Suggest prefix?

Source/Blazorise.Bootstrap/wwwroot/blazorise.bootstrap.css
```diff
@@ -156,6 +156,11 @@
   max-height: var(--autocomplete-menu-max-height, 200px);
   overflow-y: scroll;
 }
+.b-is-autocomplete .dropdown-menu .dropdown-item.focus {
```
stsrki, line 159: This is currently part of BS only. What about other providers?

David-Moreira:
Closes #3828

Source/Blazorise.Bootstrap/wwwroot/blazorise.bootstrap.css
David-Moreira: Right, need to test and adapt other providers to do the same. Have yet to start that.

Source/Extensions/Blazorise.Components/Autocomplete.razor.cs
David-Moreira: Or just `MultipleCheckbox`? I'm not sure the Suggest really applies here. What do you prefer?

Source/Extensions/Blazorise.Components/Autocomplete.razor.cs
David-Moreira: It's strange that we used `b-is-{component}` instead of `b-{component}` tough. Might be an inconsistency to fix. 
May I add to our notes that the naming for components should be `b-{component}` and we fix any inconsitency like this in 2.0?

For example:
image

Source/Extensions/Blazorise.Components/Autocomplete.razor.cs
stsrki: Well, I think checkboxes should always be visible so that they can be checked and unchecked. Not sure what is the behavior for ShowMultipleCheckbox ?

Source/Extensions/Blazorise.Components/Autocomplete.razor
```diff
@@ -62,21 +62,34 @@
...
 
                 @if ( ItemContent is not null )
                 {
-                    <DropdownItem ElementId="@DropdownItemId(i)" @key="item" Active="@(IsActiveItem(item, i))" Class="@DropdownItemClassNames(i)" Value="@value" Clicked="@OnDropdownItemClicked">@ItemContent(new (item, value, text))</DropdownItem>
+                    <DropdownItem ElementId="@DropdownItemId(i)" @key="item" Active="@isActiveItem" Class="@DropdownItemClassNames(i)" Value="@value" Clicked="@OnDropdownItemClicked">
+                        @if ( SuggestMultipleCheckbox )
+                        {
+                            <Field Horizontal>
+                                <Check @onclick:stopPropagation TValue="bool" Checked="@isActiveItem"></Check>
+                                @ItemContent(new (item, value, text))
```
stsrki, line 77: Should `@ItemContent(new (item, value, text))` be inside of Check?

David-Moreira: I don't think so... It might be harder to handle complex markup styling inside the checkbox then outside? What do you think?

But maybe for the regular @text, it makes sense to go inside instead.

stsrki: Check is made to hold the content and to place it where belongs, HTML structure-wise, based on the used provider.

David-Moreira:
> There are some conflicts so I think you should be the one to fix them. Or I could accidentally break something. Agree?

Yep

stsrki (commented):
Got the error when typed a value in the search field

image

image

stsrki:
`SuggestMultipleCheckbox` seems strange to me. I think it would be better to have something like `enum MultipleSelectionMode` with Checkbox, Default, and values. Or just `SelectMultipleWithCheckbox`.

David-Moreira:
> `SuggestMultipleCheckbox` seems strange to me. I think it would be better to have something like `enum MultipleSelectionMode` with Checkbox, Default, and values. Or just `SelectMultipleWithCheckbox`.

It's like an enhancement over `SuggestSelectedItems` : 
image
Hence maintaining same prefix.

The `enum MultipleSelectionMode` is a good idea, I'd probably prefer to rework some stuff to be:
```
enum SelectionMode{
Default,
Multiple,
SuggestSelected, (this still bases itself off multiple mode behind the covers)
Checkbox (this still bases itself off multiple + suggestselected mode behind the covers)
}
```
What do you think? It would be breaking change tough. We could schedule an issue with this, and for now leave it like it is?

stsrki:
> The `enum MultipleSelectionMode` is a good idea, I'd probably prefer to rework some stuff to be:
> 
> ```
> enum SelectionMode{
> Default,
> Multiple,
> SuggestSelected, (this still bases itself off multiple mode behind the covers)
> Checkbox (this still bases itself off multiple + suggestselected mode behind the covers)
> }
> ```
> 
> What do you think? It would be breaking change tough. We could schedule an issue with this, and for now, leave it like it is?

Having `SuggestSelected` in the enum doesn't make sense. It can easily be a boolean because it is not a new mode per-se, just a modification of the existing Multiple selection mode.

`Checkbox` can be part of the SelectionMode.

David-Moreira:
> > The `enum MultipleSelectionMode` is a good idea, I'd probably prefer to rework some stuff to be:
> > ```
> > enum SelectionMode{
> > Default,
> > Multiple,
> > SuggestSelected, (this still bases itself off multiple mode behind the covers)
> > Checkbox (this still bases itself off multiple + suggestselected mode behind the covers)
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > What do you think? It would be breaking change tough. We could schedule an issue with this, and for now, leave it like it is?
> 
> Having `SuggestSelected` in the enum doesn't make sense. It can easily be a boolean because it is not a new mode per-se, just a modification of the existing Multiple selection mode.
> 
> `Checkbox` can be part of the SelectionMode.

You could argue the same for Checkbox then... not a new mode per-se... modification/enhancement of the existing....

David-Moreira:
> Default,
> Multiple,

Eitherway it's best to keep it for a future milestone right? `Multiple` is already implemented as a bool

stsrki:
> > Default,
> > Multiple,
> 
> Eitherway it's best to keep it for a future milestone right? `Multiple` is already implemented as a bool

Do you mean to keep new enum for the future? I would add it now in v1.1, and then mark  `Multiple` as obsolete and to be removed in v2.0.

David-Moreira:
Figure out : 
image

David-Moreira:
Attempt other proposed enhancements if time available.

> There should also be a search bar in the dropdown, before the first Item in the List.

(Investigate this one, might not make sense with our current component)

> There should be an Arrow/Chevron on the right side of the mainbar, so its more Intuitive for the User, that you can open a dropdown.

(seems like it wouldn't be too hard and makes sense)

stsrki:
> > There should be an Arrow/Chevron on the right side of the mainbar, so its more Intuitive for the User, that you can open a dropdown.
> 
> (seems like it wouldn't be too hard and makes sense)
> 
> From #3867

Don't underestimate it. Adding any up-down button can be a real pain. I know about it when I did the NumericEdit step buttons. And it's really hard to do it for all providers.

David-Moreira:
> > > There should be an Arrow/Chevron on the right side of the mainbar, so its more Intuitive for the User, that you can open a dropdown.
> > 
> > 
> > (seems like it wouldn't be too hard and makes sense)
> > From #3867
> 
> Don't underestimate it. Adding any up-down button can be a real pain. I know about it when I did the NumericEdit step buttons. And it's really hard to do it for all providers.

hmm then i Might not loose anytime with this at all then haha

David-Moreira:
> > > > > There should be an Arrow/Chevron on the right side of the mainbar, so its more Intuitive for the User, that you can open a dropdown.
> > > > 
> > > > 
> > > > (seems like it wouldn't be too hard and makes sense)
> > > > From #3867
> > > 
> > > 
> > > Don't underestimate it. Adding any up-down button can be a real pain. I know about it when I did the NumericEdit step buttons. And it's really hard to do it for all providers.
> > 
> > 
> > hmm then i Might not loose anytime with this at all then haha
> 
> Hey Guys, do I get it right and the suggested Checkbox wont be implemented?
> 
> Kind regards, Silas

Checkbox is currently in testing and should be available in v1.1.0 we were wondering if we had time to squeeze more features like the arrow indicator, which we probably won't yet.

David-Moreira:
@glutio Freetype an item, go back to typing a value that results in options, first item is no longer auto pre selected
chrome-capture-2022-7-22
.

David-Moreira:
@glutio would you be willing to briefly write up your notable changes on the v1.1 release page? 
AutoSelectFirstItem && the keyboard interaction. Just let us know if you'd rather that we write them instead.

image


You can add a mention to your github tag as having contributed. :)
image




Again, appreciated your hard work on this. :)

David-Moreira:
**Note to self**
Release notes / Docs:
Touch upon the new `SelectionMode` parameter 
`SuggestMultipleCheckbox` parameter no longer exists

stsrki:
How about we make Autocomplete a separate page in the demo app? Seems like it has grown quite a bit.

David-Moreira:
> How about we make Autocomplete a separate page in the demo app? Seems like it has grown quite a bit.

Right could probably condense the options into a single autocomplete like we do for Datagrid

glutio:
Sure, will take a look at the auto select and put something into the notes

glutio:
> first item is no longer auto pre selected

Cannot repro

https://user-images.githubusercontent.com/22550674/185979878-a4749d3f-628b-42f3-a116-167521e31c48.mp4

glutio:
About release notes, on second thought, I would need to create a separate branch but one of you could probably update the release notes directly in master, so maybe for simplicity I'll leave it to you to update it. I would write something like this, feel free to edit:

- Improved keyboard support. You can now select items and remove selection using keyboard alone. Use navigation arrows to select an item in the dropdown. In multi-select mode use Backspace to remove most recently selected item or use [Shift-]Tab and Enter key to remove any previously selected item. 

- Introduced AutoSelectFirstItem property for single-select mode. Use this property to automatically select the first item from Data during Autocomplete initialization.

Thanks!

David-Moreira:
> > first item is no longer auto pre selected
> 
> Cannot repro
> 
>  Blazorise.Autocomplete.component.and.9.more.pages.-.Work.-.Microsoft.Edge.2022-08-22.10-09-15.online-video-cutter.com.mp4

I'm not home yet. But from your video I think you missed the actual commit of the item by pressing enter or leaving the field. At least your video does not look exactly like my gif but I can detail the repro better later today.

David-Moreira:
> About release notes, on second thought, I would need to create a separate branch but one of you could probably update the release notes directly in master, so maybe for simplicity I'll leave it to you to update it. I would write something like this, feel free to edit:
> 
> * Improved keyboard support. You can now select items and remove selection using keyboard alone. Use navigation arrows to select an item in the dropdown. In multi-select mode use Backspace to remove most recently selected item or use [Shift-]Tab and Enter key to remove any previously selected item.
> * Introduced AutoSelectFirstItem property for single-select mode. Use this property to automatically select the first item from Data during Autocomplete initialization.
> 
> Thanks!

It's available in this branch, there's a docs project with the release notes. But we can go ahead and edit them with the texts you provided thanks.

glutio:
> but I can detail the repro better later today

nevermind, i was looking at this branch in my fork, which has old code...

glutio:
i have the fix but i don't have permissions to push to this branch.. attached patch - git apply patch.txt
[patch.txt](https://github.com/Megabit/Blazorise/files/9397449/patch.txt)

David-Moreira:
AHHH, sorry completly forgot the branch is under the organization, haha, yes you'd need to submit a PR on top of this branch, right.

Ok thanks.

David-Moreira:
It broke a unit test, can you please take a look?
image
chrome-capture-2022-7-22 (1)

glutio:
keyboard input is broken in the demo, for readdata it works every other time, for others it just does't work.

glutio:
it passes unit tests with this patch though

[patch.txt](https://github.com/Megabit/Blazorise/files/9410512/patch.txt)

glutio:
Updated patch [patch.txt](https://github.com/Megabit/Blazorise/files/9410667/patch.txt)

stsrki:
> Updated patch [patch.txt](https://github.com/Megabit/Blazorise/files/9410667/patch.txt)

How about I give you temporary write access?

Source/Extensions/Blazorise.Components/Autocomplete.razor.cs
```diff
@@ -784,7 +787,7 @@ public async Task OpenDropdown()
             if ( HasFilteredData && AutoPreSelect )
             {
                 ActiveItemIndex = 0;
-                ExecuteAfterRender( () => ScrollItemIntoView( Math.Max( 0, ActiveItemIndex ) ) );
+                ExecuteAfterRender( () => ScrollItemIntoView( ActiveItemIndex ) );
```
stsrki, line 790: I think `Math.Max( 0, ActiveItemIndex )` is more correct. Because when `ActiveItemIndex = -1` then `0` should be the max value used for scroll.

glutio: here I assume it will be 0 because of line 789 "ActiveItemIndex = 0"

stsrki: Never assume anything when programming 😅

glutio: I have to assume at least that the computer will do what it's told :)

glutio: here's an example ;) https://docs.microsoft.com/en-us/cpp/assembler/masm/assume?view=msvc-170

glutio:
Are you guys looking into the keyboard input issue? Can you repro? Just try typing into non-ReadData autocomplete in the demo

Source/Extensions/Blazorise.Components/Autocomplete.razor.cs
David-Moreira: Who knows, maybe `ActiveItemIndex` in the future changes, and has logic in it... then your assignment might not be = 0. :)
Altough it most likelly always be haha

David-Moreira:
> Are you guys looking into the keyboard input issue? Can you repro? Just try typing into non-ReadData autocomplete in the demo

I can look into it... as soon as I have time... damn these holiday months are hard...

David-Moreira:
> Are you guys looking into the keyboard input issue? Can you repro? Just try typing into non-ReadData autocomplete in the demo

So what is broken here again? I can type in all of them.

glutio:
I wasn't able to type anything, if that's gone then great

stsrki:
When I enable FreeTyping and start typing **a** and press ENTER, I cannot include **a** in the selected list. It keeps adding the items from the dropdown menu. Is this how it needs to behave?

image

Source/Extensions/Blazorise.Components/Autocomplete.razor.cs
```diff
@@ -53,156 +55,207 @@ public partial class Autocomplete<TItem, TValue> : BaseAfterRenderComponent, ICl
...
+            return first.Equals( second );
         }
 
-        /// <summary>
-        /// Triggers the reload of the <see cref="Autocomplete{TItem, TValue}"/> data.
-        /// Makes sure not to reload if the <see cref="Autocomplete{TItem, TValue}"/> is in a loading state.
-        /// </summary>
-        /// <returns>Returns the awaitable task.</returns>
-        public async Task Reload( CancellationToken cancellationToken = default )
+        private bool SequenceEqual<T>( IEnumerable<T> first, IEnumerable<T> second )
```
stsrki, line 95: Any reason why we don't use `ArrayExtensions.AreEqual` extension? Or introduce new method `ArrayExtensions.AreEqualOrdered`

Source/Extensions/Blazorise.Components/Autocomplete.razor.cs
```diff
@@ -53,156 +55,207 @@ public partial class Autocomplete<TItem, TValue> : BaseAfterRenderComponent, ICl
...
+
+        List<string> selectedTexts = new();
+        List<string> selectedTextsParam;
 
         #endregion
 
         #region Methods
 
-        protected async Task HandleReadData( CancellationToken cancellationToken = default )
+        private bool TValueEqual<T>( T first, T second )
```
stsrki, line 84: Can we use `ObjectExtensions.IsEqual` extension method?

David-Moreira:
> When I enable FreeTyping and start typing **a** and press ENTER, I cannot include **a** in the selected list. It keeps adding the items from the dropdown menu. Is this how it needs to behave?
> 
> image

So, like we've talked, this is that new Parameter `AutoPreSelect` that controls whether the behaviour you were expecting will work or not. You can tell from the print you've sent that the first option is pre selected, therefore it will select it on `Confirm` Key Press

Source/Extensions/Blazorise.Components/Autocomplete.razor.cs
David-Moreira: @glutio I believe you've introduced this, can you elaborate?

PR #3875, opened 2022-06-07
Modal: Fix Animations

Bobbar:
Through testing I found that the full duration delay in the ClosableAdapter is not necessary when transitioning to a visible state. I also found that the modal backdrop fade can be restored by decoupling it from the parent modal visible state and only changing the backdrop visibility before or after the animation.

I tested with all providers and it seems to work well.  It definitely feels more snappy without the delay.

### DIFF
--- a/Source/Blazorise/Components/Modal/Modal.razor
+++ b/Source/Blazorise/Components/Modal/Modal.razor
@@ -5,7 +5,7 @@
         <div @ref="@ElementRef" id="@ElementId" class="@ClassNames" style="@StyleNames" role="dialog" @attributes="@Attributes">
             @if ( RenderMode == ModalRenderMode.Default || ( RenderMode == ModalRenderMode.LazyReload && IsVisible ) || ( RenderMode == ModalRenderMode.LazyLoad && lazyLoaded ) )
             {
-                @if ( ShowBackdrop )
+                @if ( ShowBackdrop && IsBackdropVisible )
                 {
                     <_ModalBackdrop />
                 }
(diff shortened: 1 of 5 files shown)


Source/Blazorise/Components/Modal/Modal.razor.cs
```diff
@@ -456,6 +467,11 @@ public Task EndAnimation( bool visible )
         /// </summary>
         protected internal bool IsVisible => State.Visible == true;
 
+        /// <summary>
+        /// Returns true if the modal backdrop should be visible.
+        /// </summary>
+        protected internal bool IsBackdropVisible = false;
```
stsrki, line 473: Rename to `BackdropVisible`. An `Is` prefix is usually used for property getters. Will be in the guidelines soon.

stsrki (commented):
Tested and works excellent.

@David-Moreira do you have any comments?

Source/Blazorise/Components/Modal/Modal.razor
```diff
@@ -5,7 +5,7 @@
         <div @ref="@ElementRef" id="@ElementId" class="@ClassNames" style="@StyleNames" role="dialog" @attributes="@Attributes">
             @if ( RenderMode == ModalRenderMode.Default || ( RenderMode == ModalRenderMode.LazyReload && IsVisible ) || ( RenderMode == ModalRenderMode.LazyLoad && lazyLoaded ) )
             {
-                @if ( ShowBackdrop )
+                @if ( ShowBackdrop && IsBackdropVisible )
```
David-Moreira: Do we need to kill all the `IsBackDropVisible` "Gibberish"? Seems to just be syncing up with the `visible` state, no? Maybe for now it isn't needed?

David-Moreira:
> Tested and works excellent.
> 
> @David-Moreira do you have any comments?

Hmm only... Great Job! If it works even better we'll take it! haha!

Source/Blazorise/Components/Modal/Modal.razor
stsrki: This is a good approach from what I can see in the code. BackdropVisible must be set at certain places so it would work as expected.

stsrki (approved):

stsrki:
This is one of those PRs that are straight to the point and solve the real problem. And small enough to be easy to review. Great job @Bobbar

Issue #3516, opened 2022-03-02
DataGrid: in VirtualizeManualReadMode, if ReadData returns synchronously, sorting would not work


hillin:
**Describe the bug**
In the DataGrid component, if Virtualize is enabled and data is read with the `ReadData` callback, and the callback returns synchronously, the sorting (by clicking on column headers) would not work correctly.

This is because in such scenario, the updated (sorted) `Data` would not be reflected to the DataGrid component in time, the DataGrid will use old `Data` to render; i.e. [`SetParametersAsync`](https://github.com/Megabit/Blazorise/blob/b8cdc71b1352010832fef3f54a6650557ad5a231/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs#L241) is called after [`VirtualizeItemsProviderHandler`](https://github.com/Megabit/Blazorise/blob/b8cdc71b1352010832fef3f54a6650557ad5a231/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs#L1044).

hillin:
A repro has been pushed to https://github.com/hillin/BlazoriseIssues (use the Blazorise3516 project):
- Click on the `Name` column header, it appears we are sorting with the `Name` field ascendingly, but the list content won't change;
- Click again, it appears we are sorting with the `Name` field descendingly, but actually data is shown ascendingly sorted. Scroll down a bit to trigger another data read, you can see now the data is sorted descendingly
- Click again, we should not be sorting with the `Name` field, but the data is still sorted descendingly. Scroll down a bit to trigger another data read, you can see now the data is unsorted.

hillin:
Returns synchronously means, despite the method is an async method which returns a `Task`, nothing in it is awaited. A method which returns `Task.CompletedTask` is exactly such a synchronous method. 

If you have cloned the repro, you can try to remove the `async` modifier in the `OnReadDataAsync` method and return `Task.CompletedTask` in it, this won't change the behavior of the bug.

stsrki:
Adding `await Task.Yield();` works.

```cs
private async Task OnReadDataAsync( DataGridReadDataEventArgs<ListItem> e )
{
    await Task.Yield();

    var query = FakeData.AsQueryable();

    foreach ( var column in e.Columns.OrderBy( c => c.SortIndex ) )
    {
        switch ( column.SortDirection )
        {
            case SortDirection.Ascending:
                query = query.OrderBy( column.Field );
                break;

            case SortDirection.Descending:
[...]

PR #3935, opened 2022-07-03
Virtualize ManualReadMode : Workaround for Stale parameter 

David-Moreira:
Closes [#3516](https://github.com/Megabit/Blazorise/issues/3516)

I couldn't find a better fix. `Virtualize` `ManualReadMode` binds to the `VirtualizeItemsProviderHandler` that expects data to be returned. This data depends upon the `Data` parameter, which in this case, like the user reported, would use stale data, due to `SetParametersAsync` not have been run yet.

I have wondered this before, and `ReadData` does work great. 
But generally, for these apis, and just like in the `Virtualize` component, they tell you to return the data (instead of basically mutating it "externally") and then they do the work internally, do you remember why the choice of `EventCallback` over the "handler" approach? Or was it not considered at the time?

### DIFF
--- a/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
+++ b/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
@@ -237,7 +237,7 @@ namespace Blazorise.DataGrid
         {
             Aggregates.Add( aggregate );
         }
-
+        
         public override async Task SetParametersAsync( ParameterView parameters )
         {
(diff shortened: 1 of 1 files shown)


stsrki:
Well, `EventCallback` is always preferred over other handler types. And I couldn't remember any other way why we wouldn't use it.

In any case, I think I will approve this PR. And if it happens to not work in the future we can always go back and _fix it_.

stsrki (approved):

PR #3937, opened 2022-07-03
ListView : Render ChildContent at the bottom of the list if provided

### LINKED ISSUE #3808: Make ChildContent generic for ListView
ChildContent for ListView is non generic. 

Guess the idea was to use TextField for customization but its pretty hard to use for interaction and data binding.

David-Moreira:
Closes #3808

### DIFF
--- a/Source/Extensions/Blazorise.Components/ListView.razor
+++ b/Source/Extensions/Blazorise.Components/ListView.razor
@@ -8,4 +8,8 @@
             <ListGroupItem Name="@text">@text</ListGroupItem>
         }
     }
+    @if ( ChildContent is not null )
+    {
+        @ChildContent
+    }
 </ListGroup>
--- a/Source/Extensions/Blazorise.Components/ListView.razor
+++ b/Source/Extensions/Blazorise.Components/ListView.razor
@@ -8,8 +8,5 @@
             <ListGroupItem Name="@text">@text</ListGroupItem>
         }
     }
-    @if ( ChildContent is not null )
-    {
-        @ChildContent
-    }
+    @ChildContent
 </ListGroup>


Source/Extensions/Blazorise.Components/ListView.razor
```diff
@@ -8,4 +8,8 @@
             <ListGroupItem Name="@text">@text</ListGroupItem>
         }
     }
+    @if ( ChildContent is not null )
```
stsrki, line 11: I don't think we need a null check. It will be either rendered or not. You can just set `@ChildContent`

David-Moreira: You know what? I get the feeling I do this everytime haha Will change.

David-Moreira: I'm not sure I think I've found some unconventional ways it'll throw. Like when you manually build or compose components.

Also, what happens when you have both 
`ChildContent`
`AnotherContent`

And you only do
```
<ListView>
<AnotherContent>
</AnotherContent>
<ListView/>
```

stsrki: Nothing will happen in that case. The problem is when you try to mix the content without specifying ChildContent. eg.

```
<ListView>
  <AnotherContent>
  </AnotherContent>
  <div>some other content</div>
<ListView/>
```

In this case, Blazor will complain. The solution is to do:

```
<ListView>
  <AnotherContent>
  </AnotherContent>
  <ChildContent>
    some other content
  </ChildContent>
<ListView/>
```

stsrki (approved):

PR #3938, opened 2022-07-03
ListView : ItemContent w/ generic support

### LINKED ISSUE #3936: ListView : ItemContent feature
So for this, make it so `ChildContent` behaves as @stsrki said in the current release.
And as a feature, make it so there's an `ItemContent` to costumize each list item.

_Originally posted by @David-Moreira in https://github.com/Megabit/Blazorise/issues/3808#issuecomment-1173042092_

David-Moreira:
Closes #3936

Haven't tested yet.
Do you agree with BaseContext Api? The idea is that we use more of these, so we have a "container" where we can easily add new stuff without doing too much of a breaking change if needed.


Source/Extensions/Blazorise.Components/ListView.razor.cs
```diff
@@ -124,6 +125,11 @@ protected string ListGroupStyleNames
         /// </summary>
         [Parameter] public RenderFragment ChildContent { get; set; }
 
+        /// <summary>
+        /// Specifies the content to be rendered inside each item of the <see cref="ListView{TItem}"/>.
+        /// </summary>
+        [Parameter] public RenderFragment<ItemContext<TItem>> ItemContent { get; set; }
```
stsrki, line 131: We have settled on the `*Template` suffix so better to stick with it. Rename it to `ItemTemplate`.

Source/Extensions/Blazorise.Components/ListViewItemContext.cs
```diff
@@ -0,0 +1,27 @@
+﻿#region Using directives
```
stsrki, line 1: Remove empty regions.

Source/Blazorise/Base/BaseContext.cs
```diff
@@ -0,0 +1,26 @@
+﻿#region Using directives
+#endregion
+
+namespace Blazorise
+{
+    /// <summary>
+    /// Holds the Context for an Item based component.
+    /// </summary>
+    /// <typeparam name="TItem">Type of an item.</typeparam>
+    public abstract class BaseContext<TItem>
```
stsrki, line 10: `BaseContext` seems too generic of a name. It should be obvious to know the Context of what it represents?

Source/Extensions/Blazorise.Components/ListView.razor.cs
David-Moreira: Right, was just because there was already a `ChildContent`

Source/Blazorise/Base/BaseContext.cs
David-Moreira: `BaseTemplateContext`?

David-Moreira: `BaseRenderFragmentContext` 
`BaseFragmentContext`

Source/Extensions/Blazorise.Components/ListView.razor.cs
stsrki: Well, I think it would confuse people. ChildContent and ItemContent seem like the same thing for an untrained eye.

stsrki (commented):
Can you also add an example to the documentation?

David-Moreira:
> Can you also add an example to the documentation?

Done

stsrki (approved):

PR #3940, opened 2022-07-04
DataGrid : SelectedRowIndex

### LINKED ISSUE #3882: Add Grid SelectedRowNumber property
With DataGrid there aren't property that give the number of row selected, so .ScrollToRow(rownumber) is little tricky to understand how to fill, much more simple have property that give this number

int rowNumber = grid.SelectedRowNumber(item)

David-Moreira:
Closes #3882


Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -2094,6 +2094,8 @@ public IEnumerable<TItem> DisplayData
         [Parameter( CaptureUnmatchedValues = true )]
         public Dictionary<string, object> Attributes { get; set; }
 
+        public int SelectedRowIndex => Data.Index( x => x.IsEqual( SelectedRow ) );
```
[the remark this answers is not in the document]
David-Moreira: I guess that if Virtualize isn't set, this makes sense? Will change.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -2097,7 +2097,9 @@ public IEnumerable<TItem> DisplayData
         [Parameter( CaptureUnmatchedValues = true )]
         public Dictionary<string, object> Attributes { get; set; }
 
-        public int SelectedRowIndex => Data.Index( x => x.IsEqual( SelectedRow ) );
+        public int SelectedRowIndex => Virtualize 
+            ? Data.Index( x => x.IsEqual( SelectedRow ) ) 
+            : Data.Index( x => x.IsEqual( SelectedRow ) ) * CurrentPage;
```
stsrki, line 2102: Isn't it `(CurrentPage - 1) * PageSize + Data.Index(...)` for manual read mode? For in-memory I think page calculation is not needed.

David-Moreira: /facepalm haha

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -2097,6 +2097,23 @@ public IEnumerable<TItem> DisplayData
         [Parameter( CaptureUnmatchedValues = true )]
         public Dictionary<string, object> Attributes { get; set; }
 
+        /// <summary>
+        /// Returns a zero-based index of the currently selected row if found; otherwise it'll return -1.
+        /// Considers the current pagination.
+        /// </summary>
+        public int SelectedRowIndex
```
stsrki, line 2103: I wonder if `SelectedRowIndex` needs to have two-way binding? What if the user would like to do `@bind-SelectedRowIndex="@selectedRowIndex"`, instead of `@bind-SelectedRow="@selectedRow"`?

stsrki: Hahaha, I know, I can't help it. I remember when I used `SelectedRowIndex` in the past when I worked with another UI library. It was really helpful. But since it is a parameter, even if we add a two-way binding in the future it will not be a breaking change. So it is good for now as it is.

stsrki (approved):

stsrki (approved):

PR #3948, opened 2022-07-09
DataGrid MultipleSortOrder

### LINKED ISSUE #3281: Initial column order with SortMode=DataGridSortMode.Multiple
Since https://github.com/Megabit/Blazorise/issues/1278 we have a `SortIndex` in the `DataGridReadDataEventArgs`. Works fine (thanks). But there seems no way to create an initial order for the sort-columns.

E.g. the data table

|A|B|
|--|--|
|1|B|
|3|A|
|2|B|

results always to

|A^|B^|
|--|--|
|1|B|
|2|B|
|3|A|

I want to have a way to set this to

|A^|B^|
|--|--|
|3|A|
|1|B|
|2|B|

In addition to the `SortDirection`, I need a `SortIndex`-Property for the initial order of the sorted columns. This number should be used as index of the list of sorted columns (more or less inserting instead of adding to the list).
Same IDs (as well as not provided values) leads to the same sort order as it is now - IDs out of the current range should not lead to an error.

David-Moreira:
Closes #3281


Source/Extensions/Blazorise.DataGrid/DataGridColumn.cs
```diff
@@ -462,6 +471,16 @@ public SortDirection CurrentSortDirection
         /// </summary>
         [Parameter] public bool PreventRowClick { get; set; }
 
+        /// <summary>
+        /// Gets or sets the order for sorting when Sorting is set to multiple. 
+        /// </summary>
+        [Parameter] public int MultipleSortOrder { get; set; }
```
stsrki, line 477: I think MultipleSortIndex is a better naming convention. What do you think?

stsrki: Also, does it makes sense to remove `Multiple` from the name?

David-Moreira: Already have `DisplayOrder` `EditOrder`, makes sense to maintain same consistency.

> Also, does it makes sense to remove Multiple from the name?


Wouldn't suprise me, we can go ahead and remove the Multiple. Just leave the comment mentioning the order only makes sense when it is multiple.

stsrki: I forgot about those. Ok Order can stay but Multiple can be removed.

stsrki (approved):

PR #3953, opened 2022-07-11
Add MouseEventArgs to all Clicked events

### LINKED ISSUE #3952: Add MouseEventArgs to Clicked events
Some of our users need extra info about the click event and that's why we need to add MouseEventArgs as an extra parameter. The change is not breaking any code as far as I am aware even though the API has changed.

stsrki:
Closes #3952

Test code

```
<Button Color="Color.Primary" Clicked="@OnClick1">1</Button>
<Button Color="Color.Primary" Clicked="@OnClick2">2</Button>
<Button Color="Color.Primary" Clicked="@(()=>Console.WriteLine("3"))">3</Button>
<Button Color="Color.Primary" Clicked="@(()=>Console.WriteLine("4"))">4</Button>
@code {
    Task OnClick1()
    {
        Console.WriteLine("1");
        return Task.CompletedTask;
    }

    Task OnClick2( MouseEventArgs e )
    {
        Console.WriteLine("2");
        return Task.CompletedTask;
    }
}
```

### DIFF
--- a/Source/Blazorise.AntDesign/Button.cs
+++ b/Source/Blazorise.AntDesign/Button.cs
@@ -2,6 +2,7 @@
 using Blazorise.Extensions;
 using Microsoft.AspNetCore.Components;
 using Microsoft.AspNetCore.Components.Rendering;
+using Microsoft.AspNetCore.Components.Web;
 #endregion
 
 namespace Blazorise.AntDesign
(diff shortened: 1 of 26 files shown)


David-Moreira:
So, I remember you already had some of this work done on other places.
You had chosen this strategy of wrapping under our own `MouseEventArgs` API (`BLMouseEventArgs`), should we not continue the trend? Might also be more beneficial in the long run if we need to add Blazorise specific stuff?
image

David-Moreira:
Other then that looks good unless we've missed something! haha

stsrki:
I tried that approach also, but Blazor doesn't recognize our custom `BLMouseEventArgs` and because of that it doesn't adopt when Clicked is used with/without the supplied `eventArgs` parameter. Meaning it would be a breaking change and all users would be required to define it.

To be honest, I'm thinking of removing `BLMouseEventArgs` completely in v2. I'm not sure if it's worth keeping.

PR #3957, opened 2022-07-12
LoadingIndicator component

glutio:
I found that there is a need for some Wrapper component that would add Busy/Loading UI to other components. I wanted to have a component that is ready to go out of the box without the need for any glue code. So here's a basic usage example

```
<BusyLoading @ref="refBusy">
   <Chart />
</BusyLoading>
<Button Click="() => refBusy.Busy(true); RefreshChart(); refBusy.Busy(false)" />
```

It supports being called via @ref, via (two way) binding, via cascading parameter, or via an external service, busy indicator and loading state can be templated but it's ready to go and has some ability to customize appearance (change color or spinner size) out of the box. 

The BusyLoadingService class that comes with the wrapper mainly used for the global application busy wrapper - you register it as a scoped service, inject into main layout and set it on the wrapper, and then inject it into wherever you want to control the wrapper from. Another option for global wrapper is cascading parameter. Multiple wrappers can share an instance of the service to get triggered simultaneously. This type of logic could be left for developers, but i wanted it to be a "batteries included" kind of component.

It can be integrated with the SpinKit, I wasn't sure whether it should be implemented based on spinkit or just have spinkit as an option for templating the busy indicator, I ended up implementing an independet component with its own default spinner so you don't have to take dependency on spinkit without need.

Everybody makes their own, but I thought it would be nice to have one ready to go in the library. Please let me know if you think the same way before I add demo and docs. thanks!

https://user-images.githubusercontent.com/22550674/178591711-5d310a13-516d-46f7-b258-e03a56afc239.mp4


glutio:
recheck

glutio:

glutio:
recheck

Source/Extensions/Blazorise.BusyLoading/BusyLoading.razor.cs
```diff
@@ -0,0 +1,216 @@
...
+        {
+            get => isBusy ?? isBusyParameter;
+            set => isBusyParameter = value;
+        }
+
+        /// <summary>
+        /// Occurs when IsLoaded state has changed
+        /// </summary>
+        [Parameter]
+        public EventCallback<bool> IsLoadedChanged { get; set; }
```
stsrki, line 154: As per put [Contributing guidelines](https://github.com/Megabit/Blazorise/wiki/Contributing), you can remove the `Is` prefix.

stsrki (commented):
`BusyLoading` name seems a little strange to me. What about changing it? Maybe a `LoadingIndicator`?

glutio:
Addressed your comments and created a demo page. Will add docs, let me know if you have any concerns. Thank you.

https://user-images.githubusercontent.com/22550674/178894118-5e9f620e-9695-43b8-921c-7cb79a6cd5a2.mp4

glutio:
Updated demo page. Still working on the docs. Please let me know if I should change anything. I added "Middle" value to Placement enum if you don't mind...

https://user-images.githubusercontent.com/22550674/179307424-f65be90e-eb5d-49fe-bbb6-252ad3f8940f.mp4

glutio:
Created a documentation page. Please let me know what I can do the help get this merged. Thanks!

image

glutio:
Ok, I think it's done, let me know what you think. Thanks!

glutio:
Some issue with antdesign, binding doesn't work. Does not repro on other demos. Any ideas what it could be?

https://user-images.githubusercontent.com/22550674/179367273-26733aa1-2ca6-4bd6-90bb-35485978254e.mp4

glutio:
Ok, now feels done.

Source/Extensions/Blazorise.LoadingIndicator/LoadingIndicator.razor.cs
```diff
@@ -0,0 +1,312 @@
...
+            builder.Append( $"z-index:{ZIndex}", ZIndex.HasValue );
+            builder.Append( $"justify-content:start", IndicatorHorizontalPlacement == Placement.Start );
+            builder.Append( $"justify-content:end", IndicatorHorizontalPlacement == Placement.End );
+            builder.Append( $"align-items:start", IndicatorVerticalPlacement == Placement.Top );
+            builder.Append( $"align-items:end", IndicatorVerticalPlacement == Placement.Bottom );
+            builder.Append( $"padding:{IndicatorMargin}", !string.IsNullOrEmpty( IndicatorMargin ) );
+            base.BuildStyles( builder );
+        }
+
+        private void DirtyClassesAndStyles()
```
stsrki, line 89: Override from base class rather than adding new method

```
protected internal override void DirtyClasses()
{
    IndicatorClassBuilder.Dirty();

    base.DirtyClasses();
}
```

Source/Extensions/Blazorise.LoadingIndicator/LoadingIndicator.razor.cs
```diff
@@ -0,0 +1,312 @@
...
+                if ( busyParameter != newBusyParameter )
+                {
+                    busy = null; // use parameter instead of local value
+                }
+            }
+
+            return base.SetParametersAsync( parameters );
+        }
+
+        void IDisposable.Dispose()
```
stsrki, line 152: Override from our dispose pattern in the base class

```
protected override void Dispose( bool disposing )
{
    if ( disposing )
    {
       ...
    }

    base.Dispose( disposing );
}
```

Source/Extensions/Blazorise.LoadingIndicator/LoadingIndicator.razor.cs
```diff
@@ -0,0 +1,312 @@
...
+        /// <summary>
+        /// Indicates whether the component should be covered with a busy screen
+        /// </summary>
+        [Parameter]
+        public bool Busy
+        {
+            get => busy ?? busyParameter;
+            set
+            {
+                busyParameter = value;
```
stsrki, line 205: It's good to check if the value has changed in the setter and exit early.

```
if ( busyParameter == value) 
  return;
```

Source/Extensions/Blazorise.LoadingIndicator/LoadingIndicatorService.cs
```diff
@@ -0,0 +1,138 @@
...
+
+        #endregion
+
+        #region Methods
+
+        /// <summary>
+        /// Set Busy state
+        /// </summary>
+        /// <param name="val">true or false</param>
+        public void SetBusy( bool val )
```
stsrki, line 24: Always prefer full naming variables instead of short names. `val` to `value`.

Source/Extensions/Blazorise.LoadingIndicator/Styles/blazorise.loadingindicator.scss
```diff
@@ -0,0 +1,24 @@
+.bl-wrapper {
```
stsrki, line 1: Use prefix `b-loading-indicator-`.

stsrki (commented):
I'm impressed by the work so far. I haven't tried it live yet, but I can see if will be a great addition to Blazorise :)

Source/Extensions/Blazorise.LoadingIndicator/LoadingIndicator.razor.cs
```diff
@@ -0,0 +1,312 @@
...
+        /// Indicator horizontal position
+        /// </summary>
+        [Parameter]
+        public Placement IndicatorHorizontalPlacement { get; set; } = Placement.Middle;
+
+        /// <summary>
+        /// Indicator CSS margin (top right bottom left)
+        /// </summary>
+        [Parameter]
+        public string IndicatorMargin { get; set; }
```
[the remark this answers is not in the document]
glutio: now using IFluentSpacing

glutio:
Thanks for spending the time to review this. I made changes per your comments. I did not want to allow explicit styling of the indicator, only color and position, any custom styling should use IndicatorTemplate I think. All this padding business is mainly for this scenario, so that i can say

```
<LoadingIndicator SpinnerHeight="80%" 
        IndicatorHorizontalPlacement="LoadingIndicatorPlacement.End" 
        IndicatorPadding="Padding.Is2.FromEnd">
    <TextEdit />
</LoadingIndicator>
```

To get a nicely positioned spinner 
image

glutio:
Addressed your comments. Thanks!

Source/Extensions/Blazorise.LoadingIndicator/LoadingIndicatorService.cs
```diff
@@ -0,0 +1,192 @@
...
+            SetLoadedAction = indicator.SetLoaded;
+            GetBusyFunc = () => indicator.Busy;
+            GetLoadedFunc = () => indicator.Loaded;
+        }
+
+        /// <summary>
+        /// Set Busy state
+        /// </summary>
+        /// <param name="value">true or false</param>
+        public void SetBusy( bool value ) => SetBusyAction( value );
```
stsrki, line 58: What if instead of `SetBusy(bool)` we have a start/end APIs? eg.

- `service.StartLoading()`
- `service.EndLoading()`

Or something with similar meaning?

Source/Extensions/Blazorise.LoadingIndicator/LoadingIndicator.razor.cs
```diff
@@ -0,0 +1,312 @@
...
+                          <animateTransform attributeName = 'transform' type='rotate' values='0 64 64;45 64 64;90 64 64;135 64 64;180 64 64;225 64 64;270 64 64;315 64 64' calcMode='discrete' dur='720ms' repeatCount='indefinite' />
+                      </g>
+                  </svg>";
+
+
+        /// <summary>
+        /// Service used to control this instance.
+        /// </summary>
+        [Parameter]
+        public LoadingIndicatorService Service
```
stsrki, line 188: Can service be injected by default without for need to manually set it through parameter?

Source/Extensions/Blazorise.LoadingIndicator/LoadingIndicatorService.cs
```diff
@@ -0,0 +1,192 @@
...
+using System.Collections.Generic;
+using System.Linq;
+#endregion
+
+namespace Blazorise.LoadingIndicator
+{
+    /// <summary>
+    /// A service to control LoadingIndicator components
+    /// </summary>
+    public class LoadingIndicatorService
```
stsrki, line 12: I think it would be "cleaner" to have an interface `ILoadingIndicatorService`. And then the interface would be used by the DI.

Documentation/Blazorise.Docs/Pages/Docs/Extensions/LoadingIndicator/Examples/LoadingIndicatorStructureExample.snippet
```diff
@@ -0,0 +1,16 @@
+﻿@if (Loaded)
+{
+    <div Wrapper>
```
stsrki, line 3: It's somewhat difficult to read this structure that is written as `<div Wrapper>`. I must confess I was confused when I first saw it 😅

stsrki (commented):
Overall features are there, but API needs a little more cleanup.

Source/Extensions/Blazorise.LoadingIndicator/LoadingIndicator.razor.cs
glutio: Yes, having to set Service feels non-standard somehow but it makes the component more generic. If a service is injected by default then every indicator will get an instance of the same service and that's not good for a scenario where you want to use loading indicator for some form on the page for example. I think an **injected** service is really only for the global app only scenario, so you can define the wrapper in the main layout but control it from elsewhere. Maybe some hybrid approach? have a boolean param Default like <LoadingIndicator Default /> then it will use the service from DI. Some syntax sugar. Let me know what you think.

Source/Extensions/Blazorise.LoadingIndicator/LoadingIndicatorService.cs
glutio: What about loadingIndicator.Show()/Hide(), loadingService.Show()/Hide()

Documentation/Blazorise.Docs/Pages/Docs/Extensions/LoadingIndicator/Examples/LoadingIndicatorStructureExample.snippet
glutio: Yes, agreed, but I'm not sure what's a better way. maybe say <wrapper-div> instead?

Source/Extensions/Blazorise.LoadingIndicator/LoadingIndicator.razor.cs
stsrki: I think by default, we can have it injected. And then, with a parameter, we can override it.

Documentation/Blazorise.Docs/Pages/Docs/Extensions/LoadingIndicator/Examples/LoadingIndicatorStructureExample.snippet
stsrki: Just having a simplest real example is enough.

Source/Extensions/Blazorise.LoadingIndicator/LoadingIndicatorService.cs
stsrki: I like it. It follows the same API approach as other components.

Source/Extensions/Blazorise.LoadingIndicator/LoadingIndicator.razor.cs
glutio: Ok, I trust your judgement here. But I feel it will be ackward in scenario like 

```
<LoadingIndicator @ref="textBusy" **Service="null"**>
   <TextEdit/>
</LoadingIndicator>
```

glutio: I removed Service property. Service is now limited to global application busy scenario. So whoever is using this Extension component will get a wrapper to use with other components and an out of the box global app wrapper with service, the multiple-indicators scenario is something the devs will have to do on their own, by inheriting from LoadingIndicator and adding Service parameter for example.

```
// Config.cs
services.AddLoadingIndicator(); // adds scoped ILoadingIndicator service

// MainLayout.cs
@Body
<ApplicationLoadingIndicator /> <!-- no need to wrap anything since it's fullscreen -->

// Page.cs
@inject LoadingIndicatorService AppBusy

<Button Clicked="() => AppBusy.Show()" />
```
https://user-images.githubusercontent.com/22550674/180569800-f6a05c54-1087-491e-8ea6-9c6acfdc443d.mp4

Source/Extensions/Blazorise.LoadingIndicator/LoadingIndicatorService.cs
```diff
@@ -0,0 +1,196 @@
...
+            SetBusyAction = indicator.SetBusy;
+            SetLoadedAction = indicator.SetLoaded;
+            GetBusyFunc = () => indicator.Busy;
+            GetLoadedFunc = () => indicator.Loaded;
+        }
+
+        /// <summary>
+        /// Show loading indicator
+        /// </summary>
+        public void Show() => SetBusyAction( true );
```
stsrki, line 57: `Show`, `Hide`, `SetLoaded` should all be `Task`-able. Long term this is better because all our other show/hide APIs are awaitable.

Source/Extensions/Blazorise.LoadingIndicator/LoadingIndicatorService.cs
```diff
@@ -0,0 +1,196 @@
...
+            }
+            return val;
+        }
+
+        #endregion
+
+        #region Properties
+
+        /// <summary>
+        /// Returns Busy state shared by all indicator instances or null if the state is not the same for all indicators.
```
stsrki, line 185: Public comments can be moved to the interface and in implementation we just add `/// <inheritdoc/>`

glutio:
About IDisposable - I'm not an expert on this one, but looking here https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose#implement-the-dispose-pattern-for-a-derived-class I think we don't need to declare IDisposable on the derived class if the base class implements it.

Source/Extensions/Blazorise.LoadingIndicator/LoadingIndicatorService.cs
stsrki: Also, having a bool meaning multiple things seems kinda strange. Maybe a better approach is to have an enum?

eg

```
LoadingBusyState
{
  All,
  Partial,
  None,
}
```

stsrki: BTW, is `Busy` name still valid? Since we now have Show/Hide methods should we rename it to `Visible`?

stsrki:
> About IDisposable - I'm not an expert on this one, but looking here https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose#implement-the-dispose-pattern-for-a-derived-class I think we don't need to declare IDisposable on the derived class if the base class implements it.

The thing is, the base class is not inheriting directly from IDisposable. Only the implementation pattern is made. We did that on purpose in the past because we had many problems having all components using dispose even though they don't need it. There was also some other problem, but I don't remember exactly, @David-Moreira might remember it.

In any case, any component that wants to use dispose of pattern must also inherit from IDisposable.

Source/Extensions/Blazorise.LoadingIndicator/LoadingIndicatorService.cs
glutio: > should we rename it to `Visible`

Yes, I was thinking about that too, I'll change it. Thanks! 


> Maybe a better approach is to have an enum?

I think it's pretty standard to use bool? for yes/no/maybe for example WPF checkbox/toggle https://docs.microsoft.com/en-us/dotnet/api/system.windows.controls.primitives.togglebutton.ischecked?view=windowsdesktop-6.0

stsrki: > I think it's pretty standard to use bool? for yes/no/maybe for example WPF checkbox/toggle

OK, maybe we can leave it `bool?` then.

glutio: for Visible, looking a bit closer, looks like there's already a Visibility property, but it's not a boolean. Maybe use "bool Loading" instead of visible? Visible woudl match show/hide better, but since there's already Visibility, it may be confusing.. Any other good names? or can leave it at Busy.. Thoughts?

stsrki: We already have Show/Hide/Visible on other components like Modal, Dropdown, etc., so it would not be problematic. But, Start/Stop/Active is also a good alternative.

stsrki:
The example in the docs seems broken now

image

Not sure if it's my fault 😅

Can you look into it by any means?

glutio:
> Can you look into it

Sure, I'll take a look, this looks like a css issue maybe related to class providers refactoring ?

stsrki:
> > Can you look into it
> 
> Sure, I'll take a look, this looks like a css issue maybe related to class providers refactoring ?

Not sure. In the demo app, it seems to be working fine.

glutio:
Does not repro for me in docs. Please try Ctrl-F5. It happens sometimes after css changes. I just synched to the branch and ran the demo app and I had the same issue on the first run, but it fixed itself after Ctrl-F5.. This is some caching issue, which actually may be a real issue that needs to be addressed, in case users will get this after an update, they will be confused...

stsrki:
We usually handle caching by appending `?v=1.0.6.0` to the CSS and JS files. It seems that is missing for the docs. I can add it into 1.0 branch and merge it back to master and this PR. Leave that to me.

I have tried clearing cache but it still doesn't work.

glutio:
> Leave that to me.

Thanks!

> I have tried clearing cache but it still doesn't work.

Trying in all browsers I have, I cannot reproduce :(

glutio:
Ok, nevermind, got it to repro now, will fix it.

glutio:
Looks like the loading indicator css has not been loaded at all (no styles for b-loading-indicator... on the right)
image

stsrki:
Seems like an API is done and ready to be merged. One more thing to confirm. We're still able to assign service individually?

```
<LoadingIndicator Service="@someService">
  ...
</LoadingIndicator>
```

glutio:
Yes, the multiple indicator demo still works fine, that uses a local instance of LoadingIndicatorService
image

David-Moreira:
Hello,

image

Does it make a difference **App.razor** vs **MainLayout.razor**?
Our existing extensions already go by recommendation into **App.razor**. Could this still be viable?
image

David-Moreira:
Should we really call it boring? Nothing Blazorise provides is boring!!

Maybe just rephrase to : the default spinner / the default look spinner / the basic default spinner ?
image

David-Moreira:
image
Can it have the `Blazorise` prefix just like the `RichTextEdit` registration? For Intelisense and using the same convention purposes.

David-Moreira:
Isn't there a way to disable all fields in a form? Sometimes users have forms with a whole damn lot of fields.

I'm not sure if this is a good idea or not, or if I'm just complicating.

Maybe some helper, that could look at form and disable/enable every element inside (probably would need js interop for this.)

image

David-Moreira:
These are missing on the docs, we still describe them there.
image

Also can we have an example of using the declarative approach which is more of the Blazor way. Just like you have in the demo?
Blazor Users should be more used to doing `@bind-Visible` then getting a ref to the component and calling `.Show()` `.Hide()`

David-Moreira:
This is kinda strange, so the `LoadingIndicator` itself has a `Loading`? And then there's an `IndicatorTemplate` and a `LoadingTemplate`? 

image

Can this just generate redundancy and confusion?
So I have `Loaded` which has `LoadingTemplate` and a default look, then I also have `Visible` with an `IndicatorTemplate` and another default look?
image

What kind of use case do we see here? I must be missing something I guess.

David-Moreira:
`Width` and `Height` don't really seem to be doing anything different from one another since it's a circle and you're always keeping it proportional?

Maybe a single `Size` parameter would be better?
image

David-Moreira:
image

You're newing up a `LoadingIndicatorService`
image
Can you please adhere to DI, which is how a regular user should be using it? 

How would it work? What if we want to have multiple Loading with different Services controlling them? Can we still do it by injecting `ILoadingIndicadorService`? Probably not right? Maybe there should be a common shared identificator instead?

glutio:
> Our existing extensions already go by recommendation into **App.razor**

Makes no difference i think, I will update the doc

> Isn't there a way to disable all fields in a form?

Yes, I think you can just disable the entire form, we can add this to the doc

> What kind of use case do we see here?

Loading is used to display some UI before component is Initialied. We can rethink naming, because originally it was Busy/Loaded instead of Visible/Loaded. In my hobby project loading indicators start with Loaded = false, then OnInitializedAsyc my component loads data via WebApi and sets Loaded to true so the real UI is shown. 

> Can you please adhere to DI

But that's the whole point here. The DI is used by the global ApplicationLoadingIndicator, because we decare indicator in one page and use it in another so we get the reference via a third party, but if you want to control a bunch of indicators on your page then you just create a local service instance (you can still use DI but that makes no sense in this scenario) and LoadingIndicator does not require ctor DI, Service is just a property

> `Width` and `Height` don't really seem to be doing anything different

Different Widht and Height allow for more precise positioning of the spinner (use ShowBorder in the demo).. if one wants only height then just leave width null. 

> Users should be more used to doing `@bind-Visible` then getting a ref to the component and calling `.Show()` `.Hide()`

Sure we can add an example. But just using this in my hobby project I found that I want to call Show/Hide, because I do this inside an action.

David-Moreira:
Just a bit confusing to read. Can we do the `indicators.Add( indicator );` at the top just like we do with `indicators.Remove( indicator );`
Then the Single and Multi mode checks will look less weird? Eitherway, does it really make a difference to have both modes? Seems like it would work just fine all the same with the default multi implementation?
image

David-Moreira:
Why are these internal?

image

The purpose of interfaces is that we can provide a contract that can be overriden by a different implementation, How would a user provide is own implementation of the `ILoadingIndicatorService` without these?

David-Moreira:
image
Should this one somehow make sure it doesn't trigger other LoadingIndicators there might be on the page? Since this does Get the scoped  `ILoadingIndicatorService` right? Or does it not?

stsrki:
> Isn't there a way to disable all fields in a form? Sometimes users have forms with a whole damn lot of fields.
> 
> I'm not sure if this is a good idea or not, or if I'm just complicating.
> 
> Maybe some helper, that could look at form and disable/enable every element inside (probably would need js interop for this.)
> 
> image

Don't do it in this PR. I have that in plan for another milestone.

glutio:
> Seems like it would work just fine all the same with the default multi implementation

Agreed, just added it out of boredom, we can remove it

> How would a user provide is own implementation of the `ILoadingIndicatorService` without these

This is an internal interface (and I don't think we need an interfce at all here, but @stsrki suggested to add one and here we are). It's public within the assembly, so technically it's fine, but outside the assembly it's just an api interface, not to be implemented. That's the decidion i made, we can change it if you think public is better.

glutio:
> Also, why use `ServiceProvider` over regular Injection

Regular injection would throw and require the global indicator service. I wanted to allow Service property to override automatically registering with DI

David-Moreira:
> > Seems like it would work just fine all the same with the default multi implementation
> 
> Agreed, just added it out of boredom, we can remove it
> 
> Edit: I recall why I did it, I thought single use is likely by far the main case so why add locking penalty.
> 
> > How would a user provide is own implementation of the `ILoadingIndicatorService` without these
> 
> This is an internal interface (and I don't think we need an interfce at all here, but @stsrki suggested to add one and here we are). It's public within the assembly, so technically it's fine, but outside the assembly it's just an api interface, not to be implemented. That's the decidion i made, we can change it if you think public is better.

Well that's the reason for the interface to exist... You can provide a custom implementation and still work.
Imagine that a user wanted to get rid of the locking mechanism you have or provide a different version altogether?

glutio:
> Well that's the reason for the interface to exist...

Like I said, I'm not invested in it and we can change it, but basically we do not intend for this interface to be implemented, there are plenty of cases where the framework exposes an interface that's not to be implemented by clients - ["should not implement this interface" site:microsoft.com](https://www.google.com/search?q=%22should+not+implement+this+interface%22+site%3Amicrosoft.com&source=hp&ei=SBrtYrjPNtqhkPIPw9mKgAM&iflsig=AJiK0e8AAAAAYu0oWI6jJZBtDAnPyOluSAlFCmsvxcu-&ved=0ahUKEwi4k8uy5q_5AhXaEEQIHcOsAjAQ4dUDCAk&uact=5&oq=%22should+not+implement+this+interface%22+site%3Amicrosoft.com&gs_lcp=Cgdnd3Mtd2l6EANQAFgAYLUEaABwAHgAgAF-iAF-kgEDMC4xmAEAoAECoAEB&sclient=gws-wiz)

David-Moreira:
> > Our existing extensions already go by recommendation into **App.razor**
> 
> Makes no difference i think, I will update the doc
> 
> > Isn't there a way to disable all fields in a form?
> 
> Yes, I think you can just disable the entire form, we can add this to the doc
> 
> > What kind of use case do we see here?
> 
> Loading is used to display some UI before component is Initialied. We can rethink naming, because originally it was Busy/Loaded instead of Visible/Loaded. In my hobby project loading indicators start with Loaded = false, then OnInitializedAsyc my component loads data via WebApi and sets Loaded to true so the real UI is shown.
> 
> > Can you please adhere to DI
> 
> But that's the whole point here. The DI is used by the global ApplicationLoadingIndicator, because we decare indicator in one page and use it in another so we get the reference via a third party, but if you want to control a bunch of indicators on your page then you just create a local service instance (you can still use DI but that makes no sense in this scenario) and LoadingIndicator does not require ctor DI, Service is just a property
> 
A local service removes the point of DI you're no longer abstracting all the newing up... 

> > `Width` and `Height` don't really seem to be doing anything different
> 
> Different Widht and Height allow for more precise positioning of the spinner (use ShowBorder in the demo).. if one wants only height then just leave width null.
> 

Yea I noticed after... Basically controls the dimensions of the div not the actual spinner. The spinner will just accommodate itself accordingingky.

> > Users should be more used to doing `@bind-Visible` then getting a ref to the component and calling `.Show()` `.Hide()`
> 
> Sure we can add an example. But just using this in my hobby project I found that I want to call Show/Hide, because I do this inside an action.

Uhm, is it not the same thing as setting _visible = true and false instead of the explicit show hide calls? That's the point at least :) 
Anyway of the two way binding seems worth it in my. Opinion.

David-Moreira:
> > Well that's the reason for the interface to exist...
> 
> Like I said, I'm not invested in it and we can change it, but basically we do not intend for this interface to be implemented, there are plenty of cases where the framework exposes an interface that's not to be implemented by clients - ["do not implement this interface" site:microsoft.com](https://www.google.com/search?q=%22do+not+implement+this+interface%22+site%3Amicrosoft.com&ei=WxvtYuTXJ6fLkPIP46KScA&ved=0ahUKEwik8cy156_5AhWnJUQIHWORBA4Q4dUDCA4&uact=5&oq=%22do+not+implement+this+interface%22+site%3Amicrosoft.com&gs_lcp=Cgdnd3Mtd2l6EANKBAhBGAFKBAhGGABQkAZYhhFgvxNoAXAAeACAAYYBiAHTB5IBAzAuOJgBAKABAcABAQ&sclient=gws-wiz)
> 
> Edit:
> 
> > Well that's the reason for the interface to exist..
> 
> It could be unit testing potentially

I stand corrected. There might be times where it does not make sense.
The link you sent seems like very very internal implementations haha so we are not comparing the same.
Eitherway to me it made sense to allow it since its more or like expected from Web libraries... Even for testing it might be needed as public anyway... Up to you guys, just my opinion.

David-Moreira:
> > Also, why use `ServiceProvider` over regular Injection
> 
> Regular injection would throw and require the global indicator service. I wanted to allow Service property to override automatically registering with DI

Won't it get the scoped service? Meaning It will have other loading indicators associated?

David-Moreira:
> Loading is used to display some UI before component is Initialied. We can rethink naming, because originally it was Busy/Loaded instead of Visible/Loaded. In my hobby project loading indicators start with Loaded = false, then OnInitializedAsyc my component loads data via WebApi and sets Loaded to true so the real UI is shown.

So it's basically something for before the ui is shown? 
Seems like just keeping the same spinner all the way would be simpler...? 

I Honestly think the component should be simpler. As Loaded which has LoadingTemplate and a default look, then I also have Visible with an Indicator Template and a default look might generate confusion....

Anyway it works good and if its well documented I guess it's fine. If @stsrki also has no problem with this.

glutio:
> A local service removes the point of DI

Exactly :) here i don't need a service from the framework, I need a local one. Service is used in two cases - global app indicator and multiple indicators. For global app indicator we have to use DI (or some other way) to get the instance because it lives elsewhere in the environment and to maintain loose coupling we use some mediator for that, ie. the DI infrastructure (alternatively one can use a cascading global indicator, without service). For multiple indicators scenario the service is just a convenience, devs can do it themselves, but since this is likely a real scenario, we provide this feature out of the box. You have a number of elements on the page that all go busy at once, to implement that you just set the same service instance on each of them. you can have two different groups with two different service instances. This is just local to your page, you don't need it injected externally. 

> So it's basically something for before the ui is shown?

Yes, for example you may have seen online when a web page loads, it shows placeholder boxes instead of the real UI, while the page content loads, so Loaded can be used for that. Basically it's the state where the component is not ready to show itself but it's already in the DOM tree.

> I Honestly think the component should be simpler.

The driving force here was a desire to implement a control that's just ready out of the box. Otherwise it's really simple and you can implement it yourself. So this one comes with a spinner that you can customize, this one can be inline or fullscreen, this one has the service that you'd need for global app indicator DI, and you can control multiple of them at once too, you can use it as cascading value, as reference, or via binding, it just does everything that you need from a simple busy spinner so you don't need to create one yourself, this one is ready for any typical use case, that's exactly how i wanted it to be, so it ended up a bit more involved.

glutio:
> Uhm, is it not the same thing as setting _visible = true

I think mostly only two-way binding makes sense, to bind to Visible to disable control. Setting Visible to false or true via binding in practice seems like an extra step, because I would likely want to bind to Visible, I already have the reference, then why create another indirection while i can use the reference directly. There maybe other scenarios, but this one seems most typical. But I'll add some example for binding.

> The link you sent seems like very very internal

Yes, I noticed afterwards, I updated the link, which may still be unsatisfactory but I know I've see these many times on MSDN

glutio:
Thinking how to rename Loaded. How about ContentReady? Any other good names?

David-Moreira:
> Thinking how to rename Loaded. How about ContentReady? Any other good names?

As usual naming is hard... I like loaded better then content ready.... Maybe something along the lines of Init? Or Initialized? Otherwise just keep loaded and have it well documented.

Oh also I forgot one thing I gotta say the component is really good it covers a alot of scenarios and adheres great to blazorise conventions, really good job!

stsrki:
> Thinking how to rename Loaded. How about ContentReady? Any other good names?

If I understand it correctly the Loaded is used until the component is fully initialized. DataGrid has a similar thing named `LoadingTemplate`. Maybe name it along something similar, a `Loading`? Or `Initializing` as David suggested?

glutio:
I'll go with Intializing, because Loading in LoadingIndicator is confising. Also, Loading is not really the right word here, Busy would be more appropriate, since I'm not always loading, maybe I'm computing, or waiting.. If we change it to BusyIndicator then we can keep Loading in line with DataGrid. Let me know if you agree and I'll rename it everywhere otherwise I'll just change Loaded to Initializing..

Source/Extensions/Blazorise.LoadingIndicator/LoadingIndicator.razor.cs
```diff
@@ -250,10 +250,10 @@ public ILoadingIndicatorService Service
         /// Indicates whether component is ready to be rendered
```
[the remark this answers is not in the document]
glutio: Ok, updated comment

glutio:
Added fixed time animation for fadein, default true for global application indicator

https://user-images.githubusercontent.com/22550674/183224636-617afe4b-1b9c-4c24-8d3c-4dee40a887a6.mp4

David-Moreira:
> Added fixed time animation for fadein, default true for global application indicator
> 
>  Bootstrap.Demo.-.Profile.1.-.Microsoft.Edge.2022-08-05.16-48-07.online-video-cutter.com.mp4

Would. Loading indicator work fine with our animate component? Including the full screen? Might be worth a shot. If it does you can additionally just document it.

glutio:
> Might be worth a shot.

Ok, I got the idea, I will take a look, but Animate demo in the doc doesn't work in this branch (also interestingly the dropdown is not recorded on video for some reason, I use Win+G)

https://user-images.githubusercontent.com/22550674/183255444-8b96d819-39fd-464e-9e65-ac826c42824d.mp4

glutio:
Looks like it is not possible to use Animate as a wrapper around LoadingIndicator. It breaks loading indicator css, i think because it wraps everything in a div, and it doesn't support inline-block. But it's possible to use Animate directly in LoadingIndicator implementation. But after playing with it and trying it in different ways I concluded that the best implementation is to leave FadeIn for the overlay screen but add FadeInDuration parameter. Animating the the overlay screen like sliding it into view from a side looks ugly. Animating spinner in that way is cool but the overlay screen's position should not be animated. And then users can user IndicatorTemplate to add animation of the spinner. I will put an example in the doc.

Documentation/Blazorise.Docs.Server/Pages/_Layout.cshtml
```diff
@@ -28,6 +28,7 @@
     <link href="_content/Blazorise.Snackbar/blazorise.snackbar.min.css?v=1.0.6.0" rel="stylesheet" />
     <link href="_content/Blazorise.SpinKit/blazorise.spinkit.min.css?v=1.0.6.0" rel="stylesheet" />
     <link href="_content/Blazorise.LoadingIndicator/blazorise.loadingindicator.min.css?v=1.0.6.0" rel="stylesheet" />
+    <link href="_content/Blazorise.Animate/aos.css" rel="stylesheet" />
```
stsrki, line 31: Don't add `aos.css` manually. It will be handled by the Animate project when needed.

Source/Extensions/Blazorise.LoadingIndicator/LoadingIndicator.razor.cs
```diff
@@ -342,6 +341,11 @@ public bool Visible
         /// </summary>
         [Parameter] public bool FadeIn { get; set; }
 
+        /// <summary>
+        /// Fade in duration
```
stsrki, line 345: Add a comment saying it is in milliseconds.

Issue #3994, opened 2022-07-21
Fix TextEdit documentation (FocusIn and FocusOut)


Jinjinov:
**Describe the bug**
Fix TextEdit documentation (FocusIn and FocusOut) EventCallback args.

**To Reproduce**
https://blazorise.com/docs/components/text

    FocusIn | Occurs when the input box gains focus. | EventCallback<KeyboardEventArgs> 
    FocusOut | Occurs when the input box loses focus. | EventCallback<KeyboardEventArgs>

**Expected behavior**

    FocusIn | Occurs when the input box gains focus. | EventCallback<FocusEventArgs> 
    FocusOut | Occurs when the input box loses focus. | EventCallback<FocusEventArgs>

stsrki:
Done

PR #4007, opened 2022-07-24
Docs: TextEdit page fixes

mbaksh01:
Changed KeyboardEventArgs to FocusEventArgs for text edit docs.

Associated Work: #3994


stsrki:
It's a good fix. But, now I'm concerned if we should allow the merge for 1.1 or for 2.0. It kinda is a breaking change and it goes against the semver. 

@David-Moreira?

David-Moreira:
Breaking change that I would say we can open an exception since currently it's just straight up wrong and people should not be using that api like that.

David-Moreira:
OK I'm on mobile. I'm confused. Are we changing an api? Or just fixing docs?

stsrki:
You're right. My bad. I'm on mobile also and was quick to judge. Yeah this is just documentation update.

stsrki:
I have recreated this PR in the 1.0 branch to go live faster. You're going to be mentioned in the release notes in any case :)

Thanks @mbaksh01

PR #4040, opened 2022-08-11
Progress supports ElementId

blake-fm:
Nothing major - just added support for HTML id attr (via `ElementId`) to `<Progress>`.

Supremely untested, outside of `Blazorise.Bulma`.  Please feel free to fix/edit as you please.

### DIFF
--- a/Source/Blazorise.AntDesign/Progress.razor
+++ b/Source/Blazorise.AntDesign/Progress.razor
@@ -6,7 +6,7 @@
                 <div class="ant-progress-inner">
                     @if ( !HasProgressBar )
                     {
-                        <div class="@ProgressBarClassNames" style="@ProgressBarStyleNames" role="progressbar" aria-valuenow="@Percentage" aria-valuemin="@Min" aria-valuemax="@Max" @attributes="@Attributes">
+                        <div id="@ElementId" class="@ProgressBarClassNames" style="@ProgressBarStyleNames" role="progressbar" aria-valuenow="@Percentage" aria-valuemin="@Min" aria-valuemax="@Max" @attributes="@Attributes">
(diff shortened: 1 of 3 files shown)


stsrki:
Thank you for the sumbission. I have made some fixes on where to apply the element id. Our general rule is to apply it to the element that has `class="@ClassNames" style="@StyleNames"` as we consider it the main target when there are many elements to consider.

Before I merge, I would ask you if that solution would work for you.

stsrki:
PS. can you comment with `I have read the CLA Document and I hereby sign the CLA`?

blake-fm:
> Thank you for the sumbission. I have made some fixes on where to apply the element id. Our general rule is to apply it to the element that has `class="@ClassNames" style="@StyleNames"` as we consider it the main target when there are many elements to consider.
> 
> Before I merge, I would ask you if that solution would work for you.

Apologies; yes, that is perfect.  I fixed this for myself (Bulma), and skim-read / hurriedly impl. for the greater Blazorise library, without verifying anything.  Bad form on my part.  Should have just raised an issue with the suggestion.

RE: signing the CLA, unfortunately right now that is going to take some time (long, legal story).  I am happy to designate this code (the few characters of it) to public domain, provided that would not cause issues for you.

Also happy for you to simply close the PR and implement an identical change; my goal here was purely to save you some time.  This suggested change is yours to do with as you please.

stsrki:
Ok, I will keep this open for a few days, and if nothing changes regarding the CLA, I will close and re-implement my own changes.

Thanks

stsrki:
All is good with the PR. but closing because of the legal issues. We will make the changes by ourselves.

Thanks @blake-fm

Issue #4044, opened 2022-08-13
How to disable 1 option in SelectList?


David-Moreira:
How to disable 1 option in SelectList? 
![image](https://user-images.githubusercontent.com/96420517/184488326-b3385f3d-f2b0-459f-914f-bdba60d12800.png)

```
<SelectList TItem="MySelectModel"
            TValue="int"
            Data="@myDdlData"
            TextField="@((item)=>item.MyTextField)"
            ValueField="@((item)=>item.MyValueField)"
            SelectedValue="@selectedListValue"
            SelectedValueChanged="@MyListValueChangedHandler"
Disable/Readonly = "@((item)=>item.IsDisabled)"
            DefaultItemText="Choose your country" />
```

Ex: is it any param that accepts functions to determine if item should be disabled, like Disable/Readonly = "@((item)=>item.IsDisabled)" ?

_Originally posted by @PavloParafiloCricut in https://github.com/Megabit/Blazorise/discussions/4043_

PR #4045, opened 2022-08-14
Added DisabledField parameter to SelectList component

Jimmys20:
Hello, this PR closes https://github.com/Megabit/Blazorise/issues/4044.

### DIFF
--- a/Source/Extensions/Blazorise.Components/SelectList.razor.cs
+++ b/Source/Extensions/Blazorise.Components/SelectList.razor.cs
@@ -81,6 +81,11 @@ namespace Blazorise.Components
         [Parameter] public Func<TItem, TValue> ValueField { get; set; }
 
         /// <summary>
+        /// Method used to determine if an item should be disabled.
+        /// </summary>
+        [Parameter] public Func<TItem, bool> DisabledField { get; set; }
+
+        /// <summary>
         /// Currently selected item value.
         /// </summary>
         [Parameter] public TValue SelectedValue { get; set; }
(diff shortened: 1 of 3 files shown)


Jimmys20:

Source/Extensions/Blazorise.Components/SelectList.razor.cs
```diff
@@ -80,6 +80,11 @@ public Task Focus( bool scrollToElement = true )
         /// </summary>
         [Parameter] public Func<TItem, TValue> ValueField { get; set; }
 
+        /// <summary>
+        /// Method used to determine if an item should be disabled.
+        /// </summary>
+        [Parameter] public Func<TItem, bool> DisabledField { get; set; }
```
stsrki, line 86: Since we're disabling a `SelectItem` I think a better name would be something along `ItemDisabled`. What do you think?

Jimmys20: I added the "Field" suffix to be consistent with `TextField`, `ValueField` which are `Func` parameters with similar purpose. Maybe `ItemDisabledField`?

stsrki: The `field` naming was meant for a field inside of a TItem.

Jimmys20: Ok, I'll rename to `ItemDisabled`.

stsrki (approved):

stsrki:
Looking good. Thank you for your contribution.

Issue #2624, opened 2021-07-10
Steps navigation


kofinn:
Hi,

Is it possible to add new parameter, that will make navigation in Steps component unclickable (navigation should be still present, but it should not be used to switch between different steps)

![image](https://user-images.githubusercontent.com/22171329/125177794-0901c680-e1df-11eb-857f-a139f0a720fe.png)

Thank you

David-Moreira:
Hello @kofinn 
I don't think you can, no.
For feature proposal, let me know if this is what you'd like:
1     - Disable Step click navigation
2     - Programmatically select the step that you'd like to navigate to. **(Already possible with `SelectStep`)**
2.1 - `NextStep` and `PreviousStep` API could be usefull, right?

kofinn:
Hi,



> Hello @kofinn
> I don't think you can, no.
> For feature proposal, let me know if this is what you'd like:
> 1 - Disable Step click navigation
> 2 - Programmatically select the step that you'd like to navigate to. **(Already possible with `SelectStep`)**
> 2.1 - `NextStep` and `PreviousStep` API could be usefull, right?

kofinn:
Hi,

ok, currently I have workaround using SelectedStepChanged and SelectStep.

Maybe with these changes it could be new component 

Here is pretty much what I wish to do.
https://bbbootstrap.com/snippets/multi-step-form-wizard-animated-progressbar-53000683

David-Moreira:
Don't think we need a new component. Just a way to disable the click on the steps themselves. And Next Step / Previous Step apis which you can then bind to the buttons as you wish obtaining the same result as in the demo you just sent. 

Feature proposal for us to consider.

Jimmys20:
Hi @stsrki, @David-Moreira, I'm thinking of giving this a try. I have 2 questions:

When there is not any step selected, should the `NextStep` method go the first step or do nothing?
How to approach disabling the click? Add `DisableClickNavigation` parameter to the `Steps` component.

PR #4049, opened 2022-08-17
Steps: added NextStep, PreviousStep methods

Jimmys20:
Closes https://github.com/Megabit/Blazorise/issues/2624

### DIFF
--- a/Documentation/Blazorise.Docs/Pages/News/2022-09-01-release-notes-110.razor
+++ b/Documentation/Blazorise.Docs/Pages/News/2022-09-01-release-notes-110.razor
@@ -182,7 +182,7 @@
 </Heading>
 
 <Paragraph>
-    This feature was created by <Anchor To="https://github.com/Jimmys20">Jimmys20</Anchor>, one of our community memberes.
+    This feature was created by <Anchor To="https://github.com/Jimmys20">Jimmys20</Anchor>, one of our community members.
 </Paragraph>
 
 <Paragraph>
--- a/Documentation/Blazorise.Docs/Pages/News/2022-09-01-release-notes-110.razor
+++ b/Documentation/Blazorise.Docs/Pages/News/2022-09-01-release-notes-110.razor
@@ -149,7 +149,7 @@
 </Paragraph>
 
 <Paragraph>
-    Introduced a much needed missing <Code>ValueField</Code> Parameter, so you may uniquely identify each item in your list. 
+    Introduced a much needed missing <Code>ValueField</Code> Parameter, so you may uniquely identify each item in your list.
     This has been introduced with a return value of <Code>String</Code> for now. However we plan to make it return a generic type <Code>TValue</Code> in a future version, much like some of our other <Code>ValueField</Code> based components, so you can choose what type does your unique value return.
 </Paragraph>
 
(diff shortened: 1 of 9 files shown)


David-Moreira:
From a brief look this seems like it. Gotta check if there's edge cases that need null checks. Selected step being null or tabs being null?

@stsrki what you think of having the possibility of the component to render the steps buttons itself?

stsrki:
> @stsrki what you think of having the possibility of the component to render the steps buttons itself?
> 
> So that would mean you'd be setting a new tabs mode instead but you would need to provide content as childcontent

I would prefer to have those as part of another component based on Steps, Like we have SelectList based on Select, ListView based on ListGroup, etc.

Source/Blazorise/Components/Steps/Steps.razor.cs
```diff
@@ -152,6 +188,11 @@ public string SelectedStep
         /// </summary>
         [Parameter] public EventCallback<string> SelectedStepChanged { get; set; }
 
+        /// <summary>
+        /// Disables navigation by clicking on step.
+        /// </summary>
+        [Parameter] public bool DisableClickNavigation { get; set; }
```
stsrki, line 194: What do you think that we make it more dynamic? For example have a delegate that returns true is step is allowed.

**Example:**

```cs
[Parameter] public Func<StepContext, bool> NavigationAllowed { get; set; }
```

where StepContext would be something like

```cs
class StepContext
{
  int CurrentStep { get; set; }
  int NextStep { get; set; } // can be anything, prev or next index
}
```

Jimmys20: It seems like a good idea. Might be useful to add step name aswell.

class StepContext
{
  int CurrentStepIndex { get; set; }
  int NextStepIndex { get; set; }
  string CurrentStepName { get; set; }
  string NextStepName { get; set; }
}

stsrki: That's good. Better name might be StepChangeContext or StepNavigationContext. I'm open to suggestions.

Jimmys20: Should indexes be one-based or zero-based and what happens if a step has overridden the index by using the `Index` parameter?

stsrki: Good question. Since we already have DataGrid as one-based, I think we should do the same for the Steps.

stsrki: > what happens if a step has overridden the index by using the `Index` parameter?

The expected override should also be one-based. We should better document this so that the user is not confused.

Jimmys20: Added `NavigationAllowed` parameter, you can check when you have time.

stsrki: Seems good to me now. Thanks.

stsrki:
Seems good to me now. @Jimmys20, if you don't have anything else to add I can merge it.

Documentation/Blazorise.Docs/Pages/News/2022-09-01-release-notes-110.razor
```diff
@@ -171,4 +171,28 @@
...
+<NewsPageSubtitle>
+    PRs
+</NewsPageSubtitle>
+
+<Heading Size="HeadingSize.Is3">
+    Steps Navigation
+</Heading>
+
+<Paragraph>
+    This feature was created by <Anchor To="https://github.com/Jimmys20">Jimmys20</Anchor>, one of our community memberes.
```
Jimmys20, line 185: Typo on `memberes`

stsrki: fixed

stsrki (approved):

PR #3115, opened 2021-11-11
WIP: Introducing WASM PreRender

David-Moreira:

PR #4065, opened 2022-08-24
DRAFT: WebAssembly prerendering

stsrki:
The Bar component still expands only after the app is fully loaded. I guess we should do a major Bar rewrite to make it work as intended.

Note: https://github.com/aspnet/AspLabs/tree/main/src/QuickGrid

### DIFF
--- a/Demos/Blazorise.Demo.Bootstrap.Host/Program.cs
+++ b/Demos/Blazorise.Demo.Bootstrap.Host/Program.cs
@@ -1,0 +1,54 @@
+using Blazorise.Bootstrap;
+using Blazorise.Icons.FontAwesome;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+
+namespace Blazorise.Demo.Bootstrap.Host
+{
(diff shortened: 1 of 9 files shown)


David-Moreira:
So is this working better now? What's different from the last time we tried it? 

Let me look up the PR.

David-Moreira:
Hmm we didn't exchange thoughts on the PR itself, we probably discussed it over Discord, I'll have to look it up. But it's similar, so same issues might be here.

https://github.com/Megabit/Blazorise/pull/3115

Demos/Blazorise.Demo.Bootstrap.Host/Program.cs
```diff
@@ -0,0 +1,54 @@
...
+{
+    public class Program
+    {
+        public static void Main( string[] args )
+        {
+            var builder = WebApplication.CreateBuilder( args );
+
+            // Add services to the container.
+
+            builder.Services.AddControllersWithViews();
```
David-Moreira, line 17: Need this or copy paste?

stsrki: Probably not.

PR #4068, opened 2022-08-25
Fluent validation

### LINKED ISSUE #4067: FluentValidation extension
https://github.com/aladotnet/BlazoriseFluentValidation/issues/5

stsrki:
Closes #4067


Source/Extensions/Blazorise.FluentValidation/FluentValidationHandler.cs
```diff
@@ -0,0 +1,117 @@
...
+            else
+            {
+                var messages = result.Errors.Select( e => e.ErrorMessage );
+                validation.NotifyValidationStatusChanged( ValidationStatus.Error, messages );
+            }
+        }
+
+        private IValidator TryGetValidatorForObjectType( Type modelType )
+        {
+            var validatorType = typeof( IValidator<> );
```
stsrki, line 110: @David-Moreira, I think for the first version, we can leave it as it is and later we can optimize this by caching types and creating objects.

David-Moreira: My guess is that you already based yourself on some quite tested version of this, right? Because I remember there was some caveat when instantiating the Fluent Validators, like it wouldn't just work for every single case with just getting the generic type for some reason. I have to take a look at the implementation I have and see if there's something of note, I don't quite remember now. Something to do with inheritance or multiple generic types?

Source/Extensions/Blazorise.FluentValidation/Config.cs
```diff
@@ -0,0 +1,21 @@
...
+    public static class Config
+    {
+        /// <summary>
+        /// Adds an implementation of the FluentValidationHandler implementation of the IValidationHandler interface
+        /// </summary>
+        /// <param name="services"></param>
+        /// <returns></returns>
+        public static IServiceCollection AddBlazoriseFluentValidation( this IServiceCollection services )
+        {
+            services.AddScoped<FluentValidationHandler>();
```
David-Moreira, line 16: Why is it not the standard, `services.AddScoped<IValidationHandler, FluentValidationHandler>();` ?

Source/Extensions/Blazorise.FluentValidation/FluentValidationHandler.cs
```diff
@@ -0,0 +1,117 @@
...
+            var formValidatorType = validatorType.MakeGenericType( modelType );
+            return serviceProvider.GetRequiredService( formValidatorType ) as IValidator;
```
David-Moreira, line 112: Nowadays everyone uses the netcore DI I guess... 
But maybe there could be a "Reflection" based mode? Where the user would give some indication of the Validators assembly, and the FluentValidationHandler would figure it out?

Source/Extensions/Blazorise.FluentValidation/FluentValidationHandler.cs
David-Moreira: Also about the caching, I'm not even sure this needs it? Are these operations even "heavy"?

Source/Extensions/Blazorise.FluentValidation/FluentValidationHandler.cs
```diff
@@ -0,0 +1,117 @@
...
+            }
+            catch ( Exception exc )
+            {
+                HandleException( exc, validation );
+            }
+        }
+
+        private void HandleException( Exception exception, IValidation validation )
+        {
+            var msg = $"An unhandled exception occurred when validating field name: '{validation.FieldIdentifier.FieldName}'";
```
David-Moreira, line 71: This seems like stuff applicable to standard blazor validation? EditForm, field identifiers, etc, is this applicable to our validation system?

Source/Blazorise/Components/Validation/Validation.razor.cs
```diff
@@ -286,15 +286,15 @@ protected virtual Type DetermineHandlerType()
 
                 if ( Validator is not null || AsyncValidator is not null )
                 {
-                    return typeof( ValidatorValidationHandler );
```
David-Moreira, line 285: I'm not sure if this is the type of stuff that nowadays the compiler is smart enough to "cache" it... Because I mean... there isn't much it could do differently at runtime is there? Making it so this implementation is redudant? Can probably take this for a spin on the benchmark project.

David-Moreira (commented):
Can you hold on this PR so I can actually download the branch and test it? I didn't have time yet.

Source/Extensions/Blazorise.FluentValidation/FluentValidationHandler.cs
stsrki: > My guess is that you already based yourself on some quite tested version of this, right? Because I remember there was some caveat when instantiating the Fluent Validators, like it wouldn't just work for every single case with just getting the generic type for some reason. I have to take a look at the implementation I have and see if there's something of note, I don't quite remember now. Something to do with inheritance or multiple generic types?

I really don't remember it.

---

> Also about the caching, I'm not even sure this needs it? Are these operations even "heavy"?

Well, it can happen with each keystroke on the input. So with blazor-server, it can be "heavy"

Source/Extensions/Blazorise.FluentValidation/Config.cs
stsrki: Because we don't inject basic `IValidationHandler` anywhere. FluentValidationHandler is a special case for this extension so that's why we will inject it directly.

Source/Extensions/Blazorise.FluentValidation/FluentValidationHandler.cs
stsrki: That's done during the setup in `Startup.cs`, eg. `services.AddValidatorsFromAssembly( typeof( App ).Assembly );";
`

Source/Extensions/Blazorise.FluentValidation/FluentValidationHandler.cs
stsrki: In this version we're only doing validation based on supplied `Model`. Similar to data annotations. TBH I'm not sure how other validators would work with fluent validation.

Source/Extensions/Blazorise.FluentValidation/FluentValidationHandler.cs
David-Moreira: No, what that does is to register it on DI (with the help of reflection it seems).
I am talking about not using DI at all to find the IValidator? 

Like, let's imagine the user for some reason has two `Person` IValidators. He could do something like:
```
<Validations Validator="EmployeeValidator"/>
<Validations Validator="VisitorValidator"/>
```
Just an idea... not sure how to elegantly implement it.

Source/Extensions/Blazorise.FluentValidation/FluentValidationHandler.cs
David-Moreira: Ahhh I had forgotten we had this boilerplate already from adding data annotations and stuff.

Source/Blazorise/Components/Validation/Validation.razor.cs
David-Moreira: I guess I had the wrong idea, there's a benefit to it it seems, altough this is the that type of stuff that's quite negligible haha 
Still an improvement eitherway.
image

Source/Extensions/Blazorise.FluentValidation/FluentValidationHandler.cs
David-Moreira: I think it was something like this.
I don't know if it's a dumb use case haha, but anyway it seems like it works.
image

Altough you have to manually go register it.
image

David-Moreira:
It seems like if I define the HandlerType at the Validations level, then I can't override the Validation at the `Validation` level. for example:

Should this not be possible?
image
image

David-Moreira:
Should we provide `HandlerType` on the `DataGridColumn`? How would one go about using Fluent Validation on `DataGrid`?

Source/Extensions/Blazorise.FluentValidation/Config.cs
stsrki: I have actually started in that direction, but there was a problem. Must remember what it was.

Source/Extensions/Blazorise.FluentValidation/FluentValidationHandler.cs
stsrki: Hmm, yeah, you're right; having more validators for a model is a valid scenario. But considering that we're close to 1.1, I don't think we will have enough time. I will try something.

stsrki:
> Should we provide `HandlerType` on the `DataGridColumn`? How would one go about using Fluent Validation on `DataGrid`?

I forgot about `DataGrid` 😅. Yeah, we should have it there too. Maybe just add a `ValidationsHandlerType` on a DataGrid level.

stsrki:
> It seems like if I define the `HandlerType` at the `Validations` level, then I can't override the Validation at the `Validation` level. for example:
> 
> Should this not be possible? image image

Validation.HandlerType should have the highest priority, always.

David-Moreira:
> > It seems like if I define the `HandlerType` at the `Validations` level, then I can't override the Validation at the `Validation` level. for example:
> > Should this not be possible? image image
> 
> Validation.HandlerType should have the highest priority, always.

So I can't mix? Why not? :D
I'd have to not define the handler type on the validations component and then define the type of validation on each single Validation component?

stsrki:
> > > It seems like if I define the `HandlerType` at the `Validations` level, then I can't override the Validation at the `Validation` level. for example:
> > > Should this not be possible? image image
> > 
> > 
> > Validation.HandlerType should have the highest priority, always.
> 
> So I can't mix? Why not? :D I'd have to not define the handler type on the validations component and then define the type of validation on each single Validation component?

Ah yes, I see what you mean now. OK I will need to change priority a little.

Source/Extensions/Blazorise.FluentValidation/FluentValidationHandler.cs
stsrki: I will do it in 1.2

Source/Extensions/Blazorise.FluentValidation/Config.cs
stsrki: So, the problem with the global option is that we would not be able to distinguish between regular DataAnnotation and custom fluent validators. They both have a similar use-case API, in a way that both require a `Model` parameter. So if we use fluent validations, there would not be a way to go back to DataAnnotation. That's why `HandlerType` on a `Validations` level is better IMO, as it is more explicit. We can try to improve it in a later version if users start to request it.

PR #4074, opened 2022-08-26
Rel1.0.0 what's coming to .net7

jamattey:
Added Code Snippets. 

Requires Compilation to generate razor file.


jamattey (commented):
All Changes pushed without local compilation

PR #4103, opened 2022-09-13
Remove Disposed Component References

### LINKED ISSUE #4100: Modal has a memory leak (Blazor Server only)
**Describe the bug**
There is a memory leak in the Modal component in Blazor Server (there is no issue with WASM).

**To Reproduce**
1. Create a new Blazor Server project
2. Add Blazorise nuget packages (I've seen the issue with bootstrap4 and bootstrap5 packages) and add to Program.cs
4. Add a Modal to any page (can literally be just `<Modal/>`)
5. Navigate to and from that page several times
6. Take a memory snapshot and there will be a modal for every time you visited the page

**Expected behavior**
No memory leak

**Screenshots**
My `Counter.razor` file (the default for this file adding only `<Modal/>`)
image

Memory snapshot taken in Jetbrains dotMemory after forcing GC showing there are 7 Modals in memory after visiting the page 7 times. 
image

**Additional context**
None

EnsignPayton:
References to disposed components were being held by the service provider forever, causing a memory leak. This will clean up already disposed components from the service provider, allowing the GC to clean up all unreferenced controls

Closes #4100

### DIFF
--- a/Source/Blazorise/ComponentActivator.cs
+++ b/Source/Blazorise/ComponentActivator.cs
@@ -14,14 +14,6 @@ namespace Blazorise
     /// </summary>
     public class ComponentActivator : IComponentActivator
     {
-        #region Members
-
-        private bool disposePossible;
-
-        private readonly IList<object> disposables;
-
-        #endregion
-
         #region Constructors
 
         /// <summary>
@@ -31,7 +23,6 @@ namespace Blazorise
         public ComponentActivator( IServiceProvider serviceProvider )
         {
             ServiceProvider = serviceProvider;
-            disposables = LoadServiceProviderDisposableList();
         }
 
         #endregion
@@ -45,9 +36,6 @@ namespace Blazorise
         /// <returns>Return the newly created component or raises an exception if the specified typo is invalid.</returns>
         public IComponent CreateInstance( Type componentType )
         {
-            if ( disposePossible )
-                RemoveDisposedComponents();
-
             var instance = ServiceProvider.GetService( componentType );
 
             if ( instance == null )
@@ -61,36 +49,6 @@ namespace Blazorise
             }
 
             return component;
-        }
-
-        /// <summary>
-        /// Loads the ServiceProvider's list of disposables
-        /// </summary>
-        /// <returns>List of object references the ServiceProvider uses to track disposables</returns>
-        private IList<object> LoadServiceProviderDisposableList()
-        {
-            var disposablesPropertyInfo = ServiceProvider.GetType().GetProperty( "Disposables", BindingFlags.Instance | BindingFlags.NonPublic );
-
-            disposePossible = disposablesPropertyInfo is not null;
-
-            return disposablesPropertyInfo?.GetValue( ServiceProvider ) as IList<object> ?? Array.Empty<object>();
-        }
-
-        /// <summary>
-        /// Removes all disposed Blazorise components from the ServiceProvider's disposables list.
-        /// This prevents a memory leak where these references were being held permanently.
-        /// </summary>
-        private void RemoveDisposedComponents()
-        {
-            var disposedComponents = disposables
-                .OfType<BaseAfterRenderComponent>()
-                .Where( x => x.Disposed || x.AsyncDisposed )
-                .ToList();
-
-            foreach ( var component in disposedComponents )
-            {
-                disposables.Remove( component );
-            }
         }
 
         #endregion
--- a/Source/Blazorise/ComponentActivator.cs
+++ b/Source/Blazorise/ComponentActivator.cs
@@ -4,7 +4,6 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
 using Microsoft.AspNetCore.Components;
-
 #endregion
 
 namespace Blazorise
@@ -16,6 +15,8 @@ namespace Blazorise
     public class ComponentActivator : IComponentActivator
     {
         #region Members
+
+        private bool disposePossible;
 
         private readonly IList<object> disposables;
 
@@ -44,7 +45,8 @@ namespace Blazorise
         /// <returns>Return the newly created component or raises an exception if the specified typo is invalid.</returns>
         public IComponent CreateInstance( Type componentType )
         {
-            RemoveDisposedComponents();
+            if ( disposePossible )
+                RemoveDisposedComponents();
 
             var instance = ServiceProvider.GetService( componentType );
 
@@ -67,8 +69,10 @@ namespace Blazorise
         /// <returns>List of object references the ServiceProvider uses to track disposables</returns>
         private IList<object> LoadServiceProviderDisposableList()
         {
-            var disposablesPropertyInfo = ServiceProvider.GetType()
-                .GetProperty( "Disposables", BindingFlags.Instance | BindingFlags.NonPublic );
+            var disposablesPropertyInfo = ServiceProvider.GetType().GetProperty( "Disposables", BindingFlags.Instance | BindingFlags.NonPublic );
+
+            disposePossible = disposablesPropertyInfo is not null;
+
             return disposablesPropertyInfo?.GetValue( ServiceProvider ) as IList<object> ?? Array.Empty<object>();
         }
 
--- a/Source/Blazorise/ComponentActivator.cs
+++ b/Source/Blazorise/ComponentActivator.cs
@@ -1,6 +1,10 @@
 ﻿#region Using directives
 using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
 using Microsoft.AspNetCore.Components;
+
 #endregion
 
 namespace Blazorise
@@ -11,6 +15,12 @@ namespace Blazorise
     /// </summary>
     public class ComponentActivator : IComponentActivator
     {
+        #region Members
+
+        private readonly IList<object> disposables;
+
+        #endregion
+
         #region Constructors
 
         /// <summary>
@@ -20,6 +30,7 @@ namespace Blazorise
         public ComponentActivator( IServiceProvider serviceProvider )
         {
             ServiceProvider = serviceProvider;
+            disposables = LoadServiceProviderDisposableList();
         }
 
         #endregion
@@ -33,6 +44,8 @@ namespace Blazorise
         /// <returns>Return the newly created component or raises an exception if the specified typo is invalid.</returns>
         public IComponent CreateInstance( Type componentType )
         {
+            RemoveDisposedComponents();
+
             var instance = ServiceProvider.GetService( componentType );
 
             if ( instance == null )
@@ -46,6 +59,34 @@ namespace Blazorise
             }
 
             return component;
+        }
+
+        /// <summary>
+        /// Loads the ServiceProvider's list of disposables
+        /// </summary>
+        /// <returns>List of object references the ServiceProvider uses to track disposables</returns>
+        private IList<object> LoadServiceProviderDisposableList()
+        {
+            var disposablesPropertyInfo = ServiceProvider.GetType()
+                .GetProperty( "Disposables", BindingFlags.Instance | BindingFlags.NonPublic );
+            return disposablesPropertyInfo?.GetValue( ServiceProvider ) as IList<object>;
+        }
+
+        /// <summary>
+        /// Removes all disposed Blazorise components from the ServiceProvider's disposables list.
+        /// This prevents a memory leak where these references were being held permanently.
+        /// </summary>
+        private void RemoveDisposedComponents()
+        {
+            var disposedComponents = disposables
+                .OfType<BaseAfterRenderComponent>()
+                .Where( x => x.Disposed || x.AsyncDisposed )
+                .ToList();
+
+            foreach ( var component in disposedComponents )
+            {
+                disposables.Remove( component );
+            }
         }
 
         #endregion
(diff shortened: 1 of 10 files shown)


EnsignPayton:

stsrki:
I have tested it, and so far, it seems to be working as expected, meaning memory is cleared as it should. What I don't like about this approach, though, is the usage of reflection, as it can negatively impact performance and, in turn, the rendering of the UI.

Source/Blazorise/ComponentActivator.cs
```diff
@@ -48,6 +61,34 @@ public IComponent CreateInstance( Type componentType )
             return component;
         }
 
+        /// <summary>
+        /// Loads the ServiceProvider's list of disposables
+        /// </summary>
+        /// <returns>List of object references the ServiceProvider uses to track disposables</returns>
+        private IList<object> LoadServiceProviderDisposableList()
+        {
+            var disposablesPropertyInfo = ServiceProvider.GetType()
```
stsrki, line 70: @David-Moreira, I think we could optimize this by compiling it to an expression.

stsrki:
I figured one more scenario that this shouldn't work. When users are using third-party DI like SimpleInjector, Autofac, or any other. Since we depend on Disposables property, it would fail with an exception.

EnsignPayton:
I agree, ideally `ServiceProviderEngineScope` would be made public so we can access it directly.

Might be able to revert https://github.com/Megabit/Blazorise/pull/2167 with this change, and in that case this should only be called once

stsrki:
One more thing. `RemoveDisposedComponents` is called only when new components are about to be created. What happens if the user closes the browser tab? From what I understand, it will stay in the memory since no new component was created and `RemoveDisposedComponents` was not called.

---

🤔

So, we need to find a better way. I think the best way is to remove component from `ServiceProvider.Disposables` every time the actual component is disposed of. For that, we will have to have a new service. I'm thinking `ServiceProviderDisposer`. It is the same thing as you have created in this PR; just moved to the new class.

And then we can do:

```diff
public void Dispose()
{
    Dispose( true );
}

protected virtual void Dispose( bool disposing )
{
    if ( !Disposed )
    {
        Disposed = true;

        if ( disposing )
        {
            executeAfterRenderQueue?.Clear();

+            ServiceProviderDisposer.Dispose( this );
        }
    }
}
```

With this approach, if anyone has third-party DI, they would be able to have their own implementation of `ServiceProviderDisposer`.

stsrki:
@EnsignPayton, I have created a new IComponentDisposer, Can you test it on your side?

EnsignPayton:
Looks like it's working

Source/Blazorise/ComponentDisposer.cs
```diff
@@ -0,0 +1,85 @@
...
+
+        public void Dispose<TComponent>( TComponent component ) where TComponent : IComponent
+        {
+            if ( !disposePossible )
+                return;
+
+            if ( component is BaseAfterRenderComponent afterRenderComponent && ( afterRenderComponent.Disposed || afterRenderComponent.AsyncDisposed ) )
+            {
+                // does it makes sense to check if component exists in the collection?
+                disposables.Remove( component );
```
EnsignPayton, line 57: This will throw if the collection is actually an array, so I think that check is necessary

stsrki:
With Contains it seems to a LOT faster

image

EnsignPayton:
Pretty huge performance gain, nice

`ServiceProviderEngineScope.Disposables` internally returns `Array.Empty<object>` if the backing field is null, so I don't think switching that to a list has much effect

stsrki:
Hmm you're right 

https://github.com/dotnet/runtime/blob/cb52e3b88177cb86929b829d92b03561ad4e806b/src/libraries/Microsoft.Extensions.DependencyInjection/src/ServiceLookup/ServiceProviderEngineScope.cs#L15

stsrki:
@EnsignPayton can you test again?

EnsignPayton:
Diff between 1 page load vs. 20

image

Nothing in Blazorise namespace is being added per-visit, this looks good

stsrki:
Excellent. Thank you.

Now the only thing that is _kinda_ of bothering me is performance. When there are many users on the Blazor Server app, is it going to slow down due to constant Contains/Remove calls?! It's hard to test it.

stsrki:
@hikalkan Sorry to bother you. I would also like your opinion on this PR. Considering that you guys at ABP also have a custom implementation of `IComponentActivator` and this PR will definitely affect you. [source](https://github.com/abpframework/abp/blob/ec593a41991b6e7dff3208d4dd3b3bdecd2d2916/framework/src/Volo.Abp.AspNetCore.Components.Web/Volo/Abp/AspNetCore/Components/Web/AbpAspNetCoreComponentsWebModule.cs)

Basically, this PR tries to fix memory leaks by removing any transient component that is being disposed but is still part of the DI. At the moment, it only works with Microsoft DI, but since we have created `IComponentDisposer` as an interface it can be made to support any DI library.

Source/Blazorise/ComponentActivator.cs
David-Moreira: Hello,
Damn this PR is exactly what we were missing on Blazorise. I tought there would be no workaround if there wasn't a fix provided implicitly by .NET.

Can't wait to try it out. I should be able to tomorrow, and I'll see if the expression makes sense.
(PS: I've seen articles a few months ago with comparisions that reflection has come a long way in recent dotnet versions and that the old knowledge we have that reflection is hella slow, is no longer 100% true, matter of testing I guess... ^^)

David-Moreira:
Hello,

Starting to look at this. Seems to work great.

I very much also prefer the `ComponentDisposer` over having the `ComponentActivator` doing this job.
Now, is this 100% foul proof? Could I have components in the page that don't inherint `BaseAfterRenderComponent` and therefore not call `IComponentDisposer.Dispose`?

Source/Blazorise/ComponentActivator.cs
David-Moreira: So this `LoadServiceProviderDisposableList` is only ever called once per client, I don't think there's any noticeable performance hit. Right?

Source/Blazorise/ComponentDisposer.cs
```diff
@@ -65,10 +65,11 @@ public ComponentDisposer( IServiceProvider serviceProvider )
         private IList<object> LoadServiceProviderDisposableList()
         {
             var disposablesPropertyInfo = ServiceProvider.GetType().GetProperty( "Disposables", BindingFlags.Instance | BindingFlags.NonPublic );
+            var disposables = disposablesPropertyInfo?.GetValue( ServiceProvider ) as IList<object> ?? Array.Empty<object>();
 
-            disposePossible = disposablesPropertyInfo is not null;
+            disposePossible = disposables is not null && !disposables.GetType().IsArray;
```
David-Moreira, line 70: Why is this `IsArray` check? Seems like an hack, what is it for? Checking for Empty? Would `!disposables.IsNullOrEmpty()` not be the same and more clear?

David-Moreira:
> @hikalkan Sorry to bother you. I would also like your opinion on this PR. Considering that you guys at ABP also have a custom implementation of `IComponentActivator` and this PR will definitely affect you. [source](https://github.com/abpframework/abp/blob/ec593a41991b6e7dff3208d4dd3b3bdecd2d2916/framework/src/Volo.Abp.AspNetCore.Components.Web/Volo/Abp/AspNetCore/Components/Web/AbpAspNetCoreComponentsWebModule.cs)
> 
> Basically, this PR tries to fix memory leaks by removing any transient component that is being disposed but is still part of the DI. At the moment, it only works with Microsoft DI, but since we have created `IComponentDisposer` as an interface it can be made to support any DI library.

hmm Why would this affect APB? We've ended not using the `IComponentActivator` for this, right? Nor we should as it is not this interface intent to clear disposables and like you said it can just be overriden by someone.

David-Moreira:
> Excellent. Thank you.
> 
> Now the only thing that is _kinda_ of bothering me is performance. When there are many users on the Blazor Server app, is it going to slow down due to constant Contains/Remove calls?! It's hard to test it.

If you're really worried, we could potentially make it a disabeable feature. 
`options.UseBlazoriseComponentDisposer` or something like that. But I would leave it on by default, as we really want this to work and make part of the product.

It's really just in case something unforseen happens and user can easily disable with an option (he could just override the interface ofc). But I don't think it'll be a problem at all. The biggest problem was in fact the components hogging memory in my opinion.

Source/Blazorise/Config.cs
```diff
@@ -27,8 +27,8 @@ public static class Config
         /// <returns></returns>
         public static IServiceCollection AddBlazorise( this IServiceCollection serviceCollection, Action<BlazoriseOptions> configureOptions = null )
         {
-            serviceCollection.Replace( ServiceDescriptor.Transient<IComponentActivator, ComponentActivator>() );
-            serviceCollection.Replace( ServiceDescriptor.Transient<IComponentDisposer, ComponentDisposer>() );
+            serviceCollection.Replace( ServiceDescriptor.Scoped<IComponentActivator, ComponentActivator>() );
+            serviceCollection.Replace( ServiceDescriptor.Scoped<IComponentDisposer, ComponentDisposer>() );
```
David-Moreira, line 31: `serviceCollection.Replace`? Should it?

Source/Blazorise/ComponentDisposer.cs
stsrki: Because internally, disposables could be an array https://github.com/dotnet/runtime/blob/cb52e3b88177cb86929b829d92b03561ad4e806b/src/libraries/Microsoft.Extensions.DependencyInjection/src/ServiceLookup/ServiceProviderEngineScope.cs#L15

stsrki:
> 



> 



> > @hikalkan Sorry to bother you. I would also like your opinion on this PR. Considering that you guys at ABP also have a custom implementation of `IComponentActivator` and this PR will definitely affect you. [source](https://github.com/abpframework/abp/blob/ec593a41991b6e7dff3208d4dd3b3bdecd2d2916/framework/src/Volo.Abp.AspNetCore.Components.Web/Volo/Abp/AspNetCore/Components/Web/AbpAspNetCoreComponentsWebModule.cs)
> > Basically, this PR tries to fix memory leaks by removing any transient component that is being disposed but is still part of the DI. At the moment, it only works with Microsoft DI, but since we have created `IComponentDisposer` as an interface it can be made to support any DI library.
> 
> hmm Why would this affect APB? We've ended not using the `IComponentActivator` for this, right? Nor should we, as it is not this interface intent to clear disposables and like you said it can just be overriden by someone.

Because

1. ABP has it's IComponentActivator 
2. They're using Blazorise for Blazor implementation

And with `IComponentDisposer` we've only focused on Microsoft DI and ServiceProviderEngineScope implementation. In the ABP framework, you as a user can choose to form multiple different DI libraries. Meaning our IComponentDisposer will not work in all cases.

stsrki:
> > Excellent. Thank you.
> > Now the only thing that is _kinda_ of bothering me is performance. When there are many users on the Blazor Server app, is it going to slow down due to constant Contains/Remove calls?! It's hard to test it.
> 
> If you're really worried, we could potentially make it a disabeable feature. `options.UseBlazoriseComponentDisposer` or something like that. But I would leave it on by default, as we really want this to work and make part of the product.
> 
> It's really just in case something unforseen happens and user can easily disable with an option (he could just override the interface ofc). But I don't think it'll be a problem at all. The biggest problem was in fact the components hogging memory in my opinion.

I would rather make it enabled by default. If for any reason someone wishes to not have it, then we can add in an opt-in feature, or they could just override IComponentDisposer that does nothing(as a workaround)

Source/Blazorise/Config.cs
stsrki: I guess we don't need to have `Replace` as not a lot of people know or use `IComponentActivator`. And if they are using them better to not stay in their ways. OK, I will register it as a regular service.

Source/Blazorise/ComponentActivator.cs
stsrki: It is called once if the service is registered as `Scoped`. If it is `Transient` then it will be called every time for every component.

David-Moreira:
I just did the dumb test of setting up master and this PR side by side and look plainly at app memory, while trying to mimick the same operations in each one, I don't see memory being reclaimed as expected, altough it does seem like stuff does not show up on the memory diagnozer I don't think we're fully releasing everything that we should?

image

David-Moreira:
> I would rather make it enabled by default. If for any reason someone wishes to not have it, then we can add in an opt-in feature, or they could just override IComponentDisposer that does nothing(as a workaround)

Well, that's exactly what I said haha. Enabled by default but with the option to turn it off by setting option property to false. My suggestion was to provide the opt-in option right away, So we'd not have to patch.
Altough the workaround for overriding `IComponentDisposer` is always there, yes.

stsrki:
> I just did the dumb test of setting up master and this PR side by side and look plainly at app memory, while trying to mimick the same operations in each one, I don't see memory being reclaimed as expected, altough it does seem like stuff does not show up on the memory diagnozer I don't think we're fully releasing everything that we should?
> 
> image

If I remember it, .NET will not release memory immediately. Maybe keep it running for a while and see what happens.

Source/Blazorise/ComponentDisposer.cs
David-Moreira: It's still a (`IList<object>`) in the end. `IsNullOrEmpty` should work always.

Source/Blazorise/ComponentActivator.cs
David-Moreira: Kk then.
tested & modified.
image

stsrki: The results speak for themselves. I'm impressed by how much of a difference it is.

stsrki:
@EnsignPayton thank you for helping with this PR.

PR #4115, opened 2022-09-19
Markdown: override toolbar buttons and dynamic modules import

### LINKED ISSUE #4060: Markdown buttons not working
image

stsrki:
Closes #4060


Source/Extensions/Blazorise.Markdown/wwwroot/markdown.js
```diff
@@ -1,3 +1,8 @@
+import "./vendors/easymde.js?v=1.1.0.0";
+import "./vendors/highlight.js?v=1.1.0.0";
```
David-Moreira, line 2: This both has the advantage that we test and make blazorise work with a certain mde version, but also the disavantage that users can't easily upgrade their markdown js to a new version, right?

Source/Extensions/Blazorise.Markdown/wwwroot/markdown.js
```diff
@@ -1,3 +1,8 @@
...
+
+document.getElementsByTagName("head")[0].insertAdjacentHTML("beforeend", "<link rel=\"stylesheet\" href=\"_content/Blazorise.Markdown/vendors/easymde.css?v=1.1.0.0\" />");
```
David-Moreira, line 4: Altough it's automatic, users loose the flexibility of configuring the link however they want. Load first, in last, with some certain setting, etc? Maybe it could inspect the head verifying if it already exists or not?

David-Moreira (approved):
LGTM, this is still pending the next mde release with the prefix fix, right?

Source/Extensions/Blazorise.Markdown/wwwroot/markdown.js
David-Moreira: Oh I was talking about the link bit. But yea it'd make sense if you could update all conditionally. I don't think it's possible to import modules conditionally either.

Source/Extensions/Blazorise.Markdown/wwwroot/markdown.js
David-Moreira: Well... some fix in the library the user really needed, :P Anyway as long as we give continuos quick support It's no biggy, we update and launch a patch with the new version I guess

stsrki:
> LGTM, this is still pending the next mde release with the prefix fix, right?

Officially yes, it is still pending release. But I have downloaded their source and built the code. That's why I switched to JS modules.

Issue #4108, opened 2022-09-14
Treeview - Expand all nodes in the TreeView by default 


saeb-panahifar:
**Is your feature request related to a problem? Please describe.**
It seems that there is no mechanism to expand all nodes in the `TreeView` by default 

A clear and concise description of what the problem is. Example: I am trying to set the below `Treeview` expand but unfortunately, there is no specific property to configure. 
<img width="272" alt="Screen Shot 2022-09-14 at 10 33 17 AM" src="https://user-images.githubusercontent.com/17474030/190211602-ad31921d-9e7b-4635-9bb9-d620bcbe7804.png">

**Describe the solution you'd like**
A clear and concise description of what you want to happen. Include any alternative solutions you've considered.

David-Moreira:
Hello,

While a good feature to have.

You might be able to do this currently.
Have you tried adding the Nodes to the `ExpandedNodes` collection? Even if its the full collection? My guess is that it can still figure out the correct structure as long as `GetChildNodes` and `HasChildNodes` are well defined.

saeb-panahifar:
Here is a PR for this: https://github.com/Megabit/Blazorise/pull/4116

saeb-panahifar:
> Hello,
> 
> While a good feature to have.
> 
> You might be able to do this currently. Have you tried adding the Nodes to the `ExpandedNodes` collection? Even if its the full collection? My guess is that it can still figure out the correct structure as long as `GetChildNodes` and `HasChildNodes` are well defined.

Yes, I agree it is a useful feature. I checked that but it didn't work.

David-Moreira:
> > Hello,
> > While a good feature to have.
> > You might be able to do this currently. Have you tried adding the Nodes to the `ExpandedNodes` collection? Even if its the full collection? My guess is that it can still figure out the correct structure as long as `GetChildNodes` and `HasChildNodes` are well defined.
> 
> Yes, I agree it is a useful feature. I checked that but it didn't work.

Hello,
I had some treeview code sitting around from helping a customer. So I went and gave it a try since I was kinda sure it would work.
It does work, here's an example.

![image](https://user-images.githubusercontent.com/22283161/191356094-a90e5332-542d-4096-8e76-0884d33f974a.png)
![image](https://user-images.githubusercontent.com/22283161/191356137-b238a5a8-4f61-4f2b-a110-43dd1db8419d.png)
[...]

PR #4116, opened 2022-09-19
Expand all nodes in the TreeView by setting

saeb-panahifar:
Hi, 

The aim of this pull request is to resolve the below issue.  

https://github.com/Megabit/Blazorise/issues/4108 

Thanks,
Saeb


saeb-panahifar:

Source/Extensions/Blazorise.TreeView/TreeView.razor.cs
```diff
@@ -7,19 +7,19 @@
 
 namespace Blazorise.TreeView
 {
-    public partial class TreeView<TNode> : BaseComponent
+    public partial class TreeView : BaseComponent
```
stsrki, line 10: Changing from generic to the non-generic component is a breaking change. That means if we did that, it would be possible only in v2.0.

The alternative would be to have an internal `TreeNode<TNode>` that would hold the node state. That way TreeView API would not change.

Source/Extensions/Blazorise.TreeView/_TreeViewNode.razor.cs
```diff
@@ -21,71 +21,40 @@ public partial class _TreeViewNode<TNode> : BaseComponent
...
+        [Parameter] public RenderFragment<TreeNode> NodeContent { get; set; }
 
-        [Parameter] public IList<TNode> ExpandedNodes { get; set; } = new List<TNode>();
+        [Parameter] public Func<TreeNode, IEnumerable<TreeNode>> GetChildNodes { get; set; }
 
-        [Parameter] public EventCallback<IList<TNode>> ExpandedNodesChanged { get; set; }
+        [Parameter] public Func<TreeNode, bool> HasChildNodes { get; set; } = node => true;
 
-        [Parameter] public Func<TNode, IEnumerable<TNode>> GetChildNodes { get; set; }
+        [Parameter] public bool IsExpanded { get; set; } = true;
```
stsrki, line 55: We prefer naming without prefix like `Is*` for parameters, eg. `Expanded`.

Source/Extensions/Blazorise.TreeView/TreeView.razor.cs
stsrki: Well, I think `AllExpanded` can also work, as long as we keep track of internal changes in `SetParametersAsync`. But having both ways is also a good approach. Just like we have the `Visible` parameter and `Show`/`Hide `methods on some other components.

saeb-panahifar: I am wondering how can we access `IsExpanded` internally.  For need, this property because of collapse/expand for each node. I was thinking about the below code but it looks like the partial class doesn't support where clause.  

```csharp
public interface ITreeNode
{
        bool IsExpanded { get; set; }
}

public partial class _TreeViewNode<TNode> : BaseComponent
        where TNode:  ITreeNode
{
}

saeb-panahifar: > Yes, we ask that you focus on the PR itself and what it implements and not changing the public facing API at all, these should not be removed. image If you have suggestions about the public facing API we can consider that for the next major version **v2.0** but on a separate PR please. Thanks!
> 
> Also I'd argue a **ExpandAll** method would be more flexible then a flag **AllExpanded** that only works on `Initialized`? What do you guys think?

I agree, I accidentally remove `[Parameter] public EventCallback<TNode> SelectedNodeChanged { get; set; }` this is not related to this PR.

saeb-panahifar: > I am wondering how can we access `IsExpanded` internally. we need this property because of collapse/expand for each node. I was thinking about the below code but it looks like the partial class doesn't support the where clause.
> 
> ```cs
> public interface ITreeNode
> {
>         bool IsExpanded { get; set; }
> }
> 
> public partial class _TreeViewNode<TNode> : BaseComponent
>         where TNode:  ITreeNode
> {
> }
> ```

Also, in this case, we will have breaking changes. Any ideas?

David-Moreira: Hmmm I can't recall the Tree View source code now and I'm on mobile now. But I think you'll have to either check the expanded nodes collection for the reference so you can tell whether it's expanded.

Or keep internal state of the isexpanded some other way, which I believe there must be a way to track.

I guess it's weird that the expanded nodes are tracked by a collection instead of the isexpanded state in each node.

stsrki:
@saeb-panahifar Thank you for your contribution, but unfortunately, we will close this PR. The main reason is that we will take a different approach with the expand feature. Internally we plan to have some changes that are not aligned with changes in this PR. And also, we need to make it a non-braking change API. 

Once again, thank you for all the work.

Issue #4093, opened 2022-09-08
DataGrid : Adjust IsDisplayable check


treg:
**Describe the bug**
On a DataGridNumericColumn, I have set Displayable="false"
The column is still included in the . DisplayableColumns  results.

**To Reproduce**

**Expected behavior**
the Column should not be included in DisplayableColumns/

David-Moreira:
That seems like a very generic condition, but the change you suggest seems right in theory gotta look at what IsDispkayabke actually does first. 

Does regular column work with Displayable set to false? That could be a workaround for you for the time being.

treg:
The work around I'm using is to add HeaderClass , FilterClass, CellClass =>
hidden css

David-Moreira:
Hello @treg 
I just tested this on our demo page. By getting Displayable="false" it correctly hides the column.
`IsDisplayable` is an internal check of ours and it should not affect the correct behaviour as it is, as it will not collide.
![image](https://user-images.githubusercontent.com/22283161/191842697-77cad15e-1720-45fb-afae-63085a3ac2a1.png)
However it seems like it'd be a good change to still make `Displayable` hide the Column / MultiSelect column types if the user whishes to do so, but that's a different matter from the one you've reported.

`Displayable `
![image](https://user-images.githubusercontent.com/22283161/191842502-21cb30fe-4ffd-4c55-94db-68fab02ce636.png)

![image](https://user-images.githubusercontent.com/22283161/191842459-1d4bd852-9776-4f0f-bdd4-87340128a3bd.png)
vs
`Displayable="false"`
![image](https://user-images.githubusercontent.com/22283161/191842866-af52ecbd-2671-45fd-a4d1-cc8bac1f8d17.png)
![image](https://user-images.githubusercontent.com/22283161/191842900-cd720caf-0d31-457e-9dff-a46a908ba867.png)


Could I request that you provide us with a repository displaying the incorrect behaviour you've described?

PR #4124, opened 2022-09-22
DataGrid : Adjust IsDisplayable check. 

David-Moreira:
While investigating issue #4093.

Noticed that the current `DisplayableColumns` check makes it so `DataGridMultiSelectColumn` / `DataGridCommandColumn` would always show even if `Displayable="false"`. 

This does not seem appropriate behaviour. User should still be able to hide them if he wants to. Only real mandatory case, is when the `EditMode` is set to inline, and user needs the `CommandColumn` to display the Save/Cancel buttons. 

This has been tested on our demo DataGridPage with the use of `CommandMode.Default`, which helps having access to the Edit command through the ButtonRow if Command Column is hidden (`Displayable="false"`).


David-Moreira:
Ok.. what the...
remote test:
image
local test:
image

figures...

stsrki:
I've rerun jobs, and it seems that tests are working.

stsrki (approved):

Issue #4125, opened 2022-09-23
disabled button - generates a shadow


drma-dev:
![image](https://user-images.githubusercontent.com/12803050/191983529-9d660258-dc51-41f5-87ed-491a8b5b7a3c.png)
when I disable a boot, it casts a shadow on the button. I think there's something wrong. shadow should only be generated when clicking the button.

stsrki:
I agree it looks wrong. What provider do you use? BS4 or BS5?

drma-dev:
@stsrki 4

PR #4126, opened 2022-09-26
Removed shadow from disabled buttons

Jimmys20:
Fixes: https://github.com/Megabit/Blazorise/issues/4125

Removed `box-shadow` from disabled buttons on `BootstrapThemeGenerator` and `Bootstrap5ThemeGenerator`. This issue does not appear on the other providers.


stsrki (approved):
Looking good. Thank you!

PR #4149, opened 2022-09-28
Fix component disposer on wasm

### LINKED ISSUE #4137: Bug with ExpressionCompiler in 1.1
Once I tried to publish demo apps, I got this error.

image

Further investigation showed me that `instance.GetType()` in `ExpressionCompiler.CreatePropertyGetter` is behaving differently:

**WebAssembly demo apps**

image

**Server demo apps**

image

---

I have added try/catch as a workaround, but we need to fix this ASAP.

```cs
private IList<object> LoadServiceProviderDisposableList()
{
    try
    {
        if ( disposablesGetter is null )
            disposablesGetter = ExpressionCompiler.CreatePropertyGetter<IList<object>>( ServiceProvider, PROPERTY_DISPOSABLES );

        var disposables = disposablesGetter( ServiceProvider );

        disposePossible = !disposables.IsNullOrEmpty();

        return disposables;
    }
    catch
    {
        disposePossible = false;
    }

    return new List<object>();
}
```

David-Moreira:
Closes #4137

(Can you first try on your demos app before releasing)

### DIFF
--- a/Source/Blazorise/ComponentDisposer.cs
+++ b/Source/Blazorise/ComponentDisposer.cs
@@ -1,13 +1,9 @@
 ﻿#region Using directives
 using System;
 using System.Collections.Generic;
-using System.Linq;
-using System.Linq.Expressions;
-using System.Reflection;
 using Blazorise.Extensions;
 using Microsoft.AspNetCore.Components;
-using Microsoft.AspNetCore.Components.Forms;
-using Microsoft.Extensions.DependencyInjection;
+using Microsoft.JSInterop;
 #endregion
 
 namespace Blazorise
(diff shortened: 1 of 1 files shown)


stsrki:
No luck. After uploading to IIS, the error still happens. And it seems to be worse since it now tries to load the disposables list for each component.

image

Source/Blazorise/ComponentDisposer.cs
```diff
@@ -95,6 +95,11 @@ private IList<object> LoadServiceProviderDisposableList()
 
         #region Properties
 
+        /// <summary>
+        /// Indicates if the current app is running in webassembly mode.
+        /// </summary>
+        protected bool IsWebAssembly => jsRuntime is IJSInProcessRuntime;
```
David-Moreira, line 101: Agreed with just reverting it back and make it so we disable this for WASM so we can stabilize the release asap.
Why this runtime based check tough? 
Microsoft provides an utility check based on a symbol : 
`OperatingSystem.IsBrowser()`
image

stsrki: I didn't know about this, that's why 🤣

PR #4165, opened 2022-10-01
Blazorise.Demo.MAUI 

David-Moreira:
Adding demo for MAUI so we can test / investigate related issues and get acquainted with MAUI.

- We should configure it so it only targets Windows by default, otherwise Build asks us for additional workloads on the build process, Android, etc...

**WIP**
**PS:** No idea why our pipeline already requests for the maui-tizen workload. It should be disabled...

 > error NETSDK1147: To build this project, the following workloads must be installed: maui-tizen


David-Moreira:
You can merge this in by me. Took the maui demo out of the regular build.

We just need to explicit use it to debug certain issues that users report us.

stsrki (approved):

Issue #4166, opened 2022-10-02
Bar component doesn't work with Maui Blazor on iPhone in release configuration when setting Mode component parameter


emanuelpaul:
**Describe the bug**
Bar component doesn't work on dotnet maui blazor on iPhones in release configuration when using `Mode` compoenent parameter, but it works on iOS simulators and on iPhones using Debug configuration.
Tested on 
- MacOS 12.6.
- XCode 13.3.1 (13E500a)
- Dotnet sdk: 6.0.401
- dotnet workloads:
Installed Workload Ids      Manifest Version      Installation Source
---------------------------------------------------------------------
wasm-tools                  6.0.9/6.0.400         SDK 6.0.400
macos                       12.3.454/6.0.400      SDK 6.0.400
maui-maccatalyst            6.0.540/6.0.400       SDK 6.0.400
maui-ios                    6.0.540/6.0.400       SDK 6.0.400
maui-android                6.0.540/6.0.400       SDK 6.0.400

It works fine on simulators:
- iphone se 3 - iOS 15.4

It doesn't work on real iphone devices. I tested on:
- iphone se 3 - iOS 16

Exception that occurs:
```
Unable to set property 'Mode' on object of type 'Blazorise.Bar'. The error was: Attempting to JIT compile method '(wrapper delegate-invoke) void <Module>:invoke_callvirt_void_Bar_BarMode (Blazorise.Bar,Blazorise.BarMode)' while running in aot-only mode. See https://docs.microsoft.com/xamarin/ios/internals/limitations for more information.
```

**To Reproduce**
- clone repo: https://github.com/emanuelpaul/BlazoriseMaui 
- use branch `bar-bug`
- replace the value of CodesignKey and CodesignProvision with your values
- run command `dotnet build -t:Run -f net6.0-ios -p:_DeviceName={replace_with_your_iphone_identifier} -r ios-arm64 -c Release`

**Expected behavior**
Bar shouldn't trigger an exception when setting the `Mode` parameter.

**Screenshots**
![image](https://user-images.githubusercontent.com/2426861/193469569-1e309f31-7957-4c68-9f06-33a33062b51b.png)

**Additional context**
Exception details: 
[...]

PR #4182, opened 2022-10-06
ComponentDisposer : Get _disposables Field instead 

David-Moreira:
So after opening an issue over at Microsoft's, we figured it was most likelly the IL trimmer that was causing our new `ComponentDisposer` to not be able to access the `Disposables` Property when the app is published. (This is because tipically the trimmer only really runs when published and it really can't tell by default what it can' trim if `Reflection` is used)

So I have been testing this... for quite a few hours... following the recommended Microsoft's articles about the matter...
Specifically this bit, where apparently you can tell the compiler, "hey this type over here, keep info about certain members".
And while trying alot of variations of it, I just couldn't get it to work. Maybe because the actual resolved `Type` isn't really known until it is actually injected. :S All you got is an `IServiceProvider`, so I'm not sure, how this feature is supposed to work for these use cases...
image
https://learn.microsoft.com/en-us/dotnet/core/deploying/trimming/fixing-warnings

Also `ServiceProviderEngineScope` itself is internal, so I don't even think it's possible to trick it, by giving it more hints.

But then I started thinking... this has to be used by `.NET` why the hell is it trimmed, and by taking a look at the source code,
`Disposables` property is apparently only used for testing (at least according to the comment, and the class itself only accesses `_disposables`).  
image

By providing a `FieldGetter` instead on `_disposables`, this seems to be fixed. However I did this testing on a separate simpler app. I did not yet test publishing a `Blazorise` Demo, but im confident it will work.
I am off for the day, could you test it by publishing on of our apps @stsrki ?

Also, I found that WASM does not yet have `_disposables` set when we call `LoadServiceProviderDisposableList` on the `ctor`. I have added a single retry, which should fix it, and not affect Blazor.Server as it will typically not even retry.


stsrki (approved):
I tested it, and it's working as expected. Good job. I bet it was not easy at all.

stsrki:
Closes #4166

PR #4204, opened 2022-10-15
FileEdit: add JS open to all providers

### LINKED ISSUE #4198: FileEdit : fileEdit.js missing "open" function
**Describe the bug**

when calling  refFileEdit?.JSFileEditModule.OpenFileDialog(refFileEdit!.ElementRef, refFileEdit!.ElementId);
the FileEdit component does not display the input open file dialog

**To Reproduce**
1. create Bootstrap or Bootstrap5 project
2. Add component: <FileEdit @ref="@refFileEdit" Multiple/>
3. call refFileEdit?.JSFileEditModule.OpenFileDialog(refFileEdit!.ElementRef, refFileEdit.ElementId);
4. nothing happens

**Expected behavior**
I expect FileEdit component to display the input'ss open file dialog

** WORKAROUND **

1) added js function:

function OpenFileDialog(elementId) {
    element = document.getElementById(elementId);

    if (!element)
        return;

    element.click();
}
from C# Code Call:

jSRuntime!.InvokeVoidAsync("OpenFileDialog", refFileEdit!.ElementId);

David-Moreira:
Closes #4198
Well,
`JSFileEditModule.Open()` is currently only available via js in AntDesign. This is certainly due to it having a slight different implementation then the other providers. However since our `JSFileEditModule` is injected with a public modifier, it can't hurt to allow the `Open` method in every provider, and it might be usefull for someone.

`[Inject] public IJSFileEditModule JSFileEditModule { get; set; }`
My guess is that it might have been a miss and this should have been protected, since most modules are only for internal logic. 

We should probably consider reviewing Injected services access modifiers in v2.0


David-Moreira:
The testing code:

```
<FileEdit @ref="@fileEditRef" Multiple />
<Button Clicked="Open">Open</Button>
@code {
    private FileEdit fileEditRef;

    private Task Open()
    {
        return fileEditRef?.JSFileEditModule.OpenFileDialog( fileEditRef!.ElementRef, fileEditRef.ElementId ).AsTask();
    }
}
```

stsrki (approved):

PR #4234, opened 2022-10-26
Add ShowPicker function to date, time, and file inputs

### LINKED ISSUE #3374: Show picker manually by using new showPicker method
https://developer.chrome.com/blog/show-picker/

stsrki:
Closes #3374


Source/Blazorise/Components/FileEdit/FileEdit.razor.cs
```diff
@@ -253,6 +253,15 @@ public ValueTask RemoveFile( int fileId )
             return JSFileEditModule.RemoveFile( ElementRef, ElementId, fileId );
         }
 
+        /// <summary>
+        /// Show a browser picker for file input.
+        /// </summary>
+        /// <returns>A task that represents the asynchronous operation.</returns>
+        public Task ShowPicker()
+        {
+            return JSUtilitiesModule.ShowPicker( ElementRef, ElementId ).AsTask();
```
David-Moreira, line 262: Isn't this the same as the regular existing `Open` for the `FileEdit` ?

Documentation/Blazorise.Docs/Pages/Docs/Components/Dates/DateEditPage.razor
```diff
@@ -64,6 +64,25 @@
     <DocsPageSectionSource Code="DateEditDateTimeExample" />
 </DocsPageSection>
 
+<DocsPageSubtitle>
+    Functions
```
David-Moreira, line 83: I think we've used the name `Methods` on other places, `Datagrid`, `Autocomplete`, I honestly have no real preference. But I think in C# the naming `Method` is used more?  We should choose one or the other.

David-Moreira (approved):
Looking good. Would we ever like to call `ClosePicker`?

Source/Blazorise/Components/FileEdit/FileEdit.razor.cs
stsrki: We don't have `Open` on FileEdit.

Documentation/Blazorise.Docs/Pages/Docs/Components/Dates/DateEditPage.razor
stsrki: Agree. Methods make more sense.

Source/Blazorise/Components/FileEdit/FileEdit.razor.cs
David-Moreira: It is not directly exposed. It's in the module. We had a user complaining he'd access the module and couldn't call `Open` before. With this, I guess we have two ways to open it?
https://github.com/Megabit/Blazorise/pull/4204

stsrki: Oh, that. Then the JSmodule version doesn't make much sense anymore.

David-Moreira:
> Looking good. Would we ever like to call `ClosePicker`?

Did you notice this? You didn't comment on this.

stsrki:
> > Looking good. Would we ever like to call `ClosePicker`?
> 
> Did you notice this? You didn't comment on this.

Yeah, I noticed it, but I was driving 😅.

There is no ClosePicker API, so it's impossible to make it.

image

PR #4238, opened 2022-10-26
DatePicker: add RangeSeparator

### LINKED ISSUE #3552: Date Range Picker settings
Hey,

just seen you guys have added a new Date Picker.
I sort of like it! But I would like to suggest some changes/features to it.

Would it be possible to change the text of the range selection for example: from "to" into "-".
This could simplify the process of manually typing a date range.

Secondly I also like the multi selection of single dates. 
But would it be possible to combine the two Date Picker selection modes into one? Where you select a range when holding shift and still can pick and add a single day?

Last but not least, congrats on the release of 1.0.0 🎉 !!!

Kind Regards
Sebastian Gerner

stsrki:
Closes #3552

### DIFF
--- a/Source/Blazorise/Components/DatePicker/DatePicker.razor.cs
+++ b/Source/Blazorise/Components/DatePicker/DatePicker.razor.cs
@@ -205,7 +205,7 @@ namespace Blazorise
                     results.Add( Formaters.FormatDateValueAsString( value, DateFormat ) );
                 }
 
-                return String.Join( SelectionMode == DateInputSelectionMode.Multiple ? ", " : " to ", results );
+                return string.Join( SelectionMode == DateInputSelectionMode.Multiple ? ", " : CurrentRangeSeparator, results );
             }
             else
             {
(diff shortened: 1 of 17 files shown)


Source/Blazorise/Components/DatePicker/DatePicker.razor.cs
```diff
@@ -381,7 +379,8 @@ private object GetLocalizationObject()
                         Localizer["December"]
                     }
                 },
-                amPM = new[] { Localizer["AM"], Localizer["PM"] }
+                amPM = new[] { Localizer["AM"], Localizer["PM"] },
+                RangeSeparator = RangeSeparator ?? Localizer["RangeSeparator"],
```
David-Moreira, line 383: Use `CurrentRangeSeparator` helper? Which also has a slightly different logic, it fallbacks to `" to "` if everything null, which probably will never happen anyway

David-Moreira (approved):
LGTM. But should we also add `CurrentRangeSeparator` to `SetParametersAsync` change detection? We never know people's use cases :D

stsrki:
> LGTM. But should we also add `CurrentRangeSeparator` to `SetParametersAsync` change detection? We never know people's use cases :D

Not for now 😅

PR #4247, opened 2022-10-28
Chart: DataLabels extension

### LINKED ISSUE #4177: Chart: add DataLabels plugin
Investigate datalabels plugins to see if it's possible to integrate into Blazorise as an extension https://github.com/chartjs/chartjs-plugin-datalabels.

stsrki:
Closes #4177


Source/Extensions/Blazorise.Charts.DataLabels/Blazorise.Charts.DataLabels.csproj
```diff
@@ -0,0 +1,24 @@
...
+  <Import Project="..\..\..\Build\Blazorise.props" />
+
+  <PropertyGroup>
+    <PackageTags>blazorise blazor components chart datalabels</PackageTags>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <NoWarn>1591</NoWarn>
+  </PropertyGroup>
```
David-Moreira, line 12: Isnt this already common to all? Or should?

Source/Extensions/Blazorise.Charts/Options/ChartArcElements.cs
```diff
@@ -0,0 +1,54 @@
+﻿#region Using directives
+using System.Text.Json.Serialization;
+#endregion
+
+namespace Blazorise.Charts
+{
+    /// <summary>
+    /// Arcs are used in the polar area, doughnut and pie charts.
+    /// </summary>
+    public class ChartArcElements
```
David-Moreira, line 10: Can we use records? Are we forgetting about them? Or is there any limitation?

Source/Extensions/Blazorise.Charts.DataLabels/ChartDataLabels.razor.cs
```diff
@@ -1,16 +1,87 @@
...
+    /// </summary>
+    /// <typeparam name="TItem">Generic dataset value type.</typeparam>
+    public partial class ChartDataLabels<TItem> : BaseComponent, IAsyncDisposable
     {
+        #region Methods
+
+        /// <inheritdoc/>
+        protected override Task OnInitializedAsync()
+        {
+            if ( JSModule == null )
```
David-Moreira, line 42: Weren't we starting to use `is` `is not` for the null checking? Or old habbits die hard?? :)
Also check might be redudant.

David-Moreira: What would happen if an actual existing module was there for some weird reason? Then it wouldn't run `SetDataLabels`? Intended?

Source/Extensions/Blazorise.Charts.DataLabels/ChartDataLabelsOptions.cs
```diff
@@ -0,0 +1,78 @@
+﻿#region Using directives
+using System.Text.Json.Serialization;
+#endregion
+
+namespace Blazorise.Charts.DataLabels
+{
+    public class ChartDataLabelsOptions
```
David-Moreira, line 7: Record?

Source/Extensions/Blazorise.Charts.DataLabels/wwwroot/chart.datalabels.js
```diff
@@ -0,0 +1,44 @@
+import { getChart } from "../Blazorise.Charts/charts.js?v=1.1.2.0";
+
+// Register the plugin to all charts:
+Chart.register(ChartDataLabels);
```
David-Moreira: This line isn't wrapped in any way. For example in a document ready state, any way it can run when it should not?

Source/Extensions/Blazorise.Charts.DataLabels/wwwroot/chart.datalabels.js
```diff
@@ -4,23 +4,27 @@ import { getChart } from "../Blazorise.Charts/charts.js?v=1.1.2.0";
 Chart.register(ChartDataLabels);
 
 export function setDataLabels(canvasId, datasets, options) {
-    const chart = getChart(canvasId);
+    Retry(function () { return getChart(canvasId) }, 100, 10).then(chart => {
```
David-Moreira: Damn... ugly... But I guess it's becaused there's nothing in the DOM yet?
I think in other places we did the await in C# instead. Something like Task.Delay(150) to give enough time to render? But this might be safer I guess.

Source/Extensions/Blazorise.Charts/wwwroot/charts.js
```diff
@@ -111,8 +111,16 @@ function createChart(dotnetAdapter, eventOptions, canvas, canvasId, type, data,
...
-    if (pluginNames && pluginNames.includes("dataLabels") && ChartDataLabels) {
-        plugins.push(ChartDataLabels);
+    if (pluginNames) {
+        if (pluginNames.includes("DataLabels") && ChartDataLabels) {
+            plugins.push(ChartDataLabels);
+        }
+
+        if (pluginNames.includes("Streaming")) {
+            plugins.push({
+                duration: 20000
```
David-Moreira, line 121: Should this be configurable?

David-Moreira (approved):
LGTM.... I went from commit to commit, maybe I left some outdated comments, please disregard them if so.
It seems like you might have forgotten to add new enum to docs?

Also if you'd like me to actual do some testing on this, levae this open for a few more days I'll get to it.

Source/Extensions/Blazorise.Charts.DataLabels/Blazorise.Charts.DataLabels.csproj
stsrki: Not for extensions. `<NoWarn>1591</NoWarn>` is used so that we don't get too many warnings. But, yeah, in the long run we will need to comment all extensions and remove NoWarn.

Source/Extensions/Blazorise.Charts/Options/ChartArcElements.cs
stsrki: I was thinking about that. But it would not make sense at the moment. Because even we have records for this extension, it still uses regular classes from the core Chart project. So, the equality operator would not work as expected. I plan to convert all options to records in v2.0.

Source/Extensions/Blazorise.Charts.DataLabels/ChartDataLabels.razor.cs
stsrki: > Or old habbits die hard?? :)

Yes, this :D

---

> What would happen if an actual existing module was there for some weird reason? Then it wouldn't run SetDataLabels? Intended?

I don't see a reason for it to be anything else but null. We make this pattern everywhere.

Source/Extensions/Blazorise.Charts.DataLabels/wwwroot/chart.datalabels.js
stsrki: You're probably looking at an older commit. This is not true anymore.

Source/Extensions/Blazorise.Charts.DataLabels/wwwroot/chart.datalabels.js
stsrki: Old commit.

Source/Extensions/Blazorise.Charts/wwwroot/charts.js
stsrki: It should. But I will make it part of the new PR.

stsrki:
> LGTM.... I went from commit to commit, maybe I left some outdated comments, please disregard them if so. It seems like you might have forgotten to add new enum to docs?
> 
> Also if you'd like me to actual do some testing on this, levae this open for a few more days I'll get to it.

If you could test it would be great.

PR #4249, opened 2022-10-31
Fix Debounce

David-Moreira:
Some simple changes seem to be enough to keep it in sync. I'm happy that hopefully we finally get to fix this feature.

**To test:**
Add to Blazorise global config:
```
options.Debounce = true;
options.DebounceInterval = 300;
```

Test by going to `Autocomplete` and `Datagrid` and see that you can write in increments of 300ms and only then it updates the component internal value triggering the `Datagrid` Filtering or the `Autocomplete` search for example...

If you have any other ideas to test the implementation, say so.

Seems to work just fine even, with 100ms for example, didn't bother to test less then that, would beat the purpose of the feature. :)

Other implementations that I tought off, but apparently we might not need to explore:
- CancellationToken based
- Use `PeriodicTimer` to setup an async awaitable timer and actually wait TaskCompletion before issuing any other update.

### DIFF
--- a/Source/Blazorise/Base/BaseTextInput.razor.cs
+++ b/Source/Blazorise/Base/BaseTextInput.razor.cs
@@ -143,6 +143,7 @@ namespace Blazorise
         private void OnInputValueDebounce( object sender, string value )
         {
             InvokeAsync( () => CurrentValueHandler( value ) );
+            InvokeAsync( StateHasChanged );
         }
 
         /// <inheritdoc/>
--- a/Source/Blazorise/Base/BaseTextInput.razor.cs
+++ b/Source/Blazorise/Base/BaseTextInput.razor.cs
@@ -143,7 +143,6 @@ namespace Blazorise
         private void OnInputValueDebounce( object sender, string value )
         {
             InvokeAsync( () => CurrentValueHandler( value ) );
-            InvokeAsync( StateHasChanged );
         }
 
         /// <inheritdoc/>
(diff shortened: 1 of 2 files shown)


Source/Blazorise/Base/BaseTextInput.razor.cs
```diff
@@ -143,6 +143,7 @@ protected override Task OnBlurHandler( FocusEventArgs eventArgs )
         private void OnInputValueDebounce( object sender, string value )
         {
             InvokeAsync( () => CurrentValueHandler( value ) );
+            InvokeAsync( StateHasChanged );
```
stsrki, line 146: Hmm, is StateHasChanged needed here?

stsrki: I'm testing without `InvokeAsync( StateHasChanged );` on my side and it seems to work just fine. Can you also test and conform yourself?

David-Moreira: If you remove it, it goes back to happening as it goes out of sync. Simply type a bunch of words as if you were actually writting some paragraph and try to do it on a slightly more paused manner.

For example with this
`<TextEdit Debounce="true" DebounceInterval="300"  />`

So I would ask you to retest on your end with and without.

stsrki: Working with StateHasChanged

stsrki (approved):

PR #4262, opened 2022-11-03
Chart streaming improvements

### LINKED ISSUE #4242: How can I stop live data streaming on a live chart, after 20 seconds for example?

### LINKED ISSUE #4258: Pause streaming chart
### Discussed in https://github.com/Megabit/Blazorise/discussions/4257

<sup>Originally posted by **herzplatinedurchgebrannt** November  2, 2022</sup>
Hey guys,

is there a way to pause streaming charts?

chartjs offers an pause bool in the options.
https://nagix.github.io/chartjs-plugin-streaming/latest/guide/options.html#onrefresh

Unfortunately I cannot find this option in Blazorise.

Thank you in advance!</div>

### LINKED ISSUE #4260: Suggestions for Blazorise Live Chart option (Stop, Pause, Rewind, Forward functions)
No I have suggestions (see below)

1) The Live Chart should be stoppable any time if required. (Let say I have just 20 measuring points (each per 1 second) and I want that the X time axis movement can be stopped programatically.

2) There should be also a "Pause" function. That means I should be able to pause the chart if I want to look to the curve in detail, but the chart should collect values in the background,so that when I restart the chart I should be able to see the actual chart streaming.

3) In our language there is a saying " requesting is free of charge" :-) So I have a last request:
It would be even better If I would be able go back to the charts streaming history. (I mean rewind and forwardbuttons after I have stopped the chart.

Sorry for requesting so much :-)

stsrki:
Closes #4258
Closes #4260
Closes #4242

### DIFF
--- a/Source/Extensions/Blazorise.Charts.Streaming/ChartStreamingOptions.cs
+++ b/Source/Extensions/Blazorise.Charts.Streaming/ChartStreamingOptions.cs
@@ -1,25 +1,51 @@
-﻿namespace Blazorise.Charts.Streaming
+﻿#region Using directives
+using System.Text.Json.Serialization;
+#endregion
+
+namespace Blazorise.Charts.Streaming
 {
     public class ChartStreamingOptions
     {
         /// <summary>
         /// Duration of the chart in milliseconds (how much time of data it will show).
         /// </summary>
-        public int Duration { get; set; } = 10000;
+        [JsonPropertyName( "duration" )]
+        [JsonIgnore( Condition = JsonIgnoreCondition.WhenWritingNull )]
+        public int? Duration { get; set; } = 10000;
+
+        /// <summary>
+        /// Duration of the data to be kept in milliseconds. If not set, old data will be automatically deleted as it disappears off the chart.
+        /// </summary>
+        [JsonPropertyName( "ttl" )]
+        [JsonIgnore( Condition = JsonIgnoreCondition.WhenWritingNull )]
+        public int? Ttl { get; set; }
 
         /// <summary>
         /// Refresh interval of data in milliseconds. onRefresh callback function will be called at this interval.
         /// </summary>
-        public int Refresh { get; set; } = 1000;
+        [JsonPropertyName( "refresh" )]
+        [JsonIgnore( Condition = JsonIgnoreCondition.WhenWritingNull )]
+        public int? Refresh { get; set; } = 1000;
 
         /// <summary>
         /// Delay added to the chart in milliseconds so that upcoming values are known before lines are plotted. This makes the chart look like a continual stream rather than very jumpy on the right hand side. Specify the maximum expected delay.
         /// </summary>
-        public int Delay { get; set; } = 0;
+        [JsonPropertyName( "delay" )]
+        [JsonIgnore( Condition = JsonIgnoreCondition.WhenWritingNull )]
+        public int? Delay { get; set; } = 0;
 
         /// <summary>
         /// Frequency at which the chart is drawn on a display (frames per second). This option can be set at chart level but not at axis level. Decrease this value to save CPU power (https://github.com/nagix/chartjs-plugin-streaming#lowering-cpu-usage).
         /// </summary>
-        public int FrameRate { get; set; } = 30;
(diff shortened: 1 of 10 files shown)


Source/Extensions/Blazorise.Charts.Streaming/ChartStreamingOptions.cs
```diff
@@ -1,25 +1,51 @@
...
         /// <summary>
         /// Frequency at which the chart is drawn on a display (frames per second). This option can be set at chart level but not at axis level. Decrease this value to save CPU power (https://github.com/nagix/chartjs-plugin-streaming#lowering-cpu-usage).
         /// </summary>
-        public int FrameRate { get; set; } = 30;
+        [JsonPropertyName( "frameRate" )]
+        [JsonIgnore( Condition = JsonIgnoreCondition.WhenWritingNull )]
+        public int? FrameRate { get; set; } = 30;
+
+        /// <summary>
+        /// If set to true, scrolling stops. Note that onRefresh callback is called even when this is set to true.
```
David-Moreira, line 45: > Note that onRefresh callback is called even when this is set to true.

Anything the user should worry about?

Source/Extensions/Blazorise.Charts.Streaming/wwwroot/charts.streaming.js
```diff
@@ -96,4 +94,47 @@ function getStreamingOptions(dotNetAdapter, vertical, chartOptions, streamOption
...
+
+function getRealtimeOptions(chart) {
+    if (chart && chart.options && chart.options.scales) {
+        const scales = chart.options.scales;
+
+        if (scales.x && scales.x.realtime) {
+            return scales.x.realtime;
+        }
+        else if (scales.x && scales.x.realtime) {
+            return scales.x.realtime;
```
David-Moreira, line 107: The else logic looks the same, is it not? Was it supposed to fallback to **y** or something?

Source/Extensions/Blazorise.Charts.Streaming/wwwroot/charts.streaming.js
```diff
@@ -96,4 +94,47 @@ function getStreamingOptions(dotNetAdapter, vertical, chartOptions, streamOption
...
+export function play(canvasId, animate) {
+    const chart = getChart(canvasId);
+
+    if (chart) {
+        const realtimeOptions = getRealtimeOptions(chart);
+
+        if (realtimeOptions) {
+            realtimeOptions.pause = false;
+
+            chart.update(animate ? null : "none");
```
David-Moreira, line 137: I don't really get why null vs "none" depending on animate... but I imagine this is correct.

Source/Extensions/Blazorise.Charts.Streaming/ChartStreamingOptions.cs
David-Moreira: For example, user wants to stop a database query or api call that is being used on refresh?

David-Moreira (approved):
Code LGTM

PS: Do we want to add a simple Play/Pause to the docs also?  (I know its a bother to replicate demo and docs everytime haha)

Source/Extensions/Blazorise.Charts.Streaming/wwwroot/charts.streaming.js
stsrki: Yep, it should be `y`.

Source/Extensions/Blazorise.Charts.Streaming/wwwroot/charts.streaming.js
stsrki: "none" is a reserver keyword from chartjs. 🤷‍♂️

Source/Extensions/Blazorise.Charts.Streaming/ChartStreamingOptions.cs
stsrki: I don't know if this is wrong. We stop scrolling, but the streaming library still does its thing behind the scene. Maybe it should call refresh.

PR #4268, opened 2022-11-04
QRCode: support for icons

### LINKED ISSUE #3748: QRCode custom icon
Add a parameter accepting the custom icon data that can be rendered in the middle of the QR code.

https://github.com/codebude/QRCoder#optional-parameters-and-overloads

stsrki:
Closes #3748

New Icon parameter supports

- png, jpeg, gif
- base64 format
- sizing and border

Also, I have changed Payload so that it doesn't directly reference QRCoder package. It is a slight breaking change, but I think it is worth it.


David-Moreira:
Leave open, I'll test asap.

David-Moreira:
Can we validate and do nothing if it is not a valid Base64? Otherwise boom:
image

Source/Extensions/Blazorise.QRCode/QRCode.cs
```diff
@@ -63,30 +67,53 @@ public string GetAsBase64Image()
...
+            {
+                using ( var image = GetImage( Icon ) )
+                {
+                    var imageType = GetImageType( IconImageType );
+
+                    return $"data:image/png;base64, {code.GetGraphic( PixelsPerModule, darkColor, lightColor, image, IconSizePercent, IconBorderWidth, DrawQuietZones, imageType )}";
+                }
+            }
+
+            return $"data:image/png;base64, {code.GetGraphic( PixelsPerModule, darkColor, lightColor, DrawQuietZones )}";
```
David-Moreira, line 89: On these lines we have : data:image/png;base64 is this relevant if it's a gif or jpeg? I did try setting a jpeg base64 and it did work fine visually.

Documentation/Blazorise.Docs/Pages/Docs/Extensions/QRCode/QRCodePage.razor
```diff
@@ -78,6 +93,14 @@
     To optimize your web for SEO, define the <Code>Alt</Code> parameter to describe the QR image content, eg. <Code>Alt="QRCode image"</Code>.
 </DocsPageParagraph>
 
+<Heading Size="HeadingSize.Is3">
+    Increase Error Level for Images
```
David-Moreira, line 97: "Error Level" seems wrong, **ECC**, stands for **Error Correction Level**, and  I think that's better for interpretation.

Source/Extensions/Blazorise.QRCode/Enums/EciMode.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace Blazorise.QRCode
+{
+    public enum EciMode
```
David-Moreira, line 3: Missing comment, No idea why this does? Some kind of encoding?
Apparently it is set but not used?

Source/Extensions/Blazorise.QRCode/QRCode.cs
```diff
@@ -133,6 +160,26 @@ private static byte[] GetColorBytes( string color )
...
+
+        /// <summary>
+        /// Defines how large the borders will be for the icon that is placed within the QRCode.
+        /// </summary>
+        [Parameter] public int IconBorderWidth { get; set; } = 6;
+
+        /// <summary>
+        /// Defines the icon image file format.
+        /// </summary>
+        [Parameter] public IconImageType IconImageType { get; set; } = IconImageType.Png;
```
David-Moreira, line 181: Would we want this to be string? Wouldn't SVG, IMG, etc... also work? What if user wants to use an img we haven't added to the enum.
While at least keeping the helper. I think we've done this in some places. It has the "helper" but in the end it can just take in a string.

Source/Extensions/Blazorise.QRCode/QRCode.cs
stsrki: Good question. I will fix it.

Source/Extensions/Blazorise.QRCode/QRCode.cs
stsrki: Probably we could make it a string. But then we need to properly detect the actual type that could be anything, and define the base64 part, eg `data:image/png;base64`. It's easier if we have only a few predefined types.

Source/Extensions/Blazorise.QRCode/Enums/EciMode.cs
stsrki: This is mainly used to describe the Payload parameter and define encoding. I have added comments.

stsrki:
> Can we validate and do nothing if it is not a valid Base64? Otherwise boom: image

Done

stsrki:
We have bigger problems now

image

David-Moreira:
> We have bigger problems now
> 
> image

haha. Let me know if you need any help.

And let me know when it's ready to re-test

Source/Extensions/Blazorise.QRCode/QRCode.cs
```diff
@@ -103,9 +106,16 @@ private static System.Drawing.Color GetColor( string color )
 
         private static Bitmap GetImage( string base64 )
         {
-            using var ms = new MemoryStream( Convert.FromBase64String( base64 ) );
+            try
+            {
+                using var ms = new MemoryStream( Convert.FromBase64String( base64 ) );
```
David-Moreira, line 111: Is there not an utility to check if string is a base64? 
Better then using the throwing mechanism as the check. But it can still stay for something actually unexpected.

stsrki: > > We have bigger problems now
> > image
> 
> haha. Let me know if you need any help.
> 
> And let me know when it's ready to re-test

I'm researching JS libraries as a fallback for WASM. Or even for all platforms. It's really dumb from MS to remove `System.Drawing` support. Not like they don't have enough resources to have a team to do it from scratch to work on any platform 🤦‍♂️

stsrki: > Is there not an utility to check if string is a base64? Better then using the throwing mechanism as the check. But it can still stay for something actually unexpected.

I'm not aware of any.

stsrki:
@David-Moreira I have removed the QRCoder completely. Now everything is done in JS with `qr-code-styling` which has much more features and works really well. The API stays the same.

David-Moreira:
> @David-Moreira I have removed the QRCoder completely. Now everything is done in JS with `qr-code-styling` which has much more features and works really well. The API stays the same.

Don't forget to mention it on the docs then.

David-Moreira:
Tested & works pretty good.
Previously we could provide the type, canvas, svg, html, etc, do we want to expose that?

If an invalid image is provided then it does not render at all. Here it could probably still render? But I don't think it's wrong as it is either. As long as it doesn't "blow up" haha

stsrki:
> Previously we could provide the type, canvas, svg, html, etc, do we want to expose that?

We only provided image type, but it is needed anymore as all is handled by the JS library. And it was supposed to be a new API so it doesn't hurt to drop the API.

PR #4270, opened 2022-11-04
Introduce Specifications / Autocomplete Specs

### LINKED ISSUE #4267: Create new Reference documentation
We need new documentation pages where we will have written all the details about how certain components are working with different settings. Inspired by a comment bellow.

---

Hello,

I can confirm that the behaviour is now different.
What I have described is the one that's in **v1.0** and we'll be readjusting.

Basically `SelectedValueChanged` should ever only trigger 
- upon finding a value in the `Data` collection
- or going back to the `default` (null) value if it is not found

This is currently not happening like you've found out. Thanks for alterting us.

I also took the time to test `SelectedTextChanged`,
in **v1.0**, 
- it would always trigger upon text input with the currently selected text even if `FreeTyping` was not on.

in **v.1.1**, 
- If `FreeTyping` is on, it will always trigger with the current text input. 
- If `FreeTyping` is off it will always trigger with the `default` (null) upon text input, and with an actual `SelectedText `only upon finding a value in the `Data` collection.

Here on **v1.1** this is the correct behaviour, however I don't see the need for `FreeTyping="false"` to keep triggering `SelectedTextChanged` with a `default` (null) value every single time. The input has changed but the `SelectedText` hasn't changed at all.

So in summary, what we'll be readjusting in this issue's scope:
- `SelectedValueChanged` only triggers when appropriate (find an actual value, or going back to `default`)
- `SelectedTextChanged` + `FreeTyping=="false"` only triggers upon an actual `SelectedText`(exists in `Data`) or going back to the `default` (null), just like the `SelectedValue`.

_Originally posted by @David-Moreira in https://github.com/Megabit/Blazorise/issues/4245#issuecomment-1298698392_

David-Moreira:
Closes #4267

So we start documenting very specific component behaviours, decisions, etc...


stsrki:
Do you think we also need to tell what happens with SelectedValueChanged? When it is initialized, when the user types, selects and clears the input.

David-Moreira:
> Do you think we also need to tell what happens with SelectedValueChanged? When it is initialized, when the user types, selects and clears the input.

hmm how so?

stsrki:
> > Do you think we also need to tell what happens with SelectedValueChanged? When it is initialized, when the user types, selects and clears the input.
> 
> hmm how so?

You know how people regularly report problems that SelectedValueChanged is called multiple times, sometimes on the initialize, sometimes on clear, etc. If we document it, then we have a way of telling them why it behaves that way.

David-Moreira:
> > > Do you think we also need to tell what happens with SelectedValueChanged? When it is initialized, when the user types, selects and clears the input.
> > 
> > 
> > hmm how so?
> 
> You know how people regularly report problems that SelectedValueChanged is called multiple times, sometimes on the initialize, sometimes on clear, etc. If we document it, then we have a way of telling them why it behaves that way.

That's what the table already does.

stsrki:
> > > > Do you think we also need to tell what happens with SelectedValueChanged? When it is initialized, when the user types, selects and clears the input.
> > > 
> > > 
> > > hmm how so?
> > 
> > 
> > You know how people regularly report problems that SelectedValueChanged is called multiple times, sometimes on the initialize, sometimes on clear, etc. If we document it, then we have a way of telling them why it behaves that way.
> 
> That's what the table already does.

Then ignore me please :D

David-Moreira:
> > > > > Do you think we also need to tell what happens with SelectedValueChanged? When it is initialized, when the user types, selects and clears the input.
> > > > 
> > > > 
> > > > hmm how so?
> > > 
> > > 
> > > You know how people regularly report problems that SelectedValueChanged is called multiple times, sometimes on the initialize, sometimes on clear, etc. If we document it, then we have a way of telling them why it behaves that way.
> > 
> > 
> > That's what the table already does.
> 
> Then ignore me please :D

Well but if you ask, you might feel that the info is not enough? :)

stsrki (approved):

PR #4272, opened 2022-11-05
Fix theme caching issue

### LINKED ISSUE #4254: Memory leak in ThemeCache
**Describe the bug**
The ThemeCache class is leaking memory because the ammount of cached themes is ever increasing.

**To Reproduce**
1. Wrap the app in a theme.
```
<Blazorise.ThemeProvider Theme="@theme">
  ...
</Blazorise.ThemeProvider>

@code{
    private Theme theme = new Theme();
}
```
2. Start the application
3. Refresh the page continuously and monitor memory usage.

**Expected behavior**
I expect memory usage to remain stable.

**Screenshots**
Before cleanup:
image

After cleanup:
image

All these themes are exactly the same.

**Additional context**
This is caused due to a flaw in the equality comparison. Whenever it is attempted to retrieve styles the provided `Theme` has 1 event handler attached to the `Changed` event. The `Theme` objects in the cache however have 2 handlers attached. Therefor no cached `Theme` can be found and a new one is inserted. This is not limited to the `Changed` event. Modifying any property can cause this issue.

```
ThemeCache cache = new ThemeCache(new(null, options => { }));
for (int i = 0; i < 12; i++)
{
    Theme theme = new Theme();
    if (!cache.TryGetStylesFromCache(theme, out _))
    {
        cache.CacheStyles(theme, "abc");
        theme.Changed += (_, _) => { }; // Comment this line and it works as expected.
    }
}
```

Another issue is that the `UncacheTheme()` method doesn't remove anything from the `variableCache` or `styleCache`.  This seems to happen when the `Theme` that was added to a dictionary is modified afterwards. This happens when a second event listener is added (not sure where). Again, this is not limited to the `Changed` event but can happen for any propery.

```
Dictionary<Theme, string> styleCache = new();
var theme = new Theme();
styleCache.Add(theme, "");
theme.Black = "not black"; // Comment this line and it works as expected.
styleCache.Remove(styleCache.FirstOrDefault().Key);
```

**Workaround**
Can be avoided by disabling the cache in the options as such:
```
services.AddBlazorise(options =>
{
     options.ThemeCacheSize = 0;
});
```

David-Moreira:
Closes #4254

- Introduced Blazorise.SourceGenerator 
- Used a Blazorise.SourceGenerator.Features project to allocate the common Attributes for the generation detection
- Theme is now partial and has Equality / HashCode methods generated. 
  - Like we've talked, the problem here was that the event field was messing up the `GetHashCode` from the `Dictionary`, so on this version at least, it was never caching anything, just wasting memory if multiple themes were used (and not cleaning up either)
  - The tests that we have are good, but unfortunately missed the event field, giving a false sense of being ok.
 - I did not see a reason for the triple variables for caching, and I have reduced the implementation to 1.

I still need to do a bit more testing later today, but all in all seems like it is now working properly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -266,3 +266,7 @@ Documentation/Blazorise.Docs.Server/wwwroot/img/blog/.DS_Store
 # Ignore VS Code
 .vscode/*
 .vscode/settings.json
+
+
+# Source Generation directory
+__SOURCEGENERATED__/
--- a/Blazorise.sln
+++ b/Blazorise.sln
@@ -108,6 +108,15 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Blazorise.LoadingIndicator"
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Blazorise.FluentValidation", "Source\Extensions\Blazorise.FluentValidation\Blazorise.FluentValidation.csproj", "{55C2652B-18A3-4B28-935A-67D78BFE6E8C}"
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "SourceGenerators", "SourceGenerators", "{0538DB67-B4F3-4D00-B969-D3874A52E405}"
+	ProjectSection(SolutionItems) = preProject
+		Source\SourceGenerators\README.md = Source\SourceGenerators\README.md
+	EndProjectSection
+EndProject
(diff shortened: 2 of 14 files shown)


stsrki (approved):
The new source generator works really well, love it. I will definitely put it to use in other places. Good job!

David-Moreira:
> The new source generator works really well, love it. I will definitely put it to use in other places. Good job!

Yep, it's pretty good. Just the exposed api isn't really great, you have to do some debugging to find the stuff you want. But I'm sure they will either improve it further or someone will come up with some abstraction on top that'll make it even easier.

David-Moreira:
PS : that using you had to revert it could. Just be a part of the source generator, since we call an extension as part of the generated code.
But you already merged, let's leave it now.

PR #4300, opened 2022-11-16
Added support for directory upload in FilePicker

mcollera:
This PR enables File Picker to support selecting a directory and keeping track of there relative path


mcollera:

mcollera:
recheck

mcollera:
recheck

mcollera:

mcollera:
recheck

mcollera:
recheck

David-Moreira:
@mcollera Thank you for the contribution.
We'll take a look at this as soon as possible.

In the meantime, 
- Can you provide a brief explanation of what this does?
- Did you only make the changes to `FilePicker`? Would this not be worth adding as a core feature of `FileEdit`? `FilePicker` actually uses `FileEdit` for the core features.

mcollera:
> @mcollera Thank you for your contribution. We'll take a look at this as soon as possible.
> 
> In the meantime,
> 
> * Can you provide a brief explanation of what this does?
> * Did you only make the changes to `FilePicker`? Would this not be worth adding as a core feature of `FileEdit`? `FilePicker` actually uses `FileEdit` for the core features.

I seen that it was. I guess I didnt understand the difference. I assumed FileEdit allowed you to edit a file. It seems like they do the same thing.

Source/Blazorise/Components/FilePicker/FilePicker.razor.cs
```diff
@@ -434,6 +440,38 @@ protected Task OnProgressed( FileProgressedEventArgs fileProgressedEventArgs )
...
+        {
+            get
+            {
+                return directory;
+            }
+            set
+            {
+                if ( value )
+                {
+                    if ( !Attributes?.ContainsKey( "webkitdirectory" ) ?? false )
```
stsrki, line 457: Any reason why we need to manually add `webkitdirectory` to the Attributes? Why not just add it directly to the file input element in HTML/razor?

Source/Blazorise/Components/FilePicker/FilePicker.razor.cs
```diff
@@ -434,6 +440,38 @@ protected Task OnProgressed( FileProgressedEventArgs fileProgressedEventArgs )
...
+        [Parameter]
+        public bool Directory
+        {
+            get
+            {
+                return directory;
+            }
+            set
+            {
+                if ( value )
```
stsrki, line 455: We should exit early if nothing changes.

eg.

```
if ( directory = value )
  return;
```

Source/Blazorise/Components/FilePicker/FilePicker.razor.cs
```diff
@@ -45,10 +50,11 @@ protected override void OnInitialized()
             ElementContainerId = IdGenerator.Generate;
             LocalizerService.LocalizationChanged += OnLocalizationChanged;
 
+            Attributes ??= new();
```
stsrki, line 53: I would like to not user `Attributes` like this. Try to assign needed HTML attributes directly to the input element where needed.

Source/Blazorise/Components/FilePicker/FilePicker.razor.cs
mcollera: I was trying not to change anything on FileEdit. I wasnt sure of what the use case is for FileEdit compared to FilePicker. I will make the changes to support Directory on FileEdit directly.

mcollera: This was removed

Source/Blazorise/Components/FilePicker/FilePicker.razor.cs
mcollera: OBE

Source/Blazorise/Components/FilePicker/FilePicker.razor.cs
mcollera: Removed

David-Moreira:
@mcollera 
why does drag & drop no longer shows?
image
vs
image

Is it not supported for this feature is that it? I can still drop files or a folder in the FileEdit, do we want this?

David-Moreira:
Any chance both can work at the same time files + folder upload?

David-Moreira:
Do you have time to had an example on the docs for this? Together with explanation of the behaviour and anything the user should know about the feature?

mcollera:
> @mcollera why does drag & drop no longer shows? image vs image
> 
> Is it not supported for this feature is that it? I can still drop files or a folder in the FileEdit, do we want this?

The JavaScript didn’t work for getting the drag and drop information for folders. I tried but couldn’t get it to work. Thought it was safer to remove.

mcollera:
> Any chance both can work at the same time files + folder upload?

No it is either our. The browser doesn’t support it.

David-Moreira:
> > @mcollera why does drag & drop no longer shows? image vs image
> > Is it not supported for this feature is that it? I can still drop files or a folder in the FileEdit, do we want this?
> 
> The JavaScript didn’t work for getting the drag and drop information for folders. I tried but couldn’t get it to work. Thought it was safer to remove.

Thing is

> > @mcollera why does drag & drop no longer shows? image vs image
> > Is it not supported for this feature is that it? I can still drop files or a folder in the FileEdit, do we want this?
> 
> The JavaScript didn’t work for getting the drag and drop information for folders. I tried but couldn’t get it to work. Thought it was safer to remove.

Well you did remove, but I can still drag & drop to the File Input itself.

Can we have the same behaviour as in here?
https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/webkitdirectory
I was just testing, and I can drag a folder into it. And dragging files is ignored.

mcollera:
@David-Moreira Added support for drag and drop. Just need to update documentation

David-Moreira:
very nice! I'll let you know once I take a look.

mcollera:
@David-Moreira documentation added

David-Moreira:
Since we've changed Choose files to Choose folders. I'd argue the same should be done for the Drag & Drop?

image

David-Moreira:
Should this be `Choose Folder`?
image

David-Moreira:
Any chance the behaviour can be the same across `FileEdit` and `FilePicker`?
When Directory is enabled. `FileEdit`, does not let me drop any regular files, but `FilePicker` does.

mcollera:
> ged Choose files to Choose folders. I'd argue the same should be done for the Drag

Bootstrap 5 does not expose the ability to change the text. Old version of bootstrap did.

mcollera:
@David-Moreira completed other changes requested.

David-Moreira:
Hello,

@mcollera 
1. Why do we now have the same text in both modes? Does this make sense? Also is it `folder` or `folders`?
image

2. When on `Directory` mode, if I use the Choose options, it shows the path
image
If I drop it does not show, is this also expected?
image

David-Moreira:
3. [Updated OnDrop for file edit](https://github.com/Megabit/Blazorise/pull/4300/commits/c98931cd02fcdbd535ae82cb5cfff9a4eb4e9053)
Is this new `JS` on `fileEdit.js` change a copy of an already existing `JS` in `filePicker.js`? If the functionality is to be the same on both, Is it possible to make it centralized if it makes sense? After all `FilePicker` composes off `FileEdit`.

4. I was able to drop a folder in `FileEdit` single selection which resulted in multiple files. This seems like a change in behaviour.
image

mcollera:
@David-Moreira 

1. I can make it where you can drop folders when not in directory mode. Is that what you prefer?
2. That is a limitation to drag and drop. 
3. I will centralize
4. Do you prefer I change? I am pretty sure I could in current version as well.

David-Moreira:
> @David-Moreira
> 
> 1. I can make it where you can drop folders when not in directory mode. Is that what you prefer?
> 2. That is a limitation to drag and drop.
> 3. I will centralize
> 4. Do you prefer I change? I am pretty sure I could in current version as well.

@stsrki can you help with these questions,

**1,** you mean, make it where you **can not** drop folders when not in directory mode? Right? 
Going to the current version
https://bootstrapdemo.blazorise.com/tests/file-picker 
it just regularly shows the folder indication. I believe we should maintain this if `Directory` is not set. Altough I'm not sure about the usefulness of this at all, I remember when implementing that for a regular `input type="file"` it seemed like you couldn't tell a folder from a file, was I wrong? Could we specifically ignore the drop of folders if `Directory` is not set? That seems like it would make most sense.
image


For **4**, go to our current version here and drop a folder. It will just show the folder indication.
https://bootstrapdemo.blazorise.com/tests/forms
In this version it would have multiple files I believe we should also not change this, right? At least not when not in `Directory` Mode. 
If we are in `Directory` mode, then we should make sure
If `multiple` is `true`
- Can drop multiple folders, which lists all files within.

If `multiple` is `false`
- Can only drop a single folder, but it'll accept all files within, I believe this is the same as the spec here : https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/webkitdirectory

stsrki:
I agree with @David-Moreira :)

mcollera:
@David-Moreira Ok I hope I caught everything. I have pushed my changes.

David-Moreira:
LGTM!
@stsrki can you go for a final look and merge it in?
Also, how do you feel about the new `fileUtilities,js`? Worth it, over just keeping these on `fileEdit.js`? or even `io.js`?

stsrki:
> Also, how do you feel about the new `fileUtilities,js`? Worth it, over just keeping these on `fileEdit.js`? or even `io.js`?

I would prefer it moved to the `fileEdit.js`. It only has one function inside.

mcollera:
It will also need to be in filepicker.js as well so two places.

mcollera:
Done

mcollera:
@David-Moreira Are any more changes needed to get this merged?

David-Moreira (approved):

David-Moreira:
Last time I checked it looked good to me. If stsrki can go ahead and merge :)

stsrki:
I'm currently busy but I will review it soon enough :)

stsrki:
OMG, Visual Studio merged the PR automatically when I pushed commit 🤦‍♂️🤦‍♂️🤦‍♂️🤦‍♂️🤦‍♂️

David-Moreira:
I guess it's reviewed & approved then!! :D

stsrki:
Haha, yeah. I think VS went nuts with both branches named `master`. Note for me. Don'' allow PRs that come from base master branches.

PS. @mcollera Thank you for the new feature!

PR #4308, opened 2022-11-17
WIP: ComponentParameterInfo

stsrki:
I have noticed on the Button component( and in other places, too ) that we have an invalid condition for the `Disabled` parameter.

`protected bool IsDisabled => ParentDropdown?.Disabled ?? Disabled;`

the order should be reversed

`protected bool IsDisabled => Disabled ?? ParentDropdown?.Disabled;`

But the problem is that `Button.Disabled` parameter is `bool` type, and if we change it to `bool?` then it would break any component that tries to set it with a simplified syntax, eg.

`<Button Disabled>`

---

And so I'm now experimenting with a new concept of receiving the parameters. That is the new `ComponentParameterInfo` type. It tells us if the parameter is being received through the Blazor pipeline. No need to have a `bool? Disabled` parameter.

We can also have some other optimizations, like only calling `DirtyClasses()` if any of the wanted parameters is really received.

At the moment, the code is relatively crude. I need to see if it makes sense at all.


David-Moreira:
I mean, it's similar to this && glutio's tripplet implementation, just that it's wrapped in a record and I guess it's easier to read & understand.
image

But seems like a good approach, I see the benefit in centralized the logic of how we detect if parameter was changed etc.

Would something like this be worth it at all?:
```
public record ButtonComponentParametersInfo
{
    public ComponentParameterInfo<bool> disabled = new ComponentParameterInfo<bool>( false );
    public ComponentParameterInfo<int> someNumber = new ComponentParameterInfo<int>( );

    event Changed; ??

        void Update( ParameterView parameters )
        {
            disabled = ResolveParameterStatus<bool>(parameters, nameof(Disabled));
            someNumber = ResolveParameterStatus<int>(parameters, nameof(SomeNumber));

            if ( disabled.Received )
            {
                Changed();
            }
        }
}
```

stsrki:
Yes, that is exactly the idea I had in mind. I did this quick prototype in half an hour to not forget it. But yeah, I think it is much cleaner to read and write than the three-value system _glutio_ had. I agree with changing it to the struct type. But as always, it might be good to benchmark it and see if it changes anything. Do you want to do that part?

David-Moreira:
> Yes, that is exactly the idea I had in mind. I did this quick prototype in half an hour to not forget it. But yeah, I think it is much cleaner to read and write than the three-value system _glutio_ had. I agree with changing it to the struct type. But as always, it might be good to benchmark it and see if it changes anything. Do you want to do that part?

Yea I can benchmark it.

David-Moreira:
@stsrki are you going to change anything, try record / struct implementation? Or want me to do some benchmarking first?

stsrki:
> @stsrki are you going to change anything, try record / struct implementation? Or want me to do some benchmarking first?

I will. Just that I had other stuff in the way. But you can try benchmarking if you want.

Source/Blazorise/Extensions/ParameterViewExtensions.cs
```diff
@@ -0,0 +1,34 @@
...
+    /// <typeparam name="T">The type of the value.</typeparam>
+    /// <param name="parameters">Dictionary of all component paremeters.</param>
+    /// <param name="parameterName">The name of the parameter.</param>
+    /// <param name="currentValue">Last known parameter value.</param>
+    /// <param name="result">Receives the result, if any.</param>
+    public static void TryGetParameter<T>( this ParameterView parameters, string parameterName, T currentValue, out ComponentParameterInfo<T> result )
+    {
+        if ( parameters.TryGetValue<T>( parameterName, out var paramNewValue ) )
+        {
+            var changed = currentValue is IEnumerable<T> array && paramNewValue is IEnumerable<T> array2
```
stsrki, line 25: @David-Moreira see if this is a good way of checking for the array types and if it's fast enough.

David-Moreira: Yea I believe pattern matching is quite optimized that seems good. Also we do have this for a more generic implementation:
image
image

But I believe you only want the `IEnumerable` there right?

stsrki: That seems like a reflection so it would be considerably slower. We just want to know that the value is `IEnumerable`.

stsrki:
Closing since a lot of work from here is done on #5732

PR #4313, opened 2022-11-22
Link: Add CustomMatch parameter

### LINKED ISSUE #4294: Allow control over Blazorise.Link.ShouldMatch(...)
I am trying to make my Blazorise.Link show as Active when in a sub-page.
Example: Link to https://localhost:44362/Users/Overview will show link active when current url = https://localhost:44362/Users/Overview.
When I click on a user, I go to page https://localhost:44362/Users/View/35faa5fb-af76-4da1-961b-2607528dcff1. Now my link is not active, which makes it confusing where you are in the application.

A clear and concise description of what the problem is. Example: I am trying to do [...] but [...]

I tried to get any control over the Blazorise.Link.ShouldMatch method, without success.
My fix would be to just check if the page starts with /Users/*.

I have two possible solutions.
One would be to add a IsMatch parameter callback to Blazorise.Link. This will have one argument, which would have three properties "IsMatch" which is a boolean. This property can be changed by the handler (public bool IsMatch {get;set;}).
"Href" which is a readonly string which is the url the Link points to.
"CurrentUrl" which is the url of the current page visiting (optional, you could also choose for the user needing to get this information from the NavigationManager themselves.).

Using this information, you can control with the IsMatch bool whether or not the link is a match by setting the value yourself. If the parameter is NOT supplied, then the current flow should be used.

Another solution I came up with is to make Blazorise.Link.ShouldMatch a `public virtual bool` instead of a `private bool`. This way you can just extend Blazorise.Link. This is probably an easier/cleaner solution to implement, but is probably not very straight forward for anyone trying to implement this behavior.

### Additional context
Option 1: 
image

Option 2:
image

I hope it made sense, if you have any questions you can either add a comment or pm me on discord.

stsrki:
Closes #4294

### DIFF
--- a/Documentation/Blazorise.Docs/Pages/Docs/Components/Links/LinkPage.razor
+++ b/Documentation/Blazorise.Docs/Pages/Docs/Components/Links/LinkPage.razor
@@ -44,7 +44,7 @@
         A <Code>#</Code> in front of a link location specifies that the link is pointing to an anchor on a page.
         (Anchor meaning a specific place in the middle of your page).
 
-        Typically <Code tag>a href="#"</Code> will cause the document to scroll to the top of page when clicked.
+        Typically <Code Tag>a href="#"</Code> will cause the document to scroll to the top of page when clicked.
         <Code Tag>Link</Code> addresses this by preventing the default action (scroll to top) when <Code>To</Code> is set to <Code>#</Code>.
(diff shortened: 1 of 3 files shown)


joost00719:
LGTM 👍
Thank you.

PR #4315, opened 2022-11-22
RichTextEdit: wait for the first render to execute methods

### LINKED ISSUE #4114: "Empty" RichTextEdit
**Describe the bug**
After setting HTML of RichTextEdit, the edit area still remains empty - most of the times.
Blazorise Version is 1.0.4 and 1.0.6 (still in 1.1.3.1)

**To Reproduce**
I tracked it down to this simple example
```razor
@using Blazorise.RichTextEdit

<RichTextEdit @ref="editor"
    Theme="RichTextEditTheme.Snow">
</RichTextEdit>

@code {
    private RichTextEdit? editor;

    private string html = "Hello world.";
    private bool isSet;

    protected override async Task OnAfterRenderAsync(bool firstRender)
    {
        if (!isSet && editor != null)
        {
            isSet = true;

            await editor.SetHtmlAsync(html);
            StateHasChanged();
        }
        await base.OnAfterRenderAsync(firstRender);
    }
}
```

The only thing this code does is to call `SetHtmlAsync` on startup.
After opening the page, "Hello world" should be visible.

**Expected behavior**
The text is changed after `SetHtmlAsync`

**Additional context**
In my tests it worked far way more often (if not always), if I
a) remove `StateHasChanged();` in `OnAfterRenderAsync`, or
b) remove `Theme="RichTextEditTheme.Snow"`

both of which are inexplainable to me (especially `b)` since `Snow` is the default value)

stsrki:
Closes #4114

### DIFF
--- a/Source/Extensions/Blazorise.RichTextEdit/BaseRichTextEditComponent.cs
+++ b/Source/Extensions/Blazorise.RichTextEdit/BaseRichTextEditComponent.cs
@@ -1,5 +1,6 @@
 ﻿#region Using directives
 using System;
+using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
 #endregion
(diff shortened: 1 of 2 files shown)


stsrki:
@WolfgangKluge, from my testing it seems to be working just fine but as always it's good to confirm it. If you can it would be best to test it. Any help is appreciated.

WolfgangKluge:
@stsrki works fine. Thanks

stsrki:
Thanks, merging it now.

PR #4316, opened 2022-11-22
Markdown: wait for the first render to execute methods

### LINKED ISSUE #4084: Markdown Component Javascript Error
**Describe the bug**
I have a few Markdown components that are loaded inside of a modal.  There is code on a main page to listen for when the modal has been initialized.  After it's been initialized, the button to launch the modal becomes available to click. 

There is a bug (pictured below) that occurs with the JS Interop that loads the Markdown component and forces a refresh of the page.  It seems to occur most often when I navigate away from the page before the modal is ready, but it can occur even after that happens sometimes.

Unfortunately, I cannot share a lot of the source code and had to modify this image a bit to hide some information, but I will share what I can.  Picture of the error in the Chrome console:
image

**To Reproduce**
1. Create a modal with the following properties: 
- If data is not loaded, do not display any html.
- If data is loaded, display a Markdown component and bind it to the data.
- Once the data has loaded, use a notifier service to notify the main page that the modal is ready.
2. On a main page, include the modal within the HTML and create a button that will launch the modal on click.
3. Load the main page, but before the modal is ready, navigate away from the main page.
4. See an error

**Expected behavior**
The page navigates away without error.

**Screenshots**
As I said before, I cannot share a lot of the source, but here is what I can share that's related to the issue:

Here is the top of the Modal file.  This prevents any html from displaying if the data has not been loaded yet:
image

Here is how the Markdown component is used within the modal:
*Quick note:* *The Blazorise.Text component above the Markdown component is used to put an object in the DOM with a specific class based on whether or not there is data in the object tied to the Markdown component.  This is then used to style the Markdown component (highlight it if it's empty).  Could a future version of the Markdown component include a Class parameter that can be used on the actual textedit part of the Markdown component?  If not, this "hack" does the job.*
image

Here's the OnAfterRenderAsync override in the modal's code-behind that notifies the application when the data for the modal has been loaded.  I have a feeling this is partly to blame for the error, but I can't figure out how.
image

Here's how the modal is added to the main page:
image

Here's the button that launches the modal (when it's ready):
image

Modal init in the main page code-behind:
image

Modal-ready-setter in the main page code-behind:
image

**Additional context**
Let me know if you need more information, but I've shared about as much of the source as I can.  Hopefully I've described well-enough how to replicate the issue.

stsrki:
Closes #4084

Added the same lifecycle logic as in #4315. We're postponing `ExecuteAfterRender` after the component is fully rendered and initialized and then forwarding it all to execution.

PS. we can consider implementing the same logic in a centralized way so that it can be reused by other components. We need to do this in `master.`

### DIFF
--- a/Demos/Blazorise.Demo/Pages/Tests/MarkdownPage.razor
+++ b/Demos/Blazorise.Demo/Pages/Tests/MarkdownPage.razor
@@ -6,9 +6,20 @@
                 <CardTitle>Markdown editor</CardTitle>
             </CardHeader>
             <CardBody>
+                <Fields>
+                    <Field ColumnSize="ColumnSize.IsAuto">
+                        <Button Color="Color.Primary" Clicked="@(()=>markdownRef.SetValueAsync("this is a new value from SetValueAsync Method"))">Set SetValueAsync Method</Button>
+                    </Field>
+                    <Field ColumnSize="ColumnSize.IsAuto">
+                        <Button Color="Color.Primary" Clicked="@(()=>markdownValue = "this is a new value from Value Parameter")">Set Value Parameter</Button>
+                    </Field>
+                </Fields>
+            </CardBody>
+            <CardBody>
                 <Row>
                     <Column>
-                        <Markdown Value="@markdownValue" ValueChanged="@OnMarkdownValueChanged"
+                        <Markdown @ref="@markdownRef"
+                                  Value="@markdownValue" ValueChanged="@OnMarkdownValueChanged"
                                   ImageUploadChanged="@OnImageUploadChanged"
                                   ImageUploadStarted="@OnImageUploadStarted"
                                   ImageUploadProgressed="@OnImageUploadProgressed"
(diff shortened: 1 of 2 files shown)


stsrki:
@WolfgangKluge, from my testing it seems to be working just fine but as always it's good to confirm it. If you can, it would be best to test it. Any help is appreciated.

WolfgangKluge:
@stsrki seems good for me - but for this I don't have a reliable test (comes up occasionally).
Maybe @ninjanerdbgm?

ninjanerdbgm:
@stsrki @WolfgangKluge Absolutely. I can replicate #4084 pretty reliably, so I'd be happy to provide additional testing.

ninjanerdbgm:
@strski my tests indicate it's resolved!  Thanks for all your help

stsrki:
@WolfgangKluge and @ninjanerdbgm, thanks guys. I appreciate your help!

PR #4325, opened 2022-11-23
Support for row overlay

stather:
Adds 2 new event callback to datagrid, and a feature to display an overlay when the mouse is over the row.

New Parameter RowOver of type EventCallback<DataGridRowMouseEventArgs<TItem>>
This is called when the mouse is over the row.

New Parameter RowLeave of type EventCallback<DataGridRowMouseEventArgs<TItem>>
This is called when the mouse leaves a row.

These are useful anyway, but used mainly to facilitate the row overlay feature.

New Parameter OverlayTemplate of type RenderFragment<TItem>.

This render fragment can be used to display additional information about the row item when the mouse is over the row. The content of the fragment is displayed over the top of the current row and is right aligned.

```
<DataGrid TItem="sometype">
                       <OverlayTemplate>
                            <Div Flex="Flex.JustifyContent.End.AlignItems.Center" Margin="Margin.Is3.FromBottom" Style="width:400px;background-color:white;">
                                <Div Padding="Padding.Is2">
                                    <Button Size="Size.Small" Color="Color.Secondary" onclick="event.stopPropagation();">Ack</Button>
                                </Div>
                                <Div Padding="Padding.Is2">
                                    <Button Size="Size.Small" Color="Color.Secondary" onclick="event.stopPropagation();">Close</Button>
                                </Div>
                                <Div Padding="Padding.Is2" @onclick:stopPropagation="true" Overflow="Overflow.Visible">
                                    <Dropdown Style="z-index:2000;">
                                        <DropdownToggle Size="Size.Small" Clicked="ToggleClicked">
                                            ...
                                        </DropdownToggle>
                                        <DropdownMenu>
                                            <DropdownItem>Add responder</DropdownItem>
                                            <DropdownItem>Delete</DropdownItem>
                                        </DropdownMenu>
                                    </Dropdown>
                                </Div>
                            </Div>
                        </OverlayTemplate>
</DataGrid>
```

The DataGrid only displays this fragment when an OverlayTemplate is present and the mouse is over a row. 
The fragment is removed when the mouse leaves the row.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -243,6 +243,9 @@ ModelManifest.xml
 # FAKE - F# Make
 .fake/
 
+# CodeRush personal settings
+.cr/personal
+
 # jekyll
 _site
 Gemfile.lock
--- a/Source/Blazorise/Components/Table/TableRow.razor
+++ b/Source/Blazorise/Components/Table/TableRow.razor
@@ -2,6 +2,8 @@
 @inherits BaseDraggableComponent
 <tr id="@ElementId" class="@ClassNames" style="@StyleNames" @onclick="@OnClickHandler"
     draggable="@DraggableString"
+    @onmouseover="@OnMouseOverHandler"
+    @onmouseleave="@OnMouseLeaveHandler"
     @ondrag="@OnDragHandler"
     @ondrag:preventDefault="@DragPreventDefault"
     @ondragend="@OnDragEndHandler"
--- a/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
+++ b/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
@@ -5,6 +5,7 @@ using System.Linq;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
+using Blazorise.DataGrid.Enums;
 using Blazorise.DataGrid.Models;
 using Blazorise.DataGrid.Utils;
 using Blazorise.Extensions;
@@ -825,6 +826,15 @@ namespace Blazorise.DataGrid
             }
         }
 
+        internal Task OnRowOverCommand( DataGridRowMouseEventArgs<TItem> eventArgs )
+        {
+            return RowOver.InvokeAsync( eventArgs );
+        }
+        internal Task OnRowLeaveCommand( DataGridRowMouseEventArgs<TItem> eventArgs )
+        {
+            return RowLeave.InvokeAsync( eventArgs );
+        }
+
         internal Task OnRowClickedCommand( DataGridRowMouseEventArgs<TItem> eventArgs )
         {
             return RowClicked.InvokeAsync( eventArgs );
@@ -1313,6 +1323,13 @@ namespace Blazorise.DataGrid
         #endregion
 
         #region Properties
+
+        /// <summary>
+        /// Template for mouse hover overlay display formatting.
+        /// </summary>
+        [Parameter] public RenderFragment<TItem> RowOverlayTemplate { get; set; }
+
+        [Parameter] public RowOverlayPosition RowOverlayPosition { get; set; } = RowOverlayPosition.End;
 
 
         /// <summary>
@@ -1870,6 +1887,16 @@ namespace Blazorise.DataGrid
         /// </summary>
         [Parameter] public EventCallback<TItem> RowRemoved { get; set; }
 
+
+        /// <summary>
+        /// Event called after the mouse leaves the row.
+        /// </summary>
+        [Parameter] public EventCallback<DataGridRowMouseEventArgs<TItem>> RowLeave { get; set; }
+
+        /// <summary>
+        /// Event called after the mouse is over the row.
+        /// </summary>
+        [Parameter] public EventCallback<DataGridRowMouseEventArgs<TItem>> RowOver { get; set; }
         /// <summary>
         /// Event called after the row is clicked.
         /// </summary>
--- a/Source/Blazorise/Components/Table/TableRow.razor.cs
+++ b/Source/Blazorise/Components/Table/TableRow.razor.cs
@@ -52,6 +52,26 @@ namespace Blazorise
             return Task.CompletedTask;
         }
 
+        /// <summary>
+        /// Handles the row mouse leave event.
+        /// </summary>
+        /// <param name="eventArgs">Supplies information about a mouse event that is being raised.</param>
+        /// <returns>A task that represents the asynchronous operation.</returns>
+        protected Task OnMouseLeaveHandler(MouseEventArgs eventArgs )
+        {
+            return MouseLeave.InvokeAsync( EventArgsMapper.ToMouseEventArgs( eventArgs ) );
+        }
+
+        /// <summary>
+        /// Handles the row mouseover event.
+        /// </summary>
+        /// <param name="eventArgs">Supplies information about a mouse event that is being raised.</param>
+        /// <returns>A task that represents the asynchronous operation.</returns>
+        protected Task OnMouseOverHandler(MouseEventArgs eventArgs )
+        {
+            return MouseOver.InvokeAsync( EventArgsMapper.ToMouseEventArgs( eventArgs ) );
+        }
+
         #endregion
 
         #region Properties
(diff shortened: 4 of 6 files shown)


stather:

stather:
recheck

stather:
recheck

stather:
@stsrki Hi, created a new PR. It's verified I have signed the CLA. But still showing an error on the CLA assistant.

stather:
yay, finally CLA bot working

stsrki:
I'm testing the PR, and it looks pretty well. There is some issue though. In our demo it shows the Check from behind, see the image.

image

Also, in the `_DataGridRow.razor`, we are adding a `TableRowCell` to the row. This might be problematic, I think because by adding this additional cell, we might break other things. For example command column or multiple select columns. Those columns take special care of calculating the available number of columns that are required for the regular DataGrudColumns. In theory, the overlay should work since it is displayed with absolute position, but I still want to be sure.

@David-Moreira can you also double-check this?

David-Moreira:
> I'm testing the PR, and it looks pretty well. There is some issue though. In our demo it shows the Check from behind, see the image.
> 
> image
> 
> Also, in the `_DataGridRow.razor`, we are adding a `TableRowCell` to the row. This might be problematic, I think because by adding this additional cell, we might break other things. For example command column or multiple select columns. Those columns take special care of calculating the available number of columns that are required for the regular DataGrudColumns. In theory, the overlay should work since it is displayed with absolute position, but I still want to be sure.
> 
> @David-Moreira can you also double-check this?

Right I don't think it affects our core functionalities because:

1- It's placed as the last cell. 
2- It's not used in the Columns iteration, which is what could actually affect the display behaviour (thead and tbody cells out of sync)
image
3- Like you said, it's absolute and the purpose is to be used as overlay not as an actual new cell.

@stather if possible could  you also look at these?

- What if it was an actual `TableRow` and not a `TableRowCell`, what do you guys think? Does this make sense? Similar to this:
image

- Missing documentation / example
- Missing an example on the demo It could maybe be somewhere around here:
image

stsrki:
> * What if it was an actual `TableRow` and not a `TableRowCell`, what do you guys think? Does this make sense? Similar to this:

It would make more sense to use a table row, if possible.

stather:
Problem is you aren’t allowed to put a div in a tr only a td. Sent from my iPhoneOn 05 Dec 2022, at 12:04, Mladen Macanović ***@***.***> wrote:﻿


What if it was an actual TableRow and not a TableRowCell, what do you guys think? Does this make sense? Similar to this:


It would make more sense to use a table row, if possible.

stsrki:
> Problem is you aren’t allowed to put a div in a tr only a td. Sent from my iPhoneOn 05 Dec 2022, at 12:04, Mladen Macanović ***@***.***> wrote:﻿ What if it was an actual TableRow and not a TableRowCell, what do you guys think? Does this make sense? Similar to this: It would make more sense to use a table row, if possible. —Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you were mentioned.Message ID: ***@***.***>

Can't the row be with the spanned cells?

```
<tr>
 <td colspan="totalColumns"></td>
</tr>
```

David-Moreira:
@stsrki Do you remember how you tested this?
image
I'm wondering how you got that white background, did you use custom css?

The default look doesn't seem very good, or am I missing something?
image

stsrki:
> @stsrki Do you remember how you tested this? image I'm wondering how you got that white background, did you use custom css?
> 
> The default look doesn't seem very good, or am I missing something? image

Just by hovering over the row. Make sure that you clear the cache.

David-Moreira:
> > @stsrki Do you remember how you tested this? image I'm wondering how you got that white background, did you use custom css?
> > The default look doesn't seem very good, or am I missing something? image
> 
> Just by hovering over the row. Make sure that you clear the cache.

Well I cleared the cache... I also don't see anything that sets the background to white, O.O hmmm...

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -2182,6 +2191,16 @@ public IEnumerable<GroupContext<TItem>> DisplayGroupedData
     /// </summary>
     [Parameter] public EventCallback<TItem> RowRemoved { get; set; }
 
+
+    /// <summary>
+    /// Event called after the mouse leaves the row.
+    /// </summary>
+    [Parameter] public EventCallback<DataGridRowMouseEventArgs<TItem>> RowLeave { get; set; }
```
stsrki, line 2198: Name it RowMouseLeave.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -2182,6 +2191,16 @@ public IEnumerable<GroupContext<TItem>> DisplayGroupedData
...
+
+    /// <summary>
+    /// Event called after the mouse leaves the row.
+    /// </summary>
+    [Parameter] public EventCallback<DataGridRowMouseEventArgs<TItem>> RowLeave { get; set; }
+
+    /// <summary>
+    /// Event called after the mouse is over the row.
+    /// </summary>
+    [Parameter] public EventCallback<DataGridRowMouseEventArgs<TItem>> RowOver { get; set; }
```
stsrki, line 2203: Name it RowMouseOver.

Source/Blazorise/Components/Table/TableRow.razor
```diff
@@ -2,6 +2,8 @@
 @inherits BaseDraggableComponent
 <tr @ref="@ElementRef" id="@ElementId" class="@ClassNames" style="@StyleNames" @onclick="@OnClickHandler"
     draggable="@DraggableString"
+    @onmouseover="@(MouseOver.HasDelegate ? OnMouseOverHandler : null)"
```
stsrki, line 5: Do we want to add `@onmouseover:stopPropagation` here?

.gitignore
```diff
@@ -243,6 +243,9 @@ ModelManifest.xml
 # FAKE - F# Make
 .fake/
 
+# CodeRush personal settings
+.cr/personal
```
stsrki, line 247: CodeRush ?

Source/Blazorise/Components/Table/TableRow.razor
David-Moreira: Do we? I don't know. What about be the benefit?

stsrki: It can prevent mouse events from bubbling up to the table and possibly slowing down the grid.

David-Moreira: It's literally a row, I'm not sure how much further it will even bubble up, and it won't be slowing down since there won't be interops which are the actual thing that can slow down blazor.

David-Moreira:
@stsrki Please review.

stsrki:
When I click on a button in the overlay, the row behind is selected. I think it should not propagate a click event behind it.

image

David-Moreira:
> When I click on a button in the overlay, the row behind is selected. I think it should not propagate a click event behind it.
> 
> image

While fixing this... I actually noticed that the suggested table row with full span isn't really that sound, this is because it's actually inside the affected table row. 
When I thought about this before, I was thinking about it sitting besides the affected table row instead, which would make sense, but would require javascript to know what row to overlay over and where to position.

So keeping the same implementation that @stather implemented makes more sense I believe since you easily get it to sit correctly on top of the row with just css.

So I decided to change to a div, I think it's easier to manage the dimensions of it in this context. Row & Cell seemed a bit inconsistent, let me know if you agree with this.

I was also thinking that maybe we should also provide a default css class to make it easier for users to style it.
Would you agree? Naming? b-table-row-overlay? or you want some other name closer to the current provider?

And I was thinking, maybe this should have been straight up implemented on the `Table` itself like we did with other features, resizable, etc..... I think it's worth moving it over. Agreed?

stsrki:
> While fixing this... I actually noticed that the suggested table row with full span isn't really that sound, this is because it's actually inside the affected table row.
When I thought about this before, I was thinking about it sitting besides the affected table row instead, which would make sense, but would require javascript to know what row to overlay over and where to position.

> So keeping the same implementation that @stather implemented makes more sense I believe since you easily get it to sit correctly on top of the row with just CSS.

The problem with div inside the table is that it isn't semantically correct, and not all browsers support it. So it can be problematic. That's why @stather chose `tr`, I believe.

> I was also thinking that maybe we should also provide a default css class to make it easier for users to style it.
Would you agree? Naming? b-table-row-overlay? or you want some other name closer to the current provider?

In this instance, we can go with `b-table-row-overlay` since it is not a standalone component.

> And I was thinking, maybe this should have been straight up implemented on the Table itself like we did with other features, resizable, etc..... I think it's worth moving it over. Agreed?

I would leave it on DataGrid.

stsrki:
PS. I like it with div more. It works better, in my opinion. We just need to make sure that it doesn't break something else.

David-Moreira:
> PS. I like it with div more. It works better, in my opinion. We just need to make sure that it doesn't break something else.

Yes I like it more also. Less hassle since it doesn't apply all the table based styles. Since it's just a position:absolute element it seems to work good even if it's semantically incorrect.

stsrki:
> I was also thinking that maybe we should also provide a default css class to make it easier for users to style it.
> Would you agree?

Do this and I think it is ready for merge.

David-Moreira:
@stsrki done.

stsrki (approved):

Issue #4075, opened 2022-08-26
NumericPicker ignores Value the first time it is created


WayneHiller:
Any page that has NumericPicker's will ignore the Value of a bound property the first time it is created (all will have zeros).  Any subsequent times that any page that is displayed with NumericPicker's it works correctly. It seems to be something global in the NumberPicker.

 ```
<NumericPicker Decimals="0" SelectAllOnFocus="true" @bind-Value="@entity.HeaderCount" />
```

.Net 6, Blazorise 1.6, Blazor Wasm. Chrome.

**Comment by stsrki** [comment]
Unable to reproduce:

**Code:**

```razor
<NumericPicker Decimals="0" SelectAllOnFocus @bind-Value="@HeaderCount" />

@code {
    decimal HeaderCount = 50;
}
```

**Open page:**

![image](https://user-images.githubusercontent.com/900302/189904673-f6d5fdca-b477-4457-be53-ae455b81e450.png)

---

Maybe you're loading the value at a different page rendering state?

PS. Closing now.

WayneHiller:
The value is being assigned from an object property returned from an API call.

WayneHiller:
I have narrowed it down to this in a standard .Net 6 Blazor Wasm with API project:
Note I am using Flurl for the API call.

Client Index.razor
```
<div>
    <NumericPicker Decimals="0" 
                    SelectAllOnFocus="true" 
                    @bind-Value="@Forecast.TemperatureC" />
</div>

@code {
    private WeatherForecast Forecast = new();

    protected override async Task OnAfterRenderAsync(bool firstRender)
    {
        await base.OnAfterRenderAsync(firstRender);

        Forecast = await NavMgr.BaseUri.AppendPathSegment("weatherforecast/getforecast").GetJsonAsync<WeatherForecast>();
        await InvokeAsync(StateHasChanged);
    }
}
```

Server WeatherForecastController.cs
```
        [HttpGet]
        [Route("getforecast")]
        public async Task<WeatherForecast> GetForecast()
        {
            return new WeatherForecast
            {
                Date = DateTime.Now,
                TemperatureC = 45,
                Summary = "Testing"
            };
        }
```

The first time you visit the Index page the NumericPicker will stay at 0. After that you can go to any page and back to Index and it will contain 45.

WayneHiller:
Definitely is a timing issue with the StateHasChanged call. Adding a Delay to the API call fixes the issue.
```
        [HttpGet]
[...]

PR #4331, opened 2022-11-25
Implement delayed execution in BaseAfterRenderComponent

stsrki:
**Test for #4075** 

This test can be run only on Blazorise.Demo.Bootstrap.Server because it only has the controller with the endpoint `weatherforecast/getforecast"`.

```razor
@using Flurl;
@using Flurl.Http;
@using System.Text.Json.Serialization;

<Row>
    <Column>
        <Field>
            <NumericPicker Decimals="0" SelectAllOnFocus="true" @bind-Value="@Forecast.TemperatureC" />
        </Field>
    </Column>
</Row>


@code {
    [Inject] NavigationManager NavMgr { get; set; }

    private WeatherForecast Forecast = new();

    protected override async Task OnAfterRenderAsync( bool firstRender )
    {
        await base.OnAfterRenderAsync( firstRender );

        if ( firstRender )
        {
            Forecast = await NavMgr.BaseUri.AppendPathSegment( "weatherforecast/getforecast" ).GetJsonAsync<WeatherForecast>();
        }
    }
}
```

---

**Test for #4114**

```razor
<Row>
    <Column>
        <RichTextEdit @ref="editor" Theme="RichTextEditTheme.Snow" />
    </Column>
</Row>

@code {
    private RichTextEdit editor;

    private string html = "Hello world.";
    private bool isSet;

    protected override async Task OnAfterRenderAsync( bool firstRender )
    {
        if ( !isSet && editor != null )
        {
            isSet = true;

            await editor.SetHtmlAsync( html );
            StateHasChanged();
        }

        await base.OnAfterRenderAsync( firstRender );
    }
}
```

### DIFF
--- a/Build/Blazorise.Demo.props
+++ b/Build/Blazorise.Demo.props
@@ -19,6 +19,7 @@
 		<PackageReference Include="Microsoft.Extensions.Http" Version="7.0.0" />
 		<PackageReference Include="FluentValidation" Version="11.2.0" />
 		<PackageReference Include="FluentValidation.DependencyInjectionExtensions" Version="11.2.0" />
+		<PackageReference Include="Flurl.Http" Version="4.0.0-pre2" />
 	</ItemGroup>
 	
 </Project>
--- a/Demos/Blazorise.Demo.Bootstrap.Server/Controllers/WeatherForecastController.cs
+++ b/Demos/Blazorise.Demo.Bootstrap.Server/Controllers/WeatherForecastController.cs
@@ -1,0 +1,30 @@
+﻿using System;
+using System.Threading.Tasks;
+using Blazorise.Shared.Models;
+using Microsoft.AspNetCore.Mvc;
+
+namespace Blazorise.Demo.Bootstrap.Server.Controllers;
+
+[Route( "weatherforecast" )]
+[ApiController]
+public class WeatherForecastController : ControllerBase
+{
+    private static readonly string[] Summaries = new[]
+    {
+        "Freezing", "Bracing", "Chilly", "Cool", "Mild", "Warm", "Balmy", "Hot", "Sweltering", "Scorching"
+    };
+
+    private static readonly Random Rand = new Random();
+
+    [HttpGet]
+    [Route( "getforecast" )]
+    public Task<WeatherForecast> GetForecast()
+    {
+        return Task.FromResult( new WeatherForecast
+        {
+            Date = DateTime.Now,
+            TemperatureC = Rand.Next( -20, 55 ),
+            Summary = Summaries[Rand.Next( Summaries.Length )]
+        } );
+    }
+}
--- a/Demos/Blazorise.Demo.Bootstrap.Server/Startup.cs
+++ b/Demos/Blazorise.Demo.Bootstrap.Server/Startup.cs
@@ -59,6 +59,7 @@ namespace Blazorise.Demo.Bootstrap.Server
 
             app.UseEndpoints( endpoints =>
             {
+                endpoints.MapControllers();
                 endpoints.MapBlazorHub();
                 endpoints.MapFallbackToPage( "/_Host" );
             } );
--- a/Shared/Blazorise.Shared/Models/WeatherForecast.cs
+++ b/Shared/Blazorise.Shared/Models/WeatherForecast.cs
@@ -1,0 +1,16 @@
+﻿using System;
+using System.Text.Json.Serialization;
+
+namespace Blazorise.Shared.Models;
+
+public class WeatherForecast
+{
+    [JsonPropertyName( "date" )]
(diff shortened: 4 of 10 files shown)


David-Moreira:
Am I missing something, isn't this basically what we had previously? `PushDelayedExecuteAfterRender` will never get called since `ShouldDelayExecution` is always `false`.
Everything works as it was working before, no?

stsrki:
> Am I missing something, isn't this basically what we had previously? `PushDelayedExecuteAfterRender` will never get called since `ShouldDelayExecution` is always `false`.

Almost, yes. ShouldDelayExecution is the only thing that separates it, but without it, it doesn't work as expected. At least I didn't find a way still.

David-Moreira:
> Almost, yes. ShouldDelayExecution is the only thing that separates it, but without it, it doesn't work as expected. At least I didn't find a way still.

I don't follow. Unless I'm missing something this **WILL NEVER** be called
image
this **WILL NEVER** be true
image

So basically the code is redudant and could be removed. Again, Am I not seeing something clearly?

David-Moreira:
Ah, I was missing some projects to be loaded and wasn't finding references, but only `RichTextEdit` uses this??
image

 Didn't `NUmericPicker` and `Markdown` had issues related to his?

stsrki:
Are you sure you're not looking at the wrong commit again? The `ShouldDelayExecution` is explicitly overridden by the components that must behave a certain way. In this vase, RTE, Markdown, and NumericPicker.

stsrki:
Haha ignore my last message. I didn't push all :D

David-Moreira:
Should `Rendered = true;` be set before `PushDelayedExecuteAfterRender()`. It's a very tiny window, but would it make sense?
image

stsrki:
> Should `Rendered = true;` be set before `PushDelayedExecuteAfterRender()`. It's a very tiny window, but would it make sense? image

I've just now hit that tiny window. So, yes.

David-Moreira:
Other than that, I'm not sure how to reliabily reproduce the previous bugs, but everything seems to be working fine, so looks good. And Implemention got way cleaner.

David-Moreira:
Now another question is.. Why would this not be on "globally" by default? For "being safe" purposes? 
Since if this would happen in any other component I guess it would be good to have this behaviour? or not?

stsrki:
> Now another question is.. Why would this not be on "globally" by default? For "being safe" purposes? Since if this would happen in any other component I guess it would be good to have this behaviour? or not?

What do you mean have it globally? You mean to always do that?

David-Moreira:
> > Now another question is.. Why would this not be on "globally" by default? For "being safe" purposes? Since if this would happen in any other component I guess it would be good to have this behaviour? or not?
> 
> What do you mean have it globally? You mean to always do that?

Yes that.

stsrki:
@David-Moreira PR description edited to include test code. Cleaned the lifecycle to not depend on `ShouldDelayExecution`, but it still requires `delayedExecuteAfterRenderQueue`.

stsrki:
PS. as long as we use `override OnFirstAfterRenderAsync` to initialize components I think we're safe with this approach.

David-Moreira:
> PS. as long as we use `override OnFirstAfterRenderAsync` to initialize components I think we're safe with this approach.

So there's a chance we might miss use the components initialize? How would we break it then?

stsrki:
> > PS. as long as we use `override OnFirstAfterRenderAsync` to initialize components I think we're safe with this approach.
> 
> So there's a chance we might miss use the components initialize? How would we break it then?

I went and double-check. Even if we don't use `OnFirstAfterRenderAsync` and just override `OnAfterRenderAsync` directly it will work. The only way that it doesn't work is if we do `await base.OnAfterRenderAsync( firstRender );` before initialization.

**THIS WILL NOT WORK**

```cs
protected override async Task OnAfterRenderAsync( bool firstRender )
{
    await base.OnAfterRenderAsync( firstRender );

    if ( firstRender )
    {
        // try initializing here
    }
}
```

**THIS WILL WORK**

```cs
protected override async Task OnAfterRenderAsync( bool firstRender )
{
    if ( firstRender )
    {
        // try initializing here
    }

    await base.OnAfterRenderAsync( firstRender );
}
```

which makes sense 100%. So we're good to go. We can make it a good convention to use `OnFirstAfterRenderAsync` but even if we don't, it will still work.

David-Moreira:
Did you commit everything, what's the weatherforecast controller for?

David-Moreira:
LGTM

stsrki:
> Did you commit everything, what's the weatherforecast controller for?

Just a thing for us if we ever need to test components with connections to the api. In its own doesn't do much but I don't want to recreate it every time. Other than that we can consider having more use cases in the demo app where rest api is used.

PR #4333, opened 2022-11-26
Markdown: NotifyImageUploadSuccess prefer string.Empty over null as mde throws on null

### LINKED ISSUE #4299: Markdown image upload not working as expected in Blazor WASM
**Describe the bug**
I'm attempting to update the `File.UploadUrl` property in the `Blazorise.Markdown.Markdown` component so it can render the image in the editor.

In my environment I cannot get the URL from `ImageUploadChanged` to `ImageUploadEnded` to set the Url.

**To Reproduce**

Environment:

- .NET 6
- ASP.NET Core hosted project
- Blazor WASM
- Blazorise & all extensions are at v1.1.3.1

Component
```html
<Markdown 
	@bind-Value="@_input.Content"
	Autofocus="true" 
	ImageMaxSize="2097152"
	ImageUploadChanged="@OnImageUploadChanged"
	ImageUploadStarted="@OnImageUploadStarted"
	ImageUploadProgressed="@OnImageUploadProgressed"
	ImageUploadEnded="@OnImageUploadEnded"/>
```

Code behind file
```csharp
private string? filePath;

[Parameter]
public string? ArticleId { get; set; }

[Inject]
private INewsManagerClient NewsManagerClient { get; set; }

async Task OnImageUploadChanged(FileChangedEventArgs e)
{

	try
	{
		foreach (var file in e.Files)
		{
			var response = await NewsManagerClient.AddNewsArticleAttachmentAsync(ArticleId, file);

			file.UploadUrl = response;

			_imageUploads[file.Name].Uri = response;
			filePath = response;

			file.Status = FileEntryStatus.Uploaded;
		}
	}
	catch (Exception exc)
	{
		Console.WriteLine(exc.Message);
	}
	finally
	{
		this.StateHasChanged();
	}
}


Task OnImageUploadEnded(FileEndedEventArgs e)
{
	if (e.Success)
	{
		e.File.UploadUrl = filePath; // filePath is always null here
	}
	else
	{
		e.File.UploadUrl = string.Empty;
	}
	
	return Task.CompletedTask;
}

```

**Expected behavior**
I would expect one of two things:
1. The `IFileEntry` would get it's update from being set after the upload completes when using the `AddNewsArticleAttachmentAsync` method. It's not. It's always null in `OnImageUploadEnded`
2. The private field `filePath` would have the update. It's always null in `OnImageUploadEnded`

**Additional context**
I have confirmed that the upload is indeed happening, I can see the files in my blob storage. I can also see my client is returning a valid URL to those resources as well.

David-Moreira:
Closes #4299

> 1— Allow to set file upload url after upload was completed. I.e receive url from api upload call.
> 2— 'null' exception

This **PR** Fixes number 2 problem.

However for number **1**, I can only think of a solution for now, which we would have to schedule for **1.2** if you agree.
Introduce a new bool `Parameter NotifyImageUploadStatusAuto`, that the user can set to `false`. 
Introduce methods that wrap the `JSModule.NotifyImageUploadSuccess` , `JSModule.NotifyImageUploadError`
And handle himself, this part, this way he can do it himself with a bit more code.
image

### DIFF
--- a/Source/Extensions/Blazorise.Markdown/Markdown.razor.cs
+++ b/Source/Extensions/Blazorise.Markdown/Markdown.razor.cs
@@ -325,7 +325,7 @@ namespace Blazorise.Markdown
                 await ImageUploadEnded.Invoke( new( fileEntry, success, fileInvalidReason ) );
 
             if ( success )
-                await JSModule.NotifyImageUploadSuccess( ElementId, fileEntry.UploadUrl );
+                await JSModule.NotifyImageUploadSuccess( ElementId, fileEntry.UploadUrl ?? string.Empty );
             else
                 await JSModule.NotifyImageUploadError( ElementId, fileEntry.ErrorMessage );
         }


stsrki (approved):
LGTM

PR #4334, opened 2022-11-26
DataGrid: Virtualize fixes and optimizations

### LINKED ISSUE #4280: Negative VirtualizeCount for Datagrids
I had problems with my `Datagrid`, so I debugged my code and saw that `VirtualizeCount`  in the `ReadData` has sometimes a negative value, as you can see in the screenshot below

image

My backend was crashing into an error because of this negative value.

I am using Blazorise 1.0.7 with .Net6

### LINKED ISSUE #4289: Datagrid virtualization with external source makes scrolling instable
I saw, that scrolling a virtual Datagrid goes very jumpy sometimes. When using an external source it is never stopping to load the data:

How to reproduce:
I used the example Blazorise project (latest version):
Changed row 90 to: `await Task.Delay(random.Next(100) + 200);` (extra delay of 200 ms)
at row 113 I inserted:
`Console.WriteLine($"Get from {e.VirtualizeOffset}, {e.VirtualizeCount} records.");` (to see what he is doing).

When scrolling a few rows it is OK, but when scrolling halfway you get this problem..

David-Moreira:
Closes #4289
Closes #4280

Problem:
- Virtualize + ReadData would get jumpy or just act all weird and doing alot of requests, specially when navigating further down the table.

Solution:
- The placeholder for the rows would be half the size the actual rows, messing up the Virtualize calculations, so we've inserted a blank space so it can calculate the same space, this seems like a fine solution I would say.

Additionally implemented the same delay as in Quickgrid's as it removes redudants calls. Copied same note, as the suggestion seems implementation worthy.
`Virtualize` seems smoother with the changes, specially on Blazor.Server.

### DIFF
--- a/Source/Extensions/Blazorise.DataGrid/DataGrid.razor
+++ b/Source/Extensions/Blazorise.DataGrid/DataGrid.razor
@@ -59,7 +59,7 @@
                                         {
                                             @column.SortDirectionTemplate( column.CurrentSortDirection )
                                         }
-                                        else 
+                                        else
                                         {
                                             if ( column.CurrentSortDirection == SortDirection.Default && ShowDefaultSortIcon )
                                             {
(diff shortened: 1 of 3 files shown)


stsrki (approved):
LGTM. Seems strange to see `&nbsp;` but if it's working I guess it's OK.

PR #4340, opened 2022-11-29
Markdown and RichTextEdit: Invoke immediately if Rendered

stsrki:
Continuation from #4331

I have noticed on our https://support.blazorise.com/ that when I try to post a comment, I can't get the result from the `Markdown` by invoking the `GetValueAsync`. The problem was that I had made it so that #4331 would only return the result if the rendered component. I only noticed it because we wrapped `Markdown` in a separate `MarkdownEditor` component, so the Clicked event didn't get triggered on an IssuesPage.

`return await ExecuteAfterRenderAsync( async () => await JSModule.GetValue( ElementId ) );`

So this PR solves it by returning the result immediately if the component it already being rendered without waiting for the new render to complete.

### DIFF
--- a/Source/Extensions/Blazorise.Markdown/Markdown.razor.cs
+++ b/Source/Extensions/Blazorise.Markdown/Markdown.razor.cs
@@ -200,6 +200,9 @@ namespace Blazorise.Markdown
         /// <returns>Markdown value.</returns>
         public async Task<string> GetValueAsync()
         {
+            if ( Rendered )
+                return await JSModule.GetValue( ElementId );
+
             return await ExecuteAfterRenderAsync( async () => await JSModule.GetValue( ElementId ) );
         }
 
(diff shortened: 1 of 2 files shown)


Source/Extensions/Blazorise.Markdown/Markdown.razor.cs
```diff
@@ -200,6 +200,9 @@ protected async Task<T> ExecuteAfterRenderAsync<T>( Func<Task<T>> action, Cancel
         /// <returns>Markdown value.</returns>
         public async Task<string> GetValueAsync()
         {
+            if ( Rendered )
+                return await JSModule.GetValue( ElementId );
+
             return await ExecuteAfterRenderAsync( async () => await JSModule.GetValue( ElementId ) );
```
David-Moreira, line 206: These `ExecuteAfterRenderASync` lines make sense for the getters also?

Choose one or another (don't think it matters which here),
`await InvokeAsync` vs `await ExecuteAfterRenderAsync`

David-Moreira (approved):

Source/Extensions/Blazorise.Markdown/Markdown.razor.cs
stsrki: We have no idea what users can think of so it's better to have them.

PR #4351, opened 2022-12-01
DragDrop: optimizations

stsrki:
Closes https://support.blazorise.com/issues/80/Reordering-Items-in-DropContainer-Takes-1-minute-to-Complete

Optimizes DropZone so that items are calculated and fetched only when requested. Also added a bunch of `key` attributes to the draggable elements.

### DIFF
--- a/Source/Blazorise/Components/Droppable/DropZone.razor.cs
+++ b/Source/Blazorise/Components/Droppable/DropZone.razor.cs
@@ -332,9 +332,19 @@ namespace Blazorise
             return (item, canBeDropped);
         }
 
-        private void OnDragStarted() => dragging = true;
+        private Task OnDragStarted()
+        {
+            dragging = true;
 
-        private void OnDragEnded() => dragging = false;
+            return Task.CompletedTask;
+        }
+
+        private Task OnDragEnded( TItem item )
+        {
+            dragging = false;
+
+            return Task.CompletedTask;
+        }
 
         private int count;
         private IEnumerable<TItem> GetItems()
--- a/Source/Blazorise/Components/Droppable/DropZone.razor.cs
+++ b/Source/Blazorise/Components/Droppable/DropZone.razor.cs
@@ -33,6 +33,12 @@ namespace Blazorise
 
         private Dictionary<TItem, int> indices = new();
 
+        private bool recalculateItems = true;
+
+        private bool shouldRerender = true;
+
+        private IEnumerable<TItem> items;
+
         #endregion
 
         #region Constructors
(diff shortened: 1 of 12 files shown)


David-Moreira:
So how much did we speed it up with these changes? Based on the 1m rendering example? 

Also 1m seems like a very long time even for if our component wasn't optimized, are we sure it was not a flaw in the logic, some key handler going off too many times, or too much data?

Source/Blazorise/Components/Droppable/DropZone.razor.cs
```diff
@@ -317,20 +332,38 @@ private async Task OnDropHandler()
...
+            dragging = false;
 
-            if ( ItemsFilter != null )
+            return Task.CompletedTask;
+        }
+
+        private int count;
+        private IEnumerable<TItem> GetItems()
+        {
+            if ( recalculateItems )
```
David-Moreira, line 352: We've used something like `Dirty`, `IsDirty` coventions on other places I believe? `Datagrid` comes to mind, do we want to maintain same convention? 
Otherwise `recalculateItems` seems good.

Source/Blazorise/Components/Droppable/_Draggable.razor.cs
```diff
@@ -56,7 +57,8 @@ private async Task OnDragEndHandler( DragEventArgs e )
             }
             else
             {
-                await DragEnded.InvokeAsync( Item );
+                if ( DragEnded is not null )
```
David-Moreira, line 60: Both `Drop `and `DragEndHandler `trigger `DragEnded`, is this correct?

David-Moreira:
> So how much did we speed it up with these changes? Based on the 1m rendering example?
> 
> Also 1m seems like a very long time even for if our component wasn't optimized, are we sure it was not a flaw in the logic, some key handler going off too many times, or too much data?

If it's related to various fragments behind added and that triggering a re.render each time, the `Defer` component from Steve Sanderson might be worth investigating, from what I understand it defers rendering until all the necessary "internal components" have been gathered and only then renders. Not sure if applicable here.

stsrki:
> > So how much did we speed it up with these changes? Based on the 1m rendering example?
> > Also 1m seems like a very long time even for if our component wasn't optimized, are we sure it was not a flaw in the logic, some key handler going off too many times, or too much data?
> 
> If it's related to various fragments behind added and that triggering a re.render each time, the `Defer` component from Steve Sanderson might be worth investigating, from what I understand it defers rendering until all the necessary "internal components" have been gathered and only then renders. Not sure if applicable here.

Hm, would you like to try and see how it would work with Defer? I can leave the PR open for a few days.

PR #4353, opened 2022-12-01
Docs: Reconnect Blazor server if connection was lost

stsrki:


David-Moreira (approved):
LGTM.
I have personally used this one which I believe it was a suggestion from Steve Sanderson before the reconnectionHandler existed:


I'll leave it documented here, if we ever need it.

```
    <script>
        // Wait until a 'reload' button appears
        new MutationObserver((mutations, observer) => {
            if (document.querySelector('#components-reconnect-modal h5 a')) {
                async function attemptReload() {
                    await fetch(''); // Check the server really is back
                    location.reload();
                }
                observer.disconnect();
                attemptReload();
                setInterval(attemptReload, 250);
            }
        }).observe(document.body, { childList: true, subtree: true });
    </script>
```
**PS:** I don't think we need to leave the `console.log("Reconnected!")`

stsrki:
> LGTM. I have personally used this one which I believe it was a suggestion from Steve Sanderson before the reconnectionHandler existed:
> 
> I'll leave it documented here, if we ever need it.
> 
> ```
>     <script>
>         // Wait until a 'reload' button appears
>         new MutationObserver((mutations, observer) => {
>             if (document.querySelector('#components-reconnect-modal h5 a')) {
>                 async function attemptReload() {
>                     await fetch(''); // Check the server really is back
>                     location.reload();
>                 }
>                 observer.disconnect();
>                 attemptReload();
>                 setInterval(attemptReload, 250);
>             }
>         }).observe(document.body, { childList: true, subtree: true });
>     </script>
> ```
> 
> **PS:** I don't think we need to leave the `console.log("Reconnected!")`

So it checks the DOM, instead of some _healthcheck_ endpoint. We can try it if our approach doesn't work.

I agree, `console.log("Reconnected!")` can be removed.

PR #3699, opened 2022-04-09
Dropdown Fix RightAligned

David-Moreira:
Closes #3682

FIxes RightAligned parameter no longer being considered, since we've introduced popper to handle dropdown menu positioning.
And like we've talked also sets up better popper overflow/out of bounds detection, moving the menu accordingly.

PR #4357, opened 2022-12-02
Dropdown: use fixed strategy

stsrki:
We changed the strategy to `absolute` in #3699. But one of our commercial clients reported an issue. It affects the scrollbars when the menu is shown, even though it has enough room on the screen.

Changing it to `fixed` seems to fix the problem. I have tested all our dropdowns on the demo app, and it seems to be working.


David-Moreira:
LGTM. I also went over the providers and did not notice anything breaking.

David-Moreira (approved):

PR #4380, opened 2022-12-09
Refactor of columns and rows used space

stsrki:
This refactor was needed for the Tailwind provider to work correctly. The TW doesn't support offset columns as other providers do. Instead of offset-3 on a column which means how much the column will move, in TW, we need to know the exact index position of the column within the row. So with this PR, we keep track of columns and their position withing the row and calculate the used space. Then we use it for the column offset when needed.

I have also made many changes to our existing base components so that they would with the new system.

References

- https://tailwindcss.com/docs/grid-column#starting-and-ending-lines
- https://getbootstrap.com/docs/5.2/layout/columns/#offsetting-columns


Source/Blazorise/Fluent/FluentColumn.cs
```diff
@@ -191,14 +192,16 @@ public class FluentColumn :
...
     {
+        public ColumnWidth ColumnWidth { get; set; }
+
         public Breakpoint Breakpoint { get; set; }
 
         public bool Offset { get; set; }
     }
 
-    private ColumnDefinition currentColumn;
+    private ColumnDefinition columnDefinition;
```
David-Moreira, line 202: When looking at the class and at what this was doing, without even looking at the changes made, I got the feeling this should be named "current" since this can change according to the fluent context. Right? 
`private ColumnDefinition columnDefinition;` to `private ColumnDefinition currentColumnDefinition;`

David-Moreira (commented):
I guess the approach is fine, you had to find some way to persist the state. 
Having a Row/Columns "connection" makes sense.
The `ResetUsedSpace` is kinda weird... I guess there must be specific uses, I didnt really get it.

Source/Blazorise/Base/BaseColumnableComponent.cs
```diff
@@ -0,0 +1,85 @@
...
+using Blazorise.Utilities;
+using Microsoft.AspNetCore.Components;
+#endregion
+
+namespace Blazorise;
+
+/// <summary>
+/// Base class for components that are containers for other components.
+/// </summary>
+public abstract class BaseColumnableComponent : BaseComponent, IColumnableComponent, IDisposable
```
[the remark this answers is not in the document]
stsrki: Yeah, I'm still thinking about that naming. There was also a `BaseRowableComponent ` but I renamed it to row context. Probably will settle for `BaseColumnComponent` too.

Source/Blazorise/Components/Row/RowContext.cs
```diff
@@ -0,0 +1,62 @@
...
+namespace Blazorise;
+
+/// <inheritdoc/>
+public class RowContext : IRowContext
+{
+    #region Members
+
+    private int spaceUsedByColumnables = 0;
+
+    private List<IColumnableComponent> columnables = new();
```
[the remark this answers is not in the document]
stsrki: Row SHOULD always have a Column placed inside of it. But others should not, like a field for example. YOu're right it is better to allocate it when needed.

Source/Blazorise/Interfaces/IRowContext.cs
```diff
@@ -0,0 +1,42 @@
...
+    /// <summary>
+    /// Forces the row to start calculating used space from start.
+    /// </summary>
+    /// <param name="column">Column that is requested the calculation.</param>
+    /// <remarks>
+    /// The used space will reset only if the column is the first columnn in a row(has a zero index). This should work
+    /// most of the time. But, in some rare cases when columns are (re)created dynamicaly(for example if it's
+    /// placed behind an if statement) then the column would have a highest index once it is recreated. For those cases
+    /// it is advised to developers to use the <see cref="BaseComponent.Display"/> parameter to show or hide the column.
+    /// </remarks>
```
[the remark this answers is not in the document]
stsrki: I have asked a few people from the community about finding a better solution, but there aren't any. There is no way of knowing an index of components within the parent. This was the closest to what we could do. And considering that it would do only in TW, we have enough time to think of better ways.

Source/Blazorise/Fluent/FluentColumn.cs
```diff
@@ -246,16 +297,13 @@ private void Dirty()
     /// <returns>Next rule reference.</returns>
     public IFluentColumnOnBreakpointWithOffsetAndSize WithColumnSize( ColumnWidth columnSize )
     {
-        HasSizes = true;
+        if ( columnSize != ColumnWidth.Default )
+            HasSizes = true;
 
-        var columnDefinition = new ColumnDefinition { Breakpoint = Breakpoint.None };
+        columnDefinition = columnDefinition = new ColumnDefinition { ColumnWidth = columnSize, Breakpoint = Breakpoint.None };
```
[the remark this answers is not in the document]
stsrki: Accidental copy paste :)

Source/Blazorise/Base/BaseRowableComponent.cs
```diff
@@ -9,58 +9,14 @@ namespace Blazorise;
 /// <summary>
 /// Base class for components that are containers for other components.
 /// </summary>
-public abstract class BaseRowableComponent : BaseComponent, IRowableComponent
+public abstract class BaseRowableComponent : BaseComponent
```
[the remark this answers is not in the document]
stsrki: Maybe BaseRowableComponent was not needed now. The RowContext was added so that there was less spaghetti code with all the `NotifyColumnableInitialized`, `NotifyColumnableRemoved` that was the same across several components. This way the logic is centralized into RowContext.

Source/Blazorise/Fluent/FluentColumn.cs
stsrki: done

Source/Blazorise/Components/Row/RowContext.cs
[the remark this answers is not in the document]
stsrki: done

Source/Blazorise/Fluent/FluentColumn.cs
```diff
@@ -211,14 +214,41 @@ private class ColumnDefinition
     #region Methods
 
     /// <inheritdoc/>
-    public string Class( IClassProvider classProvider )
+    public string Class( IClassProvider classProvider, IRowContext rowContext, IColumnableComponent currentColumnable )
     {
         if ( dirty )
         {
             void BuildClasses( ClassBuilder builder )
```
David-Moreira, line 221: Why use Closures here.
Maybe `ClassBuilder` should be able to be used without an `Action` when applicable? Meaning that you would remove this local function and just run the logic regularly using the new `ClassBuilder` instance you have built. Similar to how you use a regular `StringBuilder`.

Something similar to this?
image

stsrki:
> The `ResetUsedSpace` is kinda weird... I guess there must be specific uses, I didnt really get it.

If the page is refreshed the calculation must start from scratch. We don't know what might be changed. Otherwise without reset the space used would just grow and grow.

Source/Blazorise/Fluent/FluentColumn.cs
David-Moreira: To add to this, if `Dirty` is called alot, might be beneficial to just have a `ClassBuilder` member in the `FluentColumn`

Source/Blazorise/Base/BaseRowableComponent.cs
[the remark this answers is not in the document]
David-Moreira: I see. Great.

Source/Blazorise/Fluent/FluentColumn.cs
David-Moreira: I guess we had some other refactoring issue open for something similar to this, that we haven't gotten around to yet I think. I just happened to stumble upon this. Let me see if I find it and copy paste this there.

David-Moreira: Oh you just opened an issue. Kk

David-Moreira:
> > The `ResetUsedSpace` is kinda weird... I guess there must be specific uses, I didnt really get it.
> 
> If the page is refreshed the calculation must start from scratch. We don't know what might be changed. Otherwise without reset the space used would just grow and grow.

When does this happen exactly? If someone changes by adding a Column dynamically or changing the `ColumnSize` defintiion is that it?

stsrki:
> When does this happen exactly? If someone changes by adding a Column dynamically or changing the `ColumnSize` defintiion is that it?

Any EventCallback on a page could trigger a refresh, and then the Raw could be rerendered.

David-Moreira:
> > When does this happen exactly? If someone changes by adding a Column dynamically or changing the `ColumnSize` defintiion is that it?
> 
> Any EventCallback on a page could trigger a refresh, and then the Row could be rerendered.

Ah? But doesn't te he dirty help with not recalculating everytime?

stsrki:
Fluent builder will always create new instance. That's also one more thing that we need to figure how to optimize. Some kind of caching mechanism is needed.

stsrki:
I figured that users could chain column width for multiple breakpoints. So it would calculate the used space for each of them, and it would be wrong. I have added a dictionary for breakpoints.

Source/Blazorise/Components/Row/RowContext.cs
```diff
@@ -9,10 +9,10 @@ public class RowContext : IRowContext
 {
     #region Members
 
-    private int spaceUsedByColumns = 0;
-
     private List<IColumnComponent> columns;
 
+    private Dictionary<Breakpoint, int> usedSpaces = new();
```
David-Moreira, line 14: Same allocation problem we talked bout earlier. If no columns, no used spaces. no need for aloc.

stsrki: done

PR #4389, opened 2022-12-15
Splitter extension component

smfields:
This is a partial implementation of the Splitter component based on [BlazorSplit](https://blazor-split.smfields.net/). I haven't had the time to continue working on it, so this is only a partial implementation, and I haven't looked at it in a while.

From what I remember, I left off trying to find a way to allow users to control the styling on the component level rather than needing to create a separate CSS file that requires `::deep` CSS selectors.

Hopefully this is enough to get you going on this feature.


stsrki:
Looks great. I like how all is well documented. You did an excellent job. Shouldn't be many problems to continue from here. But if I have any questions I hope I can get back to you :)

stsrki:
Something went wrong with PR after I merged the latest changes from the `master`. Probably because your fork was too far behind. So, I have re-created it onto new PR #4412 and will continue working from there. Your contribution will still be mentioned, don't worry :)

Closing this now.

PR #4397, opened 2022-12-17
Datagrid fix empty template not rendering

### LINKED ISSUE #4390: EmptyTemplate is not rendered in DataGrid when data is empty
**Describe the bug**
EmptyTemplate is not rendered in DataGrid when data is empty. This happens in version 1.1.5, 1.1.4 seems to be ok. I'm using Bootstrap provider.

**To Reproduce**
1. Run Blazorise.Demo.Bootstrap
2. Go to Datagrid => Configurable demo
3. I didn't change any datagrid options like virtualization
4. Clear all records in data grid
5. You'll see that `No records...` is not rendered
image

**Expected behavior**
`No records...` text should be rendered.
image

I went through the repo and found out that it's working correctly under commit ebe1adf225491da7846269111823b2bf926ab26d. It breaks in the very next commit 8a605f5af766c0eaacdc83b3fdda04cf7bb8d470.

Hope I didn't overlooked some settings that might be required for this to work, but it's not working in the demo app, so I expect this is a correct setup.

David-Moreira:
Closes #4390

It was actually a combination of 8a605f5af766c0eaacdc83b3fdda04cf7bb8d470 messing with the `Rendered` flag since Datagrid does not call `await base.OnAfterRenderAsync( firstRender );` when it is first rendered.
There was two possible fixes here,
- We could make sure to call the `await base.OnAfterRenderAsync( firstRender ); `
- Make sure the Rendered flag is set on our `BaseAfterRenderComponent` even if firstRender was never true;

We decided to take the second route, as it is how it was working before from a more general pespective. From a `Datagrid` perspective we decide to not call `await base.OnAfterRenderAsync( firstRender );` just like it was previously, as to avoid any unintended/unexpected side effects.

And the 4ae8aa8909a9893a67a0a13d75968c3cc8531b87 (#3800)
Where the new added `Rendered` check to the `IsEmptyTemplateVisible` flag works for most cases, but on a first Render where the Datagrid has no `Data`, without `Virtualize` set, the flag for `IsEmptyTemplateVisible` is actually not recalculated after the component is `Rendered` and therefore would not display the `EmptyTemplate` also.

### DIFF
--- a/Source/Blazorise/Base/BaseAfterRenderComponent.cs
+++ b/Source/Blazorise/Base/BaseAfterRenderComponent.cs
@@ -80,6 +80,8 @@ namespace Blazorise
                 }
             }
 
+            Rendered = true;
+
             if ( executeAfterRenderQueue?.Count > 0 )
             {
                 while ( executeAfterRenderQueue.Count > 0 )
--- a/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
+++ b/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
@@ -1438,13 +1438,19 @@ namespace Blazorise.DataGrid
         /// Returns true if EmptyTemplate is set and Data is null or empty.
         /// </summary>
         protected bool IsEmptyTemplateVisible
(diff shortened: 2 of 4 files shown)


stsrki (approved):
LGTM

PR #4398, opened 2022-12-17
Autocomplete multiple values sync

### LINKED ISSUE #4372: Autocomplete with empty SelectedValues
**Describe the bug**
I have an Autocomplete-Element with `SelectionMode=Multiple` and want to display several badges.
When I change the `SelectedValues` property, it works as expected - except when I want to clear the values.

Then, the last value is shown.

**To Reproduce**
I've tested several ways

```razor
<Autocomplete
    AutoPreSelect
    CloseOnSelection
    Data="@(new[] {
        ("A", "a"),
        ("B", "b"),
        ("C", "c")
    })"
    SelectionMode="AutocompleteSelectionMode.Multiple"
    MinLength="0"
    TItem="(string Name, string Value)"
    TValue="string"
    TextField="@(item => item.Name)"
    ValueField="@(item => item.Value)"
    SelectedValues="@Values.Split(' ').ToList()"
    SelectedValuesChanged="@(values => Values = string.Join(" ", values))"
/>

<Autocomplete
    AutoPreSelect
    CloseOnSelection
    Data="@(new[] {
        ("A", "a"),
        ("B", "b"),
        ("C", "c")
    })"
    SelectionMode="AutocompleteSelectionMode.Multiple"
    MinLength="0"
    TItem="(string Name, string Value)"
    TValue="string"
    TextField="@(item => item.Name)"
    ValueField="@(item => item.Value)"
    @bind-SelectedValues="@Values2"
/>

<Button Clicked="@(() => { Values = "a b"; Values2.Clear(); Values2.AddRange(Values.Split(' ')); })">A, B</Button>
<Button Clicked="@(() => { Values = "a"; Values2.Clear(); Values2.AddRange(Values.Split(' ')); })">A</Button>
<Button Clicked="@(() => { Values = ""; Values2.Clear(); })">[empty]</Button>
```

(click on buttons 1 and 2 - whenever you click on \[empty\], the autocomplete elements should be empty)

Both ways will work, when I change
https://github.com/Megabit/Blazorise/blob/05b04636335cc8749cd1ce9ba7ed638b7027bcba/Source/Extensions/Blazorise.Components/Autocomplete.razor.cs#L206
to
`if ( selectedValuesParamChanged && selectedValuesParam != null && !Data.IsNullOrEmpty() )`

and
https://github.com/Megabit/Blazorise/blob/05b04636335cc8749cd1ce9ba7ed638b7027bcba/Source/Extensions/Blazorise.Components/Autocomplete.razor.cs#L218
to
`if ( texts != null && !SelectedTexts.AreEqualOrdered( texts ) )`

but I'm unsure if that's sufficient (and not too much).

-----
As a workaround, I call `Clear()` of the Autocomplete instance whenever the value changes, but that's not very convenient.

David-Moreira:
Closes #4372

I believe the commits should be self descriptive, just let me know if you have any doubts.

------
I believe we should investigate making the `List<>` Parameters we exposed, actually `IEnumerable<>`. This will probably make it easier to handle changes internally, since users won't be able to mutate the collection by calling `.Add` and `.Remove` directly.
But I'm not entirely sure it would cover all cases. 

And I believe we wouldn't need the triplet parameter thing glutio came up with. I only noticed today, that we're actually keeping an additional collection in memory with 
image
precisely because of these mutations which can't be easily detected && compared without creating new references in memory.

**Example:** User does : `SelectedValues.Add("1")`=> When it gets to `SetParametersAsync`, it can't detect a change regularly by doing
```
parameters.TryGetValue<IEnumerable<TValue>>( nameof( SelectedValues ), out var paramSelectedValues )
                && !SelectedValues.AreEqualOrdered( paramSelectedValues ) )
```
as we usually do, because it's basically the same reference and already mutated.

As for users being able to Add/Remove items programatically they would instead go trough the actual provided Autocomplete APIs or assign a new `IEnumerable<>` which, again I'm believing would make it easy to detect the change with the `ParameterView.TryGetValue` api.
Open a new issue with this if you believe it's worth exploring, otherwise I believe we've been closed most loopholes bit by bit with each issue we've had on this **v1.1** release.

### DIFF
--- a/Source/Extensions/Blazorise.Components/Autocomplete.razor.cs
+++ b/Source/Extensions/Blazorise.Components/Autocomplete.razor.cs
@@ -102,11 +102,11 @@ namespace Blazorise.Components
 
             await base.SetParametersAsync( parameters );
 
-            await SyncronizeSingle( selectedValueParamChanged, selectedTextParamChanged );
+            await SynchronizeSingle( selectedValueParamChanged, selectedTextParamChanged );
             await SynchronizeMultiple( selectedValuesParamChanged, selectedTextsParamChanged );
         }
 
-        private async Task SyncronizeSingle( bool selectedValueParamChanged, bool selectedTextParamChanged )
+        private async Task SynchronizeSingle( bool selectedValueParamChanged, bool selectedTextParamChanged )
         {
             if ( selectedTextParamChanged && !selectedValueParamChanged )
             {
--- a/Source/Extensions/Blazorise.Components/Autocomplete.razor.cs
+++ b/Source/Extensions/Blazorise.Components/Autocomplete.razor.cs
@@ -100,14 +100,14 @@ namespace Blazorise.Components
             bool selectedValuesParamChanged = false;
             bool selectedTextsParamChanged = false;
 
-            if ( parameters.TryGetValue<IEnumerable<TValue>>( nameof( SelectedValues ), out var paramSelectedValues ) 
+            if ( parameters.TryGetValue<IEnumerable<TValue>>( nameof( SelectedValues ), out var paramSelectedValues )
                 && !selectedValuesParam.AreEqualOrdered( paramSelectedValues ) )
             {
                 selectedValuesParamChanged = true;
                 selectedValues = null;
             }
 
-            if ( parameters.TryGetValue<IEnumerable<string>>( nameof( SelectedTexts ), out var paramSelectedTexts ) 
+            if ( parameters.TryGetValue<IEnumerable<string>>( nameof( SelectedTexts ), out var paramSelectedTexts )
                 && !selectedTextsParam.AreEqualOrdered( paramSelectedTexts ) )
             {
                 selectedTextsParamChanged = true;
@@ -217,6 +217,11 @@ namespace Blazorise.Components
             {
                 selectedTexts = texts;
                 await SelectedTextsChanged.InvokeAsync( texts );
+            }
+
+            if ( ( selectedValuesParamChanged && values is null && SelectedValues is null ) || ( selectedTextsParamChanged && texts is null && SelectedTexts is null ) )
+            {
+                await Task.WhenAll( ResetSelectedValues(), ResetSelectedTexts() );
             }
         }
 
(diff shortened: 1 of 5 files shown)


David-Moreira:
Testing Code:
```
<Autocomplete @ref="ac"
              AutoPreSelect
              CloseOnSelection
              Data="@(new[] {
        ("A", "a"),
        ("B", "b"),
        ("C", "c")
    })"
              SelectionMode="AutocompleteSelectionMode.Multiple"
              MinLength="0"
              TItem="(string Name, string Value)"
              TValue="string"
              TextField="@(item => item.Name)"
              ValueField="@(item => item.Value)"
@bind-SelectedValues="@Values" />

<Button Clicked="@(async () => { Values??=new(); Values.Add("a"); Values.Add("b"); })">A, B</Button>
<Button Clicked="@(async () => { Values??=new(); Values.Add("c"); })">C</Button>
<Button Clicked="@(async () => { await ac!.Clear(); })">[none]</Button>
<Button Clicked="@AlternateClear">[Alt Clear]</Button>
<Button Clicked="@AlternateClearNewCollection">[Alt Clear New Collection]</Button>
<Button Clicked="@AlternateClearNull">[Alt Clear Null]</Button>

@code {
    private List<string> Values = new();
    private Autocomplete<(string, string), string>? ac;

    private Task AlternateClear()
    {
        Values.Clear();
        return Task.CompletedTask;
    }

    private Task AlternateClearNewCollection()
    {
        Values = new();
        return Task.CompletedTask;
    }

    private Task AlternateClearNull()
    {
        Values = null;
        return Task.CompletedTask;
    }
}
```

stsrki:
There are a lot of changes. Hopefully, nothing is broken.

Regarding the mutation of the data by the user. That should be discouraged at all costs. Any Blazor component should be updated through the parameter. And if not, then e can have an exposed API in public methods. Anything else is unsupported. We should not and could not handle all scenarios.

I will try to review the PR ASAP.

David-Moreira:
> That should be discouraged at all costs. Any Blazor component should be updated through the parameter. And if not, then we can have an exposed API in public methods, just for convenience. Anything else is unsupported. We should not and could not handle all scenarios.

Right, that's why I was advocating to change to an `IEnumerable` in the future, I believe exposing a `List` might have been a bad design decision we made here.

stsrki (approved):
LGTM

PR #4406, opened 2022-12-20
Datagrid : Slightly improvements

### LINKED ISSUE #3908: Datagrid with many rows & DataGridMultiSelectColumn activated become slow when check pressed
when use @bind-SelectedRows selection of checkbox become much slower

an example in the attached example project

[BlazorTestGrid.zip](https://github.com/Megabit/Blazorise/files/8976504/BlazorTestGrid.zip)

David-Moreira:
Closes #3908

While a Datagrid without **@bind-SelectedRows** only calls `OnAfterRender` once, which is great... 
In fact, **@bind-SelectedRows** makes it way slower. 

Found out that alot of re-rendering was being provoked by the `SelectedRows.InvokeAsync` when this has a delegate. This makes it so, `SetParametersAsync` is called (twice, no idea why it calls twice) and then there were a bunch of re-renders, it's been reduced to 2x `SetParametersAsync` + 3x `OnAfterRenderAsync` currently. I couldn't figure out what more is provoking re-rendering.

Also found out, that each time `SetParametersAsync` is called, 
- it sets `Data` which would trigger `SetDirty`, and refilter Data once again. Decided to use a regular property and move the check to `SetParametersAsync` with an `IsEqual` call as we usually do.
- it would also set `SelectionMode` again and call `ExecuteAfterRender` logic multiple times too.

As usual modifying a core "component" like `Data` can be dangerous but everything seems to be working fine.

(I just did the testing by adding `Console.Writeline` on the livecycle methods and clicking the datagrid rows)
(I didn't test `@bind-SelectedRow` the same way, but I'm sure it's a similar behaviour as it will also trigger `SetParametersAsync` on row clicks)
**Testing Code:**
```
<DataGrid TItem="test" @ref="grid1"
          Data=@buff1
          Sortable Filterable Striped Narrow ShowPager Responsive Editable Bordered
          SelectionMode="DataGridSelectionMode.Multiple"
          PageSize=15>
    <DataGridColumns>
        <DataGridMultiSelectColumn TItem="test" Width="30px"></DataGridMultiSelectColumn>
        <DataGridColumn TItem="test" Field="@nameof(test.v1)" Caption="1" Sortable />
        <DataGridColumn TItem="test" Field="@nameof(test.v2)" Caption="2" Sortable />
        <DataGridColumn TItem="test" Field="@nameof(test.v3)" Caption="3" Sortable />
        <DataGridColumn TItem="test" Field="@nameof(test.v4)" Caption="4" Sortable />
        <DataGridColumn TItem="test" Field="@nameof(test.v5)" Caption="5" Sortable />
        <DataGridColumn TItem="test" Field="@nameof(test.v6)" Caption="6" Sortable />
        <DataGridColumn TItem="test" Field="@nameof(test.v7)" Caption="7" Sortable />
        <DataGridColumn TItem="test" Field="@nameof(test.v8)" Caption="8" Sortable />
    </DataGridColumns>
</DataGrid>

<DataGrid TItem="test" @ref="grid2"
          Data=@buff2
          Sortable Filterable Striped Narrow ShowPager Responsive Editable Bordered
          SelectionMode="DataGridSelectionMode.Multiple"
@bind-SelectedRows="praticheSelezionate2"
          PageSize=15>
    <DataGridColumns>
        <DataGridMultiSelectColumn TItem="test" Width="30px"></DataGridMultiSelectColumn>
        <DataGridColumn TItem="test" Field="@nameof(test.v1)" Caption="1" Sortable />
        <DataGridColumn TItem="test" Field="@nameof(test.v2)" Caption="2" Sortable />
        <DataGridColumn TItem="test" Field="@nameof(test.v3)" Caption="3" Sortable />
        <DataGridColumn TItem="test" Field="@nameof(test.v4)" Caption="4" Sortable />
        <DataGridColumn TItem="test" Field="@nameof(test.v5)" Caption="5" Sortable />
        <DataGridColumn TItem="test" Field="@nameof(test.v6)" Caption="6" Sortable />
        <DataGridColumn TItem="test" Field="@nameof(test.v7)" Caption="7" Sortable />
        <DataGridColumn TItem="test" Field="@nameof(test.v8)" Caption="8" Sortable />
    </DataGridColumns>
</DataGrid>

@code {
    DataGrid<test> grid1;
    DataGrid<test> grid2;
    public List<test> praticheSelezionate2 = new();

    public class test
    {
        public string v1 { get; set; }
        public string v2 { get; set; }
        public string v3 { get; set; }
        public string v4 { get; set; }
        public string v5 { get; set; }
        public string v6 { get; set; }
        public string v7 { get; set; }
        public string v8 { get; set; }
    }

    List<test> buff1 = new();
    List<test> buff2 = new();

    {
        for ( int cnt = 1; cnt <= 1000; cnt++ )
        {
            buff1.Add( new test() { v1 = "1", v2 = "2", v3 = "3", v4 = "4", v5 = "5", v6 = "6", v7 = "7", v8 = "8" } );
            buff2.Add( new test() { v1 = "1", v2 = "2", v3 = "3", v4 = "4", v5 = "5", v6 = "6", v7 = "7", v8 = "8" } );
        }

        base.OnInitialized();
    }
}
```

### DIFF
--- a/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
+++ b/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
@@ -246,6 +246,12 @@ namespace Blazorise.DataGrid
         {
             await CheckMultipleSelectionSetEmpty( parameters );
 
+            if ( parameters.TryGetValue<IEnumerable<TItem>>( nameof( Data ), out var changedData ) && !Data.IsEqual( changedData ) )
+                SetDirty();
+
+            if ( parameters.TryGetValue<DataGridSelectionMode>( nameof( SelectionMode ), out var changedSelectionMode ) && SelectionMode != changedSelectionMode )
+                ExecuteAfterRender( HandleSelectionModeChanged );
+
             await base.SetParametersAsync( parameters );
         }
 
@@ -322,7 +328,7 @@ namespace Blazorise.DataGrid
 
         private async Task HandleSelectionModeChanged()
         {
-            if ( selectionMode == DataGridSelectionMode.Multiple && SelectedRow != null )
+            if ( SelectionMode == DataGridSelectionMode.Multiple && SelectedRow != null )
             {
                 SelectedRows ??= new();
 
@@ -333,7 +339,7 @@ namespace Blazorise.DataGrid
                     await SelectedRowsChanged.InvokeAsync( SelectedRows );
                 }
             }
-            else if ( selectionMode == DataGridSelectionMode.Single && SelectedRows != null )
+            else if ( SelectionMode == DataGridSelectionMode.Single && SelectedRows != null )
             {
                 SelectedRows = null;
 
@@ -573,7 +579,7 @@ namespace Blazorise.DataGrid
         {
             await SelectRow( item );
 
-            await InvokeAsync( StateHasChanged );
+            await Refresh();
         }
 
         /// <summary>
@@ -743,7 +749,7 @@ namespace Blazorise.DataGrid
         protected internal Task ToggleDetailRow( TItem item, DetailRowTriggerType detailRowTriggerType, bool forceDetailRow = false, bool skipDetailRowTriggerType = false )
             => ToggleDetailRow( GetRowInfo( item ), detailRowTriggerType, forceDetailRow, skipDetailRowTriggerType );
 
-        protected internal Task ToggleDetailRow( DataGridRowInfo<TItem> rowInfo, DetailRowTriggerType detailRowTriggerType, bool forceDetailRow = false, bool skipDetailRowTriggerType = false )
+        protected internal async Task ToggleDetailRow( DataGridRowInfo<TItem> rowInfo, DetailRowTriggerType detailRowTriggerType, bool forceDetailRow = false, bool skipDetailRowTriggerType = false )
         {
             if ( rowInfo is not null )
             {
@@ -757,7 +763,7 @@ namespace Blazorise.DataGrid
                     var detailRowTriggerResult = DetailRowTrigger( detailRowTriggerContext );
 
                     if ( !skipDetailRowTriggerType && detailRowTriggerType != detailRowTriggerContext.DetailRowTriggerType )
-                        return Task.CompletedTask;
+                        return;
 
                     rowInfo.SetRowDetail( detailRowTriggerResult, detailRowTriggerContext.Toggleable );
 
@@ -774,10 +780,9 @@ namespace Blazorise.DataGrid
                     rowInfo.ToggleDetailRow();
                 }
 
-                return InvokeAsync( StateHasChanged );
+                await Refresh();
             }
 
-            return Task.CompletedTask;
         }
 
         /// <summary>
@@ -887,8 +892,7 @@ namespace Blazorise.DataGrid
             }
 
             await SelectedRowsChanged.InvokeAsync( SelectedRows );
-
-            await InvokeAsync( StateHasChanged );
+            await Refresh();
         }
 
         private async Task HandleShiftClick( MultiSelectEventArgs<TItem> eventArgs )
@@ -957,8 +961,7 @@ namespace Blazorise.DataGrid
             UnSelectAllRows = !selectAll;
 
             await SelectedRowsChanged.InvokeAsync( SelectedRows );
-
-            await InvokeAsync( StateHasChanged );
+            await Refresh();
         }
 
         // this is to give user a way to stop save if necessary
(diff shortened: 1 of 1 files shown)


Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -246,6 +246,12 @@ public override async Task SetParametersAsync( ParameterView parameters )
         {
             await CheckMultipleSelectionSetEmpty( parameters );
 
+            if ( parameters.TryGetValue<IEnumerable<TItem>>( nameof( Data ), out var changedData ) && !Data.IsEqual( changedData ) )
```
stsrki, line 249: Do we still want `.IsEqual` here, or `.AreEqual`, which is used only for lists?

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -322,7 +328,7 @@ private async Task CheckMultipleSelectionSetEmpty( ParameterView parameters )
 
         private async Task HandleSelectionModeChanged()
         {
-            if ( selectionMode == DataGridSelectionMode.Multiple && SelectedRow != null )
+            if ( SelectionMode == DataGridSelectionMode.Multiple && SelectedRow != null )
```
stsrki, line 331: Use `is not null` instead of `!= null`? I have also started to use it more lately.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -573,7 +579,7 @@ public async Task Select( TItem item )
         {
             await SelectRow( item );
 
-            await InvokeAsync( StateHasChanged );
+            await Refresh();
```
stsrki, line 582: Should we do `await InvokeAsync(Refresh);` instead of just `Refresh()`?

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -1598,15 +1601,7 @@ private bool IsMultiSelectAllIndeterminate
...
-        public IEnumerable<TItem> Data
-        {
-            get { return data; }
-            set
-            {
-                SetDirty();
-                data = value;
-            }
-        }
+        public IEnumerable<TItem> Data { get; set; }
```
stsrki, line 1604: I like this change.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
David-Moreira: Right. Its not new code. Sometimes I forget to change. We should do a wide PR that changes this in one of the next releases.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
David-Moreira: No. Its same thing as far as I know. The Statehaschanged is the thing that should be synced with the ui thread and it is called with the invokeasync already.
In fact this change that I did doesn't change much other then centralised the call.

I was trying to see if many state changes were making a difference in the amount of OnAfterRender changes. And trying to mitigate these by having a centralized flag that would keep track of the refresh call and only call statehaschanged once. However being called once or multiple times in a row click did not seem to matter.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
David-Moreira: Right I think we should disregard doing logic in the parameter setters and move it to the SetParameters, where we even have async support.
I think there's even an analyser that throws a warning to not do stuff on blazor parameters.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
David-Moreira: Honestly I forgot. From the tests it seemed to have the correct behaviour. I can change and retest.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
stsrki: Yeah, we can clean this after or for 1.2.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
stsrki: At least the logic that will trigger rerender. Other props that set `DirtyClasses()` are OK, I think.

stsrki (approved):
LGTM

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
David-Moreira: Actually I'd argue that those should also go there and it would even be more clear & concise what triggers DirtyClasses, since those would all be in the same place.

stsrki: Well, tbf, that was the idea for the new ComponentParameterInfo that is still in prototype limbo. I agree, we should refactor it slowly by moving to v2.

PR #4426, opened 2023-01-02
Added missing ant-display-none

Magehernan:
In AntDesign scss was missing ant-display-none so Fluid wasn't working fine.


stsrki (approved):
LGTM. Thank you.

PR #4427, opened 2023-01-02
Tailwind fix dropdown nesting

David-Moreira:


stsrki (approved):
Working great. As always it was a simple fix :)

Thank you!

Issue #4431, opened 2023-01-04
DataGrid: Select the sort direction on first click of sorting


pDakov:
**Is your feature request related to a problem? Please describe.**
Hi team,

I am curious if it is possible at the moment and I miss seeing how to do it, or if it can be implemented in the future.

I would like to have the opportunity to change the direction on the first click of sorting.

I do not mean this SortDirection="SortDirection.Descending", this is for the default sorting when the screen is loaded.

At the moment when I click for the first time on the header I get ASC sorting, and when clicking it again (the second time) I get DESC sorting.

**Describe the solution you'd like**
My understanding is if we have some attribute like this FirstClickSortDirection="SortDirection.Descending" to select the sort direction, we can achieve this result without any custom logic.

This will be useful for pages that show any financial report and will be more intuitive and user-friendly.

stsrki:
Instead of `FirstClickSortDirection` it seems to me that it would make more sense to have a sort direction logic reversed when clicked. Or, make the sort icon clickable and then show the drop-menu with an extra feature to select the desired sort mode.

pDakov:
> Instead of `FirstClickSortDirection` it seems to me that it would make more sense to have a sort direction logic reversed when clicked. Or, make the sort icon clickable and then show the drop-menu with an extra feature to select the desired sort mode.

I see it to be simple, without any visible changes for the user, only for the developer to set for the first click DESC, for the second to be ASC.

For the moment have no option, always is ASC for the first click and DESC for the second.

injDakov:
@stsrki
please take a look at this https://github.com/Megabit/Blazorise/pull/4432

Here is the implementation of my suggestion.

I will be thankful if you accepted it.
I am open to discussing the naming if you have some concerns.

PR #4432, opened 2023-01-04
#4431 DataGrid's Column extension: Added ability to change on first click sort direction

injDakov:
This is a simple improvement with a small impact on the existing functions.

This improvement adds the property FirstClickSortDirection gives the ability of the developer to select which will be sort direction on the first click, instead of always being ASC and after that, on the second click be DESC.

[Feature #4431](https://github.com/Megabit/Blazorise/issues/4431)

### DIFF
--- a/Demos/Blazorise.Demo/Pages/Tests/DataGrid/DataInMemoryPage.razor
+++ b/Demos/Blazorise.Demo/Pages/Tests/DataGrid/DataInMemoryPage.razor
@@ -33,7 +33,7 @@
                         <DataGridColumn TItem="Employee" Field="@nameof( Employee.Zip )" Caption="Zip">
                         </DataGridColumn>
                         <DataGridDateColumn TItem="Employee" Field="@nameof( Employee.DateOfBirth )" DisplayFormat="{0:dd.MM.yyyy}" Caption="Date Of Birth" Editable />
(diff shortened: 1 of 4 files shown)


David-Moreira:
> `*Click*` feels strange. I think naming it `bool ReverseSorting` will make more sense to the users.
> 
> @David-Moreira what do you think?
> 
> PS. @injDakov can you please accept the CLA?

I agree, since this does not affect only click, but sorting in general. I believe programatically calling the Sort method will have the same effect right?

I like the new naming, `ReverseSorting` or use the same naming as in the new bool flag, `InverseSorting`. 
Eitherway keep the flag and the new Parameter naming consistent.

Also this line should read better in the reverse order
`return !isInverse ? SortDirection.Ascending : SortDirection.Descending;`
**Prefer** 
`return isInverse ? SortDirection.Descending : SortDirection.Ascending;`

injDakov:

injDakov:
recheck

jwilcken (approved):

injDakov:
@stsrki  @David-Moreira  

I implemented the requested changes.
But I am not sure how to accept the CLA, I added comments "I have read the CLA Document and I hereby sign the CLA" and "recheck"  but looks is still not OK. 

Also see that have a mess with my local GIT settings, but hope that this is not related to CLA issue.

David-Moreira:
Hello,

Unfortunately it seems you didn't add proper metadata to the commits. Could you ammend these? It's probably easier to submit a new PR if you don't mind. Your user and email should be configured when submitting a commit.
image

PR #4434, opened 2023-01-04
#4431 DataGrid: Added ability to inverse sort direction

### LINKED ISSUE #4431: DataGrid: Select the sort direction on first click of sorting

injDakov:
This is a simple improvement with a small impact on the existing functions.

This improvement adds the property FirstClickSortDirection InverseSorting gives the ability of the developer to select which will be sort direction on the first click, instead of always being ASC and after that, on the second click be DESC.

Closes #4431

### DIFF
--- a/Demos/Blazorise.Demo/Pages/Tests/DataGrid/DataInMemoryPage.razor
+++ b/Demos/Blazorise.Demo/Pages/Tests/DataGrid/DataInMemoryPage.razor
@@ -33,7 +33,7 @@
                         <DataGridColumn TItem="Employee" Field="@nameof( Employee.Zip )" Caption="Zip">
                         </DataGridColumn>
                         <DataGridDateColumn TItem="Employee" Field="@nameof( Employee.DateOfBirth )" DisplayFormat="{0:dd.MM.yyyy}" Caption="Date Of Birth" Editable />
-                        <DataGridNumericColumn TItem="Employee" Field="@nameof( Employee.Childrens )" Caption="Childrens" InverseSorting="true" Editable Filterable="false" />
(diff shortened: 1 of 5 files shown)


injDakov:

injDakov:
recheck

stsrki:
I prefer the `Reverse`. Google also seems to agree

> In general, "inverse" usually means when all the individual qualities of something are opposite and "reverse" refers to direction or things going from the end/middle toward the beginning

David-Moreira:
> I prefer the `Reverse`. Google also seems to agree
> 
> > In general, "inverse" usually means when all the individual qualities of something are opposite and "reverse" refers to direction or things going from the end/middle toward the beginning

I see, then I will also have to agree. @injDakov if you'd be so kind... :)

injDakov:
> > I prefer the `Reverse`. Google also seems to agree
> > > In general, "inverse" usually means when all the individual qualities of something are opposite and "reverse" refers to direction or things going from the end/middle toward the beginning
> 
> I see, then I will also have to agree. @injDakov if you'd be so kind... :)

The names are improved. 
Documentation is updated.

:-)

stsrki (approved):
It took a while, but it is looking good now :)

Nice job. Thank you for the contribution @injDakov

PR #4435, opened 2023-01-05
DataGrid: grouping

### LINKED ISSUE #3120: GroupBy in DataGrid
I would like to be able to group by a property in my DataGrid. I don't think there's a way to do this currently.

Something like this:

image

David-Moreira:
**WIP**
Closes #3120

For now just drafting how the api would be. Rendering also does not seem too complex. We might want to use an Accordion or something.

@stsrki Can you help me with a few questions?

Isn't Groupby a single function only? Where were we going thinking about defining multiple Funcs in each column? I can't see how would that work?

Maybe it should be:
- a `GroupBy` Func at the Datagrid level, if defined it overrides everything else.
- Groupable defined at column level, if multiple columns are defined as groupable, then these will all be considered in the groupby (probably need to create an expression here that targets the multiple fields)

No need for the GroupBy Func at the column level, unless i'm missing something elementar... 

------
(Also, should we even attempt groupby in Virtualize? Not sure how feasible it would be.)

---------
image


stsrki:
From my understanding, by looking at how others made their APIs., they all have grouping on a column level.

- `Groupable` means that the column CAN be grouped, but it doesn't mean that it is grouping at the moment. Think of this parameter as a way of defining for the END USER that he can drag and drop the column header to the grouping area, which will then group the data. For now, we can ignore this field and make it in the future as all this drag and drop will take more time ATM.
- `Grouping` parameter. While it seems the same as `Groupable`, they are different. The `Grouping` means that the component will automatically be grouped as since it is set to `true`. The default grouping `key` is the `Field` parameter.
- This leads us to the last thing. The `GroupBy` needs to be on the column level. The `GroupBy` is only optional. If undefined column will be grouped by the `Field`. If defined, the `GroupBy` can override the `Field` and it can be grouped by anything, even by multiple keys, eg `.GroupBy(x => new { x.Column1, x.Column2 })`.

One thing, though. Since `GroupBy` return `object`, I think that the expression `Func<TItem, object>` will need to be compiled.

Source/Extensions/Blazorise.DataGrid/DataGridColumn.cs
```diff
@@ -154,6 +154,13 @@ internal bool CanSort()
     internal string GetFieldToSort()
         => string.IsNullOrEmpty( SortField ) ? Field : SortField;
 
+    /// <summary>
+    /// Gets the GroupBy Func to be applied.
+    /// </summary>
+    /// <returns></returns>
+    internal Func<TItem, object> GetGroupByFunc()
+        => GroupBy is not null ? GroupBy : GroupBy;
```
stsrki, line 162: It returns GroupBy in all cases?

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -1312,9 +1318,22 @@ private IEnumerable<TItem> FilterViewData()
...
             }
-            return filteredData.Skip( skipElements ).Take( PageSize );
+
+            var pagedData = filteredData.Skip( skipElements ).Take( PageSize );
+
+            foreach ( var column in Columns )
+            {
+                if ( column.Groupable )
+                {
+                    //TODO : Need to think about this? Can we really have multiple GroupBy Funcs? Does that make sense at all? There should only be a single Func?
```
stsrki, line 1328: We can have nested grouping. This is why grouping is on a column level.

eg

image

Source/Extensions/Blazorise.DataGrid/DataGridColumn.cs
```diff
@@ -407,6 +414,16 @@ public SortDirection CurrentSortDirection
     /// </summary>
     [Parameter] public bool Filterable { get; set; } = true;
 
+    /// <summary>
+    /// Gets or sets whether the column should be used as a Group Field. A custom <see cref="GroupBy"/> function can however be provided.
+    /// </summary>
+    [Parameter] public bool Groupable { get; set; }
```
stsrki, line 420: Also, add the Grouping parameter.

stsrki:
We will also need to make groups collapsible. And make groups bold. Just wrap the text with a `Strong` component.

image

PS. what about making the collapsed groups not render any content or rows that belong to them?

Source/Extensions/Blazorise.DataGrid/DataGridColumn.cs
David-Moreira: Right I didn't do the field getter yet.

David-Moreira:
> From my understanding, by looking at how others made their APIs., they all have grouping on a column level.
> 
> * `Groupable` means that the column CAN be grouped, but it doesn't mean that it is grouping at the moment. Think of this parameter as a way of defining for the END USER that he can drag and drop the column header to the grouping area, which will then group the data. For now, we can ignore this field and make it in the future as all this drag and drop will take more time ATM.
> * `Grouping` parameter. While it seems the same as `Groupable`, they are different. The `Grouping` means that the component will automatically be grouped as since it is set to `true`. The default grouping `key` is the `Field` parameter.
> * This leads us to the last thing. The `GroupBy` needs to be on the column level. The `GroupBy` is only optional. If undefined column will be grouped by the `Field`. If defined, the `GroupBy` can override the `Field` and it can be grouped by anything, even by multiple keys, eg `.GroupBy(x => new { x.Column1, x.Column2 })`.
> 
> One thing, though. Since `GroupBy` return `object`, I think that the expression `Func<TItem, object>` will need to be compiled.

I just slept on it. I wasn't thinking about it right. Yes you can have several group by Func. But there should also be an order right?

So for example I can group by gender and then on each gender bucket group by type of employee.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
David-Moreira: You re right I did not think right about the problem yesterday. See my comment regarding possible ordering to define the order the grouping / nesting happens?

stsrki:
> So for example I can group by gender and then on each gender bucket group by type of employee.
> Or the opposite first by type then by gender. There has to be some ordering to define these right?

Yes that's right. The order of grouping is in the order of how they are added to the "bucket". First in, first to group by.

David-Moreira:
> > So for example I can group by gender and then on each gender bucket group by type of employee.
> > Or the opposite first by type then by gender. There has to be some ordering to define these right?
> 
> Yes that's right. The order of grouping is in the order of how they are added to the "bucket". First in, first to group by.

I'm thinking about this one programmatically. If it's on the ui then the user can just go as he pleases. If for example the developer is doing a grouped Datagrid where the user has no interaction? He might want to define the order each group is applied? Even from an initial start point of view.

Source/Extensions/Blazorise.DataGrid/DataGridColumn.cs
David-Moreira: `Groupable` vs `Grouping` seems so confusing. 
What about `Grouped` or `StartGrouped` instead of `Grouping`?

David-Moreira:
@stsrki  I'm not sure it's going to be easy to integrate the accordion, we'd need to heavily style it, what do you think?

image

stsrki:
> @stsrki I'm not sure it's going to be easy to integrate the accordion, we'd need to heavily style it, what do you think?

It seems I missed your last comment about accordions. Don't do it with an accordion. Make it all with regular table rows. Just add padding where needed. It will be a lot faster.

Source/Extensions/Blazorise.DataGrid/DataGridColumn.cs
stsrki: > `Groupable` vs `Grouping` seems so confusing.
> What about `Grouped` or `StartGrouped` instead of `Grouping`?

We already have a similar naming scheme, `Editable`, `Resizable`, `Filterable`, so `Groupable` will fit. We can think of renaming `Grouping` to `Group` but I don't think it is necessary. Just document it well and it will be ok.

stsrki:
It was just now that I thought about it. People will definitely want to have `GroupTemplate` to customize the group UI. Whould we just add `GroupTemplate`, or do we need more?

David-Moreira:
> It was just now that I thought about it. People will definitely want to have `GroupTemplate` to customize the group UI. Should we just add `GroupTemplate`, or do we need more?

Nice catch. Will add.

stsrki:
I guess you will integrate the new TableRowGroup next?

David-Moreira:
> I guess you will integrate the new TableRowGroup next?

Yes I'd like that and some way to select columns to group by.

stsrki:
> some way to select columns to group by.

So you want to have that feature in 1.2?

David-Moreira:
> > some way to select columns to group by.
> 
> So you want to have that feature in 1.2?

Yea I'd like to if we make it on time.

stsrki:
In that case add a drop area like in the image. The columns that are marked as `Groupable` can be dragged. Ir did you had some other idea?

image

Source/Extensions/Blazorise.DataGrid/Contexts/GroupContext.cs
```diff
@@ -0,0 +1,43 @@
...
+using Microsoft.AspNetCore.Components;
+#endregion
+
+namespace Blazorise.DataGrid;
+
+/// <summary>
+/// Context for a group.
+/// </summary>
+/// <typeparam name="TItem"></typeparam>
+public class GroupContext<TItem>
```
stsrki, line 14: Name it DataGridGroupContext

Source/Extensions/Blazorise.DataGrid/Contexts/GroupContext.cs
```diff
@@ -0,0 +1,43 @@
...
+
+    /// <summary>
+    /// Gets the group key.
+    /// </summary>
+    public object Key { get; }
+
+    /// <summary>
+    /// Gets the group values.
+    /// </summary>
+    public IEnumerable<TItem> Values { get; }
```
stsrki, line 40: Items?

Source/Extensions/Blazorise.DataGrid/DataGrid.razor
```diff
@@ -267,27 +267,39 @@
...
-                {
-                    @rowFragment( item )
-                }
-            }
-            else
-            {
-                var groupedData = group.GroupBy( x => nextGroupableColumn.GetGroupByFunc().Invoke( x ) );
-                @groupFragment((x.iteration + 1, x.groupableColumns, groupedData))
-            }
+            <TableRowGroup @key="group.Key.ToString()">
```
stsrki, line 304: Do we need `.ToString()`?

Source/Extensions/Blazorise.DataGrid/_DataGridRow.razor
```diff
@@ -22,6 +22,20 @@
...
               HoverCursor="@GetHoverCursor()">
+
+        @if ( ParentDataGrid.IsGroupable() )
+        {
+            var count = ParentDataGrid.GroupableColumnsCount();
+            if ( count > 1 )
+            {
+                for ( int i = 0; i < count; i++ )
+                {
+                    <TableRowCell Style="width:25px;" />
```
stsrki, line 33: What is the need for a cell with `Style="width:25px;"`?

Source/Extensions/Blazorise.DataGrid/DataGrid.razor
David-Moreira: I believe so. The key is an object, we are just expecting that it always is a string. We could cast it sooner somewhere I guess?

Source/Extensions/Blazorise.DataGrid/Contexts/GroupContext.cs
David-Moreira: Uhm I believe that for Datagrid and since its on its own namespace we didn't do Datagrid prefix for the context? See other existing contexts... Filter, edit etc?

stsrki: Hmm you're right. OK leave it without prefix.

Source/Extensions/Blazorise.DataGrid/_DataGridRow.razor
David-Moreira: I was doing that so the width was predictable and then we could use width times the number of nesting to calculate the indenting.
But I guess I'm a bit lost with the indenting part. If you can take a look.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor
stsrki: Make it a string, and convert it to a string when you assign it in a constructor.

Source/Extensions/Blazorise.DataGrid/_DataGridRow.razor
stsrki: I can look at it once you settle on the almost-finished solution.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor
David-Moreira: I guess if we go the state tracking route then we'll have the groups "pre-calcukated" we can cast there.

Source/Extensions/Blazorise.DataGrid/_DataGridRow.razor
David-Moreira: Right maybe I can think of something in the meantime. Thanks.

stsrki: Try with Margin or Padding utility.

David-Moreira: It didn't work because it applies to the TR which has no effect.

David-Moreira: I mean, at least from what I tried

Source/Extensions/Blazorise.DataGrid/Contexts/GroupContext.cs
```diff
@@ -13,31 +13,66 @@ namespace Blazorise.DataGrid;
...
+    #endregion
+
+    #region Methods
+
+    internal void SetExpanded( bool expanded )
+    {
+        Expanded = expanded;
+    }
+
+    internal void SetNestedGroup( List<GroupContext<TItem>> nestedGroup )
```
stsrki, line 45: Can we have IEnumerable here, so that we don't need to to `.AsEnumerable()`?

Source/Extensions/Blazorise.DataGrid/DataGrid.razor
David-Moreira: Added state tracking.

David-Moreira:
Current TODO  
- Docs
- Identing 
- Expanded State tracking is lost when sorting

stsrki:
What do you think we add some kind of icon to the groupable columns header? To indicate that it can be dragged.

image

stsrki:
I get the error when I navigate here

image

David-Moreira:
> I get the error when I navigate here
> 
> image

Oops since I changed the core implementation now it lacks a null check :D we should consider using the new nullable feature in c# it should help with these.

David-Moreira:
> What do you think we add some kind of icon to the groupable columns header? To indicate that it can be dragged.
> 
> image
> 
> PS. https://fontawesome.com/search?q=grip&o=r

Agree. And change the cursor to drag. Do you think we need more styling elsewhere?

stsrki:
> we should consider using the new nullable feature in c#

I hate it. Once you enable it, you need to mark eeeeeevrything as nullable.

stsrki:
> Do you think we need more styling elsewhere?

Add more padding on group Badges. Also change color to Secondary or Light.

David-Moreira:
> > we should consider using the new nullable feature in c#
> 
> I hate it. Once you enable it, you need to mark eeeeeevrything as nullable.

You need to mark stuff as nullable THAT make sense to be nullable otherwise you are not using the feature at all 😅 but yes for an ongoing huge project its a hummer we probably have like a thousand of warnings if we enable on blazorise 😅
But you can use a predecessor that activated it only on certain parts of the code.

stsrki:
I don't like that you need to mark reference types as nullable. And reference types are already nullable by themselves.

> But you can use a predecessor that activated it only on certain parts of the code.

In that case, we have a lot of code smell going around the codebase. Which again is too much for me.

David-Moreira:
> I don't like that you need to mark reference types as nullable. And reference types are already nullable by themselves.
> 
> > But you can use a predecessor that activated it only on certain parts of the code.
> 
> In that case, we have a lot of code smell going around the codebase. Which again is too much for me.

The point is to pretend they are not nullable by default. As such analyzers will help you and give you warnings if you're gonna do something that's gonna provoke a null ref exception. It's pretty good if you start a new project with that in mind.
But yea I'm not gonna keep arguing about this :D

Source/Extensions/Blazorise.DataGrid/DataGrid.razor
```diff
@@ -63,8 +63,13 @@
                             }
                             else
                             {
-                                <TableHeaderCell Draggable="@column.Groupable" DragStart="() => ColumnDragStart(column)" DragEnd="ColumnDragEnd"
+                                <TableHeaderCell Draggable="@(IsGroupableByColumn && column.Groupable)" DragStart="() => ColumnDragStart(column)" DragEnd="ColumnDragEnd"
                                                  Class="@column.HeaderCellClass" Style="@column.BuildHeaderCellStyle()" Cursor="@(Sortable && column.Sortable ? Cursor.Pointer : Cursor.Default)" Clicked="@(() => Sort( column ) )" TextAlignment="@column.HeaderTextAlignment" VerticalAlignment="@column.HeaderVerticalAlignment" Display="@column.HeaderDisplay" Background="@column.HeaderCellBackground()">
+                                    
+                                    @if ( IsGroupableByColumn && column.Groupable )
+                                    {
+                                        <Icon Name='("fas fa-grip-vertical")' Margin="@Blazorise.Margin.Is1.FromEnd" Style="cursor:move" />
```
stsrki, line 71: We should not use hardcoded icon names. Use `Name="IconName.SomeName"`.

Source/Extensions/Blazorise.DataGrid/_DataGridColumnDropZone.razor
```diff
@@ -1,17 +1,28 @@
...
+    {
+        <span>@Localizer.GetString("Drag a column header here to group by that column")</span>
     }
 </Div>
 
 @code {
+    /// <summary>
+    /// Gets or sets the DI registered <see cref="ITextLocalizer{_DataGridColumnDropZone}"/>.
+    /// </summary>
+    [Inject] protected ITextLocalizer<_DataGridColumnDropZone<TItem>> Localizer { get; set; }
```
stsrki, line 24: Make localization on the `DataGrid` level, `_DataGridColumnDropZone`.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor
```diff
@@ -5,17 +5,28 @@
...
+
 <CascadingValue Value="@this" IsFixed>
     @if ( IsPagerVisible && ( PagerPosition == DataGridPagerPosition.Top || PagerPosition == DataGridPagerPosition.TopAndBottom ) )
     {
         @paginationFragment
     }
     <Form>
+        @if ( IsGroupableByColumn )
+        {
+            <_DataGridColumnDropZone TItem="TItem"
```
stsrki, line 17: The drag area must be opt-in. Some users would like not to show it and just have grouping enabled explicitly on columns.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor
```diff
@@ -247,8 +275,60 @@
         }
     };
 
+    protected RenderFragment groupFragment( (int iteration, IEnumerable<GroupContext<TItem>> groupedData) x ) => __builder =>
+    {
+        @foreach ( var group in x.groupedData )
+        {
+            <TableRowGroup @key="group.Key" Expanded="@group.Expanded" ExpandedChanged="@group.SetExpanded" IndentTableCells="x.iteration">
```
stsrki, line 282: Do we want to show one empty cell when we have only one group? `IndentTableCells="x.iteration + 1"`

Source/Blazorise/Components/Table/TableRowGroup.razor
```diff
@@ -19,6 +19,15 @@
     @oncontextmenu="@OnContextMenuHandler"
     @oncontextmenu:preventDefault="@ContextMenuPreventDefault"
     @attributes="@Attributes">
+    @for ( int i = 0; i < IndentTableCells; i++ )
+    {
+        <td>
```
stsrki, line 24: Add empty cell class name here?

Source/Extensions/Blazorise.DataGrid/DataGrid.razor
David-Moreira: When there's only one group. There are no empty cells at all.
So the number of indented cells is x — 1 where x is the number of groups.

Would you think it looks better to do an indent cell per group in the header and row content but on the group row it is still x - 1?

Source/Blazorise/Components/Table/TableRowGroup.razor
David-Moreira: Suggestions for naming? just empty? b-cell-empty?

Source/Extensions/Blazorise.DataGrid/DataGrid.razor
David-Moreira: Agree. Tought about it but then forgot.
Should we do a specific parameter like GroupDrag or a more generic one like GroupEdit? Or GroupAssign?

Source/Extensions/Blazorise.DataGrid/DataGrid.razor
stsrki: From my investigation, some are indenting the first group; some are not. We can leave it as this for now. The first group without indent.

Source/Blazorise/Components/Table/TableRowGroup.razor
stsrki: image

Add a third class name and inherit for each provider. Can be named `table-group-indentcell`.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor
stsrki: ShowGrouping on DataGrid?

David-Moreira: Kk

Source/Blazorise/Components/Table/TableRowGroup.razor.cs
```diff
@@ -49,10 +50,21 @@ private void BuildRowCellClasses( ClassBuilder builder )
         builder.Append( ClassProvider.TableRowGroupCell() );
     }
 
+    /// <summary>
+    /// Builds the classnames for an indent cell.
+    /// </summary>
+    /// <param name="builder">Class builder used to append the classnames.</param>
+    private void BuildRowIndentCellClasses( ClassBuilder builder )
+    {
+        builder.Append( ClassProvider.TableRowGroupCell() );
```
stsrki, line 59: I think TableRowGroupCell should not be here.

David-Moreira: I could see it being there, it's still a regular group cell. But I'll remove then.

David-Moreira:
Feature is now technically ready. As we talked going to investigate if there's possible performance improvements we can do in the feature itself.
Otherwise we'll do those on a seperate PR.

Documentation/Blazorise.Docs/Pages/News/2023-01-15-release-notes-120.razor
```diff
@@ -197,22 +197,24 @@
 </Paragraph>
 
 <Heading Size="HeadingSize.Is3">
-    DataGrid 
+    DataGrid
 </Heading>
 
 <Paragraph>
-    A new <Code Tag>Groupable</Code> feature is introduced. You are now able to group the data in the <Code Tag>DataGrid</Code>.
+    The new grouping feature for <Link To="docs/extensions/datagrid">Blazorise DataGrid</Link> allows you to easily group and organize your data by specific columns. This can be useful for analyzing large sets of data and finding patterns or trends.
```
David-Moreira, line 204: might want to use these phrasing on the seo too :D

stsrki: Exactly haha

Source/Blazorise/wwwroot/dragDrop.js
```diff
@@ -1,4 +1,6 @@
 ﻿import { getRequiredElement } from "./utilities.js?v=1.1.5.0";
+let throttledDragHandler;
```
stsrki, line 2: Feels strange that you save the event reference on a module level. What happens when multiple draggable components are initialized and all point to the same variable?

David-Moreira: I guess if you'd keep "proper" state, you would push into an array, but from my tests this was enough for it to work right, and use the named function to remove the listener. Would you like to change it?

stsrki: I've changed it. Seems to be working as expected.

David-Moreira: Don't forget that to test it, you now need to set Draggable + (Drag or DragOver) Did you do that?

New implementation does look way better.

stsrki: Yes, I have tested with an example from your throttle PR.

stsrki:
I think the PR is ready. Anything else you want to add?

David-Moreira:
> I think the PR is ready. Anything else you want to add?

Nope, lgtm, excited to have this. :) And maybe we can now run the benchmarks you talked about?

stsrki:
> And maybe we can now run the benchmarks you talked about?

I will try to publish the preview NuGet tomorrow and then test it.

stsrki (approved):
As always, you did a great job :)

PR #4436, opened 2023-01-05
TreeView: ExpandAll, async loading and other improvements

### LINKED ISSUE #1233: Load TreeView child nodes on demand
I would like to load child nodes on demand from an API

### LINKED ISSUE #4108: Treeview - Expand all nodes in the TreeView by default

stsrki:
Closes #4108 
Closes #1233

**New APIs**
- `GetChildNodesAsync` and `HasChildNodesAsync`
- `ExpandAll` and `CollapseAll`
- `AutoExpandAll`

**New features**

- multiple selections with checkboxes
- internal refactor to hold tree states


Source/Extensions/Blazorise.TreeView/Internal/_TreeViewNode.razor.cs
```diff
@@ -0,0 +1,207 @@
...
+                } );
+            }
+        }
+
+        DirtyClasses();
+
+        await InvokeAsync( StateHasChanged );
+    }
+
+    private Action<TNode, NodeStyling> ResolveNodeStylingAction( Action<TNode, NodeStyling> action )
```
David-Moreira, line 114: Is this because you can't pass nulls around or something? Wouldn't it be better to pass a null and do nothing in the other side?

Source/Extensions/Blazorise.TreeView/Internal/_TreeViewNodeContent.razor
```diff
@@ -0,0 +1,13 @@
+﻿@typeparam TNode
+@inherits BaseComponent
+<div class="b-tree-view-node-title">
+    @if ( ParentTreeViewState.SelectionMode == TreeViewSelectionMode.Multiple )
+    {
+        <div class="b-tree-view-node-check">
+            <Check TValue="bool" Checked="@Checked" CheckedChanged="@OnCheckedChanged" />
```
David-Moreira, line 7: Did you have any problems in syncing stuff up? Check selection vs regular selection? Tested all providers? I believe Datagrid wasn't easy to get right at the time. But I can't recall the exact problems we had.

Source/Extensions/Blazorise.TreeView/Internal/_TreeViewNodeContent.razor.cs
```diff
@@ -1,113 +1,146 @@
...
+        else
+            builder.Append( $"{ClassProvider.BackgroundColor( nodeStyling.Background )} {ClassProvider.TextColor( nodeStyling.TextColor )} {nodeStyling.Class}" );
+
+        base.BuildClasses( builder );
+    }
+
+    protected Task OnClick()
+    {
+        if ( SelectionMode != TreeViewSelectionMode.Single )
+            return Task.CompletedTask;
```
David-Moreira, line 60: Are we disabling regular selection when multiple? I think people still like that the rows can be selected either on checkbox or row?

Source/Extensions/Blazorise.TreeView/TreeView.razor.cs
```diff
@@ -11,21 +17,130 @@ public partial class TreeView<TNode> : BaseComponent
...
+
+        if ( selectedNodeChanged )
+        {
+            await SelectedNodeChanged.InvokeAsync( treeViewState.SelectedNode );
+        }
+
+        if ( selectedNodesChanged )
+        {
+            await SelectedNodesChanged.InvokeAsync( treeViewState.SelectedNodes );
+        }
```
David-Moreira, line 70: If we are on `SetParametersAsync`, and either of these have been changed then it means the Parameter was already changed / mutated and there is no need to invoke these? We didn't really do any logic to mutate anything did we?

Source/Extensions/Blazorise.TreeView/TreeView.razor.cs
```diff
@@ -11,21 +17,130 @@ public partial class TreeView<TNode> : BaseComponent
...
+    /// Selects the node when in single selection mode.
+    /// </summary>
+    /// <param name="node">Node to select.</param>
     public void SelectNode( TNode node )
     {
-        SelectedNode = node;
+        if ( treeViewState.SelectionMode == TreeViewSelectionMode.Multiple )
+            return;
+
+        if ( treeViewState.SelectedNode.IsEqual( node ) )
```
David-Moreira, line 89: can be null?

Source/Extensions/Blazorise.TreeView/TreeView.razor.cs
```diff
@@ -11,21 +17,130 @@ public partial class TreeView<TNode> : BaseComponent
...
+        builder.Append( "b-tree-view" );
+
+        base.BuildClasses( builder );
+    }
+
+    /// <summary>
+    /// Selects the node when in single selection mode.
+    /// </summary>
+    /// <param name="node">Node to select.</param>
     public void SelectNode( TNode node )
```
David-Moreira, line 84: Why not task?

Source/Extensions/Blazorise.TreeView/TreeView.razor.cs
```diff
@@ -11,21 +17,130 @@ public partial class TreeView<TNode> : BaseComponent
...
+
+        DirtyClasses();
         InvokeAsync( StateHasChanged );
     }
 
+    /// <summary>
+    /// Toggles the checked state of the node when in multiple selection mode.
+    /// </summary>
+    /// <param name="node">Node to toggle.</param>
+    public void ToggleCheckNode( TNode node )
```
David-Moreira, line 104: Why not task?

David-Moreira (commented):
Only CR. I still need to test if you don't mind leaving it open.

Documentation/Blazorise.Docs/Pages/News/2023-01-15-release-notes-120.razor
```diff
@@ -183,4 +183,41 @@
...
+</Paragraph>
+
+<Paragraph>
+    <UnorderedList>
+        <UnorderedListItem>
+            <Code>GetChildNodesAsync</Code>: allows you to retrieve the child nodes of a specific node asynchronously.
+        </UnorderedListItem>
+        <UnorderedListItem>
+            <Code>HasChildNodesAsync</Code>: allows you to check if a specific node has child nodes asynchronously.
+        </UnorderedListItem>
```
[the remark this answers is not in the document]
stsrki: We will remove these two anyway in v2.0. GetChildNodes and HasChildNodes will become taskable.

Source/Blazorise.Tailwind/Check.razor
```diff
@@ -2,6 +2,9 @@
 @inherits Blazorise.Check<TValue>
 <Control Role="ControlRole.Check" Inline="@Inline">
     <input @ref="@ElementRef" id="@ElementId" type="checkbox" class="@ClassNames" style="@StyleNames" disabled="@Disabled" readonly="@ReadOnly" checked="@Checked" tabindex="@TabIndex" @onchange="@OnChangeHandler" @onkeydown="@OnKeyDownHandler" @onkeypress="@OnKeyPressHandler" @onkeyup="@OnKeyUpHandler" @onblur="@OnBlurHandler" @onfocus="@OnFocusHandler" @onfocusin="@OnFocusInHandler" @onfocusout="@OnFocusOutHandler" @attributes="@Attributes" />
-    <Label Type="LabelType.Check" For="@ElementId" Style="@Style" Cursor="@Cursor">@ChildContent</Label>
+    @if ( ChildContent is not null )
+    {
+        <Label Type="LabelType.Check" For="@ElementId" Style="@Style" Cursor="@Cursor">@ChildContent</Label>
+    }
```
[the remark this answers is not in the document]
stsrki: It seems that it is OK from what I can see. We might do the same for other providers and see how that will work. It can be usable on DataGrid for multi select where we also don't need the label.

Source/Extensions/Blazorise.TreeView/Internal/_TreeViewNode.razor.cs
```diff
@@ -0,0 +1,207 @@
...
+
+            await ExpandedNodesChanged.InvokeAsync( ExpandedNodes );
+
+            if ( nodeState.HasChildren )
+            {
+                var childNodes = GetChildNodesAsync is not null
+                    ? await GetChildNodesAsync( nodeState.Node )
+                    : GetChildNodes( nodeState.Node );
+
+                if ( !nodeState.Children.Select( x => x.Node ).AreEqual( childNodes ) )
```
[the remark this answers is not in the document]
stsrki: We want to initialize states only if child nodes have actually changed. Not if the nodes are loaded as that would reinitialize states every time.

Source/Extensions/Blazorise.TreeView/Internal/_TreeViewNode.razor.cs
```diff
@@ -0,0 +1,207 @@
...
+                {
+                    nodeState.Children = new();
+
+                    await foreach ( var childNodeState in childNodes.ToNodeStates( HasChildNodesAsync, HasChildNodes, ExpandedNodes.Intersect( childNodes ?? Enumerable.Empty<TNode>() ).Any() ) )
```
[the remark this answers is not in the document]
stsrki: It seems that your link indicates `Intersect` is faster than `Any`. Or am I wrong?

Source/Extensions/Blazorise.TreeView/Internal/_TreeViewNode.razor.cs
```diff
@@ -0,0 +1,207 @@
...
+    }
+
+    public async Task CollapseAll()
+    {
+        foreach ( var nodeState in NodeStates ?? Enumerable.Empty<TreeViewNodeState<TNode>>() )
+        {
+            if ( HasChildNodes( nodeState.Node ) )
+            {
+                if ( nodeState.Expanded )
+                    await ToggleNode( nodeState, false );
```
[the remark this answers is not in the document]
stsrki: To reuse the same `ToggleNode` method. I don't think we will have many problems with race conditions here. The `CollapseAll()` is meant to be run only once, usually per user click.

Source/Extensions/Blazorise.TreeView/Internal/_TreeViewNode.razor.cs
stsrki: You're right. This was done by the original TreeView creator, but it seems to not be necessary, so I have removed it.

Source/Extensions/Blazorise.TreeView/Internal/_TreeViewNodeContent.razor
stsrki: Yeah, I have checked with all providers, and it seems to work with the current code structure. I must admit it wasn't easy to do, though.

Source/Extensions/Blazorise.TreeView/Internal/_TreeViewNodeContent.razor.cs
stsrki: Well, it comes to my mind. But maybe later on, people would want to click on a node without making it checked. So I left it only to be selectable with a checkbox.

Source/Extensions/Blazorise.TreeView/TreeView.razor.cs
stsrki: This is an existing API. We can do it in v2.

Source/Extensions/Blazorise.TreeView/TreeView.razor.cs
stsrki: Changed to Task.

Source/Extensions/Blazorise.TreeView/TreeView.razor.cs
stsrki: `.IsEqual` will check for nulls.

Source/Extensions/Blazorise.TreeView/Internal/_TreeViewNode.razor
```diff
@@ -0,0 +1,60 @@
...
+        {
+            <div @key="@nodeState.Key">
+                @if ( nodeState.HasChildren )
+                {
+                    @if ( AutoExpandAll && !nodeState.Expanded && !nodeState.AutoExpanded )
+                    {
+                        nodeState.AutoExpanded = true;
+
+                        InvokeAsync( () => ToggleNode( nodeState ) );
+                    }
```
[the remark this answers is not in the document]
stsrki: Moved to OnInitializedAsync.

stsrki:
All comments addressed. You can check PR again.

Source/Extensions/Blazorise.TreeView/Internal/_TreeViewNode.razor.cs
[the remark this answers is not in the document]
David-Moreira: Reason why I searched and linked is because I believe any is more intuitive and readable here. As for performance considerations it is said that it depends on the data, plus intersect has an additional memory footprint. We could possibly benchmark but I guess it's no big problem to maintain one or the other.

Source/Extensions/Blazorise.TreeView/TreeView.razor.cs
David-Moreira: That's not my point. My point is that we don't need to call the Changed at all for either. Because at this time, the change has already been done and the consumer already knows about it.

Source/Extensions/Blazorise.TreeView/TreeView.razor.cs
David-Moreira: Ahhh! Haha I forgot! 😅

Source/Extensions/Blazorise.TreeView/TreeView.razor.cs
stsrki: 🤔 I guess you're right.

stsrki: We need these. If the user updates `SelectedNode` though the code and through the parameter, maybe he also listens for changes with `SelectedNodeChanged`. If we don't do this, then the user will never get the notification.

David-Moreira: Well exactly. He updated he doesn't need notification, it's redundant I believe? The notification is there due to the declarative nature of blazor. If you go to a regular Blazor input, a text for example and programmatically set text to something does it notify changed back? I doubt so, would be interesting to test. 

I believe this will trigger an unnecessary re-render that's why I was advocating to not having it. But if you feel we should still treat as notification worthy.

stsrki: I still think notification must be invoked.

stsrki: I admit the defeat

```razor
<div>
    <Button Color="Color.Primary" Clicked="@(()=>Description = "TEST")">TEST</Button>
</div>
<div>
    <input type="text" value="@Description" @oninput="@OnInput" />
</div>
<div>
    Value : @Description
</div>
@code {
    string Description;

    Task OnInput( ChangeEventArgs eventArgs )
    {
        Description = eventArgs?.Value?.ToString();
        Console.WriteLine( $"onchanged: {Description}" );
        return Task.CompletedTask;
    }
}
```

David-Moreira: haha noone's defeating anyone, this ain't a fight 🗡️🗡️ 

I actually meant for us to test the Native blazor input component, that'd be a closer comparisition I believe. But I believe the result would be the same.

stsrki: > I actually meant for us to test the Native blazor input component

Can you try it?

stsrki: OK I have tried it. The event is not called after the binded value is changed.

```razor
<div>
    <Button Color="Color.Primary" Clicked="@(()=>exampleModel.Name = "TEST")">TEST</Button>
</div>
<div>
    <EditForm Model="@exampleModel">
        <DataAnnotationsValidator />

        <MyInputText id="name" Value="@exampleModel.Name" ValueChanged="@OnInput" ValueExpression="@(()=>exampleModel.Name)" />
    </EditForm>

</div>
<div>
    Value : @exampleModel.Name
</div>

@code {
    public class ExampleModel
    {
        [Required]
        [StringLength( 10, ErrorMessage = "Name is too long." )]
        public string? Name { get; set; }
    }
    private ExampleModel exampleModel = new();
    Task OnInput( string value )
    {
        exampleModel.Name = value;
        Console.WriteLine( $"onchanged: {exampleModel.Name}" );
        return Task.CompletedTask;
    }
}
```

and

**MyInputText.razor**

```razor
@inherits Microsoft.AspNetCore.Components.Forms.InputText
<input @attributes="@AdditionalAttributes" class="@CssClass" @bind="@CurrentValueAsString" @bind:event="oninput" />
```

Source/Extensions/Blazorise.TreeView/Internal/_TreeViewNode.razor
```diff
@@ -27,9 +27,9 @@
                 }
 
                 <_TreeViewNodeContent TNode="TNode" NodeState="@nodeState"
-                                      NodeStyling="@(ResolveNodeStylingAction(NodeStyling))"
+                                      NodeStyling="@NodeStyling"
                                       SelectionMode="@SelectionMode"
-                                      SelectedNodeStyling="@(ResolveSelectedNodeStylingAction(SelectedNodeStyling))">
+                                      SelectedNodeStyling="@SelectedNodeStyling">
```
David-Moreira, line 25: The commit only has this replacement, is there nothing else we need to do? Because previously it would have an Action with an empty return? If everything ok in case these are null?

Source/Extensions/Blazorise.TreeView/Internal/_TreeViewNode.razor.cs
```diff
@@ -15,6 +15,21 @@ public partial class _TreeViewNode<TNode> : BaseComponent
...
+    {
+        foreach ( var nodeState in NodeStates ?? Enumerable.Empty<TreeViewNodeState<TNode>>() )
+        {
+            if ( nodeState.HasChildren && AutoExpandAll && !nodeState.Expanded && !nodeState.AutoExpanded )
+            {
+                nodeState.AutoExpanded = true;
+
+                await ToggleNode( nodeState );
+            }
+        }
```
David-Moreira, line 28: Could possible wrap in a method like `AutoExpandNodes` or something just so it's a bit more descriptive of what's it doing, Otherwise I guess this is indeed better.

David-Moreira (changes requested):

Source/Extensions/Blazorise.TreeView/Internal/_TreeViewNode.razor
stsrki: We don't need it as far as I can tell. I really don't know why the person who created the first TreeView version did that but as long as we don't invoke on an event that is null we are good.

stsrki:
I have done all the changes.

Source/Extensions/Blazorise.TreeView/Internal/_TreeViewNode.razor.cs
```diff
@@ -44,16 +44,9 @@ private async Task ToggleNode( TreeViewNodeState<TNode> nodeState, bool refresh
...
             if ( !ExpandedNodes.Contains( nodeState.Node ) )
             {
-                nodeAdded = true;
                 ExpandedNodes.Add( nodeState.Node );
-            }
-
-            if ( nodeAdded )
-            {
                 await ExpandedNodesChanged.InvokeAsync( ExpandedNodes );
             }
```
David-Moreira, line 59: I'm super confused, this went back to the original version no? What I mean is 
```
if ( !ExpandedNodes.Contains( nodeState.Node ) )
            {
                ExpandedNodes.Add( nodeState.Node );
                await ExpandedNodesChanged.InvokeAsync( ExpandedNodes );
            }
```

instead of

```
if ( !ExpandedNodes.Contains( nodeState.Node ) )
            {
                ExpandedNodes.Add( nodeState.Node );
            }
            await ExpandedNodesChanged.InvokeAsync( ExpandedNodes );
```

David-Moreira:
Ok, gotta take some time to actually test eventually if you don't mind waiting. :P

Source/Extensions/Blazorise.TreeView/Internal/_TreeViewNode.razor.cs
stsrki: > I'm super confused, this went back to the original version no? What I mean is
> 
> ```
> if ( !ExpandedNodes.Contains( nodeState.Node ) )
>             {
>                 ExpandedNodes.Add( nodeState.Node );
>                 await ExpandedNodesChanged.InvokeAsync( ExpandedNodes );
>             }
> ```
> 
> instead of
> 
> ```
> if ( !ExpandedNodes.Contains( nodeState.Node ) )
>             {
>                 ExpandedNodes.Add( nodeState.Node );
>             }
>             await ExpandedNodesChanged.InvokeAsync( ExpandedNodes );
> ```

I did that

image

David-Moreira:
Arrows don't seem right in TailWind:
image

David-Moreira:
Are we sure we don't want to still have this selection in multiple mode?
image

David-Moreira:
Not much to say, it's working really good. :)

David-Moreira:
Could we get a working example for this? I'm sure users will appreciate it. 
With this, we do not need `ReadData`, right?
image

stsrki:
> Arrows don't seem right in TailWind, please there's no indentation image

Refresh browser cache.

stsrki:
> Could we get a working example for this? I'm sure users will appreciate it. With this, we do not need `ReadData`, right? image
> 
> PS: And make this PR close it, if applicable, #1233

That's right. ReadData is not needed.

stsrki:
> Are we sure we don't want to still have this selection in multiple mode? image

Pretty sure. If someone wants it we can consider it.

David-Moreira:
> Refresh browser cache.

You're right. My bad. Then it's also looking good. :)

David-Moreira (approved):

Issue #4440, opened 2023-01-06
Add dependabot notificaftions for updated js libs


GeertvanHorrik:
**Is your feature request related to a problem? Please describe.**

It's about outdated js library versions.

**Describe the solution you'd like**

For internal projects, I've set up dependabot for the js libs. At least there is a signal when a new version is shipped so it doesn't always leave the library with outdated libs.

### Additional context

Happy to provide a PR that:

1. Uses npm to manage the js library
2. Build steps to copy the correct scripts from the npm lib folder into the source folders
3. Automatically checks for updates and create PRs using dependabot

We use this in our own projects and we really like this workflow since it's easy to "forget" about checking new updates for js libraries (in a .NET world).

stsrki:
It would be nice, but I don't believe that it would be possible for all JS libraries. For some libraries, we had to modify them slightly so that they would work properly as JS modules. If in the meantime, those libraries are fixed then it is great. 

Although I would really move away from NPM completely 😅

GeertvanHorrik:
Would you like me to work on a PR?

stsrki:
Try it on a small scale first. On one or two JS libraries to see how it goes. Then submit a WIP PR so we can look and see how to proceed. Is that OK?

stsrki:
I have merged the PR. We'll see how it goes before moving forward with other changes.

Thank you for your help in making it happen!

stsrki:
PS. it worked great once I merged and got reports. I have patched all the reported files to the latest supported, nonbreaking version, and so far, all is working fine.

Thank you once again @GeertvanHorrik

GeertvanHorrik:
Glad you like it :)

Shall I do one for the NuGet ecosystem as well? Shouldn't be this big

stsrki:
> Shall I do one for the NuGet ecosystem as well? Shouldn't be this big

I would appreciate it ☺

PR #4445, opened 2023-01-07
Dependabot for js dependencies

GeertvanHorrik:
As discussed in #4440 , this is a PoC to add dependabot for the js dependencies.

_Note that this is a PoC to show how it can be used and is not a full solution._

- [x] Add dependabot configuration file so it is aware of the npm projects to check
- [x] Add npm package.config for the _Blazorise_ project. I tried finding the correct packages, but **please verify that these are the correct packages / versions**
- [x] Add "copy js dependencies" in debug mode step so the software is always shipped with the intended packages

### DIFF
--- a/Source/Blazorise/Blazorise.csproj
+++ b/Source/Blazorise/Blazorise.csproj
@@ -41,8 +41,8 @@
 
   <!-- Note: only copy files in debug mode, we don't want automatic actions in release mode / during publishing -->
 
-  <Target Name="PreBuild - Copy javascript dependencies" 
-		  AfterTargets="NpmInstall" 
+  <Target Name="PreBuild - Copy javascript dependencies"
+		  AfterTargets="NpmInstall"
 		  Condition="'$(Configuration)'=='Debug'">
     <!-- In case original, non-modified libraries are used, it's recommended to always overwrite with the correct version -->
     <!--<Exec Command="xcopy /y &quot;$(ProjectDir)node_modules\autonumeric\dist\autoNumeric.min.js&quot; &quot;$(ProjectDir)wwwroot\vendors\autoNumeric.js&quot;" />
@@ -50,10 +50,10 @@
 	<Exec Command="xcopy /y &quot;$(ProjectDir)node_modules\inputmask\dist\inputmask.min.js&quot; &quot;$(ProjectDir)wwwroot\vendors\inputmask.js&quot;" />-->
   </Target>
 
-  <Target Name="NpmInstall" 
-		  Inputs="package.json" 
-		  Outputs="node_modules/.install-stamp" 
-		  BeforeTargets="BeforeBuild" 
+  <Target Name="NpmInstall"
+		  Inputs="package.json"
+		  Outputs="node_modules/.install-stamp"
+		  BeforeTargets="BeforeBuild"
 		  Condition="'$(Configuration)'=='Debug'">
     <!--
       Use npm install or npm ci depending on RestorePackagesWithLockFile value.
@@ -69,5 +69,5 @@
     <!-- Write the stamp file, so incremental builds work -->
     <Touch Files="node_modules/.install-stamp" AlwaysCreate="true" />
   </Target>
-	
+
 </Project>
--- a/Source/Blazorise/Blazorise.csproj
+++ b/Source/Blazorise/Blazorise.csproj
@@ -13,11 +13,13 @@
 
   <ItemGroup>
     <Content Remove="compilerconfig.json" />
+    <Content Remove="package.json" />
   </ItemGroup>
 
   <ItemGroup>
     <None Include="..\..\LICENSE.md" Pack="true" Visible="false" PackagePath="" />
     <None Include="..\..\NuGet\Blazorise.png" Pack="true" Visible="false" PackagePath="" />
+    <None Include="package.json" />
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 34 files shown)


GeertvanHorrik:
As an extra comment:

even when using customized versions of these libraries, it's important to be aware of updates. Therefore you could consider not using the "copy files part" but still be made aware of updates by dependabot.

stsrki:
I appreciate the effort. But I think something else will be needed. As I already mentioned, this PR solution will override our existing JS files on each build, which are slightly modified. So it breaks the code.

GeertvanHorrik:
Yes, so that's the 2nd comment I made: even if you decide not to auto-copy the dependencies, it's good to be aware of changes and updates. So if you remove the copy steps from the csproj, you will still have notifications, but it won't overwrite the files you use. But it's a signal that something (probably) needs an update.

stsrki:
Notification can be good to have. I agree with that.

GeertvanHorrik:
Pushed an update to the PoC to disable copying of files (but still allows dependabot to create pull requests to make you aware of updates). Also added some more examples, hopefully this gives you an idea what's possible and will make it "easier" to stay up to date with js dependencies and updates.

Source/Blazorise/Blazorise.csproj
```diff
@@ -39,4 +39,35 @@
     <CompilerGeneratedFilesOutputPath>__SOURCEGENERATED__</CompilerGeneratedFilesOutputPath>
   </PropertyGroup>
 
+  <!-- Note: only copy files in debug mode, we don't want automatic actions in release mode / during publishing -->
+
+  <Target Name="PreBuild - Copy javascript dependencies"
```
stsrki, line 44: Do we need all this NPM stuff in Blazorise.csproj? Can it work without it, just with dependabot.yml?

Source/Blazorise/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "version": "1.0.0",
+  "name": "blazorise",
+  "private": true,
+  "devDependencies": {},
+  "dependencies": {
+    "autonumeric": "4.6.0",
```
stsrki, line 7: I'm not familiar with `dependabot`. Should this be defined with a wildcard? eg `4.6.*`?

Source/Blazorise/Blazorise.csproj
GeertvanHorrik: You don't need it, it's what we use in our projects because we use unmodified versions. But you can remove it if you never use unmodified javascript versions.

Source/Blazorise/package.json
GeertvanHorrik: This is actually an NPM package.json. Dependabot will, depending on the infrastructure (nuget, npm, etc) check for updates in the right package manager.

The real "magic" for dependabot is in the .github directory. This file contains the actual versions being used by Blazorise so it knows whether you are up to date or not.

stsrki: And it knows about the major, minor, and patch versioning?

GeertvanHorrik: Yes. The "only" thing you need to do is make sure the initial versions are correct in package.json and your customized files so they are not out of sync.

GeertvanHorrik:
A great example of what happened is that all examples of Blazorise use luxon 1.27.0. However, a security vulnerability was found and everyone is recommended to update to 1.28.1.

This would have been noted because:

1. Dependabot will auto explain that the update is available
2. GitHub will do dependency security vulnerability scanning via dependabot (but it needs a package configuration for this)

stsrki:
What kind of vulnerability? I'm curious. Can't find it.

GeertvanHorrik:
See https://github.com/advisories/GHSA-3xq5-wjfh-ppjc

stsrki:
> This causes a noticeable slowdown for inputs with lengths above 10k characters

I guess it is highly unlikely to happen but yeah, I agree it is better to upgrade.

GeertvanHorrik:
Totally agree with you about the impact. I am just trying to prove that it's very handy to have these notifications. At least you are aware and you can decide whether it will have an impact or not. Right now it feels too many hidden dependencies becoming stale.

stsrki:
I have removed changes from the .csproj files and left only the `package.json` files. If you think it will work, I can merge the PR so the action can run.

stsrki:
It would be great if you could update this PR and add package.json for other project.

GeertvanHorrik:
I'll add some notes on the side here:

* popper.js seems to be replaced by floating-ui (source: https://github.com/floating-ui/floating-ui)

GeertvanHorrik:
I think this PR is done. If you are interested, I can prepare another one for dependabot for .NET packages (but only after this one is approved & merged).

stsrki:
Thanks.

> I can prepare another one for dependabot for .NET packages

Can you explain what .NET packages?

GeertvanHorrik:
When creating a new PR for NuGet package references, I would:

1. Use * on .NET package references (e.g. 7.0.*) so you don't have to update every Patch Tuesday
2. Ensure these (for example) packages are kept up to date (at least notify you):
* FluentValidation
* Moq
* etc

stsrki:
> 1. Use * on .NET package references (e.g. 7.0.*) so you don't have to update every Patch Tuesday

As long as this works on the `master` branch and not on production branches, eg `rel-1.x`, then I guess it is fine. Forcing users to go from 7.0.0 to 7.0.1 is a breaking change and it forces them to update their systems. We want to avoid that.

> 2. Ensure these (for example) packages are kept up to date (at least notify you):

This part is perfectly fine to update as it shouldn't break Blazorise API or get in the way for the user.

stsrki:
@David-Moreira what do you think of this PR?

GeertvanHorrik:
> As long as this works on the master branch and not on production branches, eg rel-1.x, then I guess it is fine. Forcing users to go from 7.0.0 to 7.0.1 is a breaking change and it forces them to update their systems. We want to avoid that.

Since this wildcard is only added on the patch level, it should never be a breaking change (e.g. these are important security fixes, not behavioral changes). This will mostly only contain security fixes released by Microsoft.

If people really want to, they can enforce an older version (e.g. 7.0.0) on their project, but I think this is really unwise.

stsrki (approved):

PR #4448, opened 2023-01-08
DataGrid: add EditFieldColumnSize

### LINKED ISSUE #4365: On the DataGrid component, is there a 'FormFieldColumnSize' parameter?
On the DataGrid component, is there a 'FormFieldColumnSize' parameter similar to the 'PopupFieldColumnSize' that would allow fields on the Form Edit Mode to be set to a custom width instead of defaulting to 6 or half?

I saw the PopupFieldColumnSize parameter used in one of the demo samples and assumed there would be something similar for the Form Edit Mode but there doesn't seem to be.

Thanks!

stsrki:
Closes #4365

### DIFF
--- a/Source/Extensions/Blazorise.DataGrid/_DataGridModal.razor
+++ b/Source/Extensions/Blazorise.DataGrid/_DataGridModal.razor
@@ -31,7 +31,7 @@
                     <Fields>
                         @foreach ( var column in OrderedEditableColumns )
                         {
-                            <Field ColumnSize="@column.PopupFieldColumnSize" Padding="Blazorise.Padding.Is2.OnX.Is2.OnY">
+                            <Field ColumnSize="@(column.EditFieldColumnSize ?? column.PopupFieldColumnSize)" Padding="Blazorise.Padding.Is2.OnX.Is2.OnY">
                                 <FieldLabel>
                                     @column.Caption
                                 </FieldLabel>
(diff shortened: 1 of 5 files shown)


Source/Extensions/Blazorise.DataGrid/_DataGridModal.razor
```diff
@@ -31,7 +31,7 @@
                     <Fields>
                         @foreach ( var column in OrderedEditableColumns )
                         {
-                            <Field ColumnSize="@column.PopupFieldColumnSize" Padding="Blazorise.Padding.Is2.OnX.Is2.OnY">
+                            <Field ColumnSize="@(column.EditFieldColumnSize ?? column.PopupFieldColumnSize)" Padding="Blazorise.Padding.Is2.OnX.Is2.OnY">
```
David-Moreira, line 34: This might not quite result in correct reverse compatibility. Better do some sync logic, I believe we need an internal field for that. Since both Edit and Popup have defined values, it's currently a breaking change for people that are currently using `PopupFieldColumnSize`

David-Moreira (changes requested):

Source/Extensions/Blazorise.DataGrid/_DataGridModal.razor
David-Moreira: Or just introduce `EditFieldColumnSize` as  null so the coalescing operators act ok.

stsrki: How did I not notice the default values?! 🤦‍♂️

I have removed them, and I think the logic is now OK.

David-Moreira (approved):
LGTM

PR #4452, opened 2023-01-09
Autocomplete: improvements

### LINKED ISSUE #3727: Set Background Color in Autocomplete (single and multiple)
Hello, so far I really love the autocomplete component and I wonder is it possible to add option to set the autocomplete background color like in TextEdit?

I know that the background color is used for autocomplete values in multiple mode, but it would be nice if there are options to adjust the background color both for the TextEdit(autocomplete) itself and the value in multiple mode

Thank you very much

### LINKED ISSUE #3791: Match highlighting for Autocomplete component
When using an `Autocomplete `component there should be the possibility to enable match highlighting for Autocomplete items.
When a search string is entered, and matches are shown, the matching parts should be highlighted when this Attribute is set.

Obviously this would only work if no `<ItemContent>` is given. To mitigate that the current search string could be entered into the `Context `of the `Autocomplete`, then the match highlighting could be done "manually". Still this would be complicated.

The matching parts of the `Autocomplete `items could be made bold.

### Additional context
image

Currently there is no possibility to do match highlighting manually via `ItemContent `and matching the search string again to the data, as the `SearchText` is not available afaik, as only `SearchChanged ` is exposed but not `SearchText`. Also one would essentially have to implement the matching again, and then it is also executed twice, only to get the highliting. Not so nice, if even possible.

stsrki:
Closes #3791
Closes #3727

Also fixed dropodown classnames.

### DIFF
--- a/Source/Extensions/Blazorise.Components/Autocomplete.razor.cs
+++ b/Source/Extensions/Blazorise.Components/Autocomplete.razor.cs
@@ -1248,6 +1248,16 @@ public partial class Autocomplete<TItem, TValue> : BaseAfterRenderComponent, IAs
     [Parameter] public bool HighlightSearch { get; set; }
 
     /// <summary>
+    /// Defines the background color of the search field.
+    /// </summary>
+    [Parameter] public Background SearchBackground { get; set; }
+
+    /// <summary>
+    /// Defines the text color of the search field.
+    /// </summary>
+    [Parameter] public TextColor SearchTextColor { get; set; }
+
+    /// <summary>
     /// Currently selected items values.
     /// Used when multiple selection is set.
     /// </summary>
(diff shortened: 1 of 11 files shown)


Source/Extensions/Blazorise.Components/Autocomplete.razor.cs
```diff
@@ -1242,6 +1242,11 @@ public string Search
     /// </summary>
     [Parameter] public EventCallback<string> SearchChanged { get; set; }
 
+    /// <summary>
+    /// If true, the searched text will be highlighted in the dropdown list items based on <see cref="Search"/> value.
+    /// </summary>
+    [Parameter] public bool HighlightSearch { get; set; }
```
David-Moreira, line 1248: If user uses ItemContent, he looses Highlight capabilities right, basically he should just used the Highlighter himself, everything should work ok, right?

David-Moreira (approved):
LGTM, good feature.

Source/Extensions/Blazorise.Components/Autocomplete.razor.cs
stsrki: Yes, that is right. We have no way of knowing what the user is doing in the template. So he needs to highlight manually.

PR #4453, opened 2023-01-10
Modal: dynamically calculate ZIndex

### LINKED ISSUE #3878: Multiple Modals/ MessageService on top of Modal should  "Hide" the Modal beneath
**Describe the bug**
Message Service/Modal won't hide a Modal beneath.
**To Reproduce**
Open a Modal and then an MessageService.Confirm

**Expected behavior**
See right screenshot.

**Screenshots**
Left how it looks. Right how it should look(made with paint.net).
2022-06-09 16_03_41-is jpg - IrfanView

stsrki:
Closes #3878

The ZIndex is calculated based on the difference between zindex of the modal and backdrop. The difference is then raised based on the open index to ensure the new index is always higher than the previous one.

Also, I fixed some minor bugs with the styles `;` delimiter.

### DIFF
--- a/Source/Blazorise.AntDesign/AntDesignStyleProvider.cs
+++ b/Source/Blazorise.AntDesign/AntDesignStyleProvider.cs
@@ -7,13 +7,25 @@ public class AntDesignStyleProvider : StyleProvider
 {
     #region Modal
 
+    public override int DefaultModalZIndex => 1000;
+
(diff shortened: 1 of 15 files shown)


David-Moreira (approved):
LGTM. Maybe SharedContext is not Thread safe, but I don't think it matters for modals.

PR #4455, opened 2023-01-11
DataGrid: better control over header, group and filter cells

### LINKED ISSUE #3884: Responsive hiding columns of a Datagrid
**Describe the bug**
I want to hide some columns of a DataGrid on mobile devices.
For that I use:
```           
<DataGridColumn TItem="DeviceVM"
             Field="@nameof(DeviceVM.Side)"
              Caption="Kant"
              Display="Display.None.OnMobile.TableCell.OnTablet">
```
This makes the content of the grid hide for small screens and show for larger screens. That's doing well. But the header of the grid is **not** hidden together with the content.
I see that the content cell gets decorated with the class: "d-sm-table-cell", which does the trick, but the header is **not** decorated with this class. Since the DataGridColumn and the CaptionTemplate component do not allow to manually add a class I can't get a workaround very easy.

**My request**
Can you make the component to also decorate the header with the appropriate class?

stsrki:
Closes #3884

I have introduced new and updated existing API for the group, header, and filter cells. I also added a few "determine" properties that check for those new APIs if they are defined and then fall back to default. For example, ` => HeaderTextAlignment ?? TextAlignment;`


Source/Extensions/Blazorise.DataGrid/_DataGridAggregateRow.razor
```diff
@@ -20,7 +20,7 @@
         }
         else
         {
-            <TableRowCell Class="@column.GroupCellClass" Style="@column.BuildGroupCellStyle()" TextAlignment="@column.TextAlignment" VerticalAlignment="@column.VerticalAlignment" Display="@column.Display" Flex="@column.Flex" Gap="@column.Gap">
+            <TableRowCell Class="@column.GroupCellClass" Style="@column.BuildGroupCellStyle()" TextAlignment="@column.GroupCellTextAlignment" VerticalAlignment="@column.GroupCellVerticalAlignment" Display="@column.GroupCellDisplay" Flex="@column.GroupCellFlex" Gap="@column.GroupCellGap">
```
David-Moreira, line 23: So the **Group** prefix is to be applied on **Aggregates**? I think it can be mistaken with the new group feature we are working on, isn't it more clear to straight up use the **Aggregate** prefix?

David-Moreira (changes requested):

Source/Extensions/Blazorise.DataGrid/_DataGridAggregateRow.razor
stsrki: You're right. I forgot about the new group feature. I will change it.

stsrki: Done. Renamed to Aggregate. And made existing GroupCellClass and GroupCellStyle obsolete.

David-Moreira (approved):

Issue #4460, opened 2023-01-13
Datagrid Sort behaviour regression


David-Moreira:
@stsrki Regression for Sorting on master.
I believe there was a non sorted state right? It now only switches between asc/desc, I believe it might have been added here : #4434

@injDakov would you have time to take a look at it?

stsrki:
Yes, there should be a non-sorting state when looping the states.

injDakov:
@David-Moreira 
My changes here only give the opportunity to reverse the first direction of sorting. 

Non-Sorting (on init) -> ASC (first click)-> DESC (second click) and etc
or 
Non-Sorting (on init) -> DESC (first click)-> ASC (second click) and etc

David-Moreira:
But does it go back to the non sorting again?
Ie:
Non-sorting -> asc -> desc -> non-sorting so on...

stsrki:
As David mentione need to go

Non sorting -> asc -> desc -> non sorting

and in reverse

Non sorting -> desc -> asc -> non sorting

injDakov:
I got it.
Will prepare a new PR with a fix for this in one hour.

Need to remove these lines https://github.com/Megabit/Blazorise/pull/4434/files#diff-6d764d902c8a96989eb7ab34c9aa0681559b3a1c059e96fffc62cae2171baf16R29

injDakov:
Here is it: https://github.com/Megabit/Blazorise/pull/4464

David-Moreira:
Closing issue. Thanks alot for taking the time @injDakov

PR #4464, opened 2023-01-13
#4460 Datagrid Sort behaviour regression

injDakov:
Improvements into NextDirection() to keep again the sort direction sequence:

- "Non sorting -> asc -> desc -> non sorting" 
or
- "Non sorting -> desc -> asc -> non sorting".

[#4460](https://github.com/Megabit/Blazorise/issues/4460)

### DIFF
--- a/Source/Extensions/Blazorise.DataGrid/ExtensionMethods.cs
+++ b/Source/Extensions/Blazorise.DataGrid/ExtensionMethods.cs
@@ -25,9 +25,9 @@ public static class ExtensionMethods
             case SortDirection.Default:
                 return isReverse ? SortDirection.Descending : SortDirection.Ascending;
             case SortDirection.Ascending:
-                return SortDirection.Descending;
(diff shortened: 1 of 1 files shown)


stsrki:
I'm trying to test this. I've set the `ReverseSorting="true"`, but when I run and debug, it always stays at `false`.

image

injDakov:
@stsrki If passing the attribute's value is OK in the master branch, there is no reason to have an issue in this branch. 

How did you test it? 

Here is the field with it I tested it.

https://github.com/Megabit/Blazorise/blob/master/Demos/Blazorise.Demo/Pages/Tests/DataGrid/DataInMemoryPage.razor#L36

And for me works fine.

stsrki:
That's precisely what I did. Then run in VS in debug mode. But for some reason, ReverseSorting is ignored.

stsrki (approved):

PR #4473, opened 2023-01-16
DataGrid PageSelectorTemplate And PageSizesTemplate

### LINKED ISSUE #4295: Datagrid pagination : "PageSelectorTemplate"
> Changes are:
> 
> * Consider globally defined Theme's `PaginationOptions.Size` if `PagerOptions.ButtonSize` is not defined.
>   
>   * If neither are defined, the default is applied, ofc
> * The "Page Selector" which only shows below Tablet, should also consider the defined Size.

Sounds good to me.

> I was thinking, should we provide a Template for the "PageSelector" in v1.2?

We already have all other page templates so it is a good idea to have PageSelectorTemplate also.

_Originally posted by @stsrki in https://github.com/Megabit/Blazorise/issues/4290#issuecomment-1313347716_

David-Moreira:
Closes #4295


stsrki (approved):
LGTM

PR #4474, opened 2023-01-17
Markdown - Able to provide uploadUrl after the actual Upload has happened

### LINKED ISSUE #4335: Markdown : Able to set upload url after processing a file
Example print :
OnImageUploadedEnded is called in the exact moment upload is ended, **but before an URL can be set,** for example, as a result of a remote api call that returns the UploadedUrl
image

Investigate:
image

Or @stsrki's suggestion with TaskCompletionSource

David-Moreira:
Closes #4335

As a brief reminder, the problem here, is that when uploading images on `Markdown`, you can't set the `File.UploadUrl` before the internal `JSModule.NotifyImageUploadSuccess` (notifies markdown of image success + url) occurs, since it occurs as part of the final process of our file write operations (`FileEndedEvent`).

I don't think there's a good way to do this without introducing any new api. You could probably do some tricks with delays, etc...? But that doesn't seem right. Better have an explicit api in my opinion.
Do you like this one? As long as we properly document it isn't too hard for the user to use. Also, was this where you were going with the `TaskCompletionSource`? I couldn't find another way.

Testing code:
```
<Markdown ImageUploadChanged="Changed" ImageUploadEnded="ImageUploadEnded">

</Markdown>

@code {
    private string result = string.Empty;
    private async Task Changed( FileChangedEventArgs e )
    {
        var file = e.Files[0];
        file.UploadUrlCallback = new TaskCompletionSource<string>();
        result = await Upload( file );
        file.UploadUrlCallback.SetResult(result);
    }

    private async Task<string> Upload( IFileEntry file )
    {
        var stream = new MemoryStream();
        await file.WriteToStreamAsync( stream );
        return "xpto";
    }

    Task ImageUploadEnded( FileEndedEventArgs e )
    {
        Console.WriteLine( "Ended" );
        if ( e.Success )
        {
            e.File.UploadUrl = result;
        }
        else
        {
            e.File.UploadUrl = "FAIL";
        }
        return Task.CompletedTask;
    }
}
```

### DIFF
--- a/Source/Extensions/Blazorise.Markdown/Markdown.razor.cs
+++ b/Source/Extensions/Blazorise.Markdown/Markdown.razor.cs
@@ -270,15 +270,18 @@ public partial class Markdown : BaseComponent,
     [JSInvokable]
     public async Task NotifyImageUpload( FileEntry file )
     {
+        file.FileUploadEndedCallback = new();
+
         // So that method invocations on the file can be dispatched back here
         file.Owner = (IFileEntryOwner)(object)this;
 
         if ( ImageUploadChanged is not null )
             await ImageUploadChanged.Invoke( new( file ) );
 
+        file.FileUploadEndedCallback.SetResult();
         await InvokeAsync( StateHasChanged );
     }
-
+    private TaskCompletionSource taskCompletionSource;
     /// <inheritdoc/>
     public Task UpdateFileStartedAsync( IFileEntry fileEntry )
     {
@@ -306,7 +309,7 @@ public partial class Markdown : BaseComponent,
         }
 
 
-        if ( fileEntry.UploadUrlCallback is null )
+        if ( fileEntry.FileUploadEndedCallback is null )
         {
             await JSModule.NotifyImageUploadSuccess( ElementId, fileEntry.UploadUrl ?? string.Empty );
         }
@@ -315,8 +318,8 @@ public partial class Markdown : BaseComponent,
 #pragma warning disable CS4014 // We want to let execution complete but wait for TaskCompletionSource on the background.
             Task.Run( async () =>
             {
-                var uploadUrl = await fileEntry.UploadUrlCallback.Task;
-                await InvokeAsync( async () => await JSModule.NotifyImageUploadSuccess( ElementId, uploadUrl ?? string.Empty ) );
+                await fileEntry.FileUploadEndedCallback.Task;
+                await InvokeAsync( async () => await JSModule.NotifyImageUploadSuccess( ElementId, fileEntry.UploadUrl ?? string.Empty ) );
             } );
 #pragma warning restore CS4014 
         }
--- a/Source/Extensions/Blazorise.Markdown/Markdown.razor.cs
+++ b/Source/Extensions/Blazorise.Markdown/Markdown.razor.cs
@@ -297,32 +297,28 @@ public partial class Markdown : BaseComponent,
     }
 
     /// <inheritdoc/>
-    public async Task UpdateFileEndedAsync( IFileEntry fileEntry, bool success, FileInvalidReason fileInvalidReason )
+    public Task UpdateFileEndedAsync( IFileEntry fileEntry, bool success, FileInvalidReason fileInvalidReason )
     {
-        if ( ImageUploadEnded is not null )
-            await ImageUploadEnded.Invoke( new( fileEntry, success, fileInvalidReason ) );
-
-        if ( !success )
-        {
-            await JSModule.NotifyImageUploadError( ElementId, fileEntry.ErrorMessage );
-            return;
-        }
-
-
-        if ( fileEntry.FileUploadEndedCallback is null )
-        {
-            await JSModule.NotifyImageUploadSuccess( ElementId, fileEntry.UploadUrl ?? string.Empty );
-        }
-        else
-        {
 #pragma warning disable CS4014 // We want to let execution complete but wait for TaskCompletionSource on the background.
-            Task.Run( async () =>
-            {
+        InvokeAsync( async () =>
+        {
+            if ( fileEntry.FileUploadEndedCallback is not null )
                 await fileEntry.FileUploadEndedCallback.Task;
-                await InvokeAsync( async () => await JSModule.NotifyImageUploadSuccess( ElementId, fileEntry.UploadUrl ?? string.Empty ) );
-            } );
-#pragma warning restore CS4014 
-        }
+
+            if ( ImageUploadEnded is not null )
+                await ImageUploadEnded.Invoke( new( fileEntry, success, fileInvalidReason ) );
+
+            if ( !success )
+            {
+                await JSModule.NotifyImageUploadError( ElementId, fileEntry.ErrorMessage );
+                return;
+            }
+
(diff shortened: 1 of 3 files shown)


stsrki:
Can we have TSC internally around here?

image

The user would set the URL result as

```
private async Task Changed( FileChangedEventArgs e )
{
    var file = e.Files[0];
    
    e.UploadUrl = await Upload( file );
}

```
Then only proceed with ImageUploadEnded if the TSC result is set.

David-Moreira:
> Can we have TSC internally around here?
> 
> image
> 
> The user would set the URL result as
> 
> ```
> private async Task Changed( FileChangedEventArgs e )
> {
>     var file = e.Files[0];
>     
>     e.UploadUrl = await Upload( file );
> }
> ```
> 
> Then only proceed with ImageUploadEnded if the TSC result is set.

Hmm I think I got a possible idea on how to do it. Don't know if it's the same as yours but I'll try it later today.

David-Moreira:
Yep, you were right. My bad. :)
I still left the TCS as part of the IFileEntry, since I believe we should have this per file instead of Markdown. What do you think of that?

Source/Extensions/Blazorise.Markdown/Markdown.razor.cs
```diff
@@ -315,8 +318,8 @@ public async Task UpdateFileEndedAsync( IFileEntry fileEntry, bool success, File
 #pragma warning disable CS4014 // We want to let execution complete but wait for TaskCompletionSource on the background.
             Task.Run( async () =>
```
stsrki, line 319: Do we still need this `Task.Run`?

Source/Extensions/Blazorise.Markdown/Markdown.razor.cs
```diff
@@ -315,8 +318,8 @@ public async Task UpdateFileEndedAsync( IFileEntry fileEntry, bool success, File
...
             {
-                var uploadUrl = await fileEntry.UploadUrlCallback.Task;
-                await InvokeAsync( async () => await JSModule.NotifyImageUploadSuccess( ElementId, uploadUrl ?? string.Empty ) );
+                await fileEntry.FileUploadEndedCallback.Task;
```
stsrki, line 321: What if the TSC was canceled?

Source/Extensions/Blazorise.Markdown/Markdown.razor.cs
David-Moreira: Maybe we don't, I'll check

Source/Extensions/Blazorise.Markdown/Markdown.razor.cs
David-Moreira: Shouldn't even get to this point then?

Source/Extensions/Blazorise.Markdown/Markdown.razor.cs
David-Moreira: So, it doesn't work, because **UpdateFileEndedAsync** is part of the File Write Operation (which is called in **ImageUploadChanged**) and this is the big original problem. 
So **ImageUploadChanged** won't finish until **UpdateFileEndedAsync** has finished. So that's why we set up the background task, to make everything finish and we secretly wait for TCS which is right after **ImageUploadChanged** (**UpdateFileEndedAsync / ImageUploadEnded**) which means URL should already been set.

David-Moreira: It doesn't work by removing the **Rask.Run** I mean.

David-Moreira:
Didn't see any necessary changes, let me know otherwise.

Source/Extensions/Blazorise.Markdown/Markdown.razor.cs
stsrki: Ok

Source/Extensions/Blazorise.Markdown/Markdown.razor.cs
stsrki: Not a big fan of `Task.Run` but if there is no other way I guess we can leave it.

David-Moreira: Well if you got other ideas I'm all ears 😅

stsrki: I dont haha

David-Moreira:
> I don't think this works as expected.
> 
> Here is my test code
> 
> ```
> <Markdown ImageUploadChanged="Changed" ImageUploadEnded="ImageUploadEnded" />
> 
> @code {
>     private string result = string.Empty;
>     private async Task Changed( FileChangedEventArgs e )
>     {
>         var file = e.Files[0];
> 
>         result = await Upload( file );
>     }
> 
>     private async Task<string> Upload( IFileEntry file )
>     {
>         var stream = new MemoryStream();
>         await file.WriteToStreamAsync( stream );
>         return "xpto";
>     }
> 
>     Task ImageUploadEnded( FileEndedEventArgs e )
>     {
>         Console.WriteLine( "Ended" );
>         if ( e.Success )
>         {
>             e.File.UploadUrl = result;
>         }
>         else
>         {
>             e.File.UploadUrl = "FAIL";
>         }
> 
>         Console.WriteLine( e.File.UploadUrl );
> 
>         return Task.CompletedTask;
>     }
> }
> ```
> 
> The first time I try to upload, I get an empty result, and the second-time it works.
> 
> image

Actually my latest tests were with file.uploadUrl = await Upload( file ); could there be a race condition if we set it in ImageUploadEnded instead? I'll take a look later today.

stsrki:
Still no luck

```
<Markdown ImageUploadChanged="Changed" ImageUploadEnded="ImageUploadEnded" />

@code {
  private async Task Changed( FileChangedEventArgs e )
  {
    var file = e.Files[0];
    
    file.UploadUrl = await Upload( file );
  }
  
  private async Task<string> Upload( IFileEntry file )
  {
    var stream = new MemoryStream();
    await file.WriteToStreamAsync( stream );
    return "xpto";
  }
  
  Task ImageUploadEnded( FileEndedEventArgs e )
  {
    Console.WriteLine( "Ended" );
    
    Console.WriteLine( e.File.UploadUrl );
    
    return Task.CompletedTask;
  }
}
```

David-Moreira:
> Still no luck
> 
> ```
> <Markdown ImageUploadChanged="Changed" ImageUploadEnded="ImageUploadEnded" />
> 
> @code {
>   private async Task Changed( FileChangedEventArgs e )
>   {
>     var file = e.Files[0];
>     
>     file.UploadUrl = await Upload( file );
>   }
>   
>   private async Task<string> Upload( IFileEntry file )
>   {
>     var stream = new MemoryStream();
>     await file.WriteToStreamAsync( stream );
>     return "xpto";
>   }
>   
>   Task ImageUploadEnded( FileEndedEventArgs e )
>   {
>     Console.WriteLine( "Ended" );
>     
>     Console.WriteLine( e.File.UploadUrl );
>     
>     return Task.CompletedTask;
>   }
> }
> ```

what the ... :D

stsrki (approved):

PR #4592, opened 2023-02-11
DatePicker: disable input on mobile

### LINKED ISSUE #4577: DatePicker DisableMobile not working. Keyboard still opens on selection on Mobile. (Blazorise 1.2.0)
**Describe the bug**
like title.
grafik
**To Reproduce**

I Just tested it with the new DisableMobile (My Field is inside A Row and a Modal if this changes anything)
And it still behaves like in my original screenshot
Tested on Android Huawei P30 Android 10

```
<Field ColumnSize="ColumnSize.Is12" Margin="Margin.Is0">
    <FieldLabel>Stuff</FieldLabel>
    <FieldBody>
        <Addon AddonType="AddonType.Body">
            <DatePicker @ref="@datePicker" DisableMobile TValue="DateTime?" @bind-Date="@Date" Placeholder="Stuff" DisplayFormat="MM.yy" InputMode="DateInputMode.Date" Min="@MinDate">

            </DatePicker>
        </Addon>
        <Addon AddonType="AddonType.End">
            <Button Color="Color.Danger" Clicked="@(()=>DoStuff= null)">
                <Icon Name="IconName.Delete" />
            </Button>
        </Addon>

    </FieldBody>
</Field>
```

**Expected behavior**
Keyboard should be hidden.

stsrki:
Closes #4577

This PR disables the input it DisableMobile is enabled. I have also set DisableMobile with default `false`, par flatpickr documentation.

The only problem is that the input is shown as grayed like it has a disabled state. I don't know if that is the right behavior for the [flatpickr](https://github.com/flatpickr/flatpickr).

**Example**

image

@Mr-Pearce do you think this is OK?

### DIFF
--- a/Source/Blazorise/wwwroot/datePicker.js
+++ b/Source/Blazorise/wwwroot/datePicker.js
@@ -39,8 +39,8 @@ export function initialize(element, elementId, options) {
     const defaultOptions = {
         enableTime: options.inputMode === 1,
         dateFormat: options.inputMode === 1 ? 'Y-m-d H:i' : 'Y-m-d',
-        allowInput: true,
-        altInput: true,
+        allowInput: !(options.disableMobile || false),
+        altInput: !(options.disableMobile || false),
         altFormat: options.displayFormat ? options.displayFormat : (options.inputMode === 1 ? 'Y-m-d H:i' : 'Y-m-d'),
         defaultDate: options.defaultDate,
         minDate: options.min,
@@ -52,7 +52,7 @@ export function initialize(element, elementId, options) {
         clickOpens: !(options.readOnly || false),
         disable: options.disabledDates || [],
         inline: options.inline || false,
-        disableMobile: options.disableMobile || true,
+        disableMobile: options.disableMobile || false,
         static: true
     };
 
--- a/Source/Blazorise/wwwroot/datePicker.js
+++ b/Source/Blazorise/wwwroot/datePicker.js
@@ -69,7 +69,7 @@ export function initialize(element, elementId, options) {
 
     const picker = flatpickr(element, Object.assign({}, defaultOptions, pluginOptions));
 
-    if (options) {
+    if (options && picker.altInput) {
         picker.altInput.disabled = options.disabled || false;
         picker.altInput.readOnly = options.readOnly || false;
     }
--- a/Source/Blazorise/wwwroot/datePicker.js
+++ b/Source/Blazorise/wwwroot/datePicker.js
@@ -39,8 +39,8 @@ export function initialize(element, elementId, options) {
     const defaultOptions = {
         enableTime: options.inputMode === 1,
         dateFormat: options.inputMode === 1 ? 'Y-m-d H:i' : 'Y-m-d',
-        allowInput: !(options.disableMobile || false),
-        altInput: !(options.disableMobile || false),
+        allowInput: !(isMobile() && (options.disableMobile || false)),
+        altInput: !(isMobile() && (options.disableMobile || false)),
         altFormat: options.displayFormat ? options.displayFormat : (options.inputMode === 1 ? 'Y-m-d H:i' : 'Y-m-d'),
         defaultDate: options.defaultDate,
         minDate: options.min,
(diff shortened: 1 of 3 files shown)


Mr-Pearce:
It will probably confuse users when the form is greyed out as if it is disabled but is enabled just that the keyboard wont show up.

David-Moreira:
I have to agree with Mr.Pearce, If I'm a user I would think its disabled by some reason. 
Isn't there a better way to represent that text is disabled but you can interact with it?

Source/Blazorise/wwwroot/datePicker.js
```diff
@@ -52,7 +52,7 @@ export function initialize(element, elementId, options) {
         clickOpens: !(options.readOnly || false),
         disable: options.disabledDates || [],
         inline: options.inline || false,
-        disableMobile: options.disableMobile || true,
+        disableMobile: options.disableMobile || false,
```
David-Moreira, line 55: maybe I'm dumb, what is this? options.disableMobile || false? is this some kind of way to coalesce?

Source/Blazorise/wwwroot/datePicker.js
```diff
@@ -39,8 +39,8 @@ export function initialize(element, elementId, options) {
     const defaultOptions = {
         enableTime: options.inputMode === 1,
         dateFormat: options.inputMode === 1 ? 'Y-m-d H:i' : 'Y-m-d',
-        allowInput: true,
-        altInput: true,
+        allowInput: !(options.disableMobile || false),
+        altInput: !(options.disableMobile || false),
```
David-Moreira: Is this some kind of way to coalesce is that it? I don't get the usage of the or logic here.

Source/Blazorise/wwwroot/datePicker.js
```diff
@@ -39,8 +39,8 @@ export function initialize(element, elementId, options) {
     const defaultOptions = {
         enableTime: options.inputMode === 1,
         dateFormat: options.inputMode === 1 ? 'Y-m-d H:i' : 'Y-m-d',
-        allowInput: !(options.disableMobile || false),
-        altInput: !(options.disableMobile || false),
+        allowInput: !(isMobile() && (options.disableMobile || false)),
+        altInput: !(isMobile() && (options.disableMobile || false)),
```
David-Moreira, line 43: My mind can't think today :D Is this right? Does it handle every case?
Might be a better read if you just bring the isMobile() out.
I.E 
!IsMobile() || (isMobile()  && !options.disableMobile)

But I think it's those pipes that are killing me, again is it a coalesce operation?

David-Moreira (approved):

Source/Blazorise/wwwroot/datePicker.js
stsrki: The logic is actually broken on flatpickr side. They don't detect the mobile device properly. So I was trying to come up with a workaround, even pushed a PR on their repo, but it still seems not to detect all of the mobile devices.

David-Moreira:
@stsrki friendly reminder, this PR sitting here

stsrki:
> @stsrki friendly reminder, this PR sitting here

I know. I still didn't have much time to fix an error on flatpickr side. Which we must do before even applying it here.

David-Moreira:
Just a friendly reminder, @stsrki

stsrki:
> Friendly reminder, What is the status of this one?

Unfortunately, it seems that flat pick is not actively maintained. So, in the long run, we need to either fork it or maintain it ourselves. Or move to another library.

Mr-Pearce:
> > Friendly reminder, What is the status of this one?
> 
> Unfortunately, it seems that flat pick is not actively maintained. So, in the long run, we need to either fork it or maintain it ourselves. Or move to another library.

FYI @strski:
I Just had an "bug" with the DatePicker where when in oninizializedasync an async call is made.
the binding wouldn't update the ui values afterwards, even with StateHasChanged.
I changed to DateEdit and it worked out of the box.

Maybe more stuff is broken with flatpick.

stsrki:
> > > Friendly reminder, What is the status of this one?
> > 
> > 
> > Unfortunately, it seems that flat pick is not actively maintained. So, in the long run, we need to either fork it or maintain it ourselves. Or move to another library.
> 
> FYI @stsrki : I Just had an "bug" with the DatePicker where when in oninizializedasync an async call is made. the binding wouldn't update the ui values afterwards, even with StateHasChanged. I changed to DateEdit and it worked out of the box.
> 
> Maybe more stuff is broken with flatpick.

Might be something with the js interop. Can you open new issue and make a reproducible?

stsrki:
Closing in favor of #6107

PR #4598, opened 2023-02-12
TreeView: check SelectedNode against internal state value

### LINKED ISSUE #4413: TreeView: `SelectedNodeChanged`-Event is fired when there should be no change
**Describe the bug**
I have a TreeView and whenever the user want's to change the current node, I want to ask the user before taking any action.

**To Reproduce**
Here is a simple example

```razor
<TreeView TNode="string" Nodes="nodes" SelectedNode="selectedNode" SelectedNodeChanged="SelectedNodeChanged">
    <NodeContent>@context</NodeContent>
</TreeView>
Selected: @selectedNode
@code {
    [Inject] IMessageService MessageService { get; set; } = default!;

    string?[] nodes = new[] { "A", "B", "C" };
    string? selectedNode = "A";

    private async Task SelectedNodeChanged(string? newSelectedNode)
    {
        // is fired twice (so fast you'll never see the first message)
        if (!await MessageService.Confirm($"Navigate from {selectedNode} to {newSelectedNode}?")) return;
        selectedNode = newSelectedNode;
    }

    private async Task SelectedNodeChangedSwitchBack(string? newSelectedNode)
    {
        var prevNode = selectedNode;
        selectedNode = newSelectedNode;
        if (!await MessageService.Confirm($"Navigate from {prevNode} to {newSelectedNode}?")) {
            selectedNode = prevNode; // fires the event again (since it differs from the `state.SelectedNode`)
        }
    }
}
```

Whenever you click on any other node than the selected, a message pops up.

I expect the message shown states something like `Navigate from A to B?` and if you debug that, there is really such a message - but right after the first message, there's a second one stating `Navigate from A to A?`. At the end nothing happens.

If you use `SelectedNodeChangedSwitchBack`, the behavior is somewhat different and it seems to work - but then you get two messages in a row.

I think, the main reason for the whole problem is, that you invoke `SelectedNodeChanged` whenever `SelectedNode` is changed - even if that change comes from outside.
https://github.com/Megabit/Blazorise/blob/019c5803cf201f452172fe47329e64727c6472f0/Source/Extensions/Blazorise.TreeView/TreeView.razor.cs#L77-L92, especially the line `SelectedNodeChanged.InvokeAsync( state.SelectedNode );`

If I move that logic (except `DirtyClasses`) to `SelectNode` I get the expected behavior.

**Expected behavior**
I expect to change `selectedNode` without getting `SelectedNodeChanged` fired.

You should only fire events, if you are changing the values (not when the value is changed from outside your component). `InvokeAsync` inside the property setter of `[Parameters]` seems no good idea (for all components, not only the TreeView).
https://github.com/dotnet/aspnetcore/issues/26230

**Additional context**
My workaround is to test, if the `newSelectedNode` is changed at all - but this seems unexpected and unnecessary.

### LINKED ISSUE #4572: TreeView is not acting correctly when setting selected node and expanded nodes in code.
I just updated from version 1.0.6 to 1.2.0.

I have a TreeView where I load the nodes, then set the selected node and expand it in OnParametersSetAsync. This worked fine under 1.0.6, but under 1.2.0 it does not.

What happens is the tree is shown unexpanded, but the icon for all branches is the expanded icon. To open each branch, you have to click the tree icon twice. After that, things seem to work just fine.

image
image

If you comment out the code that sets the selected and expanded nodes, everything works correctly.

My examples are here:
https://github.com/JMoodyUMC/TreeViewTester.1.0.6
https://github.com/JMoodyUMC/TreeViewTester.1.2.0

stsrki:
Closes #4413
Closes #4572

### DIFF
--- a/Source/Extensions/Blazorise.TreeView/Internal/_TreeViewNode.razor.cs
+++ b/Source/Extensions/Blazorise.TreeView/Internal/_TreeViewNode.razor.cs
@@ -21,6 +21,16 @@ public partial class _TreeViewNode<TNode> : BaseComponent
         {
             await AutoExpandNodes();
         }
+        else if ( ExpandedNodes?.Count > 0 )
+        {
+            foreach ( var nodeState in NodeStates ?? Enumerable.Empty<TreeViewNodeState<TNode>>() )
+            {
+                if ( nodeState.HasChildren && ExpandedNodes?.Contains( nodeState.Node ) == true )
+                {
+                    await LoadChildNodes( nodeState );
+                }
+            }
+        }
 
         await base.OnInitializedAsync();
     }
@@ -60,21 +70,7 @@ public partial class _TreeViewNode<TNode> : BaseComponent
 
             if ( nodeState.HasChildren )
             {
-                var childNodes = GetChildNodesAsync is not null
-                    ? await GetChildNodesAsync( nodeState.Node )
-                    : GetChildNodes is not null
-                        ? GetChildNodes( nodeState.Node )
-                        : null;
-
-                if ( !nodeState.Children.Select( x => x.Node ).AreEqual( childNodes ) )
-                {
-                    nodeState.Children.Clear();
-
-                    await foreach ( var childNodeState in childNodes.ToNodeStates( HasChildNodesAsync, HasChildNodes, ExpandedNodes.Intersect( childNodes ?? Enumerable.Empty<TNode>() ).Any() ) )
-                    {
-                        nodeState.Children.Add( childNodeState );
-                    }
-                }
+                await LoadChildNodes( nodeState );
             }
         }
         else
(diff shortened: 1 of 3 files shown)


Source/Extensions/Blazorise.TreeView/Internal/_TreeViewNode.razor.cs
```diff
@@ -21,6 +21,16 @@ protected override async Task OnInitializedAsync()
         {
             await AutoExpandNodes();
         }
+        else if ( ExpandedNodes?.Count > 0 )
+        {
+            foreach ( var nodeState in NodeStates ?? Enumerable.Empty<TreeViewNodeState<TNode>>() )
+            {
+                if ( nodeState.HasChildren && ExpandedNodes?.Contains( nodeState.Node ) == true )
```
David-Moreira, line 28: Can replace `ExpandedNodes?.Contains( nodeState.Node ) == true` with `ExpandedNodes.Contains( nodeState.Node )` if you want to bother, as you've null checked above.

David-Moreira (approved):
LGTM

PR #4611, opened 2023-02-15
Enhanced cropper example

njannink:
@stsrki I'm trying to integrate the Blazorise cropper with our solution. We use the cropper in combination with a file input element, to be able to crop a profile image.

I implemented this example in the Blazorise examples to show how the 2 can work together and for us to think how we can improve a few things. As you can see I need to add some javascript to be able to create an ObjectURL from a file in the file picker and this URL also needs to be cleaned up at the end.

Maybe a good improvement to Blazorise would be if the `IFileEntry` has a call for creating an ObjectURL that returns a wrapper with the URL and that implements the `IAsyncDisposable` interface for cleanup.

### DIFF
--- a/Demos/Blazorise.Demo/Pages/Tests/CropperPage.razor
+++ b/Demos/Blazorise.Demo/Pages/Tests/CropperPage.razor
@@ -1,5 +1,9 @@
 ﻿@page "/tests/cropper"
-@using System.Security.Cryptography.X509Certificates
+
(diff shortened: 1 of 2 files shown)


stsrki:
Thanks for the contribution. I tried to rebase to 1.2 so that it can be uploaded live sooner but there were some wrong commits. So I will close the PR, as I have already recreated it and pushed it.

PR #4620, opened 2023-02-23
DataGrid: place command buttons inside of inline div

### LINKED ISSUE #4618: Fix DataGrid command cell spacing
https://blazorise.com/support/issues/94

stsrki:
Closes #4618 

Fixes the spacing of the command cell when the table content is too large. The downside is now that the users cannot override the default spacing of the inlined `div`.


David-Moreira (approved):
Bit hard to read the repositioning on github but I'm sure you got it right. :) LGTM!

PR #4624, opened 2023-02-24
Add Cropper center method

njannink:
Add the missing $center method to the Cropper extension


stsrki:
Looking good. But since it is adding a new API I think it should be better if it goes to the `master`, for the next major release.

njannink:
thats a pity since its a small enhancement and looking up the `cropper-image` in my own javascript is only possible by calling `getElementsByTagName`

stsrki (approved):
Looking good. But I will have to move it to the 1.3 release.

PR #4633, opened 2023-02-28
 DataGrid: New sorting method and events

danm-de:
**Introduction**

As a happy Blazorise customer, we use the DataGrid very often in our applications. We support deep-linking on our pages (which our employees really like).

An example deep link looks like this:

`https://medic.server.com/drugs?page=2&pageSize=50&sort=drugName:Asc&sort=amount:Desc`

**Problem**

Til today, we haven't found a way to track and restore the user's original sort order. The DataGrid provides a single `Sort` method which gives us no way to replace (or initially set) the sort order of all columns.

Additionally, the `Sort` method requires specifying the `DataGridColumn.Field` name, but the `DataGridSortChangedEventArgs` event contains the `DataGridColumn.SortField` name (mismatch).

**Solution**

This pull requests adds

- An additional property `ColumnFieldName` to the `DataGridSortChangedEventArgs`.
- A new event `DataGridSortOrderChangedEventArgs` that contains a list of all sort columns + its sort direction.
- A new method `DataGrid.ApplySorting(..)` that can be used to programmatically reset or replace all sort columns.
- Two new methods `DataGrid.GetColumns()` and `DataGrid.GetSortColumns()` to query all columns programmatically without having to define an @ref in the razor file for each `DataGridColumn`.
- bUnit tests for basic sorting features.
- A bit of documentation.


danm-de:

David-Moreira:
Sorry I haven't been able to take a look at this. I'm planning to take a look at this by tomorrow latest!

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -244,7 +264,28 @@ public void AddRow( DataGridRowInfo<TItem> row )
...
+                SortOrderChanged.InvokeAsync(
+                    new DataGridSortOrderChangedEventArgs(
+                        SortMode,
+                        SortByColumns.Select( c => new DataGridSortInfo(
+                                c.Field,
+                                c.GetFieldToSort(),
+                                c.CurrentSortDirection,
+                                c.SortOrder ) )
+                            .OrderBy( c => c.SortOrder )
+                            .ToList() ) ) );
```
David-Moreira, line 284: Do we think this is expected? If a column is removed should it realy trigger a Sort Order Changed?

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -244,7 +264,28 @@ public void AddRow( DataGridRowInfo<TItem> row )
     /// </summary>
     /// <param name="column">Column to link with this datagrid.</param>
     public bool RemoveColumn( DataGridColumn<TItem> column )
-        => Columns.Remove( column );
+    {
+        var removed = Columns.Remove( column );
+        
+        if ( column.SortDirection != SortDirection.Default )
+        {
+            SortByColumns.Remove( column );
```
David-Moreira, line 272: So, wasn't this already handled before? Would it remain in the SortByColumn list?
I guess this is only really going to matter if someone plays with dynamic columns though.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -820,6 +861,79 @@ public Task Sort( DataGridColumn<TItem> column, SortDirection? sortDirection = n
         return Task.CompletedTask;
     }
 
+
+    /// <summary>
+    /// Replaces the current sorting with the given sort order
```
David-Moreira, line 866: Do you think this would explain it better?
"Applies a new sorting definition to the Datagrid. Replaces the current sorting."

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -820,6 +861,79 @@ public Task Sort( DataGridColumn<TItem> column, SortDirection? sortDirection = n
...
+    /// <remarks>
+    /// Note that <see cref="DataGridColumn{TItem}.Sortable"/> and <see cref="Sortable"/> must be enabled to be able to sort!
+    /// If more than one column is specified, <see cref="SortMode"/> must be <see cref="DataGridSortMode.Multiple"/>
+    /// </remarks>
+    public async Task ApplySorting(params DataGridSortColumn[] columns ) {
+        if ( columns.Length > 1 && SortMode == DataGridSortMode.Single )
+        {
+            throw new ArgumentException(
+                $"Cannot sort on multiple columns when {nameof(SortMode)} is {nameof(DataGridSortMode.Single)}." );
+        }
```
David-Moreira, line 879: @stsrki do we like this check? 
I think we generally disregard from throwing too much exceptions in Blazorise.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -820,6 +861,79 @@ public Task Sort( DataGridColumn<TItem> column, SortDirection? sortDirection = n
...
+
+        if ( SortMode == DataGridSortMode.Single )
+        {
+            if ( columns.Length == 1 )
+            {
+                await Sort( columns[0].Field, columns[0].SortDirection );
+            }
+            else if ( SortByColumns.Count == 1 )
```
David-Moreira, line 887: Can you explain?
You've done a sanity check at the top for columns > 1 + Single. 
How would the **else if** trigger here and why? If it was empty? Does that make sense?

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -820,6 +861,79 @@ public Task Sort( DataGridColumn<TItem> column, SortDirection? sortDirection = n
...
+        }
+
+        // reset sorting
+        foreach ( var column in SortByColumns )
+        {
+            column.CurrentSortDirection = SortDirection.Default;
+            await column.ResetSortOrder();
+        }
+
+        SortByColumns.Clear();
```
David-Moreira, line 902: Isn't there an api that does the cleanign automatically?

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -820,6 +861,79 @@ public Task Sort( DataGridColumn<TItem> column, SortDirection? sortDirection = n
...
+
+        // apply new sorting
+        var columnTuples = columns
+            .Select( ( x, idx ) => (
+                Column: Columns.SingleOrDefault( c => c.Field == x.Field ),
```
David-Moreira, line 907: Why **SingleOrDefault**? User could define two columns with same Field for some reason. Maybe better not throw in that case.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -820,6 +861,79 @@ public Task Sort( DataGridColumn<TItem> column, SortDirection? sortDirection = n
...
+            }
+            else if ( SortByColumns.Count == 1 )
+            {
+                await Sort( SortByColumns[0].Field, SortDirection.Default );
+            }
+
+            return;
+        }
+
+        // reset sorting
```
David-Moreira, line 895: whenever you find youself applying a comment like this, it's probably because there might be a chance to just move the logic to a method. 
I.E **ResetSorting**.
You get self explanatory code and less lines of code in a single method.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -820,6 +861,79 @@ public Task Sort( DataGridColumn<TItem> column, SortDirection? sortDirection = n
...
+            .Select( ( x, idx ) => (
+                Column: Columns.SingleOrDefault( c => c.Field == x.Field ),
+                Direction: x.SortDirection,
+                SortOrder: idx ) )
+            .Where( x => x.Column is { Sortable: true } &&
+                         x.Direction != SortDirection.Default );
+
+        foreach ( var (column, direction, sortOrder) in columnTuples )
+        {
+            if ( SortByColumns.All( x => x.GetFieldToSort() != column.GetFieldToSort() ) )
```
David-Moreira, line 915: Wasn't this cleared above?
`SortByColumns.Clear();`

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -1307,41 +1421,73 @@ protected async ValueTask<ItemsProviderResult<TItem>> VirtualizeItemsProviderHan
...
+            column.SetSortOrder( nextOrderToSort );
+            SortByColumns.Add( column );
+        }
+        else if ( column.CurrentSortDirection == SortDirection.Default )
+        {
+            SortByColumns.Remove( column );
+            column.ResetSortOrder();
+        }
+
+        Task RaiseSortChanged() =>
```
David-Moreira, line 1462: Are these local functions here? Aren't we capturing variables needlessly?

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -2195,10 +2341,15 @@ public IEnumerable<GroupContext<TItem>> DisplayGroupedData
     [Parameter] public EventCallback<DataGridReadDataEventArgs<TItem>> ReadData { get; set; }
 
     /// <summary>
-    /// Occurs after the column sort direction has changed.
+    /// Occurs after the sort direction of a single column has changed.
     /// </summary>
     [Parameter] public EventCallback<DataGridSortChangedEventArgs> SortChanged { get; set; }
 
+    /// <summary>
+    /// Occurs after the column sorting has changed.
```
David-Moreira, line 2349: After the column sorting or sort order?

Source/Extensions/Blazorise.DataGrid/EventArguments/DataGridSortOrderChangedEventArgs.cs
```diff
@@ -0,0 +1,31 @@
...
+public class DataGridSortOrderChangedEventArgs
+{
+    /// <inheritdoc cref="DataGrid{TItem}.SortMode"/>
+    public DataGridSortMode SortMode { get; init; }
+
+    /// <summary>
+    /// A collection of <see cref="SortOrder"/> items. Contains all columns by which to sort.
+    /// </summary>
+    /// <remarks>Returns an empty list if no sorting is applied.</remarks>
+    public IReadOnlyCollection<DataGridSortInfo> SortOrder { get; init; }
```
David-Moreira, line 19: I don't get this recurring `SortOrder` naming. Do you mean it's just regular info on how a column should be sorted, and not related to the order? In which case the naming `SortInfo` or `SortDefinition` or something else more general seems better?

Source/Extensions/Blazorise.DataGrid/Models/DataGridSortColumn.cs
```diff
@@ -0,0 +1,34 @@
...
+
+    /// <summary>
+    /// Gets or sets direction of the column used for sorting (see <see cref="DataGridColumn{TItem}.SortDirection"/>).
+    /// </summary>
+    public SortDirection SortDirection { get; init; }
+
+    /// <summary>
+    /// Creates a new instance
+    /// </summary>
+    public DataGridSortColumn()
```
David-Moreira, line 21: Is an empty ctor valid?

Tests/Blazorise.Tests/Components/DataGridComponentTest.cs
```diff
@@ -39,19 +45,330 @@ public void NoData_Should_Render_EmptyTemplate()
...
+        var sortChanged = new List<DataGridSortChangedEventArgs>();
+        var comp = RenderComponent<DataGridComponent>( parameters =>
+        {
+            parameters.Add(
+                parameterSelector: x => x.SortChanged,
+                callback: e => sortChanged.Add( e ) );
+        } );
+
+        // test
+        await Task.Factory.StartNew(
```
David-Moreira, line 91: Can you explain briefly why we need all of these `Task.Factory.StartNew`

David-Moreira (changes requested):
Looking good, I like the new ApplySorting api.
Very nice complete work, even demo, docs and tests were done, thank you.

I'd like to understant that `DataGridSortOrderChangedEventArgs` and `SortOrder` api better, as the naming seems like it implies it's something to do with the sorting order being changed. But it's just a more generic sorting change notification that's being triggered right?

**PS:** I still need to take it for a spin. Only done CR.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
stsrki: I'm generally against throwing exceptions. In this case, it could _blow_  to the user by mistake if they try to sort in "somewhat unsuported"  way,

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
stsrki: I agree that we should use only `FirstOrDefault`.

danm-de:
Thanks for the comments, I'll have another look at it.

Documentation/Blazorise.Docs/Pages/Docs/Extensions/DataGrid/GettingStartedPage.razor
```diff
@@ -249,9 +249,15 @@
...
         Defines the default handler type that will be used by the validation, unless it is overriden by <Code>Validation.HandlerType</Code> property.
     </DocsAttributesItem>
+    <DocsAttributesItem Name="SortChanged" Type="EventCallback<DataGridSortChangedEventArgs>" Default="null">
+        Occurs after the sort direction of a single column has changed.
+    </DocsAttributesItem>
     <DocsAttributesItem Name="SortMode" Type="DataGridSortMode" Default="Multiple">
         Defines whether the user can sort only by one column or by multiple.
     </DocsAttributesItem>
+    <DocsAttributesItem Name="SortOrderChanged" Type="EventCallback<DataGridSortOrderChangedEventArgs>" Default="null">
+        Occurs after the column sorting has changed.
```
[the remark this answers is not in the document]
danm-de: Fixed in commit f42c2d155d67ee5d870346548e318f02f8755b0d

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -201,19 +201,39 @@ private string GetFixedTableHeaderMaxHeight()
             return FixedHeaderDataGridMaxHeight;
     }
 
+    /// <summary>
+    /// Creates a list of all the columns that are currently associated with this datagrid.
+    /// </summary>
+    /// <returns>A read-only list of all columns</returns>
+    public IReadOnlyList<DataGridColumn<TItem>> GetColumns() => Columns.ToList();
```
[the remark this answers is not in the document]
danm-de: I know the `List<T>.AsReadOnly()` method. This method returns a `ReadOnlyCollection<TItem>` containing the internal list of columns (it's just a wrapper).

This means the returned list would change as the user adds or removes columns. I figured a copy would be safer here.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -201,19 +201,39 @@ private string GetFixedTableHeaderMaxHeight()
...
+    /// Creates a list of all the columns that are currently associated with this datagrid.
+    /// </summary>
+    /// <returns>A read-only list of all columns</returns>
+    public IReadOnlyList<DataGridColumn<TItem>> GetColumns() => Columns.ToList();
+
+    /// <summary>
+    /// Creates a list of all columns currently used to sort this datagrid's data.
+    /// </summary>
+    /// <returns>A read-only list of all sort columns, or an empty list.</returns>
+    public IReadOnlyList<DataGridColumn<TItem>> GetSortColumns() => SortByColumns.ToList();
```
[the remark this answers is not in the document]
danm-de: Changed in ceb3804152e7b7c41f727ef4079c58609505bc72

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
danm-de: Yes, thanks.

Fixed in 01f883a751824450f29f02939fea1726012fcfe0

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
danm-de: Fixed in c625994d8165eaffb7e33c3d6c66b72a3126b8f7

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
danm-de: I added code comments to clarify the if/else statement. Basically, the ELSE branch checks whether the DataGrid is already sorted by a specific column. If this is the case, the sorting must be **reset** here (the user has not passed any sorting columns).

Is this okay?

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
danm-de: I only found cleanup in `HandleSortColumn`. Maybe I did not understand what you mean.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
danm-de: I want to make sure that no duplicates end up in `SortByColumns`. New suggestion: `DistinctBy(x)` on the list passed by the user.

Fixed in e3e51e1686c3a99a22468e12fab5646c4a061d2b

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
danm-de: Changed the local function to `static` in bf531e2209cfcf20d3ec1467fb1adf5c272dbb90.

Source/Extensions/Blazorise.DataGrid/EventArguments/DataGridSortOrderChangedEventArgs.cs
danm-de: I guess I am using the wrong term :-( . 

This event contains a list of all columns in their sort order and sort direction (ascending, descending). It also contains the column's `Field` name and (if available) the `SortField` name.

danm-de: Fixed naming in 8bdd62859a36e46ebe0cf812f8f6f3e0e4c13544

Source/Extensions/Blazorise.DataGrid/Models/DataGridSortColumn.cs
danm-de: Removed, thanks. 95e004f2e153bb28fc55d1d2adcf60cf1a77de68

Tests/Blazorise.Tests/Components/DataGridComponentTest.cs
danm-de: During development, I often launch the existing unit tests. It seems that there are currently some "unstable" tests. My colleagues call these tests "blink tests" because they constantly alternate between green (ok) and red (fail).

The experience we've had with our own bUnit tests is that these tests are often timing issues of async/parallel code.

In *DataGrid.razor.cs* there are a few methods that call `InvokeAsync(..)` without awaiting the method call (e.g. the events that are raised inside `HandleSortColumn` -I haven't investigated why this method returns `void` instead of `Task`). 

Not awaiting the `InvokeAsync` method results in parallel processing or out-of-order processing. You cannot write a stable unit test without having control over when **a method actually finishes executing**.

I thought about how to get synchronous method execution. The only idea I had was to write my own TaskScheduler. It's called `CurrentThreadTaskScheduler` and has a maximum concurrency level of 1. It doesn't allow scheduling of tasks and (hopefully) enforces synchronous execution.

Some examples of currently unstable tests:
image
image
image

danm-de: What do you think about this "hack"?

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -872,58 +872,47 @@ public Task Sort( DataGridColumn<TItem> column, SortDirection? sortDirection = n
     /// If more than one column is specified, <see cref="SortMode"/> must be <see cref="DataGridSortMode.Multiple"/>
     /// </remarks>
     public async Task ApplySorting(params DataGridSortColumn[] columns ) {
-        if ( columns.Length > 1 && SortMode == DataGridSortMode.Single )
-        {
-            throw new ArgumentException(
-                $"Cannot sort on multiple columns when {nameof(SortMode)} is {nameof(DataGridSortMode.Single)}." );
-        }
```
David-Moreira: I think it would be best if we just exit the function if not columns have been provided, what do you think?

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -872,58 +872,47 @@ public Task Sort( DataGridColumn<TItem> column, SortDirection? sortDirection = n
...
-            if ( columns.Length == 1 )
+            if ( columns.Length >= 1 )
             {
+                // Sort the DataGrid based on the first column passed
                 await Sort( columns[0].Field, columns[0].SortDirection );
             }
             else if ( SortByColumns.Count == 1 )
             {
+                // If the user has not passed any columns and the DataGrid is currently sorted
+                // by a column, use the data-source's default sort order.
```
David-Moreira, line 887: I think it would be best if we just exit the function if not columns have been provided, what do you think?

Documentation/Blazorise.Docs/Pages/Docs/Extensions/DataGrid/GettingStartedPage.razor
```diff
@@ -255,8 +255,8 @@
     <DocsAttributesItem Name="SortMode" Type="DataGridSortMode" Default="Multiple">
         Defines whether the user can sort only by one column or by multiple.
     </DocsAttributesItem>
-    <DocsAttributesItem Name="SortOrderChanged" Type="EventCallback<DataGridSortOrderChangedEventArgs>" Default="null">
-        Occurs after the column sorting has changed.
+    <DocsAttributesItem Name="SortDefinitionChanged" Type="EventCallback<DataGridSortDefinitionChangedEventArgs>" Default="null">
```
David-Moreira, line 258: @stsrki do you also like "**SortDefintion**" prefix? I'm kinda on the fence about it. But i can't think of any other... naming is hard... :d Maybe **SortInfo**? Or you think this is fine?

Tests/Blazorise.Tests/Components/DataGridComponentTest.cs
David-Moreira: Fair enough, you're right we've been having our issues with bunit also due to async/parallel code and therefore unstable/unreliable tests.
I think in some places we just opted for retry for expected state, or wait longer, but this seems like at least an alternative approach. Thank you.

Documentation/Blazorise.Docs/Pages/Docs/Extensions/DataGrid/GettingStartedPage.razor
David-Moreira: Or maybe event `Sorting`Changed?

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
David-Moreira: Or upon further though, if you call `ApplySorting` with an empty collection, then no sorting is applied and any sorting is removed, I think that would be expected? What do you think?

David-Moreira:
Tested, works good, and intuitive api.
I was thinking about the naming, `ApplySorting` , should we just make it an overload of `Sort` instead? 

How do you guys think it's more understable to the user?

Documentation/Blazorise.Docs/Pages/Docs/Extensions/DataGrid/GettingStartedPage.razor
stsrki: If the sort, in general, has changed, then `Sorting`Changed is a good name.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
danm-de: I was talking to my colleague about what should happen if no columns are passed for sorting (empty array). We both agreed that in this case the "default sorting" of the data source should be active again. Otherwise a public "ResetSorting" method would be required.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
danm-de: I was talking to my colleague about what should happen if no columns are passed for sorting (empty array). We both agreed that in this case the "default sorting" of the data source should be active again.

David-Moreira: I'm sorry I am still not convinced. you explicitly pass in an empty array.
I would expect the behaviour of calling `ApplySorting()` and `ApplySorting(Array.Empty)` to be different. 
(Granted we don't have the version without an argument)

On the first option I could expect the behaviour you guys are mentioning, on the second, I'm explicitly passing in the collection I want to sort on, which is empty, therefore I will sort on no columns, which is basically an "empty" sorting, or no sorting at all.

That's my thought process, what do you think?
@stsrki you have any thoughts on this?

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
David-Moreira: I'm sorry I am still not convinced. you explicitly pass in an empty array.
I would expect the behaviour of calling ApplySorting() and ApplySorting(Array.Empty) to be different.
(Granted we don't have the version without an argument)

On the first option I could expect the behaviour you guys are mentioning, on the second, I'm explicitly passing in the collection I want to sort on, which is empty, therefore I will sort on no columns, which is basically an "empty" sorting, or no sorting at all.

That's my thought process, what do you think?
@stsrki you have any thoughts on this?

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
David-Moreira: I'm sorry I am still not convinced. you explicitly pass in an empty array.
I would expect the behaviour of calling ApplySorting() and ApplySorting(Array.Empty) to be different.
(Granted we don't have the version without an argument)

On the first option I could expect the behaviour you guys are mentioning, on the second, I'm explicitly passing in the collection I want to sort on, which is empty, therefore I will sort on no columns, which is basically an "empty" sorting, or no sorting at all.

That's my thought process, what do you think?
@stsrki you have any thoughts on this?

Documentation/Blazorise.Docs/Pages/Docs/Extensions/DataGrid/GettingStartedPage.razor
danm-de: I'll change this.

danm-de: Fixed in acef337e31e4427a9bf736801eb5f16c9624a59d

danm-de:
> Tested, works good, and intuitive api. I was thinking about the naming, `ApplySorting` , should we just make it an overload of `Sort` instead?
> 
> How do you guys think it's more understable to the user?

The `Sort` and `ApplySorting` methods have different semantic behavior. The existing `Sort` method adds (new) columns for sorting, while `ApplySorting` always replaces the complete sort.

David-Moreira (approved):
LGTM, please give your final approval when you have time @stsrki

David-Moreira (changes requested):
Apologies, actually please review this bit:
image

David-Moreira:
> Apologies, actually please review this bit: image

@danm-de have you gone through this?

David-Moreira:
@stsrki LGTM Please review & merge.

David-Moreira:
Also don't want to bother with 
image
No idea what's up with this error
image


But danme already signed / is in our cla list
image

David-Moreira (approved):

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -193,19 +193,39 @@ private string GetFixedTableHeaderMaxHeight()
             return FixedHeaderDataGridMaxHeight;
     }
 
+    /// <summary>
+    /// Returns a list of all the columns that are currently associated with this datagrid.
+    /// </summary>
+    /// <returns>A read-only list of all columns</returns>
+    public IReadOnlyList<DataGridColumn<TItem>> GetColumns() => Columns.AsReadOnly();
```
stsrki, line 200: Should we just make the existing Columns API public instead of adding a new `GetColumns()` API?

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -193,19 +193,39 @@ private string GetFixedTableHeaderMaxHeight()
...
+    /// Returns a list of all the columns that are currently associated with this datagrid.
+    /// </summary>
+    /// <returns>A read-only list of all columns</returns>
+    public IReadOnlyList<DataGridColumn<TItem>> GetColumns() => Columns.AsReadOnly();
+
+    /// <summary>
+    /// Returns a list of all columns currently used to sort this datagrid's data.
+    /// </summary>
+    /// <returns>A read-only list of all sort columns, or an empty list.</returns>
+    public IReadOnlyList<DataGridColumn<TItem>> GetSortColumns() => SortByColumns.AsReadOnly();
```
stsrki, line 206: Should we just make the existing SortByColumns API public instead of adding a new `GetSortColumns()` API?

Also, rename it to `GetSortByColumns`?

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -2213,10 +2366,15 @@ public IEnumerable<GroupContext<TItem>> DisplayGroupedData
...
     /// <summary>
-    /// Occurs after the column sort direction has changed.
+    /// Occurs after the sort direction of a single column has changed.
     /// </summary>
     [Parameter] public EventCallback<DataGridSortChangedEventArgs> SortChanged { get; set; }
 
+    /// <summary>
+    /// Occurs when the sort order, sort direction, or number of columns to sort changes.
+    /// </summary>
+    [Parameter] public EventCallback<DataGridSortingChangedEventArgs> SortingChanged { get; set; }
```
stsrki, line 2376: We have two similar named APIs. `SortChanged` and `SortingChanged`. What is the difference?

Source/Extensions/Blazorise.DataGrid/EventArguments/DataGridSortColumnInfo.cs
```diff
@@ -0,0 +1,42 @@
+﻿namespace Blazorise.DataGrid;
+
+/// <summary>
+/// Holds the sort order and sort direction information of a specific column
+/// </summary>
+public class DataGridSortColumnInfo
```
stsrki, line 6: The class should be moved under the `./Models` subfolder instead of `./EventArguments`.

Source/Extensions/Blazorise.DataGrid/Models/DataGridSortColumn.cs
```diff
@@ -0,0 +1,28 @@
+﻿namespace Blazorise.DataGrid;
+
+/// <summary>
+/// A column sort definition
+/// </summary>
+public class DataGridSortColumn
```
stsrki, line 6: What is the difference between `DataGridSortColumnInfo` and `DataGridSortColumn`?

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
David-Moreira: I don't mind either way. 
But I prefer this approach because:
-  This just conveys a better meaning that they shouldn't really be mutated (altought we know it's possible)
- Makes it harder to mutate it (again we know it's unfortunately possible anyway)

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
David-Moreira: Same answer as for `GetColumns`. I'm good with the renaming to be consistent with the underlying collection.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
David-Moreira: I think I had this question myself before. Let me look it up, see if @danm-de has clarified it before.

Source/Extensions/Blazorise.DataGrid/EventArguments/DataGridSortColumnInfo.cs
David-Moreira: Agreed, good catch.

Source/Extensions/Blazorise.DataGrid/Models/DataGridSortColumn.cs
David-Moreira: Again I believe I've had the same question & @danm-de has clarified it before, I'll re-check the reviews above. It does seem like a bit of uneeded duplication though on first look.

David-Moreira: From what I can tell, the `DataGridSortColumn` is used as an input for `ApplySorting`, as such it only has the necessary information for this operation.
The `DataGridSortColumnInfo` holds concrete useful information about the column sorting attributes.

This seems fine to me. However I have doubts about the `DataGridSortColumn` naming since it's more like an input dto and not really an actual `Column` like it would seem. I think we've gonne through renaming before on this though haha

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
David-Moreira: I couldn't find anything about this. But from what I can tell it's slightly different.
`SortChanged` is invoked on a single sorting.
`SortingChanged` is invoked when various sorting is applied.

image

So on `SortOrder` you find specific single column attributes:
image
It still triggers the same as before, on `HandleSortColumn`.

On `SortingOrder` you find multiple column attributes:
image
It triggers on:
- `HandleSortColumn`
- `ApplySorting`
- `RemoveColumn`

Altought I'm thinking, if we really need this event... 
- On `ApplySorting` seems redudant, since it's an api explicitly called. (Altought there can be a benefit when used in generic places.)
- On `RemoveColumn`, kinda makes sense, that if a sort column gets removed, a notification happens, but this seems like an off case and could be handled by the regular `SortChanged`
- `HandleSortColumn` -  Redudant...Can be benefical since it notifies of all the columns sort state...

What are your thoughts?

stsrki: > On ApplySorting seems redudant, since it's an api explicitly called. (Altought there can be a benefit when used in generic places.)

I think we can leave this. It is still good to have.

> On RemoveColumn, kinda makes sense, that if a sort column gets removed, a notification happens, but this seems like an off case and could be handled by the regular SortChanged

Makes sense to call `SortChanged` when column is removed since it can and will affect the query that the user build for their Rest APIs.

> HandleSortColumn - Redudant...Can be benefical since it notifies of all the columns sort state...

Not sure what you mean by this. Be already had it before.

Source/Extensions/Blazorise.DataGrid/Models/DataGridSortColumn.cs
David-Moreira: So do you feel we should make up a convention for this type of input based arguments? Just that the name is similar to `DataGridColumn`, `DataGridDateColumn`, `DataGridCommandColumn`, etc which hold a different meaning, they are actual child `DataGridColumns` types of `DataGrid`

`DataGridSortColumnArgs`? `DataGridSortColumnDefinition`? `DataGridSortColumnInput`?

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
David-Moreira: Please resolve this if you're ok with this.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
David-Moreira: Renamed. Please resolve this if you're ok with the API.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
David-Moreira: > Makes sense to call SortChanged when column is removed since it can and will affect the query that the user build for their Rest APIs.

User should get correct sort columns whenever `ReadData` is called in the `ReadData` arguments. They usually don't base themselves on this event, I assume...
Anyway I think it's fine. But again, the event called is not the existing `SortChanged`, but the new `SortingChanged` (Maybe it would make sense to call the existing one also actually)

What we are considering here, is whether `SortingChanged` is an unrelevant redudant event to be called. Can lead to user confusion, `SortChanged` ... `SortingChanged` I just got myself, I had to go check the code when you mentioned these.

stsrki: The users will definitely be confused. I mean, both of us are already confused. What would happen in half a year from now?

If possible there should only be one event.

Source/Extensions/Blazorise.DataGrid/Models/DataGridSortColumn.cs
stsrki: In the past, I have created an `*Info` class. For example `DataGridColumnInfo` for `DataGridColumn` have simple data to be used by the grid without having full reference to the DataGridColumn. So the _Info_ is an OK suffix to have. The `Definition` is also used in some places.

David-Moreira: I believe we are using the `Info` as actual `*.Info` provided by events etc... That's how the `DataGridSortColumnInfo` was being used also.
image

in this case you're providing a method with an argument / input. So using Info here could be slightly different, but I guess we can do that. We're removing `SortingChanged` and can repurpose `DataGridSortColumnInfo`

Source/Extensions/Blazorise.DataGrid/DataGridColumn.cs
```diff
@@ -51,7 +51,7 @@ protected override void OnInitialized()
 
         if ( ParentDataGrid is not null )
         {
-            ParentDataGrid.AddColumn( this );
+            ParentDataGrid.AddColumn( this, true );
```
danm-de, line 54: Although tools like Resharper or Visual Studio  can display the argument names - GitHub cannot (yet). Does it maybe make sense to work with named arguments here?

```suggestion
            ParentDataGrid.AddColumn( this, suppressSortChangedEvent: true );

```

Source/Extensions/Blazorise.DataGrid/EventArguments/DataGridSortColumnInfo.cs
danm-de: There are currently three files in /Models:
- DataGridColumnInfo, namespace `Blazorise.DataGrid`
- DataGridRowInfo, namespace `Blazorise.DataGrid.Models`
- DataGridSortColumnInfo, namespace `Blazorise.DataGrid`

Is it intentional that the three files have partially different namespaces? If not - what would be the desired namespace?

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -894,7 +894,7 @@ public async Task ApplySorting( params DataGridSortColumnInfo[] columns )
 
         if ( SortMode == DataGridSortMode.Single )
         {
-            if ( columns.Length >= 1 )
+            if ( !columns.IsNullOrEmpty() )
```
David-Moreira: Isn't this different now? you'll never get to the `Count==1` You still need to check the Length **>=1**

David-Moreira: Nvm, it's a different collection, `SortByColumns` haha. Yea nice catch. :)

Source/Extensions/Blazorise.DataGrid/DataGridColumn.cs
David-Moreira: I personally don't really mind it. But we don't generally make a point to do it on the code base, so that'd be a bit inconsistent.
I agree it does provide more readability though. 

I personally disregard using because any argument name change then leads anyone who was using the named argument to also having to rename.

@stsrki any thoughts?

David-Moreira:
> ou

Hmm I see so it's a specific case you found where you're trying to sync up the page url with the grid sorting. However, I believe you can sync it up by using the new `GetSortByColumns()`, no? 

In order to fullfill more specific cases we were creating too much redudancy & possibly api confusion with this new one, at least in my opinion. But let's work to have the best of both worlds, can you see if `GetSortByColumns()` would be satisfactory?

Source/Extensions/Blazorise.DataGrid/DataGridColumn.cs
stsrki: I've been using it in the past on different projects.

But now, it is best to be consistent with the existing codebase.

Source/Extensions/Blazorise.DataGrid/EventArguments/DataGridSortColumnInfo.cs
David-Moreira: `DataGridRowInfo` namespace is wrong. we just keep everything under the base namespace, generally at least.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
danm-de: Double checked the `IsNullOrEmpty` implementation:

```csharp
    public static bool IsNullOrEmpty<T>( this IEnumerable<T> collection )
    {
        return collection is null || !collection.Any();
    }
``` 

The predicate `!columns.IsNullOrEmpty()` will be true if the column has at least one item.

David-Moreira: Yes yes I gotcha, I miss-read the `elseif` I tought it was accessing the same collection.

Source/Extensions/Blazorise.DataGrid/DataGridColumn.cs
David-Moreira: In that case let's disregard this suggestion, but thanks for pointing it out @danm-de

danm-de:
> > ou
> 
> Hmm I see so it's a specific case you found where you're trying to sync up the page url with the grid sorting. However, I believe you can sync it up by using the new `GetSortByColumns()`, no?
> 
> In order to fullfill more specific cases we were creating too much redudancy & possibly api confusion with this new one, at least in my opinion. But let's work to have the best of both worlds, can you see if `GetSortByColumns()` would be satisfactory?

Yes, the method `GetSortByColumns` will work. In terms of API design, I don't like the fact that I get multiple events (one per column) when calling `ApplySorting`.

David-Moreira:
> > > ou
> > 
> > 
> > Hmm I see so it's a specific case you found where you're trying to sync up the page url with the grid sorting. However, I believe you can sync it up by using the new `GetSortByColumns()`, no?
> > In order to fullfill more specific cases we were creating too much redudancy & possibly api confusion with this new one, at least in my opinion. But let's work to have the best of both worlds, can you see if `GetSortByColumns()` would be satisfactory?
> 
> Yes, the method `GetSortByColumns` will work. In terms of API design, I don't like the fact that I get multiple events (one per column) when calling `ApplySorting`.

That's what I believe makes sense given the current design. I.E not refactoring the api. 
A `SortChanged` is invoked everytime a sort happens on a column.

An option I can see is if we introduce the sorted columns in the `SortChanged` event args instead and in the case of `ApplySorting` call it only once at the end, in order to avoid single `ApplySorting` call leading to multiple calls of `SortChanged`.
But this will only be of benefit once `ApplySorting` is called which is already a consumer only callable api.

danm-de:
> > > > ou
> > > 
> > > 
> > > Hmm I see so it's a specific case you found where you're trying to sync up the page url with the grid sorting. However, I believe you can sync it up by using the new `GetSortByColumns()`, no?
> > > In order to fullfill more specific cases we were creating too much redudancy & possibly api confusion with this new one, at least in my opinion. But let's work to have the best of both worlds, can you see if `GetSortByColumns()` would be satisfactory?
> > 
> > 
> > Yes, the method `GetSortByColumns` will work. In terms of API design, I don't like the fact that I get multiple events (one per column) when calling `ApplySorting`.
> 
> That's what I believe makes sense given the current design. I.E not refactoring the api. A `SortChanged` is invoked everytime a sort happens on a column.
> 
> An option I can see is if we introduce the sorted columns in the `SortChanged` event args instead and in the case of `ApplySorting` call it only once at the end, in order to avoid single `ApplySorting` call leading to multiple calls of `SortChanged`. But this will only be of benefit once `ApplySorting` is called which is already a consumer only callable api.

Then let's try the current implemented approach. I think we can work around this :-)

Source/Extensions/Blazorise.DataGrid/Models/DataGridRowInfo.cs
```diff
@@ -48,7 +48,7 @@ public DataGridRowInfo( TItem item, IEnumerable<DataGridColumn<TItem>> columns )
     /// <param name="hasDetailRow">DetailRow evaluation result.</param>
     /// <param name="toggleable">If true toggles the detail row.</param>
     public void SetRowDetail( bool hasDetailRow, bool toggleable )
-        => this.hasDetailRow = ( toggleable && !this.hasDetailRow & hasDetailRow ) || ( !toggleable && hasDetailRow );
+        => this.hasDetailRow =  toggleable && !this.hasDetailRow & hasDetailRow  ||  !toggleable && hasDetailRow ;
```
David-Moreira, line 51: Wth? Did vs just remove parentheses here? Isn't this probably now logically diff? I'll put them back just in case... O.o

David-Moreira:
@stsrki merged with newest master, please review again. :)

Source/Extensions/Blazorise.DataGrid/EventArguments/DataGridSortChangedEventArgs.cs
```diff
@@ -12,19 +14,37 @@ public class DataGridSortChangedEventArgs : EventArgs
...
     /// <summary>
-    /// Gets the field name of the column that is being sorted.
+    /// Gets the field name of the column that is being sorted. The name is either <see cref="BaseDataGridColumn{TItem}.Field"/> or <see cref="DataGridColumn{TItem}.SortField"/> if specified.
     /// </summary>
     public string FieldName { get; }
 
+    /// <summary>
+    /// Gets the field name of the column that is being sorted. The name is always the same as in <see cref="BaseDataGridColumn{TItem}.Field"/>.
+    /// </summary>
+    public string ColumnFieldName { get; set; }
```
stsrki, line 46: What is the difference between `ColumnFieldName` and `FieldName`. Aren't they essentially the same thing? If `ColumnFieldName` is used for sorting then it should be named `SortByFieldName`.

David-Moreira: Is this a new api? If I remember correctly this api already exists so we shouldn't be making design decisions about jt now. We could however for 2.0
I think one represents the name of the column and the other the actual field used for sorting.

stsrki: Yes, it is a new API. It is marked as _add_ in the GitHub review mode

image

David-Moreira: I was under the impression that already existed.... Strange...
Still, what we'd like to actually rename is the one that already exists. Naming seems fine since we can't rename the other one wouldn't you agree?

stsrki: `SortByFieldName`?

David-Moreira: No, I'm pretty sure that's the `FieldName` concern. In which case we could rename it in **2.0**.
Original current api in master : 
image
image


`ColumnFieldName` seems fine to me, as it wants to convey the meaning that's the original column name. We might want to reword the documentation so it's clearer.

David-Moreira: @stsrki can you recheck if at least the documentation change for these two makes it more clearer now?

stsrki (approved):

PR #4638, opened 2023-03-01
DataGrid : Add StateHasChanged after IsLoading is set to true

David-Moreira:
https://blazorise.com/support/issues/96/DataGrid-LoadingTemplate-not-displayed-during-OnReadData

Missing `StateHasChanged` after `IsLoading` is set to true.
However, from my tests an async execution would still trigger the `LoadingTemplate` correctly. So Blazor was still smart in that regard. (Otherwise we'd certainly have more open issues because of `LoadingTemplate` not working)

Only way I could reproduce the **LoadingTemplate** not working was to block the async thread by appling a `Thead.Sleep`.
image
I guess the developer should make sure he's writting correct async code.

But by adding `Task.Yield` after `StateHasChanged` call we can force it to be definitely displayed for good measure. 

I'd rather not touch `HandleVirtualizeReadData` as it seems to be working fine and with no complaints.

### DIFF
--- a/Source/Extensions/Blazorise.DataGrid/DataGrid.razor
+++ b/Source/Extensions/Blazorise.DataGrid/DataGrid.razor
@@ -207,7 +207,7 @@
                 }
                 @if ( IsLoadingTemplateVisible )
                 {
-                    <_DataGridFullColumnSpanRow TItem="TItem" Columns="@Columns">
+                    <_DataGridFullColumnSpanRow TItem="TItem" Columns="@Columns" RenderUpdates>
                         @LoadingTemplate
                     </_DataGridFullColumnSpanRow>
                 }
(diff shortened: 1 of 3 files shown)


David-Moreira:
I decided to add onto this PR, the other issue.
User noted that the `ProgressBar` inside `LoadingTemplate` was not updating. That is because we were assuming the `LoadingTemplate` would be just static, and had `ShouldRender` set to false in `_DataGridFullColumnSpanRow`

Decided to add a new Parameter **RenderUpdates** where we can tell whether we want it to actually re-render.
If I remember correctly this was an optimization mainly because of the **Virtualize** component. I believe it's enough to just add it here as people can sometimes have progress bars.

stsrki:
How can a user use the new `RenderUpdates` parameter when `_DataGridFullColumnSpanRow` is an internal component?

David-Moreira:
> How can a user use the new `RenderUpdates` parameter when `_DataGridFullColumnSpanRow` is an internal component?

A User? This is for internal usage, if we want to set it ourselves. If you notice we set it ourselves for the standard `IsLoading`. Others, i.e in virtualize remain the same.

stsrki:
Gotcha. I had to expand the code review in the PR to see more code usage. I understand it now.

stsrki (approved):

PR #4650, opened 2023-03-04
Update CLA assistant to 2.2.1

David-Moreira:
Let's go ahead and update the cla assistant, see if it gets better.
One recurring issue is that it seems that recheck is not working properly. We always have to manually retrigger the action after the contributor has signed.

### DIFF
--- a/.github/workflows/cla.yml
+++ b/.github/workflows/cla.yml
@@ -12,7 +12,7 @@ jobs:
       - name: "CLA Assistant"
         if: (github.event.comment.body == 'recheck' || github.event.comment.body == 'I have read the CLA Document and I hereby sign the CLA') || github.event_name == 'pull_request_target'
         # Beta Release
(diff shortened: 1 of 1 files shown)


David-Moreira:
Actually was taking a look at the changes, there's the v2.3.0.
image

stsrki:
This is a different assistant, right? Does it use the same configuration settings and the same CLA file?

David-Moreira:
It's the same, they just renamed it.

David-Moreira:
https://github.com/marketplace/actions/cla-assistant-lite

David-Moreira:
The thing I noticed though, is that on the marketplace it doesn't show the **v2.3.0**. But it's been released 
image
 and on the marketplace example they do show **v2.3.0**.  
image

So there's a chance it might not be published to the marketplace? In which case we'll just set the **v2.2.1**.

stsrki (approved):
Let's merge and see how it goes 💪

PR #4662, opened 2023-03-09
Add IModalSharedContext interface and make it public

### LINKED ISSUE #4660: Unable to create tests for Modals due to dependency on injected internal ModalSharedContext
**Describe the bug**
Writing tests for own components using Modal is not possible due to the dependency on the injected class ModalSharedContext since it is marked as internal.

Ran into this when trying suggestion from @David-Moreira in issue: https://github.com/Megabit/Blazorise/issues/3327

**To Reproduce**
1. Create a component using Modal
2. In a test project using bUnit (like in Blazorise.Tests) create a test for the component in 1
3. Notice 3 different outcomes when rendering
a. If ModalSharedContext has not been added to services:
"System.InvalidOperationException : Cannot provide a value for property 'ModalContext' on type 'Blazorise.Modal'. There is no registered service of type 'Blazorise.ModalSharedContext'." 
b. If ModalShareContext is added to services:
"'ModalSharedContext' is inaccessible due to its protection level"
c. If Services.AddBlazorise() is used:
"System.InvalidOperationException : Cannot consume scoped service 'Microsoft.AspNetCore.Components.IComponentActivator' from singleton 'Bunit.Rendering.WebTestRenderer'."

**Expected behavior**
Be able in a test scenario to render a component that uses the Modal component.

**Screenshots**
Component:
image

Tests for cases A, B and C:
image

**Additional context**
Notice that all dependencies presented in Blazorise.Tests' BlazoriseConfig is either based on public interfaces or public classes. ModalSharedContext is the seemingly only dependency marked as internal and not behind an interface.

ModalSharedContext marked as internal: https://github.com/Megabit/Blazorise/blob/master/Source/Blazorise/Components/Modal/ModalSharedContext.cs#L6
Modal dependency on ModalSharedContext: https://github.com/Megabit/Blazorise/blob/master/Source/Blazorise/Components/Modal/Modal.razor.cs#L485
Blazorise allowing internals visible to Blazorise.Tests: https://github.com/Megabit/Blazorise/blob/master/Source/Blazorise/Properties/AssemblyInfo.cs#L10
Blazorise.Tests' setup of Blazorise Configuration: https://github.com/Megabit/Blazorise/blob/master/Tests/Blazorise.Tests/Helpers/BlazoriseConfig.cs#L40

stsrki:
Closes #4660

### DIFF
--- a/Source/Blazorise/Components/Modal/IModalSharedContext.cs
+++ b/Source/Blazorise/Components/Modal/IModalSharedContext.cs
@@ -3,33 +3,22 @@
 /// <summary>
 /// Holds the shared context for all the modals in current application state.
 /// </summary>
-public class ModalSharedContext : IModalSharedContext
+public interface IModalSharedContext
 {
     /// <summary>
     /// Raises the modals <see cref="OpenIndex"/> and returns the new index.
     /// </summary>
     /// <returns>Returns the new open index.</returns>
-    public int RaiseModalOpenIndex()
-    {
-        return ++OpenIndex;
-    }
+    int RaiseModalOpenIndex();
 
     /// <summary>
     /// Decrease the modals <see cref="OpenIndex"/> and returns the new index.
     /// </summary>
     /// <returns>Returns the new open index.</returns>
-    public int DecreaseModalOpenIndex()
-    {
-        --OpenIndex;
-
-        if ( OpenIndex < 0 )
-            OpenIndex = 0;
-
-        return OpenIndex;
-    }
+    int DecreaseModalOpenIndex();
 
     /// <summary>
     /// Gets or sets the modal open index.
     /// </summary>
-    public int OpenIndex { get; private set; }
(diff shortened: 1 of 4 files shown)


David-Moreira:
I guess the real problem here, is that we were using DI for an internal member. But I guess it's necessary to easily get the benefit of scoping it. So from a api design perspective maybe we shouldn't have to change it to public on purpose. 
But I guess it's the easier route and there's no real harm on it, so LGTM.

David-Moreira (approved):

stsrki:
> I guess the real problem here, is that we were using DI for an internal member. But I guess it's necessary to easily get the benefit of scoping it. So from a api design perspective maybe we shouldn't have to change it to public on purpose. But I guess it's the easier route and there's no real harm on it, so LGTM.

This should get easier once we do the Blazorise packages for testing.

PR #4678, opened 2023-03-16
Playwright E2E tests

### LINKED ISSUE #4652: Investigate playwright as replacement for Selenium
Investigate playwright as replacement for Selenium.

Make sure it can be setup on github actions successfully.

David-Moreira:
Closes #4652


David-Moreira:
Note to self:
Pipeline seems to be logging too much info from the wasm app. Investigate if it can be removed.

Tests/Blazorise.E2E.Tests/ReadMe.md
```diff
@@ -0,0 +1,39 @@
+# Playwright E2E Test Project
+
+This test project uses Playwright with NUnit.
+NUnit was chosen since there is already an helpful playwright helper package for NUnit.
+
+Below you will find instructions on how to use Playwright. This assumes you will be using the command line and that you will be positioned in the root of the E2E tests project.
+
+## Install
+- Build the project, either by using the dotnet build command or Visual Studio.
+- Execute the playwright powershell script to install necessary dependencies (browser dependencies, etc...), i.e: `powershell .\bin\Debug\net7.0\playwright.ps1 install`
```
stsrki, line 10: Can this be done automatically without needing to run powershell?

David-Moreira: Well, this is a one time thing to initially install the dependencies, but maybe you can add a build post step in the csproj, no idea if those are still a thing in dotnet core.

David-Moreira: But I can see a problem on having it run automatically. If it throws an error running the script (i.e needs admin or policy modification) then the user might be puzzled to what's happening.

David-Moreira:
> Can we make it so that running tests is easy as clicking on _Run All_ in the test explorer?
> 
> image
> 
> I would prefer not to deal with any terminal at all.

And it should be. Only manual step is the one for the dependencies.
Then uou just need to click **RunAll** everytime. And if you know the API by heart/docs you don't even need to use playwright tools, just write the tests. It's basically the same as happens on the CICD pipeline.
By the way, why are they failing for you? Can you post the error?

David-Moreira:
hmm time to test the trace :D
Can you activate it for that particular test and send it to me over discord?
// Start tracing before creating / navigating a page.
await context.Tracing.StartAsync(new()
{
  Screenshots = true,
  Snapshots = true,
  Sources = true
});

// Test code

// Stop tracing and export it into a zip archive.
await context.Tracing.StopAsync(new()
{
  Path = "trace.zip"
});

https://playwright.dev/dotnet/docs/trace-viewer#recording-a-trace

Tests/Blazorise.E2E.Tests/ReadMe.md
David-Moreira: Don't think it's worth chasing this. let me know if you agree.

stsrki: Yes, leave it for now.

stsrki:
Still failing

image

David-Moreira:
> Still failing
> 
> image

Can you send an updated trace? Ovr o ndiscord? thanks!

stsrki:
Should be skip the NumericEdit test that gives us a problem?

David-Moreira:
> Should be skip the NumericEdit test that gives us a problem?

Well I'd like to fix it or still test it the same way, but since I can't reproduce the problem you have on my machine or on the pipeline, I can't really figure it out.

But maybe we can disable for now and attempt again in a few months, maybe some update or some configuration might be fixed by then, whatever it is that causes the problem... (I think we cover the same test in bunit)

It will be helpful merging this in soon, as I'd like to try and start using this as a tool when investigating / bug fixing.

stsrki:
Ok we can disable it.

What do you think we remade all of the tests on 1.2 branch instead of master? We still do the support and 1.2 pushes fail all the time for bUnit.

David-Moreira:
> Ok we can disable it.
> 
> What do you think we remade all of the tests on 1.2 branch instead of master? We still do the support and 1.2 pushes fail all the time for bUnit.

Oh right, that'll probably be helpful.
I'll just rebase and retarget later.

Yea I'm hopefully fixing those as I see them, so let me just know if you see a new one happening after it's merged in.

David-Moreira:
recheck

David-Moreira:
Hmm, I'm gonna redo PR.

PR #4693, opened 2023-03-23
Feature: RichTextEdit based on roosterjs

njannink:


njannink:
Still lot of work to do, but it's easy to create a rooster editor. I added a utility function for loading scripts on demand. Some similar logic is also present in the old RichTextEdit, but I think this way is more elegant

stsrki:
By a quick glance at the code, it seems like a clean solution.

njannink:
I will do another push soon to show how it works. roosterjs doesn't have any toolbar functionality. So you have to build your own bar. But that will give all the styling and customization options to the user so I think its actually good. Keybindings do work out of the box like Ctrl+B etc

```
<Buttons>
   <Button Clicked="() => rte.Format(Action.Bold)">B</Button>
   <Button Clicked="() => rte.Format(Action.Italic)">I</Button>
   <Button Clicked="() => rte.Format(Action.Underline)">U</Button>
</Buttons>
<RichTextEdit @ref="rte" />
```

njannink:
this is still on my todo list. I hope to be able to pick this up again sometime soon to be able to merge away from QuillJS

stsrki:
@njannink What do you think of this? https://github.com/Megabit/Blazorise/issues/5077

njannink:
@stsrki I have been working a bit on this again. Have a look at the command system that I added. This will allow users to bind buttons, drop-downs etc to formatting command and gives you the option to build your own toolbar. Maybe we could ship a few predefined toolbars/layouts to make it easier to start using.

Still quite some work to do and some commands don't work somehow

njannink:
> You made a lot of work this time. Thanks. Will look into it.
> 
> Btw, I was thinking a lot about RTE, and I think for the future, maybe in Blazorise v2 it would be best to organize RTE similar to core Blazorise. And then have each implementation separate of the core project.
> 
> eg.
> 
> * Blazorise.RichTextEdit
> * Blazorise.RichTextEdit.Quill
> * Blazorise.RichTextEdit.Rooster

I have looked at this a bit, but there is limited common ground between the 2 editors. QuillJS really comes with a buildin toolbar etc and Rooster doesn't. So yes it might be good to rename the current RichTextEdit to RichTextEdit.Quill, but I think that a common library doesn't add that much.

njannink:
closing this old PR. will start new one when all issues resolved

PR #4702, opened 2023-03-29
Observe list in TreeView

### LINKED ISSUE #4635: Treeview: Cannot dynamically refresh the tree when adding or removing nodes
**Describe the bug**

When I try to add a new node, the tree doesn't refresh

test

**To Reproduce**

```csharp
<button class="btn btn-primary" @onclick="AddNewNode">Add new item</button>

<TreeView Nodes="Items"
          GetChildNodes="@(item => item.Children)"
          HasChildNodes="@(item => item.Children?.Any() == true)"
          @bind-SelectedNode="selectedNode"
          @bind-ExpandedNodes="expandedNodes">
    <NodeContent>
        <Icon Name="IconName.Folder" />
        @context.Text
    </NodeContent>
</TreeView>
@code{
    public class Item
    {
        public string Text { get; set; }
        public IEnumerable<Item> Children { get; set; }
    }

    async Task AddNewNode()
    {
        Items.Add(new Item
            {
                Text = "newNode"
            });

       await InvokeAsync(StateHasChanged);
    }

    List<Item> Items = new List<Item>
    {
        new Item { Text = "Item 1" },
        new Item
        {
            Text = "Item 2",
            Children = new []
            {
                new Item { Text = "Item 2.1" },
                new Item
                {
                    Text = "Item 2.2",
                    Children = new []
                    {
                        new Item { Text = "Item 2.2.1" },
                        new Item { Text = "Item 2.2.2" },
                        new Item { Text = "Item 2.2.3" },
                        new Item { Text = "Item 2.2.4" }
                    }
                },
                new Item { Text = "Item 2.3" },
                new Item { Text = "Item 2.4" }
            }
        },
        new Item { Text = "Item 3" },
    };

    IList<Item> expandedNodes = new List<Item>();
    Item selectedNode;
```

stsrki:
Closes #4635

**Test**

```razor
@using System.Collections.ObjectModel;
<TreeView Nodes="Items"
          GetChildNodes="@(item => item.Children)"
          HasChildNodes="@(item => item.Children?.Any() == true)"
          @bind-SelectedNode="selectedNode"
          @bind-ExpandedNodes="expandedNodes">
    <NodeContent>
        <Icon Name="IconName.Folder" />
        @context?.Text
    </NodeContent>
</TreeView>

<Button Clicked="OnClick" Color="Color.Primary">Add node</Button>


<p><b>Tree top items</b></p>

@foreach (var item in Items)
{
    <p>@item.Text</p>
}

@code {

    public class Item
    {
        public string? Text { get; set; }
        public IEnumerable<Item> Children { get; set; }
    }

    ObservableCollection<Item> Items = new()
    {
        new Item { Text = "Item 1" },
        new Item
        {
            Text = "Item 2",
            Children = new []
            {
                new Item { Text = "Item 2.1" },
                new Item
                {
                    Text = "Item 2.2",
                    Children = new []
                    {
                        new Item { Text = "Item 2.2.1" },
                        new Item { Text = "Item 2.2.2" },
                        new Item { Text = "Item 2.2.3" },
                        new Item { Text = "Item 2.2.4" }
                    }
                },
                new Item { Text = "Item 2.3" },
                new Item { Text = "Item 2.4" }
            }
        },
        new Item { Text = "Item 3" },
    };

    IList<Item> expandedNodes = new List<Item>();
    Item selectedNode;

    private Task OnClick()
    {
        Items.Add(
            new Item { Text = "Buggy item!" }
        );

        return Task.CompletedTask;
    }
}
```


Source/Extensions/Blazorise.TreeView/Extensions/NodeStateExtensions.cs
```diff
@@ -24,5 +25,23 @@ internal static class NodeStateExtensions
                 yield return new TreeViewNodeState<TNode>( node, hasChildren, expanded( node ) );
             }
         }
+
+        public static async IAsyncEnumerable<TreeViewNodeState<TNode>> ToNodeStates<TNode>( this IList nodes,
```
David-Moreira, line 29: Shouldn't just check / cast the IList to an IList of TNode and call the `IEnumerable` version?

Source/Extensions/Blazorise.TreeView/TreeView.razor.cs
```diff
@@ -58,6 +57,39 @@ await foreach ( var nodeState in paramNodes.ToNodeStates( HasChildNodesAsync, Ha
                 treeViewNodeStates.Add( nodeState );
             }
         }
+
+        if ( paramNodes is INotifyCollectionChanged observableCollection )
```
David-Moreira, line 61: Great, seems like it's easy to support Obesrvables, we should do the same for DataGrid. Try for 1.3?

Source/Extensions/Blazorise.TreeView/TreeView.razor.cs
```diff
@@ -58,6 +57,39 @@ await foreach ( var nodeState in paramNodes.ToNodeStates( HasChildNodesAsync, Ha
...
+        {
+            observableCollection.CollectionChanged -= OnCollectionChanged;
+            observableCollection.CollectionChanged += OnCollectionChanged;
+        }
+    }
+
+    private void OnCollectionChanged( object sender, System.Collections.Specialized.NotifyCollectionChangedEventArgs e )
+    {
+        if ( e.Action == NotifyCollectionChangedAction.Add || e.Action == NotifyCollectionChangedAction.Remove || e.Action == NotifyCollectionChangedAction.Reset )
```
David-Moreira, line 70: My knowledge of `INotifyCollectionChanged` is limited, but should even the `NotifyCollectionChangedAction.Reset` run the same logic? Is that not something like calling .Clear on a List? Or am I wrong?

Source/Extensions/Blazorise.TreeView/TreeView.razor.cs
```diff
@@ -58,6 +57,39 @@ await foreach ( var nodeState in paramNodes.ToNodeStates( HasChildNodesAsync, Ha
                 treeViewNodeStates.Add( nodeState );
             }
         }
+
+        if ( paramNodes is INotifyCollectionChanged observableCollection )
+        {
+            observableCollection.CollectionChanged -= OnCollectionChanged;
+            observableCollection.CollectionChanged += OnCollectionChanged;
```
David-Moreira, line 64: Why do you need to remove and add? Isn't there something more elegant for this?
Can we improve this to where it just does this work once?

David-Moreira (changes requested):

David-Moreira:
You don't need to explicitly call `StateHasChanged();` here:
image
Right? It was just out of habbit? Should work regularly as part of `Eventcallback` execution?

Source/Extensions/Blazorise.TreeView/TreeView.razor.cs
stsrki: There is `|| e.Action == NotifyCollectionChangedAction.Reset` check also.

stsrki:
> You don't need to explicitly call `StateHasChanged();` here: image Right? It was just out of habbit? Should work regularly as part of `Eventcallback` execution?

This was code from the user that I just C/P.

Source/Extensions/Blazorise.TreeView/TreeView.razor.cs
stsrki: This still needs improvements. We need to figure out a way to unsubscribe events if the `observableCollection` reference changes.

Source/Extensions/Blazorise.TreeView/TreeView.razor.cs
stsrki: Why not.

Source/Extensions/Blazorise.TreeView/TreeView.razor.cs
David-Moreira: I guess you'd need to check the backing field always for the old collection vs the new :S.
You're saying an unlikely case where people start with an obervable collection but change it midway to non observable rght?

stsrki: I'm saying if a user creates a new list.

`<TreeView Nodes="myList">`

and

then

`myList = new ();`

David-Moreira: What if you always unsubscribe before `await base.SetParametersAsync( parameters );` is called and the collection is mutated with a different one altogether? Doesn't that fix that use case?

stsrki: 🤔🤔... that's... actually a good idea.

stsrki: Updated

Source/Extensions/Blazorise.TreeView/TreeView.razor.cs
stsrki: Clear is still the same as removing elements, just that it removes them all at once. So yes, it should be here.

Source/Extensions/Blazorise.TreeView/Extensions/NodeStateExtensions.cs
stsrki: It's not a generic type

image

David-Moreira: That's strange, what's the original type the debugger tells you? O.o I'm pretty sure you should be able to get it to work with a generic collection.

You also have stuff like this:
https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.cast?view=net-8.0

Anyway it isn't a big deal otherwise both methods are easily viewable in the same file, doubt we'd make the mistake to only update one in the future if any refactor was needed.

stsrki: image

David-Moreira: Ahh, it stems from the collection notifier already casting down. Damn those APIs from Windows forms eras... :D 

So in my opinion the change should be something like 
`e.NewItems.Cast<TNode>().ToNodeStates`

or `OfType`

Source/Extensions/Blazorise.TreeView/TreeView.razor.cs
```diff
@@ -46,6 +46,13 @@ public override async Task SetParametersAsync( ParameterView parameters )
             treeViewState = treeViewState with { SelectedNodes = paramSelectedNodes };
         }
 
+        // We don't want to have memory leak so we need to unsubscribe any previous event if it exist.
+        // The unsubscribe must happen before SetParametersAsync.
+        if ( Nodes is INotifyCollectionChanged observableCollection1 )
```
David-Moreira, line 51: Maybe the collection variable name could be obseravableCollectionBeforeChange and observableColleftionAfterChange. Makes it a bit more clear if someone does. Not understand they might eb two different collections.

David-Moreira (approved):
Lgtm

PR #4715, opened 2023-04-04
Add ProductToken to BlazoriseOptions

### LINKED ISSUE #4712: ProductToken in BlazoriseOptions
We need to introduce new **ProductToken** in **BlazoriseOptions**. And make **LicenseKey** property obsolete.

stsrki:
Closes #4712

### DIFF
--- a/Demos/Blazorise.Demo.AntDesign/Program.cs
+++ b/Demos/Blazorise.Demo.AntDesign/Program.cs
@@ -17,7 +17,7 @@ public class Program
         var builder = WebAssemblyHostBuilder.CreateDefault( args );
 
         builder.Services
-            .SetupDemoServices( builder.Configuration["Licensing:LicenseKey"] )
+            .SetupDemoServices( builder.Configuration["Licensing:ProductToken"] )
             .AddAntDesignProviders()
             .AddFontAwesomeIcons();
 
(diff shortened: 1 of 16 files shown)


David-Moreira:
LGTM, doesn't seem like you've missed anything.
I was just thinking from looking at the code, do we actually stop the background worker / dispose of it? Would that not be interesting?

stsrki:
As far as I know, it stops by itself. It is a fire once operation.

David-Moreira:
> As far as I know, it stops by itself. It is a fire once operation.

Maybe I'm confusing with some other api? I was under the impression it would keep calling back until you explicitly tell it to stop. otherwise why would you have the initialized flag checked if it's one time run?

David-Moreira:
Well it does seem to be one time only from the docs, so nvm, I must be confusing.
image

Thanks.

David-Moreira:
I was confusing with this one:
https://learn.microsoft.com/en-us/aspnet/core/fundamentals/host/hosted-services?view=aspnetcore-7.0&tabs=visual-studio

image

stsrki:
I think I tried that. But WASM has limited support for async code during the startup. That's why I had to fallback into the old BackgroundWorker

PR #4722, opened 2023-04-06
Fix DropZone index

### LINKED ISSUE #4383: Incorrect `IndexInZone` in Drag & Drop (reorder)
**Describe the bug**
I want to sort my elements in a list based on the reordering of elements. The `ItemDropped`-event fires but the index seems not always correct.

**To Reproduce**
I took your example and simplified it a bit.

```razor
<DropContainer TItem="DropItem" Items="@items" ItemsFilter="@((_, _) => true)" ItemDropped="@ItemDropped" Flex="Flex.Wrap.Grow.Is1">
    <ChildContent>
        <DropZone TItem="DropItem" Name="d1" AllowReorder Padding="Padding.Is3" Margin="Margin.Is3" Flex="Flex.Grow.Is1">
            <Heading Size="HeadingSize.Is4" Margin="Margin.Is3.FromBottom">Drop Zone</Heading>
        </DropZone>
    </ChildContent>
    <ItemTemplate>
        <Card Shadow="Shadow.Default" Margin="Margin.Is3.OnY">
            <CardBody>
                @context.Name
            </CardBody>
        </Card>
    </ItemTemplate>
</DropContainer>

@lastIndexInDropZone

@code {
    private int? lastIndexInDropZone;

    public class DropItem
    {
        public string Name { get; init; } = "";
    }

    private List<DropItem> items = new()
    {
        new DropItem() { Name = "Item 1" },
        new DropItem() { Name = "Item 2" },
        new DropItem() { Name = "Item 3" },
        new DropItem() { Name = "Item 4" },
        new DropItem() { Name = "Item 5" },
    };

    private void ItemDropped(DraggableDroppedEventArgs<DropItem> dropItem)
    {
        lastIndexInDropZone = dropItem.IndexInZone;
    }
}
```

Take a look at `lastIndexInDropZone` while moving the parts around.

E.g. take "__Item 1__" and move it to the very beginning of the Drop Zone (heading). The placeholder appears (this should not be the case while reordering since it's already the first element).
`lastIndexInDropZone = 0`

Then move "__Item 1__" to the middle of itself. Here, (correctly) no placeholder element shows up (but the event is triggerd nonetheless).
`lastIndexInDropZone = 1`

Move "__Item 1__" to the space between "__Item 2__" and "__Item 3__"
`lastIndexInDropZone = 1`
And that's the error (since it's the same index as above). I have no chance to distinguish between these two.

The code is https://github.com/Megabit/Blazorise/blob/6d63d2a62360df91ad74798b815adc8fe9c3db46/Source/Blazorise/Components/Droppable/DropContainer.razor.cs#L78-L81
and I think I can imagine what you try to do - but I also think it's not fully working.
In my opinion, you should just leave the index as is (in my tests, that will work). The `IndexInDropZone` is then _always_ the index of the space between the elements _before_ any reordering (or insertion) happens. Like this
image

Then I have a chance to handle the different variations in the event.

Alternatively, you have to stop triggering the event whenever the element is moving to itself. Don't know exactly if this is a sufficient fix or not! While I know you don't want to break things and this seems like a legit and easy fix, I think the current index handling is at least a bit confusing. Why should "__Index = 2__" (from the image above) be `IndexInDropZone = 1` if I reorder downwards and `IndexInDropZone = 2` if I reorder upwards while not knowing in which direction the reordering happens at all? Thus, this alternative would be "ok", I guess, but still leaves us the confusing part.

**Expected behavior**
- The index should be consistent over all operations
- While reordering, the placeholder should not appear above and below the current element (even for the first element)
- Consider to not trigger events if no reordering happens
  this could easily be handled if the index is consistent - and maybe there's even a use case where one wants to know if the element is sorted to the same position
- Consider to add a source and target index property (so at end, it's obvious where the element comes from (source drop zone name and index) and where it should be later on (target drop zone name and index) - see also additional context).

In order to not break things, my suggestion is to add new properties, so at the end there are `SourceDropZoneName`, `SourceDropZoneIndex` (new), `DropZoneName`, and `DropZoneIndex` (the new, unmodified index from above code).
`IndexInZone` is provided as is - maybe marked as obsolete.

**Additional context**
Beside that, in order to reorder elements you need the current index of the element in the list. You can get that with `List<T>.IndexOf`, but that will only work if each element appears only once in the list (otherwise you always only get the first instance (think of value types or a list with mulitple equal references)).
So `List<>.IndexOf` is no good solution - therefore it would be great if you could provide the source index as well.

stsrki:
Closes #4383


Source/Blazorise/Components/Droppable/DropContainer.razor.cs
```diff
@@ -75,7 +75,7 @@ public async Task CommitTransaction( string dropZoneName, bool reorderIsAllowed
         {
             index = GetTransactionIndex() + 1;
 
-            if ( transaction.SourceZoneName == transaction.CurrentZoneName && IsItemMovedDownwards )
+            if ( transaction.SourceZoneName == transaction.CurrentZoneName && IsItemMovedDownwardsOrSame )
             {
                 index -= 1;
```
David-Moreira, line 80: There isn't much to review. I'm not sure what this code does but just to make sure...
On line 76 you have increased the index by 1 and on line 80 you decrease it again by 1. Will it not pretty much stay the same as GetTransactionIndex? Or is that expected? Also if it stays the same. Maybe the logic should change to be more clear where it really only increases by 1 given a certain condition. 
And does it need to be set to some internal state?

stsrki: `+1` is because once the item is dropped, it is assumed it will be placed after the currently focused item.

David-Moreira: but if you do 
**index = GetTransactionIndex() + 1;**
and then
**index -= 1;**
It will just go back to the original transaction index, and seems harder to reason? Which means, maybe the logic could be simplified to a single if, or something?

i.e: 
`index = GetTransactionIndex();`
`if (logic)`
`  index += 1; `

stsrki: Yeah, that seems better.

PR #4733, opened 2023-04-12
Prevent large performance hits after using the DropContainer/DropZone component

danm-de:
We noticed that after using the `DropContainer`/`DropZone` component all pages having (large) tables became very slow. As soon as the JS module has been initialized all table rows call the destroyThrottledDragEvents method (drapDrop.js:37) on `DisposeAsync`. If the table has 200 rows, the method is called at least 200 times (via RPC/SignalR Blazor server side <-> Browser).

The problem persists until the user opens a new session - reason: `JSDragDropModule` is registered as a scoped service and therefore holds state.

image

### DIFF
--- a/Source/Blazorise/Base/BaseDraggableComponent.cs
+++ b/Source/Blazorise/Base/BaseDraggableComponent.cs
@@ -22,6 +22,11 @@ public abstract class BaseDraggableComponent : BaseComponent, IDisposable, IAsyn
     /// </summary>
     private DotNetObjectReference<BaseDraggableComponent> dotNetObjectRef;
 
+    /// <summary>
+    /// Is <c>true</c> if the component needs to clean up the JS ThrottleDragEvent handler on <see cref="Dispose"/> or <see cref="DisposeAsync"/>.
+    /// </summary>
+    private bool jsEventsInitialized;
+
     #endregion
 
     #region Methods
@@ -33,6 +38,7 @@ public abstract class BaseDraggableComponent : BaseComponent, IDisposable, IAsyn
         {
             dotNetObjectRef = CreateDotNetObjectRef( this );
             await JSDragDropModule.InitializeThrottledDragEvents( ElementRef, ElementId, dotNetObjectRef );
+            jsEventsInitialized = true;
         }
         await base.OnFirstAfterRenderAsync();
     }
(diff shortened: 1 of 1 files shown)


David-Moreira:
Hello @danm-de can you briefly explain how does this solve the problem.
If I understand correctly this will not invoke the destroy if initialize hasn't been called which is a good approach, and we could take the idea and apply it globally.

However, isn't the problem still there if the initialize has already been called?

danm-de:
Hi, thanks for the fast reply. I'm trying to sum it up. My colleague and I actually had a lot of trouble figuring out exactly what was happening.

1. The `DropZone` component calls `JSModule.Initialize`. The property `JSModule` is an injected type of `IJSDragDropModule` (registered as scoped service).
2. The `JSDragDropModule` calls `BaseJSModule.InvokeSafeVoidAsync`. From here it gets complicated. The method awaits its own property called `Module`. The property getter calls `GetModule()` which caches a `IJSObjectReference` in a private field `moduleTask`. This private field holds state and is therefore shared across all instances that use the `JSDragDropModule`.

The Blazorise `Table` component is inherited from `BaseDraggableComponent`. Normally it does not initialize the `JSDragDropModule`. The problem starts during `Dispose` / `DisposeAsync`:
1. When `Dispose/DisposeAsync` is called -> `JSDragDropModule.DestroyThrottledDragEvents` is invoked.
2. The `DestroyThrottledDragEvents` has a pre-check -> It checks the `IsUnsafe` property **and does nothing if the javascript module has not been initialized**. 

When our users visit a page that contains the drap&drop component, this will initialize the `JSDragDropModule`. At this point, ALL tables suddenly start calling JS methods during Dispose. 
We're using Blazor server-side and these Invoke-JS calls result in high CPU peaks and heavy SignalR traffic between the server and the user's browser. This additional network traffic (RPC calls for each table row) makes it worse, especially on slow network connections. It makes a difference if you have a latency of a few milliseconds or a multiple of 100ms (bandwidth != latency).

danm-de:
The check we added calls `JSDragDropModule.DestroyThrottledDragEvents` only if the `<Table>` has been initialized with `Draggable=true` (or `Drag/DragOver` is set).

David-Moreira:
Ahhh the key thing here is:

>  Normally it does not initialize the JSDragDropModule.

Note that in fact, it is not initialized tipically if you've not set it to draggable or bound drag events that need js.
image

The dispose process, should indeed not care about trying to dispose of something that's certainly not been initialized.
The PR is looking sound in that case.
Any objections @stsrki ?

danm-de:
> Ahhh the key thing here is:
> 
> > Normally it does not initialize the JSDragDropModule.

Yes. I guess all components inherited from `BaseDraggableComponent` have the same problem, right?

stsrki (approved):
Looks good to me. Thanks for your contribution @danm-de.

PR #4747, opened 2023-04-20
Fixed Autocomplete Disabled state

### LINKED ISSUE #4720: Disabling Autocomplete does not work correctly
**Describe the bug**
Disabling an Autocomplete does not "turn it grey", only disables being clickable. Also if Multiselect is enabled, the x on the selected values can still be clicked

**To Reproduce**
see example on index page in repo
https://github.com/oaldrian/BlazoriseReproductions/tree/AutocompleteDisabling

**Expected behavior**
Greying out the autocomplete, and disabling (or removing) the x on selected values.

steal4life:
Closes #4720


oaldrian (commented):
This would change the colour of the selected text badges, which is already great. But the input itself does stay default colored, which is a disconnect from other inputs which are grey when disabled. Or am I missing something in the code entirely? 
Fyi: just commenting as I reported the issue

David-Moreira:
Hey,

I think @oaldrian is on to something here,
can we understand why don't we follow the same look for disabled on the text input for the autocomplete? Might be conflicting css?

**Standard disabled:**
image

**Current Autocomplete disabled:**
image

I'm guessing the badge color is fine though.

---

@stsrki I was just going through the apps and the default look for tailwind seems off, is this expected?
https://tailwinddemo.blazorise.com/
image

stsrki:
My guess is that since TextEdit is part of the Dropdown, it somehow overrides its CSS styles. It definitely need to look more like regular disabled inputs. 

---

> @stsrki I was just going through the apps and the default look for tailwind seems off, is this expected?

I haven't worked much on dark themes so it needs some adjustments.

David-Moreira:
@stsrki news on this?

David-Moreira:
@stsrki Heyia, should probably take a look at this anytime soon. Shouldn't keep bug fixes waiting if possible.

steal4life:
I tested it and it seems to me that only when it comes to the autocomplete multiple disabled parameter it doesn't work as it should, also unrelated to that I think maybe the badges should be dark because you can see the contrast better, @stsrki what do you think?

steal4life:
> I tested it and it seems to me that only when it comes to the autocomplete multiple disabled parameter it doesn't work as it should, also unrelated to that I think maybe the badges should be dark because you can see the contrast better, @stsrki what do you think?

when autocomplete is disabled ofc

Source/Extensions/Blazorise.Components/Autocomplete.razor.cs
```diff
@@ -1091,6 +1091,9 @@ protected string CssStyle
             if ( Style != null )
                 sb.Append( Style );
 
+            if ( Disabled != false )
+                sb.Append( "background-color: #e9ecef;" );
```
stsrki, line 1095: Why exactly is this color hardcoded for the disabled state? What will happen when the user has dark-mode, for example in Tailwind?

steal4life:

stsrki (approved):

PR #4769, opened 2023-05-14
WIP | DataGrid | Observable

### LINKED ISSUE #4706: Datagrid : Observable support | Do Not mutate Dirty Data directly in parameter.
Great, seems like it's easy to support Obesrvables, we should do the same for DataGrid. Try for 1.3?

_Originally posted by @David-Moreira in https://github.com/Megabit/Blazorise/pull/4702#discussion_r1152519913_

David-Moreira:
Closes #4706

--- 
**Notes:** 

Started by taking a look at #4406 which is where we first wanted to go with the `Data` auto property to avoid `DirtyFilter` every single time.
I noticed that the original improvement to reduce number of renders was slightly higher again. Due to now calling FocusAsync when selecting a row. This is for the `Navigable` feature, and it does not make sense to be called if it's not active.

While doing this, since it's not pratical to go test manually everytime, I wanted to figure a way to have tests in place so we can figure out this kind of perf regressions, bunit does have an OnAfterRender event, but since it mocks js calls out, it's not 100% representative of real world usage... so went with playwright to test it... so trying to prototype a simple approach to keep track of renders, it's kinda messy & not sure if this is the best approach...but probably better then nothing.

From researching I couldn't find someone who has a good approach at this yet.

### DIFF
--- a/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
+++ b/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
@@ -775,9 +775,12 @@ public partial class DataGrid<TItem> : BaseDataGridComponent
     {
         await SelectRow( item );
 
-        var selectedTableRow = GetRowInfo( item )?.TableRow;
-        if ( selectedTableRow is not null )
-            await selectedTableRow.ElementRef.FocusAsync();
+        if ( Navigable )
+        {
+            var selectedTableRow = GetRowInfo( item )?.TableRow;
+            if ( selectedTableRow is not null )
+                await selectedTableRow.ElementRef.FocusAsync();
+        }
 
         await Refresh();
     }
--- a/Tests/BasicTestApp.Client/DataGridPerfComponent.razor
+++ b/Tests/BasicTestApp.Client/DataGridPerfComponent.razor
@@ -1,0 +1,110 @@
+﻿@using BasicTestApp.Client.PerfComponents;
+
+<DataGridPerf @ref="PerfRef"
+              TItem="Employee"
+              Data="@Data"
+                ShowPager
+                ShowPageSizes
+                Sortable
+              SortMode="DataGridSortMode.Single"
+                Editable
+              EditMode="@DataGridEditMode">
+    <DataGridColumns>
+        <DataGridCommandColumn TItem="Employee">
+            <NewCommandTemplate>
+                <Button ElementId="btnNew" Color="Color.Success" Clicked="@context.Clicked">@context.LocalizationString</Button>
+            </NewCommandTemplate>
+            <EditCommandTemplate>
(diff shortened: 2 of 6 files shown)


stsrki:
I stumbled on a Reddit post that has tracking for a number of rerenders. Maybe it can help you 

https://www.reddit.com/r/Blazor/comments/13gc0rv/i_made_a_render_tracking_dialog_and_i_think_it_is/

David-Moreira:
> I stumbled on a Reddit post that has tracking for a number of rerenders. Maybe it can help you
> 
> https://www.reddit.com/r/Blazor/comments/13gc0rv/i_made_a_render_tracking_dialog_and_i_think_it_is/

I've taken a brief look. There's no web project included, so I can't really tell how it's being used, and it also uses Fluxor which I'm not used to. 

So I'm not really sure how applicable it is. But the example gif the guy shows is nice, if it's a simple setup.

But I would say it acts more kinda like a logging system, so it would be somewhat manual...
image

I suspect you add `Dispatch(AddEntryAction)` to each method which is kinda cumbersome, anyway might just be my ignorance on how fluxor lib works. Let's watch the thread see if he adds better info.

David-Moreira:
@stsrki Please review at your earliest convenience.

I just added those tests for the regular selections for now.
I also took a brief look and I feel like the underlying components also can be improved re-render wise. We'll investigate further in the future, `DataGrid` already feels pretty fast I feel like.

`Observable` working great based on your `TreeView` implementation.
Seems an easy enough migration for anyone using .Add/.Remove type of implementations.

stsrki (commented):
Seems ready. Basically the same as what we have on TreeView.

Can you just check why we now have a failing unit test?

David-Moreira:
> Seems ready. Basically the same as what we have on TreeView.
> 
> Can you just check why we now have a failing unit test?

Basically the edit test is mutating an item from the data. Since we now don't magically `DirtyFilter()`, an explicit call to `Reload` needs to be done.

David-Moreira:
@stsrki I was now thinking about the case where we mutate items and have an observable collection, do you think we should be listening to more notifications?

image

David-Moreira:
I.e I just tried the RowUpdate with an observable collection and it triggers the Replace Notification:
image

I think we should add it, agree? In fact we should probably just straight up listen to every single one of them, as they notifcate valid changes:
image

stsrki:
Yeah, let's listen to all of them.

David-Moreira:
> Yeah, let's listen to all of them.

Done, please review & merge. By the way, think we should do the same to `TreeView`?

stsrki (approved):

stsrki:
> > Yeah, let's listen to all of them.
> 
> Done, please review & merge. By the way, think we should do the same to `TreeView`?

Do it. Shouldn't take too long I guess.

PR #4771, opened 2023-05-15
Add DeepCloner Package and basic tests

### LINKED ISSUE #4388: Datagrid : Investigate Deep Clone implementation
https://github.com/force-net/DeepCloner

David-Moreira:
Closes #4388


Source/Extensions/Blazorise.DataGrid/EventArguments/CancellableRowChange.cs
```diff
@@ -13,14 +14,27 @@ public class CancellableRowChange<TItem> : CancelEventArgs
...
+        OldItem = oldItem;
+        NewItem = newItem;
     }
 
     /// <summary>
     /// Gets the model that was saved.
     /// </summary>
+    [Obsolete( "CancellableRowChange: The Item is deprecated, please use the OldItem instead." )]
+    [EditorBrowsable( EditorBrowsableState.Never )]
     public TItem Item { get; }
```
stsrki, line 31:
```suggestion
    public TItem Item => OldItem;
```

Source/Extensions/Blazorise.DataGrid/EventArguments/CancellableRowChange.cs
```diff
@@ -13,14 +14,27 @@ public class CancellableRowChange<TItem> : CancelEventArgs
...
     /// Initializes a new instance of the cancelable event argument.
     /// </summary>
-    /// <param name="item">Saved item.</param>
-    public CancellableRowChange( TItem item )
+    /// <param name="oldItem">Old Saved item.</param>
+    /// <param name="newItem">New Saved item.</param>
+    public CancellableRowChange( TItem oldItem, TItem newItem )
     {
-        Item = item;
+        Item = oldItem;
```
stsrki, line 21:
```suggestion        
```

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -716,7 +717,10 @@ public async Task Save()
 
         var rowSavingHandler = editState == DataGridEditState.New ? RowInserting : RowUpdating;
 
-        if ( await IsSafeToProceed( rowSavingHandler, editItem, editedCellValues ) )
+        var editItemClone = editItem.DeepClone();
+        SetItemEditedValues( editItemClone );
+
+        if ( await IsSafeToProceed( rowSavingHandler, editItem, editItemClone, editedCellValues ) )
```
stsrki, line 723: Should we also close the `OldItem` to prevent users from accidentally changing the original object values?

stsrki (commented):
By the look at the code it looks like a great feature.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
David-Moreira: Can you elaborate?

stsrki: When user receives an eventArgument through the callback they could override one of the values

eg.

```cs
Task OnUpdating(eventArgs)
{
   eventArgs.Item.Name = "new name";
}
```

David-Moreira: Well, that always has been like that. What do you propose? Even the `OldItem` to be cloned?

stsrki: It was. We only added `.Item` because users kept requesting it. And I was always reluctant to add it because it could interfere with a dictionary of editing Values. If we clone that object then there would be no danger of overriding anything in the original Data.

David-Moreira: I see, well we can add the clone there too I guess. protecting users from themselves. 🗡️

David-Moreira:
Review if you have time, but actually don't merge this yet. I just remembered there is now a flaw in the demo code because I was using the actual reference to find item in collection and also to do deletion I will have to fix it later.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
David-Moreira: Another note, is that cloning the .Item will be a breaking change for whover who was using it as a reference. Are we okay with this? 
**Note to self:** Update docs with this.

stsrki: > Another note, is that cloning the .Item will be a breaking change for whover who was using it as a reference. Are we okay with this?

Damn, I need to think about this.

David-Moreira: Honestly, I would prefer if we didn't clone it. Having the reference still holds some utility.

We'll be removing `.Item` in the future, so users will be less likely to try and mutate something called `OldItem`. And anyway, if they do that, they kinda do it at their own risk, can't protect every possible miss-usage...

stsrki: 🤔 OK, then leave the Item as before. Better not to break anything in 1.x.

David-Moreira: By utility, I mean, something like this:
image
And maybe other use cases which can require the reference.

PR #4776, opened 2023-05-17
Fix warnings

### LINKED ISSUE #4653: Warnings
Go over the solution warnings, making sure to keep it tidy, otherwise we just ignore warnings if the are alot of them. And generally they can list usefull stuff.

David-Moreira:
Closes #4653
Keeping warnings to a minimum so we actually notice them.

Any idea about this one?

Apparently this example does not exist?
image
image

QRCode Library has just too many warnings and I just got rid of them without explicitly fixing them.
Most of them are missing docs, a few obsolete apis which i'm not sure would be worth changing or not as I'm not too knowledgeable of these.
image


stsrki (approved):
LGTM

PR #4779, opened 2023-05-19
Fieldlabel option to show red asterix

### LINKED ISSUE #4683: FieldLabel : Option to show red asterix
image

David-Moreira:
Closes #4683


Source/Blazorise.Bootstrap/wwwroot/blazorise.bootstrap.css
```diff
@@ -762,6 +762,11 @@ input[readonly=readonly][type=range]::-ms-thumb {
   display: contents;
 }
 
+.b-required:before {
+  content: "* ";
+  color: #dc3545;
```
stsrki, line 767:
```suggestion
  color: var(--b-theme-danger, #dc3545);
```

Source/Blazorise.Bootstrap/BootstrapClassProvider.cs
```diff
@@ -310,6 +310,11 @@ public override string LabelType( LabelType labelType )
 
     public override string FieldLabel( bool horizontal ) => horizontal ? "col-form-label" : null;
 
+    public override string ValidationIndicator( bool validationIndicator )
```
stsrki, line 313:
```suggestion
    public override string FieldLabelValidationIndicator( bool validationIndicator )
    
```

Source/Blazorise.Bootstrap/BootstrapClassProvider.cs
```diff
@@ -310,6 +310,11 @@ public override string LabelType( LabelType labelType )
 
     public override string FieldLabel( bool horizontal ) => horizontal ? "col-form-label" : null;
 
+    public override string ValidationIndicator( bool validationIndicator )
+        => validationIndicator
+            ? "b-required"
```
stsrki, line 315: name it as "b-field-required", or "b-field-label-required", or "b-input-required"?

Source/Blazorise.Bootstrap/wwwroot/blazorise.bootstrap.css
David-Moreira: You're looking at compiled css, either way makes sense to prioritize the b-theme var.
`    color: var(--b-theme-danger, $danger);`

Source/Blazorise.Bootstrap/BootstrapClassProvider.cs
David-Moreira: Makes sense.

Source/Blazorise.Bootstrap/BootstrapClassProvider.cs
David-Moreira: Initial I thought b-required could be used generally whenever we wanted to tell something was required be it a label, input ,etc... 
But we can specify it, in that case, I like `b-field-label-required`

stsrki: Actually, I think we should try to follow each of the provider naming strategies. For example, in BS4 `form-group` represents a Field, and then it could be `form-group-required`, or `form-control-required` if we want to target inputs. You get the idea.

image

David-Moreira: hmm You're just looking at bs, each have their own. What would be the benefit? 
Additionally we would to potentially have more code / ceremony if we want to target other components from the `FieldLabel`? 
And that gives the sense that a lot more can happen since the class is more general, while we're just adding functionality specific to the label.

stsrki: Well, I tend to use our custom `.b-` classnames as little as possible. Also, I want to stick to the naming of each provider as much as I can.

David-Moreira: Well we need to differentiate what's custom made, for both our sanity and the users...
Imagine in a few years time, we have a bug in the feature, and you read the class `form-group-required`, you're going to think it's a bs class right away.

stsrki: That's a future us problem 🤣

David-Moreira: Ok then. So just to reiterate we'll be setting `ValidationIndicator` on the `FieldLabel` and then adding the class to the parent div if it exists (which should be the Parent `Field`), `<div class="form-group form-group-required"`. This is it right?

stsrki: Considering an API that you started

```razor
<Field>
    <FieldLabel ValidationIndicator>
        Name
    </FieldLabel>
    <FieldBody>
        <TextEdit Placeholder="Name" />
    </FieldBody>
</Field>
```

I would say that it will need to be on the `<label>` instead. But then again, I don't like that API. I am not sure if I want to have it on a `FieldLabel`.

What if we add `Required` on the `Validation` component? So the API would be:

```razor
<Validation Required>
    <Field>
        <FieldLabel ValidationIndicator>
            Name
        </FieldLabel>
        <FieldBody>
            <TextEdit Placeholder="Name" />
        </FieldBody>
    </Field>
</Validation>
```

If the required asterisk is only going to be used along with the Validation, than it makes sense. Otherwise, if it can be used without validation, then it can stay on FieldLabel.

David-Moreira: Well in my mind It's just a visual indicator whether there's validation or not or even if the user does validation some other way. That's why it's in the label.

If you put Required in the Validation component then one could assume that it would automatically validate emptyness?
And you are saying that in this case there would be no ValidationIndicator parameter?

stsrki: > Well in my mind It's just a visual indicator whether there's validation or not or even if the user does validation some other way. That's why it's in the label.
> 
> If you put Required in the Validation component then one could assume that it would automatically validate emptyness? And you are saying that in this case there would be no ValidationIndicator parameter?

Makes sense. You're right, it should stay on FieldLabel. I would just rename it to `RequiredIndicator`.

stsrki: > * Adds the class to the parent div if it exists (which should be the Parent `Field`),
>   `<div class="form-group form-group-required">`. This is it, right?

No, don't add it to the parent. That would complicate it as you would have to know if there is a child component Field >FieldLabel

It's easier to have `RequiredIndicator` on the `FieldLabel`, and assign the class to its element.

> The naming must be in par with whatever's defined for the `Field` for the current provider:
> i.e for bs: `form-group` => `form-group-required`

See what is already on the label element. If there is `form-label` like in BS5, then you can add `form-label-required`. If there is no label class available, like on some provider then think of a name that could be best for that provider.

David-Moreira: > It's easier to have RequiredIndicator on the FieldLabel, and assign the class to its element.

Well my point all along that we were adding unecessary complexity... that's what's already implemented... Will change just the Parameter and class name's then.

David-Moreira: Done. Please review again at your earliest convenience.

stsrki: I was trying to come up with a way to automate it. But in the end, you were right all along :)

Source/Blazorise.Tailwind/Styles/components/_form.scss
```diff
@@ -0,0 +1,4 @@
+﻿.label-required:before {
```
stsrki, line 1: What do you think of doing it in the tailwind style? https://tailwindcss.com/docs/content

David-Moreira: Is that good for overriding? 
Since you have more experience would it be something like : `before:content-['*']` and how do we color style it?

Source/Blazorise.Bootstrap/BootstrapClassProvider.cs
stsrki: Yeah, the FieldLabel could somehow detect if its placed inside of a `Validation` component and if the validated field is required. But not sure if it would take too long to do. Might be something for the future.

Source/Blazorise.Tailwind/Styles/components/_form.scss
stsrki: I believe you can still use `var` for colors.

stsrki: ps. https://stackoverflow.com/questions/64872861/how-to-use-css-variables-with-tailwind-css

David-Moreira: before:content-['*'] before:[color:red]

David-Moreira: However it doesn't seem like it supports the blank space do you know anything about it? `before:content-['* ']`

David-Moreira: Seems decent enough?
image

stsrki: Shouldn't it need to be after the label?

image

David-Moreira: Well I don't know from an ux perspective... . I think by being the first thing you can tell the required fields right away. While if they are at the end it can be something that you don't notice right away? The bigger the label is the further it gets pushed.

David-Moreira: But yes I have seen both ways of doing it.

stsrki: > Well I don't know from an ux perspective... . I think by being the first thing you can tell the required fields right away. While if they are at the end it can be something that you don't notice right away? The bigger the label is the further it gets pushed.

It seems from a UX perspective it is better to have on the end

https://uxpickle.com/required-fields-in-forms/
https://uxdesign.cc/form-field-required-vs-optional-9b4d7cdbf400

David-Moreira: They don't seem to talk about the design decision **before vs after** do they ? 
We can't just base ourselves in some articles...

I see arguments for both sides...
image

Anyway as long as its consistent.. Would you prefer to switch to after then?

David-Moreira: W3C (Altough it seems old) In here they deem important to define the meaning of it. And in the example its after the table.

https://www.w3.org/TR/2012/NOTE-WCAG20-TECHS-20120103/H90
image

David-Moreira: (plus they have recommended markup with abbr that we don't comply, nor do I think we need to bother? :D:D)

stsrki: 🤔well, we have in plan to support AODA, so I think it is worth doing it with `abbr`.

David-Moreira: oh well... There goes all the dev haha. Alright Let's follow these acessibility guidelines... but if you want to support some other guidelines, AODA? Then we should read those up?
Is it this? https://aoda.ca/ There does not seem to be web related guidelines?

David-Moreira: This seems newer standard definitions
https://www.w3.org/WAI/tutorials/forms/validation/
image
I'm not sure this is a rabbit hole we want to go into just yet...?

stsrki: haha, leave it for some in the future. Better to go with the other 1.3 tasks.

David-Moreira: Ye, I would just do a release focused on this. It seems like we need to do alot of research.
We can however move the * to after for now.

stsrki (approved):

PR #4783, opened 2023-05-23
Autocomplete: Introduce callbacks for the events currently being listened on TextEdit by Autocomplete

David-Moreira:
- Only exposed the current events that we are actually listening to
- Took the approach of calling the events always at the bottom as to not conflict with any of our processing.
Any thoughts about this approach?

I just noticed this, 
> Occurs when the input box gains or loses focus.

I wonder if we have an uneeded call or dumb "bug" here since it assumes it's only on focus in? Or the documentation is just wrong... hmm?
image


David-Moreira:
Also I've seen the SF component vendor do something like this: 
I wonder if this separation is actually a nice idea? Or just straight up unedeed, since it adds more classes/components
image

stsrki:
> Also I've seen the SF component vendor do something like this: I wonder if this separation is actually a nice idea? Or just straight up unedeed, since it adds more classes/components image

It's not a bad approach, but it would mean a breaking API. Maybe for the v2, we could think of making it more declarative. I already have something similar in mind for the Charts component.

Source/Extensions/Blazorise.Components/Autocomplete.razor.cs
```diff
@@ -2,6 +2,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Reflection.Metadata.Ecma335;
```
stsrki, line 5: What is this namespace for? 😕

Source/Extensions/Blazorise.Components/Autocomplete.razor.cs
```diff
@@ -1524,5 +1534,22 @@ public List<string> SelectedTexts
     /// </remarks>
     [Parameter] public int? TotalItems { get; set; }
 
+    /// <summary>
+    /// Occurs after text has changed.
+    /// </summary>
+    [Parameter] public EventCallback<string> TextChanged { get; set; }
+    /// <summary>
```
stsrki, line 1541: I don't want to sound too harsh. But always leave one empty line between properties

image

stsrki (commented):
Looking good overall.

Source/Extensions/Blazorise.Components/Autocomplete.razor.cs
David-Moreira: LOL, I honestly have no idea? Might have gotten imported somehow.

Source/Extensions/Blazorise.Components/Autocomplete.razor.cs
David-Moreira: I like it also, Sometimes I just don't notice.

stsrki: Do you maybe have CodeLens enabled in Visual Studio?

David-Moreira: Yes, all sorts of stuff enabled... that sometimes you just have too many visual jargon that you stop noticing formatting as easily hehe

Documentation/Blazorise.Docs/Pages/Docs/Extensions/Autocomplete/AutocompletePage.razor
```diff
@@ -325,4 +325,16 @@
     <DocsAttributesItem Name="Virtualize" Type="bool" Default="false">
         Gets or sets whether the Autocomplete will use the <Code>Virtualize</Code> functionality.
     </DocsAttributesItem>
+    <DocsAttributesItem Name="TextChanged" Type="EventCallback<string>">
+        Occurs after text has changed.
+    </DocsAttributesItem>
+    <DocsAttributesItem Name="KeyDown" Type="EventCallback<KeyboardEventArgs>">
```
stsrki, line 331: We have all these new events KeyDown, OnFocus, and Blur, but none is directly mentioning which part of the Autocomplete it behaves to. Maybe we need to name them all:

- TextKeyDown, TextFocus, and TextBlur, or
- SearchKeyDown, SearchFocus, and SearchBlur, or
- InputKeyDown, InputFocus, and InputBlur

We will most possibly add more APIs in the future that will control parts of the Autocomplete so we need to decide on naming now.

David-Moreira: Like any other component we have, the events pertain to the shown input... the shown input for an autocomplete is a text box... any of these interactions I have with that visible input will trigger the events as expected.
I can't see how it isn't clear... but if you feel like it isn't, we can rename if you'd like. 

I like the `Search` prefix. Since we already have a `Search` / `SearchChanged` parameters.

stsrki: Don't forget that in multiple mode, the input will resize accordingly to available space. 

image

Proceed with Search as it makes the most sense.

stsrki (approved):

PR #4784, opened 2023-05-23
Autocomplete TagTemplate

David-Moreira:


Source/Extensions/Blazorise.Components/Contexts/AutoCompleteBadgeContext.cs
```diff
@@ -0,0 +1,39 @@
...
+public class BadgeContext<TItem, TValue> : BaseTemplateContext<TItem>
+{
+    /// <summary>
+    /// Initializes BadgeContext.
+    /// </summary>
+    /// <param name="item">Holds current Autocomplete Item Data.</param>
+    /// <param name="value">Holds current Autocomplete Item Value.</param>
+    /// <param name="text">Holds current Autocomplete Item Text.</param>
+    /// <param name="removeCallback">Removes Item from selection callback.</param>
+    public BadgeContext( TItem item, TValue value, string text, EventCallback removeCallback ) : base( item )
```
stsrki, line 17: Rename it to `AutocompleteBadgeContext`.

Documentation/Blazorise.Docs/Pages/Docs/Extensions/Autocomplete/AutocompletePage.razor
```diff
@@ -325,4 +325,7 @@
     <DocsAttributesItem Name="Virtualize" Type="bool" Default="false">
         Gets or sets whether the Autocomplete will use the <Code>Virtualize</Code> functionality.
     </DocsAttributesItem>
+    <DocsAttributesItem Name="BadgeTemplate" Type="RenderFragment<BadgeContext<TItem, TValue>>">
```
stsrki, line 328: I know we use the Badge for the autocomplete tags. But I don't think we should expose that naming outside of it. The parts that are selected are usually named Tags, so maybe we should follow that convention.

Source/Extensions/Blazorise.Components/Autocomplete.razor
```diff
@@ -11,7 +11,17 @@
...
-                <Badge Color="@MultipleBadgeColor" Margin="Margin.Is2.FromStart" CloseClicked="async () => await RemoveMultipleTextAndValue( selectedText )">@selectedText</Badge>
+                @if ( BadgeTemplate is null )
+                {
+                    <Badge Color="@MultipleBadgeColor" Margin="Margin.Is2.FromStart" CloseClicked="async () => await RemoveMultipleTextAndValue( selectedText )">@selectedText</Badge>
+                }
+                else
+                {
+                    var item = GetItemByText( selectedText );
+                    var value = GetItemValue( selectedText );
+                    var removeCallback = EventCallback.Factory.Create( this, async () => await RemoveMultipleTextAndValue( selectedText ) );
```
stsrki, line 22: Is there any performance hits for doing `EventCallback.Factory.Create` on each render?

Source/Extensions/Blazorise.Components/Contexts/AutoCompleteBadgeContext.cs
David-Moreira: I thought about it, but for some reason we didn't do that here:
image
So I just used the same approach for consistency, because they are both namespace to **Autocomplete** already.
image

Should we still rename anyway?

Documentation/Blazorise.Docs/Pages/Docs/Extensions/Autocomplete/AutocompletePage.razor
David-Moreira: So `TagTemplate`? It still requires the user to have knowledge of what a Tag would be in this context, since we don't really mention it anywhere...
Maybe `MultipleSelectionTemplate` that would be agnostic?

stsrki: Yeah, I think `Tag` is the most practiced name in this context, Although something with `Selected` is also a good idea.

Source/Extensions/Blazorise.Components/Autocomplete.razor
David-Moreira: Well, same hit as getting the item and value. It has to be recalculated if it goes through the loop and collection hasn't changed.

It is of note that `Eventcallback` is a struct, the biggest hit is from the anonymous delegate.

Only workaround I know is this : 
(https://learn.microsoft.com/en-us/aspnet/core/blazor/performance?view=aspnetcore-7.0#avoid-recreating-delegates-for-many-repeated-elements-or-components)
image
But since the collection is dynamic we would still need to recalculate each time and to make it efficient try to do it diff based, So I'm not sure there would be a significant benefit in relation to the added complexity, `Autocomplete` is already huge, eh.

---

stsrki: Ok. it can stay in that case. Thank you for the clarification.

Documentation/Blazorise.Docs/Pages/Docs/Extensions/Autocomplete/AutocompletePage.razor
David-Moreira: `MultipleSelectedTemplate` ? I don't really think we can run much from that since everything is named around "`Multiple`"
image

Even though `Tag` can be common, it would still be a guessing name, some people might know some might not. Specifially since we don't have any other mentions anywhere.

Source/Extensions/Blazorise.Components/Contexts/AutoCompleteBadgeContext.cs
David-Moreira: @stsrki friendly reminder

Documentation/Blazorise.Docs/Pages/Docs/Extensions/Autocomplete/AutocompletePage.razor
David-Moreira: @stsrki friendly reminder

Source/Extensions/Blazorise.Components/Contexts/AutoCompleteBadgeContext.cs
stsrki: Yes, rename it to AutocompleteTagContext.

stsrki (approved):

stsrki:
There was one test failing for DataGrid. We need to look it up.

PR #4788, opened 2023-05-26
DataGrid: Use ValidationsHandlerType from ParentDataGrid

### LINKED ISSUE #4787: Datagrid FluentValidationHandler support unavailable or undocumented
**Describe the bug**
The datagrid `ValidationsHandlerType` does not appear to be respected when set to FluentValidationHandler. The validation handler type defined on a DataGrid is not used when validating columns during edit/create flows. Specifically this prevents the use of FluentValidationHandler on the top-level datagrid.

**To Reproduce**

minimal code example:
https://github.com/dsmiller95/data-grid-fluent-validation-example/blob/fe93ed4ac1a4211a3ab490f1f4d8946a789a801f/DataGridFluentValidation/Pages/DataGridPage.razor

1. Follow setup steps for Blazorise.FluentValidation : https://blazorise.com/docs/extensions/fluent-validation
2. Copy example for DataGrid validation : https://blazorise.com/docs/extensions/datagrid/validations
3. Set `ValidationsHandlerType="typeof(FluentValidationHandler)"`
4. Implement `EmployeeValidator : AbstractValidator<Employee>`
5. Run the app, edit the Employee row
6. Observe that the validations are not applied

**Expected behavior**
The validations in EmployeeValidator are checked over the data before saving, or while editing.

**Screenshots**
Data grid allows saving an employee with no name, when a FluentValidator is defined to explicitly deny this:
image

**Additional context**
I believe the `ValidationsHandlerType` is completely ignored in this use case. In [this revision of the code](https://github.com/dsmiller95/data-grid-fluent-validation-example/blob/166a9514dde611923ef018f23032862ad8fa0747/DataGridFluentValidation/Pages/DataGridPage.razor) , we set `ValidationsHandlerType="typeof(TestValidationHandler)"`. `TestValidationHandler` throws an exception on all methods, but no exceptions are thrown when running the app. And a breakpoint set in `TestValidationHandler` will not be hit.

This may be intentional, but the documentation implies it is not, as I have not overriden the HandlerType anywhere else in the code:

Attributes
DataGrid
| Name | Description |
|--------|--------|
| ValidationsHandlerType | Defines the default handler type that will be used by the validation, unless it is overriden by Validation.HandlerType property. |

stsrki:
closes #4787

### DIFF
--- a/Source/Extensions/Blazorise.DataGrid/_DataGridCell.razor.cs
+++ b/Source/Extensions/Blazorise.DataGrid/_DataGridCell.razor.cs
@@ -44,6 +44,9 @@ public abstract class _BaseDataGridCell<TItem> : ComponentBase
             else if ( HasValidator )
                 return typeof( ValidatorValidationHandler );
 
+            if ( ParentDataGrid?.ValidationsHandlerType is not null )
+                return ParentDataGrid.ValidationsHandlerType;
+
             // default is always data-annotations
             return typeof( DataAnnotationValidationHandler );
         }


David-Moreira (approved):
LGTM. Add some simple datagrid validation tests if you care to bother.

PR #4791, opened 2023-05-29
Dropdown : Checkbox support

David-Moreira:


David-Moreira:
@stsrki do you think regular clicks should change multiple selection? They close the dropdown after all, seems more clear to only allow selection on checkboxes since they do not close the dropdown?

stsrki:
> @stsrki do you think regular clicks should change multiple selection? They close the dropdown after all, seems more clear to only allow selection on checkboxes since they do not close the dropdown?

When in multiple selection mode, the click should toggle the checkbox state, no matter where the user clicks on an item.

David-Moreira:
@stsrki However for regular `Dropdown`, it does make sense that a regular click still closes it, right? But clicking the checkbox selects/unselects?

stsrki:
I would really prefer if we could have a real checkbox instead of fake ones, like on Autocomplete. Even if it is hard to do.

> @stsrki However for regular Dropdown, it does make sense that a regular click still closes it, right? But clicking the checkbox selects/unselects?

I'm divided on this one TBH. Both ways make sense, but I tend to go more along that it needs to toggle on the entire width of an item. Like https://flowbite.com/docs/components/dropdowns/#dropdown-with-checkbox for example.

David-Moreira:
> I would really prefer if we could have a real checkbox instead of fake ones, like on Autocomplete. Even if it is hard to do.

I will see what we can do then for having "real " checkboxes then.

> 
> > @stsrki However for regular Dropdown, it does make sense that a regular click still closes it, right? But clicking the checkbox selects/unselects?
> 
> I'm divided on this one TBH. Both ways make sense, but I tend to go more along that it needs to toggle on the entire width of an item. Like https://flowbite.com/docs/components/dropdowns/#dropdown-with-checkbox for example.

Thing is that on that example the dropdown does not close. maybe it's worth adding an `AutoClose` parameter?
But if we do then we have to think about `Multiple` selection behaviour, does it auto close ever? If this new parameter that means it will `AutoClose` by default.

But I guess clicking the whole "row" selects/unselects makes sense then for both `Dropdown` / `DropdownList`. In fact, whatever is implemented in `Dropdown` check wise, should be how `DropdownList` works.

stsrki:
Clicking outside of the menu or pressing an escape key should close it.

David-Moreira:
@stsrki 
It actually worked out fine. I remember having trouble binding checkboxes, properly and re-rendering out of order. Maybe misused it on the past. 
What do you think in considering changing Autocomplete checkbox to use this new feature in **2.0**?

stsrki:
> What do you think in considering changing Autocomplete checkbox to use this new feature in **2.0**?

Absolutely yes. If there aren't any problems.

Source/Blazorise/Components/Dropdown/DropdownItem.razor.cs
```diff
@@ -108,5 +142,20 @@ public bool Disabled
     /// </summary>
     [CascadingParameter] protected Dropdown ParentDropdown { get; set; }
 
+    /// <summary>
+    /// The dropdown renders a checkbox.
+    /// </summary>
+    [Parameter] public bool Checkbox { get; set; }
```
stsrki, line 148: Rename to ShowCheckbox?

Source/Extensions/Blazorise.Components/DropdownList.razor
```diff
@@ -28,6 +28,9 @@
         var text = TextField?.Invoke( item );
         var value = ValueField != null ? ValueField.Invoke( item ) : default;
 
-        <DropdownItem @key="@item" Clicked="@HandleDropdownItemClicked" Value="@value">@text</DropdownItem>
+        <DropdownItem @key="@item" Clicked="@HandleDropdownItemClicked" Value="@value"
+        Checkbox="@(SelectionMode == DropdownListSelectionMode.Checkbox)"
```
stsrki, line 32: Align with other params.

Source/Blazorise/Components/Dropdown/DropdownItem.razor.cs
```diff
@@ -17,10 +18,23 @@ public partial class DropdownItem : BaseComponent
 
     private bool disabled;
 
+    private bool check;
```
stsrki, line 21: Renamed to `@checked`?

Source/Extensions/Blazorise.Components/Enums/AutocompleteFilter.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace Blazorise.Components;
+
+/// <summary>
+/// The Autocomplete's filter mode.
+/// </summary>
+public enum AutocompleteFilter
```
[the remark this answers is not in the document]
David-Moreira: What? I haven't touched that? have I haha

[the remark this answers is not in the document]
David-Moreira: Ahh this is a new file. `AutocompleteFilter` was moved to its own file. `AutocompleteSelectionMode`, remains. They are distinct features.

Source/Extensions/Blazorise.Components/Enums/DropdownListSelectionMode.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace Blazorise.Components;
+
+/// <summary>
+/// The DropdownList's selection mode.
+/// </summary>
+public enum DropdownListSelectionMode
```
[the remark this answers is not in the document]
David-Moreira: Don't think so, `DropdownListSelectionMode` is the same convention naming as `AutocompleteSelectionMode`. Both will exist.

stsrki:
Don't forget to update all `DropdownItem` for each of the providers.

**Tailwind**

image

Source/Blazorise/Components/Dropdown/DropdownItem.razor.cs
David-Moreira: fair enough, I thought we wouldn't like the necessary @ in `@checked`

David-Moreira:
> Don't forget to update all `DropdownItem` for each of the providers.
> 
> **Tailwind**
> 
> image

Nice catch, I forgot to check if there were dedicated files.

Source/Blazorise/Components/Dropdown/DropdownItem.razor.cs
stsrki: We already have it in several places. So why not.

stsrki:
For tailwind, we might need to adjust the HTML structure based on https://flowbite.com/docs/components/dropdowns/#dropdown-with-checkbox

image

stsrki:
@David-Moreira  Do you think there is anything else you wish to add to this PR?

David-Moreira:
> @David-Moreira Do you think there is anything else you wish to add to this PR?

We'll if you have added the suggested changes for tailwind. No not really.

stsrki (approved):

PR #4795, opened 2023-05-30
TimePicker: fix blur event does not fire TimeChanged event

### LINKED ISSUE #4786: TimePicker does not fire TimeChanged event
When you edit a time in a ```<TimePicker>``` control by typing with the keyboard, the TimeChanged event is fire when you leave the control by using the <tab> key. But when you leave the control after typing by clicking on another control, the TimeChanged event is **not** fired.

**Expected behavior**
That the TimeChanged event als fires when leaving by loosing focus.

I use the latest Blazorise controls (Wasm), .Net7.

stsrki:
Closes #4786

Fixed

- TimePicker
- DatePicker

**Test code**

```razor
<TimePicker TValue="DateTime?" Time="@dt" TimeChanged="@OnTimeChanged" />
<DatePicker TValue="DateTime?" Date="@dt2" DateChanged="@OnDateChanged" />
@code {
    DateTime? dt;
    DateTime? dt2;

    Task OnTimeChanged( DateTime? v )
    {
        Console.WriteLine( $"OnTimeChanged {v}" );

        dt = v;

        return Task.CompletedTask;
    }

    Task OnDateChanged( DateTime? v )
    {
        Console.WriteLine( $"OnDateChanged {v}" );

        dt2 = v;

        return Task.CompletedTask;
    }
}
```

### DIFF
--- a/Source/Blazorise/wwwroot/datePicker.js
+++ b/Source/Blazorise/wwwroot/datePicker.js
@@ -75,6 +75,16 @@ export function initialize(dotnetAdapter, element, elementId, options) {
         picker.altInput.disabled = options.disabled || false;
         picker.altInput.readOnly = options.readOnly || false;
         picker.altInput.placeholder = options.placeholder;
+
+        picker.altInput.addEventListener("blur", (e) => {
+            const isInput = e.target === picker._input;
+
+            // Workaround for: onchange does not fire when user writes the time and then click outside of the input area.
+            if (isInput && picker.isOpen === false) {
+                picker.input.dispatchEvent(utilities.createEvent("change"));
+                picker.input.dispatchEvent(utilities.createEvent("input"));
+            }
+        });
     }
 
     picker.customOptions = {


Source/Blazorise/wwwroot/timePicker.js
```diff
@@ -57,6 +57,16 @@ export function initialize(element, elementId, options) {
         picker.altInput.disabled = options.disabled || false;
         picker.altInput.readOnly = options.readOnly || false;
         picker.altInput.placeholder = options.placeholder;
+
+        picker.altInput.addEventListener("blur", (e) => {
+            const isInput = e.target === picker._input;
+
+            // Workaround for: onchange does not fire when user writes the time and then click outside of the input area.
+            if (isInput && picker.isOpen === false) {
+                picker.input.dispatchEvent(utilities.createEvent("change"));
```
David-Moreira, line 66: There's also this other option that's in the code base, that's not too verbose: `new Event("change")` still would be pretty good. 
image
image

David-Moreira (approved):

Source/Blazorise/wwwroot/timePicker.js
David-Moreira: No I actually meant that the utility might not be necessary since we could just do `new Event("change")` Anyway it's fine.

David-Moreira: But eitherway, we should just have every single one of them use the same implementation.

PR #4796, opened 2023-05-30
Docs: mention Modal fullscreen workaround

### LINKED ISSUE #4777: Fullscreen Modal does not render background correctly if content higher than screen size
**Describe the bug**
For a Modal that is shown Fullscreen, the background gets not correctly rendered, if the modal content is higher than the screen.

**To Reproduce**
1. reproduction repo: https://github.com/oaldrian/BlazoriseReproductions/tree/ModalBackground
2. on index page there is a broken modal

reproducing the problem is maybe dependent on the screen size, for me it happened with a FullHD monitor.

**Expected behavior**
Modal backgrounds get rendered correctly.

**Screenshots**
"first page" - aka initially shown content

image

after a bit of scrolling

image

**Additional context**
It might very well be the case that this is some sort of misconfiguration, so not really a bug, but something that needs documentation. 
Happens in both, WebAssembly and ServerSide

stsrki:
Closes #4777

### DIFF
--- a/Documentation/Blazorise.Docs/Pages/Docs/Components/Modals/ModalPage.razor
+++ b/Documentation/Blazorise.Docs/Pages/Docs/Components/Modals/ModalPage.razor
@@ -103,6 +103,51 @@
     Modal dialogs are disruptive by nature and should be used sparingly. Do not use them to communicate nonessential information, such as success messages like "Logged in", "Copied", etc. Instead, use <Anchor To="docs/services/notification-provider">Notifications</Anchor> when appropriate.
 </DocsPageParagraph>
 
+<Heading Size="HeadingSize.Is3">
+    Fullscreen Mode
+</Heading>
+
+<DocsPageParagraph>
+    When using the Fullscreen mode for the Modal component, it's crucial to be aware of how you structure your layout within the <Strong>ModalBody</Strong>.
+</DocsPageParagraph>
+
+<DocsPageParagraph>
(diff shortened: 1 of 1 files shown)


Documentation/Blazorise.Docs/Pages/Docs/Components/Modals/ModalPage.razor
```diff
@@ -103,6 +103,51 @@
...
 
+<Heading Size="HeadingSize.Is3">
+    Fullscreen Mode
+</Heading>
+
+<DocsPageParagraph>
+    When using the Fullscreen mode for the Modal component, it's crucial to be aware of how you structure your layout within the <Strong>ModalBody</Strong>.
+</DocsPageParagraph>
+
+<DocsPageParagraph>
```
David-Moreira, line 114: Generally the base documentation pages show case examples of the features available & usage. Sometimes with well delimited notes about certain quirks (usage of Alerts).

It seems like we have much to explain here, so using just an Alert would not be enough. Would this be best on  "Best practices" section or something?

David-Moreira (approved):

Documentation/Blazorise.Docs/Pages/Docs/Components/Modals/ModalPage.razor
stsrki: Yes, this is part of **Best Practices**. While it is somewhat long I think it gives enough information about the context of a problem.

---

In the long run, we should expand all our existing examples with more descriptive use cases and best practices. Not related to this PR, but the usage of TextEdit and other inputs without wrapping them with Field comes to mind.

PR #4806, opened 2023-06-03
Docs: add best practices for Carousel component

### LINKED ISSUE #3641: Carousel slides are not uniform size
**Describe the bug**
Carousel slides with differing size content is causing the carousel to resize on the page

**To Reproduce**
Evident on the documentation page: https://blazorise.com/docs/components/carousel

**Expected behavior**
Probably expect the carousel size to be determined by default by the size of the largest element, that way the individual slides will always fit within the size of the carousel

stsrki:
Closes #3641


David-Moreira (approved):
> This can be prevented by defining consistent widths and heights through additional CSS styles.

Do we need to give a more concrete example, or this should be enough?

stsrki:
I don't think it is necessary.

PR #4812, opened 2023-06-08
DataGrid : DataGridSelectColumn dynamic loading of data

David-Moreira:
Here's an initial approach for `DataGridSelectColumn` actually be more useful and act like a `SelectList`

We took the approach of boxing the values. This means that the user will have to cast the items whenever he's using stuff like ValueField or TextField. I believe this is acceptable, given we already have a few datagrid apis like this.

We could however think of introducing a **TSelectItem** to make the experience better, what do you think?

Other than that, agree with the api? (I also introduced the `DefaultItem*` stuff, but that api almost seems like its redudant?)


stsrki (commented):
The API looks good to me. If you think you can add TValue, then go for it. Look into _DataGridCellEdit.razor, you might need to check for every common type, similar to NumericPicker.

David-Moreira:
> The API looks good to me. If you think you can add TValue, then go for it. Look into _DataGridCellEdit.razor, you might need to check for every common type, similar to NumericPicker.

If you provide **TValue** you don't need to check for every type. You can just propagate it.

stsrki:
Well, I did try it in the past, and I know how difficult it is. We can go with the object for now, and then in v2, we can try to introduce TValue.

David-Moreira:
> Well, I did try it in the past, and I know how difficult it is. We can go with the object for now, and then in v2, we can try to introduce TValue.

Feels like we're pushing everything to v2, haha

Source/Extensions/Blazorise.DataGrid/DataGridSelectColumn.cs
```diff
@@ -8,27 +8,26 @@ public class DataGridSelectColumn<TItem> : DataGridColumn<TItem>
...
-
-
     /// <summary>
     /// Gets or sets the select data-source.
     /// </summary>
     [Parameter] public IEnumerable<object> Data { get; set; }
 
     /// <summary>
     /// Method used to get the display field from the supplied data source.
+    /// You can unbox the object by using the following example: TextField="(x) => ((Model)x).Text"
```
stsrki, line 18: A suggestion. When you have comments into new lines, maybe try to separate them with paragraphs. eg `<para></para>` .

David-Moreira: Usually do that, didn't remember in this PR. :)

Source/Extensions/Blazorise.DataGrid/DataGrid.razor
```diff
@@ -124,14 +124,23 @@
...
                                     {
                                         @(column.FilterTemplate( column.Filter ))
                                     }
                                     else
                                     {
-                                        <TextEdit Text="@column.Filter.SearchValue?.ToString()" TextChanged="@(async (newValue) => await OnFilterChanged(column, newValue))" />
+                                        //If this grows, move to a seperate component.
+                                        if ( column is DataGridSelectColumn<TItem> selectColumn )
+                                        {
+                                            <_DataGridCellSelectEdit Column=selectColumn CellValue="@selectColumn.Filter.SearchValue" CellValueChanged="@(async (newValue) => await OnFilterChanged(column, newValue))"></_DataGridCellSelectEdit>
```
stsrki, line 138:
```suggestion
                                            <_DataGridCellSelectEdit Column="@selectColumn" CellValue="@selectColumn.Filter.SearchValue" CellValueChanged="@(async (newValue) => await OnFilterChanged(column, newValue))"></_DataGridCellSelectEdit>

```

Tend to assign parameters inside of '"', eg. `"@selectColumn"`

stsrki (approved):

PR #4834, opened 2023-06-24
Prevents the JS call on each Dispose() when the JSDragDropModule has not Init

David-Moreira:
Cherry pick from #4733


stsrki (approved):
LGTM

PR #4846, opened 2023-06-29
Fixes typos in date picker localizer

InzeNL:
I had meant to make this pull request forever ago, but I never got to it.

I should note that this *may* break implemented custom localizers (as they will localize the typoed "Thurday" and "May!" strings). I personally think this would be an acceptable break, but I'll trust your judgement in this

It fixes two typos that show when using the default localizer for the DatePicker, where "Thursday" is missing an 's' (spelling "Thurday") and "May" (the non-shorthand) has an exclamation mark at the end (spelling "May!")


InzeNL:

InzeNL:
recheck

stsrki (approved):
Looking good. Thank you!

PR #4856, opened 2023-07-01
Validation: add support for IValidatableObject

### LINKED ISSUE #4360: Validation: add support for IValidatableObject
Sorry if this is a total newbie question...

Does Blazorise validation support IValidatableObject ?

When I use `DataAnnotationsValidator` the `IValidatableObject.Validate` works fine.
I couldn't make Blazorise Validation component to call `IValidatableObject.Validate`

### LINKED ISSUE #4869: Validations duplicating error messages
User over at discord reported the following:

image

stsrki:
Closes #4360
Closes #4869


stsrki:
Coping comment from the ticket here:

---

Just a quick update. I already have a working code in my private branch. But I cannot decide on the right behavior. With my current implementation, it works automatically if validation detects the IValidatableObject.

The problem comes when circular validation references other fields and the validation rule changes dynamically.

Like in the following example, if we switch the Company.UseAlphaCode true and false. It will not trigger re-validation unless we go and type into the input field.

```razor
<Validations Model="@Company" Mode="ValidationMode.Auto">
    <Validation>
        <Field>
            <FieldLabel>Name</FieldLabel>
            <FieldBody>
                <TextEdit @bind-Text="@Company.Name" />
            </FieldBody>
        </Field>
    </Validation>
    <Validation>
        <Field>
            <FieldLabel>Description</FieldLabel>
            <FieldBody>
                <TextEdit @bind-Text="@Company.Description" />
            </FieldBody>
        </Field>
    </Validation>
    <Field>
        <Switch @bind-Checked="@Company.UseAlphaCode">Use AlphaCode</Switch>
    </Field>
    <Fields>
        <Validation>
            <Field>
                <FieldLabel>AlphaCode</FieldLabel>
                <FieldBody>
                    <TextEdit @bind-Text="@Company.AlphaCode" />
                </FieldBody>
            </Field>
        </Validation>
        <Validation>
            <Field>
                <FieldLabel>BetaCode</FieldLabel>
                <FieldBody>
                    <TextEdit @bind-Text="@Company.BetaCode" />
                </FieldBody>
            </Field>
        </Validation>
    </Fields>
</Validations>

@code {
  CompanyInfo Company = new CompanyInfo()
  {
      UseAlphaCode = true,
  };
  
  public class CompanyInfo : IValidatableObject
  {
      [Required( ErrorMessage = "Name is required" )]
  
      [Required( ErrorMessage = "Description is required" )]
      public string Description { get; set; }
  
      public bool UseAlphaCode { get; set; }
  
      public string AlphaCode { get; set; }
  
      public string BetaCode { get; set; }
  
      [Range( 0, 999.99 )]
      public decimal Price { get; set; }
  
      public IEnumerable<ValidationResult> Validate( ValidationContext validationContext )
      {
          if ( UseAlphaCode )
          {
              if ( String.IsNullOrWhiteSpace( AlphaCode ) )
              {
                  yield return new ValidationResult( "AlphaCode is required", new[] { "AlphaCode" } );
              }
          }
          else
          {
              if ( String.IsNullOrWhiteSpace( BetaCode ) )
              {
                  yield return new ValidationResult( "BetaCode is required", new[] { "BetaCode" } );
              }
          }
      }
  }
}
```

David-Moreira:
> The problem comes when circular validation references other fields and the validation rule changes dynamically.

Isn't the actual problem that the `UseAlphaCode` / `Switch` is not triggering validation when it's checked/unchecked?

Source/Blazorise/Components/Validation/Validations.razor.cs
```diff
@@ -127,6 +128,11 @@ internal void NotifyValidationStatusChanged( IValidation validation )
         // Try to come up with solution that StatusChanged will be called only once while it will
         // still provide all of the failed messages.
 
+        if ( Model is IValidatableObject validatableObject )
+        {
+            var result = validatableObject.Validate( new ValidationContext( Model ) );
```
David-Moreira, line 133: We don't need to do anything with the result here? use discard instead then?

David-Moreira (changes requested):
Can you try testing with this:
```
<Validations @ref=Validations Model="@Company" StatusChanged="StatusChanged" Mode="ValidationMode.Auto">
    <Validation>
        <Field>
            <FieldLabel>Name</FieldLabel>
            <FieldBody>
                <TextEdit @bind-Text="@Company.Name" />
            </FieldBody>
        </Field>
    </Validation>
    <Validation>
        <Field>
            <FieldLabel>Description</FieldLabel>
            <FieldBody>
                <TextEdit @bind-Text="@Company.Description" />
            </FieldBody>
        </Field>
    </Validation>
    <Field>
        <Switch @bind-Checked="@Company.UseAlphaCode">Use AlphaCode</Switch>
    </Field>
    <Fields>
        <Validation>
            <Field>
                <FieldLabel>AlphaCode</FieldLabel>
                <FieldBody>
                    <TextEdit @bind-Text="@Company.AlphaCode" />
                </FieldBody>
            </Field>
        </Validation>
        <Validation>
            <Field>
                <FieldLabel>BetaCode</FieldLabel>
                <FieldBody>
                    <TextEdit @bind-Text="@Company.BetaCode" />
                </FieldBody>
            </Field>
        </Validation>
    </Fields>
</Validations>

@code {
    Validations Validations;
    CompanyInfo Company = new CompanyInfo()
        {
            UseAlphaCode = true,
        };

    public void StatusChanged(ValidationsStatusChangedEventArgs e)
    {
        Console.WriteLine( "EditContext MEssages" );
        Validations.EditContext.GetValidationMessages().ToList().ForEach(Console.WriteLine);

        Console.WriteLine( "Status MEssages" );
        e.Messages.ToList().ForEach(Console.WriteLine);
    }

    public class CompanyInfo : IValidatableObject
    {
        [Required( ErrorMessage = "Name is required" )]

        [Required( ErrorMessage = "Description is required" )]
        public string Description { get; set; }

        public bool UseAlphaCode { get; set; }

        public string AlphaCode { get; set; }

        public string BetaCode { get; set; }

        [Range( 0, 999.99 )]
        public decimal Price { get; set; }

        public IEnumerable<ValidationResult> Validate( ValidationContext validationContext )
        {


            if ( UseAlphaCode )
            {
                if ( String.IsNullOrWhiteSpace( AlphaCode ) )
                {
                    yield return new ValidationResult( "AlphaCode is required", new[] { "AlphaCode" } );
                }
            }
            else
            {
                if ( String.IsNullOrWhiteSpace( BetaCode ) )
                {
                    yield return new ValidationResult( "BetaCode is required", new[] { "BetaCode" } );
                }
            }
        }
    }
}
```

There's also this similar open issue : #4869
The EditContext validation messages don't seem to quite match our own validations messages (our own are correct)
image

Source/Blazorise/Components/Validation/Validations.razor.cs
stsrki: It was an unnecessary call, so I removed it.

stsrki:
> The EditContext validation messages don't seem to quite match our own validations messages (our own are correct)

As far as I know, I might be wrong since this was true in the past, but EditContext has some very limited support for IValidatableObject. Nevertheless, I will see what can be done.

stsrki:
Ok, with some reflection magic, I was able to fix the duplicate message from the EditContext. Essentially I have called EditContext internal API and checked against the validating field. If there is a match, I clear the store.

@David-Moreira You're better at it then me. Can you optimize it so that we precompile reflection calls to expression trees? That way we would make sure it is fast enough.

David-Moreira:
> <Validations @ref=Validations Model="@Company" StatusChanged="StatusChanged" Mode="ValidationMode.Auto">
>     <Validation>
>         <Field>
>             <FieldLabel>Name</FieldLabel>
>             <FieldBody>
>                 <TextEdit @bind-Text="@Company.Name" />
>             </FieldBody>
>         </Field>
>     </Validation>
>     <Validation>
>         <Field>
>             <FieldLabel>Description</FieldLabel>
>             <FieldBody>
>                 <TextEdit @bind-Text="@Company.Description" />
>             </FieldBody>
>         </Field>
>     </Validation>
>     <Field>
>         <Switch @bind-Checked="@Company.UseAlphaCode">Use AlphaCode</Switch>
>     </Field>
>     <Fields>
>         <Validation>
>             <Field>
>                 <FieldLabel>AlphaCode</FieldLabel>
>                 <FieldBody>
>                     <TextEdit @bind-Text="@Company.AlphaCode" />
>                 </FieldBody>
>             </Field>
>         </Validation>
>         <Validation>
>             <Field>
>                 <FieldLabel>BetaCode</FieldLabel>
>                 <FieldBody>
>                     <TextEdit @bind-Text="@Company.BetaCode" />
>                 </FieldBody>
>             </Field>
>         </Validation>
>     </Fields>
> </Validations>
> 
> @code {
>     Validations Validations;
>     CompanyInfo Company = new CompanyInfo()
>         {
>             UseAlphaCode = true,
>         };
> 
>     public void StatusChanged(ValidationsStatusChangedEventArgs e)
>     {
>         Console.WriteLine( "EditContext MEssages" );
>         Validations.EditContext.GetValidationMessages().ToList().ForEach(Console.WriteLine);
> 
>         Console.WriteLine( "Status MEssages" );
>         e.Messages.ToList().ForEach(Console.WriteLine);
>     }
> 
>     public class CompanyInfo : IValidatableObject
>     {
>         [Required( ErrorMessage = "Name is required" )]
>         public string Name { get; set; }
> 
>         [Required( ErrorMessage = "Description is required" )]
>         public string Description { get; set; }
> 
>         public bool UseAlphaCode { get; set; }
> 
>         public string AlphaCode { get; set; }
> 
>         public string BetaCode { get; set; }
> 
>         [Range( 0, 999.99 )]
>         public decimal Price { get; set; }
> 
>         public IEnumerable<ValidationResult> Validate( ValidationContext validationContext )
>         {
> 
> 
>             if ( UseAlphaCode )
>             {
>                 if ( String.IsNullOrWhiteSpace( AlphaCode ) )
>                 {
>                     yield return new ValidationResult( "AlphaCode is required", new[] { "AlphaCode" } );
>                 }
>             }
>             else
>             {
>                 if ( String.IsNullOrWhiteSpace( BetaCode ) )
>                 {
>                     yield return new ValidationResult( "BetaCode is required", new[] { "BetaCode" } );
>                 }
>             }
>         }
>     }
> }

Did you take a look at this?

David-Moreira:
> Ok, with some reflection magic, I was able to fix the duplicate message from the EditContext. Essentially I have called EditContext internal API and checked against the validating field. If there is a match, I clear the store.
> 
> @David-Moreira You're better at it then me. Can you optimize it so that we precompile reflection calls to expression trees? That way we would make sure it is fast enough.

I think I've used very similar code for something similar I need before haha.

I think we have some helpers for that. Anyway it isn't an hot path, so it would be fine.

David-Moreira:
> > Ok, with some reflection magic, I was able to fix the duplicate message from the EditContext. Essentially I have called EditContext internal API and checked against the validating field. If there is a match, I clear the store.
> > @David-Moreira You're better at it then me. Can you optimize it so that we precompile reflection calls to expression trees? That way we would make sure it is fast enough.
> 
> I think I've used very similar code for something similar I need before haha.
> 
> I think we have some helpers for that. Anyway it isn't an hot path, so it would be fine.

Done.

David-Moreira:
image

Maybe a `FirstOrDefault` is clearer that we are iterating the collection just to make an operation on a single record.

stsrki:
> Did you take a look at this?

I did. I tested duplicate messages with it.

David-Moreira:
> > Did you take a look at this?
> 
> I did. I tested duplicate messages with it.

I'm almost positive that click the switch does not trigger validation, is it not suppoesd to?

stsrki:
> > > Did you take a look at this?
> > 
> > 
> > I did. I tested duplicate messages with it.
> 
> I'm almost positive that clicking the switch does not trigger validation, is it not suppoesd to?

For the time being, it is not supported. First, we will have to expose some internal APIs in case users wish to do so. #4877 might be enough.

If we would support re-triggering validation, it would lead to more mess. Possible circular re-validations. How do we even tackle that? How do we know when to stop re-validating fields between? I will just leave it as it is for now.

David-Moreira:
> > > > Did you take a look at this?
> > > 
> > > 
> > > I did. I tested duplicate messages with it.
> > 
> > 
> > I'm almost positive that clicking the switch does not trigger validation, is it not suppoesd to?
> 
> For the time being, it is not supported. First, we will have to expose some internal APIs in case users wish to do so. #4877 might be enough.
> 
> If we would support re-triggering validation, it would lead to more mess. Possible circular re-validations. How do we even tackle that? How do we know when to stop re-validating fields between? I will just leave it as it is for now.

I don't think I'm getting the problem. If I just change the `Switch` it should trigger the validation once, currently it does not trigger even once. Is that a limitation? Am I missing something? For example the check auto validates successfully upon click:
image

How would it be validating multiple times?

stsrki:
It does not trigger validation because nothing changes on the Input components that should revalidate. Our validation happens when the user types something or when the Input component receives the new value through a parameter. In this case, none of that happens.

David-Moreira:
> It does not trigger validation because nothing changes on the Input components that should revalidate. Our validation happens when the user types something or when the Input component receives the new value through a parameter. In this case, none of that happens.

I still don't get it. Maybe I'm just not thinking about it correctly.
`UseAlphaCode` is part of the `Model` that's bound to the `Validations`. `UseAlphaCode` is used on a `Switch` component, therefore any change to the `Switch` component, on or off, means the `UseAlphaCode` now has a different value and needs to trigger **Validation**, no?

stsrki:
But ` public bool UseAlphaCode { get; set; }` is not bound to any validation; therefore it is not tracked by anything.

David-Moreira:
> But ` public bool UseAlphaCode { get; set; }` is not bound to any validation; therefore it is not tracked by anything.

AHHHHHHHHH, I'm dumb. I forgot it needs to be wrapped by a `Validation` component to track haha.
Well what's the behaviour if it's tracked, that what really matters, no?

stsrki:
Even if `UseAlphaCode` is validated, it wouldn't know how to trigger AlphaCode or BetaCode validations.

David-Moreira:
So I just made it validate. Yes, I would expect that if I make changes to `UseAlphaCode`, the fields that are outside of it are updated. This is the problem you're finding right?

If that's a "breaking change", at least logically, then we could add some kind of method or attribute so the user can explicitly tell it to refresh the validation status of the whole model. 

i.e:
image

David-Moreira:
Well you're more intimate with the validations, so you would know what's best.
I do believe this should work, specially since the contract provided allows to refer to other member's names.

However it should not trigger a validation in my opinion, you already have the required validation information there, because as you've implied this would probably just loop indefinitely.

If you would like to trigger a validation, then it would need to track whether it has gonne through every field itnernally in a single "validation session", as to not go through them again.

But I'm talking about this without having so much intricate knowledge of our validation system internals. So again you know best.

Let me know if you need me to give it a go ; see if I have a diff perspective.

stsrki:
The only time when we could have a full validation state for all fields is when we ValidateAll in Manual mode. In all other situation every validation is running in their own context. Knowing when validation starts ans stops between all other validations is hard to do in this way. 

But, you got me thinking. We could create a new validation transaction system. You mentioned that we have full model with all fields already known. With these information we could keep track of all fields, and their validation state.

...l will need to continue thinking about this some more...

stsrki:
> The only time when we could have a full validation state for all fields is when we ValidateAll in Manual mode. In all other situation every validation is running in their own context. Knowing when validation starts ans stops between all other validations is hard to do in this way.
> 
> But, you got me thinking. We could create a new validation transaction system. You mentioned that we have full model with all fields already known. With these information we could keep track of all fields, and their validation state.
> 
> ...l will need to continue thinking about this some more...

There is not enough time to do this in 1.3. It would require some major refactorings on the validation system. So I will do it later in 1.4, or 2.0. At least we now have basic support for IValidatableObject, which is enough I believe. I you think that's OK we can proceed with merging.

David-Moreira:
> > The only time when we could have a full validation state for all fields is when we ValidateAll in Manual mode. In all other situation every validation is running in their own context. Knowing when validation starts ans stops between all other validations is hard to do in this way.
> > But, you got me thinking. We could create a new validation transaction system. You mentioned that we have full model with all fields already known. With these information we could keep track of all fields, and their validation state.
> > ...l will need to continue thinking about this some more...
> 
> There is not enough time to do this in 1.3. It would require some major refactorings on the validation system. So I will do it later in 1.4, or 2.0. At least we now have basic support for IValidatableObject, which is enough I believe. I you think that's OK we can proceed with merging.

Go ahead, please open an issue for 1.4 so we don't forget. :)

stsrki:
> Go ahead, please open an issue for 1.4 so we don't forget. :)

Done

David-Moreira (approved):

PR #4866, opened 2023-07-05
Fixes more typos in date picker localization

InzeNL:
Addition to: https://github.com/Megabit/Blazorise/pull/4846

Found another typo in "Sunday" being "Sunsay"

I reverted the "fix" from "May!" to "May" since I misinterpreted the function of it. It was not a typo, but used to separate "May" as a shorthand from "May" as a completely written out month (Different in Czech for example, "Kvě" against "Květen")


stsrki (approved):
LGTM. Thank you!

PR #4868, opened 2023-07-05
DataGrid: add HeaderGroupCaption implementation

David-Moreira:
Here's an initial implementation for this feature.

Got it to work without changing too much internals.

However it is of note that we changed DisplayableColumns to keep everything in sync. And it does a little bit more work.
I'm not sure we can optimize it? We always need to figure out if a column with `HeaderGroupCaption` exists.

Do you think 
shortcircuting the logic with
`Columns.Any(x=> !string.IsNullOrWhiteSpace(x.HeaderGroupCaption))`
would be better? It still goes through every item most times, but it will probably save allocations from the new logic otherwise?

I've left an example on the Dashboard, but basically we get this.
image


Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -1788,12 +1788,85 @@ internal IEnumerable<DataGridColumn<TItem>> DisplayableColumns
...
+
+            return newOrderedDisplayColumns;
+
+        }
+    }
+
+    /// <summary>
+    /// Gets or sets whether user can see group header column captions.
+    /// </summary>
+    internal bool IsGroupHeaderCaptionsEnabled
```
stsrki, line 1867: What if we have `ShowHeaderGroupCaptions`, just as we have `ShowCaptions`? I know the user would have to enable it strictly, but then we have less work to check when rendering.

David-Moreira: I guess it's okay...

stsrki:
Do we want to have `HeaderGroupCaptionTemplate`?

David-Moreira:
> IsGroupHeaderCaptionsEnabled

Sky is the limit!

David-Moreira:
> Do we want to have `HeaderGroupCaptionTemplate`?

Actually that does not mesh well, because you want to have a single Caption per Group. While defining the new **HeaderGroupCaption** in every column we want to be part of a group and having it act as a key, the same does not seem possible with **HeaderGroupCaptionTemplate**. 
We'd need to look for a defined Template in that group, but then there can be multiple, seems like it would be kind of wonky.

Maybe if we've had it defined at **DataGrid** level instead? But then you can't really define it per group.
What are your thoughts?

stsrki:
> > Do we want to have `HeaderGroupCaptionTemplate`?
> 
> Actually that does not mesh well, because you want to have a single Caption per Group. While defining the new **HeaderGroupCaption** in every column we want to be part of a group and having it act as a key, the same does not seem possible with **HeaderGroupCaptionTemplate**. We'd need to look for a defined Template in that group, but then there can be multiple, seems like it would be kind of wonky.
> 
> Maybe if we've had it defined at **DataGrid** level instead? But then you can't really define it per group. What are your thoughts?

It should be on the DataGrid level. And it can receive group names as `context` so that the user can decide how to render the content based on the condition. But, we can make it in the future if it takes too long.

David-Moreira:
> > > Do we want to have `HeaderGroupCaptionTemplate`?
> > 
> > 
> > Actually that does not mesh well, because you want to have a single Caption per Group. While defining the new **HeaderGroupCaption** in every column we want to be part of a group and having it act as a key, the same does not seem possible with **HeaderGroupCaptionTemplate**. We'd need to look for a defined Template in that group, but then there can be multiple, seems like it would be kind of wonky.
> > Maybe if we've had it defined at **DataGrid** level instead? But then you can't really define it per group. What are your thoughts?
> 
> It should be on the DataGrid level. And it can receive group names as `context` so that the user can decide how to render the content based on the condition. But, we can make it in the future if it takes too long.

That's a good idea.

David-Moreira:
@stsrki Friendly reminder

stsrki:
> However it is of note that we changed DisplayableColumns to keep everything in sync. And it does a little bit more work.
> I'm not sure we can optimize it? We always need to figure out if a column with `HeaderGroupCaption` exists.

I'm not a big fan of doing this, to always allocate a new list and iterate to return displayable columns. But as long as it works we can always optimize it later.

stsrki:
All in all. The feature works as expected. I like it.

David-Moreira:
> All in all. The feature works as expected. I like it.

Well, go ahead and merge if you're okay with the feature then!

PR #4883, opened 2023-07-15
Video quality options

stsrki:
https://github.com/Megabit/Blazorise/discussions/4881

### DIFF
--- a/Demos/Blazorise.Demo/Pages/Tests/VideoPage.razor
+++ b/Demos/Blazorise.Demo/Pages/Tests/VideoPage.razor
@@ -19,21 +19,32 @@
                 </Buttons>
                 <Button Color="Color.Primary" Clicked="@OnChangeVideoUrl">Change Video</Button>
                 <Button Color="Color.Primary" Clicked="@OnChangeCurrentTime">Go to 20 seconds</Button>
+                <Button Color="Color.Secondary" Clicked="@OnChangeQuality">Change Quality to 1080</Button>
             </CardBody>
             <CardBody>
-                <Video @ref="@videoRef" Source="@videoUrl" CurrentTime="@currentTime" />
+                <Video @ref="@videoRef" Source="@videoSource" CurrentTime="@currentTime" DefaultQuality="@quality" />
             </CardBody>
         </Card>
     </Column>
 </Row>
 @code {
     Video videoRef;
-    string videoUrl = "http://commondatastorage.googleapis.com/gtv-videos-bucket/sample/BigBuckBunny.mp4";
+    VideoSource videoSource = new VideoSource()
+    {
+        Medias = new ValueEqualityList<VideoMedia>
+        {
+            new VideoMedia("https://cdn.plyr.io/static/demo/View_From_A_Blue_Moon_Trailer-576p.mp4", "video/mp4", 576),
+            new VideoMedia("https://cdn.plyr.io/static/demo/View_From_A_Blue_Moon_Trailer-720p.mp4", "video/mp4", 720),
+            new VideoMedia("https://cdn.plyr.io/static/demo/View_From_A_Blue_Moon_Trailer-1080p.mp4", "video/mp4", 1080),
+        }
+    };
+
     double currentTime = 10;
+    int quality = 576;
 
     Task OnChangeVideoUrl()
     {
-        videoUrl = "http://commondatastorage.googleapis.com/gtv-videos-bucket/sample/ElephantsDream.mp4";
+        videoSource = "http://commondatastorage.googleapis.com/gtv-videos-bucket/sample/ElephantsDream.mp4";
         currentTime = 0;
 
         return Task.CompletedTask;
@@ -42,6 +53,13 @@
     Task OnChangeCurrentTime()
     {
         currentTime = 20;
+
+        return Task.CompletedTask;
+    }
+
+    Task OnChangeQuality()
+    {
+        quality = 1080;
 
         return Task.CompletedTask;
     }
--- a/Source/Extensions/Blazorise.Video/Video.razor.cs
+++ b/Source/Extensions/Blazorise.Video/Video.razor.cs
@@ -632,7 +632,7 @@ public partial class Video : BaseComponent, IAsyncDisposable
     /// <summary>
     /// If the default controls are used, you can specify which settings to show in the menu.
     /// </summary>
-    [Parameter] public string[] SettingsList { get; set; } = new string[] { VideoSettingsType.Captions, VideoSettingsType.Quality, VideoSettingsType.Speed, VideoSettingsType.Loop };
+    [Parameter] public VideoSettingsType[] SettingsList { get; set; } = new VideoSettingsType[] { VideoSettingsType.Captions, VideoSettingsType.Quality, VideoSettingsType.Speed, VideoSettingsType.Loop };
 
     /// <summary>
     /// Sent periodically to inform interested parties of progress downloading the media. Information about the current amount of the media that has been downloaded is available in the media element's buffered attribute.
--- a/Source/Extensions/Blazorise.Video/Video.razor.cs
+++ b/Source/Extensions/Blazorise.Video/Video.razor.cs
@@ -735,7 +735,10 @@ public partial class Video : BaseComponent, IAsyncDisposable
     [Parameter] public Func<Task> Ready { get; set; }
 
     /// <summary>
-    /// The quality of playback has changed.
+    /// The quality of playback has changed. 
+    /// <para>
+    /// If the quality argument is <c>null</c> it is considered that a default quality value was selected.
+    /// </para>
     /// </summary>
     [Parameter] public Func<int?, Task> QualityChanged { get; set; }
 
--- a/Source/Extensions/Blazorise.Video/Video.razor.cs
+++ b/Source/Extensions/Blazorise.Video/Video.razor.cs
@@ -70,6 +70,7 @@ public partial class Video : BaseComponent, IAsyncDisposable
             {
                 Controls,
                 ControlsList,
+                SettingsList,
                 AutomaticallyHideControls,
                 AutoPlay,
                 AutoPause,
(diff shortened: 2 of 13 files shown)


Demos/Blazorise.Demo/Pages/Tests/VideoPage.razor
```diff
@@ -19,21 +19,32 @@
...
             </CardBody>
         </Card>
     </Column>
 </Row>
 @code {
     Video videoRef;
-    string videoUrl = "http://commondatastorage.googleapis.com/gtv-videos-bucket/sample/BigBuckBunny.mp4";
+    VideoSource videoSource = new VideoSource()
+    {
+        Medias = new ValueEqualityList<VideoMedia>
```
David-Moreira, line 34: What api is this? `ValueEqualityList`. a user is supposed to use it?

Demos/Blazorise.Demo/Pages/Tests/VideoPage.razor
```diff
@@ -19,21 +19,32 @@
...
+        {
+            new VideoMedia("https://cdn.plyr.io/static/demo/View_From_A_Blue_Moon_Trailer-576p.mp4", "video/mp4", 576),
+            new VideoMedia("https://cdn.plyr.io/static/demo/View_From_A_Blue_Moon_Trailer-720p.mp4", "video/mp4", 720),
+            new VideoMedia("https://cdn.plyr.io/static/demo/View_From_A_Blue_Moon_Trailer-1080p.mp4", "video/mp4", 1080),
```
David-Moreira, line 38: Can we get some static helpers for the quality? Or we shouldn't bother? 720p ; 1080p ; etc...

Source/Extensions/Blazorise.Video/Video.razor.cs
```diff
@@ -613,11 +629,26 @@ public Task NotifyReady()
     /// </summary>
     [Parameter] public string[] ControlsList { get; set; } = new string[] { VideoControlsType.PlayLarge, VideoControlsType.Play, VideoControlsType.Progress, VideoControlsType.CurrentTime, VideoControlsType.Mute, VideoControlsType.Volume, VideoControlsType.Captions, VideoControlsType.Settings, VideoControlsType.Pip, VideoControlsType.Airplay, VideoControlsType.Fullscreen };
 
+    /// <summary>
+    /// If the default controls are used, you can specify which settings to show in the menu.
+    /// </summary>
+    [Parameter] public string[] SettingsList { get; set; } = new string[] { VideoSettingsType.Captions, VideoSettingsType.Quality, VideoSettingsType.Speed, VideoSettingsType.Loop };
```
David-Moreira, line 635: Well, would be cool to have enum flags for this maybe? Would that not make the user api better?

Source/Extensions/Blazorise.Video/VideoSource.cs
```diff
@@ -64,14 +61,23 @@ public VideoSource( string[] sources )
...
+
     #endregion
 
     #region Properties
 
     /// <summary>
     /// True if the source contains the single address.
     /// </summary>
-    public bool Indexed { get; private set; }
+    public bool Indexed => Medias?.Count > 1;
```
David-Moreira, line 80: True if source contains single address?

What is this for, and why not > 0?

Demos/Blazorise.Demo/Pages/Tests/VideoPage.razor
stsrki: It's a custom-overloaded list that is used to compare only values instead of the list references.

Demos/Blazorise.Demo/Pages/Tests/VideoPage.razor
stsrki: I'm not sure what that helper would do. Any idea?

Source/Extensions/Blazorise.Video/Video.razor.cs
```diff
@@ -481,6 +484,19 @@ public Task NotifyReady()
...
+    /// </summary>
+    /// <returns>A task that represents the asynchronous operation.</returns>
+    [JSInvokable]
+    public Task NotifyQualityChange( int? quality )
+    {
+        if ( QualityChanged != null )
+            return QualityChanged.Invoke( quality );
+
+        return Task.CompletedTask;
+    }
```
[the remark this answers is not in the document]
stsrki: Do you mean `quality` parameter? Well, I think in this context `null` as a valid value, so I would say yes.

Source/Extensions/Blazorise.Video/Video.razor.cs
stsrki: I guess its a good idea.

Source/Extensions/Blazorise.Video/Video.razor.cs
```diff
@@ -703,6 +734,11 @@ public Task NotifyReady()
     /// </summary>
     [Parameter] public Func<Task> Ready { get; set; }
 
+    /// <summary>
+    /// The quality of playback has changed.
+    /// </summary>
+    [Parameter] public Func<int?, Task> QualityChanged { get; set; }
```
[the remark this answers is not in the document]
stsrki: I will write the comment explaining it.

Source/Extensions/Blazorise.Video/VideoSource.cs
stsrki: Because if there is more than 1 element, it is considered as an array, and it will be serialized as an array. Otherwise, it is serialized as a primitive type.

David-Moreira: Maybe let's explain that? 
Maybe IsArray better conveys that meaning?

Demos/Blazorise.Demo/Pages/Tests/VideoPage.razor
David-Moreira: Just something like 
VideoPlayer.Quality.480p
VideoPlayer.Quality.720p
VideoPlayer.Quality.1080p
VideoPlayer.Quality.1440p
VideoPlayer.Quality.2160p
with some predefined common values.

Demos/Blazorise.Demo/Pages/Tests/VideoPage.razor
David-Moreira: Ok then, I guess it's fine, they'll have to new it up anyway.

Source/Extensions/Blazorise.Video/Video.razor.cs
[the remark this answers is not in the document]
David-Moreira: So if I have a quality set to 1080p and it is invoked as null nothing happens? 
Is there like a default? And if the default is 720p should it trigger with that?

Source/Extensions/Blazorise.Video/VideoSource.cs
David-Moreira: Or `SerializeAsArray`

Demos/Blazorise.Demo/Pages/Tests/VideoPage.razor
stsrki: Video can have any size, so it is impossible to make it generic. Skipping this.

Source/Extensions/Blazorise.Video/Video.razor.cs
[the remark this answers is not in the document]
stsrki: It would trigger with anything that `plyr` fires. But since the user picks quality from the dropdown menu on the video player it is highly unlikely it will be null.

stsrki:
@David-Moreira done with the changes.

Demos/Blazorise.Demo/Pages/Tests/VideoPage.razor
David-Moreira: It's an helper class with predefined int values. I'm not talking about making anything generic.

Source/Extensions/Blazorise.Video/Video.razor.cs
```diff
@@ -632,7 +632,7 @@ public Task NotifyQualityChange( int? quality )
     /// <summary>
     /// If the default controls are used, you can specify which settings to show in the menu.
     /// </summary>
-    [Parameter] public string[] SettingsList { get; set; } = new string[] { VideoSettingsType.Captions, VideoSettingsType.Quality, VideoSettingsType.Speed, VideoSettingsType.Loop };
+    [Parameter] public VideoSettingsType[] SettingsList { get; set; } = new VideoSettingsType[] { VideoSettingsType.Captions, VideoSettingsType.Quality, VideoSettingsType.Speed, VideoSettingsType.Loop };
```
David-Moreira, line 635: Actually I meant this to be flags that you can set like: 
`SettingsList = VideoSettingsType.Captions | VideoSettingsType.Quality | VideoSettingsType.Speed | VideoSettingsType.Loop`

David-Moreira (approved):

Source/Extensions/Blazorise.Video/Video.razor.cs
David-Moreira: I mean, it's just a bit more elegant api and one less allocation. 
I don't care for it, if internally it's harder to handle. (Another option would be to have a custom internal property that would convert the values into valid strings the js expects, anyway, if it's more complex to do it, not worth it in relation to the benefits I believe.)

stsrki: I would argue that using flags is even harder. For example, how would you dynamically combine flags? With an array type, you can just `Add` or `Remove` values from a temporarily lost and set it like `SettingsList="someList.ToArray()"`.

David-Moreira: It's just the usage of bitwise operators:
Add
`|=`
Remove
`^=`
And there's an helper `HasFlag` so you don't have to use a bitwise operation to calc whether it exists.

I guess there's the chance that devs might not be quite as familiar with flags and how they work, yea let's keep the array, nvm.

stsrki: I am familiar with flags. But not all devs are. That's my concern.

PR #4900, opened 2023-07-28
Snackbar: new positions and animation support

### LINKED ISSUE #1451: Snackbar extension of the Location property for top of the screen.
Wanted the ability to pick other locations where the snackbar could be.  Ultimately would like to put it in the top right corner. 

To not be a breaking change, a second property could be added to pick between Top and Bottom and default to Bottom.

steal4life:
Closes #1451


Source/Extensions/Blazorise.Snackbar/Snackbar.razor.cs
```diff
@@ -65,7 +69,8 @@ public partial class Snackbar : BaseComponent, IDisposable
     protected override void BuildClasses( ClassBuilder builder )
     {
         builder.Append( "snackbar" );
-        builder.Append( "snackbar-show", Visible );
+        builder.Append( Visible && !closeAnimation ? "snackbar-show" : null );
```
stsrki, line 72:
```suggestion
        builder.Append( "snackbar-show", Visible && !closeAnimation );
```

Source/Extensions/Blazorise.Snackbar/Snackbar.razor.cs
```diff
@@ -82,6 +87,13 @@ protected override void OnInitialized()
             countdownTimer.Elapsed += OnCountdownTimerElapsed;
         }
 
+        if ( closeAnimationTimer == null )
+        {
+            closeAnimationTimer = new( 500 );
```
stsrki, line 92: Add parameter CloseAnimationInterval so that it isn't hardcoded at 500.

Source/Extensions/Blazorise.Snackbar/Snackbar.razor.cs
```diff
@@ -130,12 +139,23 @@ public Task Show()
         return InvokeAsync( StateHasChanged );
     }
 
+    async Task ProposeHide( SnackbarCloseReason closeReason )
+    {
+        closeAnimation = true;
+        closeAnimationTimer.Delay( 500 );
```
stsrki, line 145: Add parameter `CloseAnimationInterval` so that it isn't hardcoded at 500.

Source/Extensions/Blazorise.Snackbar/Snackbar.razor.cs
```diff
@@ -130,12 +139,23 @@ public Task Show()
         return InvokeAsync( StateHasChanged );
     }
 
+    async Task ProposeHide( SnackbarCloseReason closeReason )
```
stsrki, line 142: Rename `ProposeHide` to something like `BeginHide`, or `ExecuteHide`.

Source/Extensions/Blazorise.Snackbar/Utils/Extensions.cs
```diff
@@ -20,8 +20,9 @@ public static string GetName( this SnackbarLocation snackbarLocation )
                 return "bottom-right";
             case SnackbarLocation.Default:
             case SnackbarLocation.Bottom:
+                return "bottom";
             default:
-                return null;
+                return "";
```
stsrki, line 25: Always return null instead of empty string.

Source/Extensions/Blazorise.Snackbar/Utils/Extensions.cs
```diff
@@ -43,8 +44,9 @@ public static string GetName( this SnackbarStackLocation snackbarStackLocation )
                 return "bottom-right";
             case SnackbarStackLocation.Center:
             case SnackbarStackLocation.Bottom:
+                return "bottom";
             default:
-                return null;
+                return "";
```
stsrki, line 49: Always return null instead of empty string.

Documentation/Blazorise.Docs/Pages/Docs/Extensions/Snackbar/SnackbarPage.razor
```diff
@@ -104,9 +104,12 @@
     <DocsAttributesItem Name="Multiline" Type="bool" Default="false">
         Allow snackbar to show multiple lines of text.
     </DocsAttributesItem>
-    <DocsAttributesItem Name="DefaultInterval" Type="double" Default="5000">
+    <DocsAttributesItem Name="Interval" Type="double" Default="5000">
```
David-Moreira, line 107: Are we changing the naming of an existing **Parameter**?

Demos/Blazorise.Demo/Pages/Tests/SnackbarsPage.razor
```diff
@@ -157,15 +177,19 @@
...
+
+ @code {
     Snackbar snackbar;
     Snackbar snackbarAction;
     Snackbar snackbarMultiline;
     Snackbar snackbarLeft;
     Snackbar snackbarRight;
+    Snackbar snackbarTop;
+    Snackbar snackbarTopEnd;
+    Snackbar snackbarTopStart;
```
David-Moreira, line 192: Could almost just have a configurable location parameter?

Source/Extensions/Blazorise.Snackbar/Enums/SnackbarLocation.cs
```diff
@@ -0,0 +1,59 @@
...
+    [Obsolete( "Use BottomEnd instead." )]
+    End,
+
+    /// <summary>
+    /// Show the snackbar on the top side of the screen.
+    /// </summary>
+    Top,
+
+    /// <summary>
+    /// Show the snackbar on the top-left side of the screen.
```
David-Moreira, line 51: the documentation refers to left & right. it should refer to start or end. we should not assume the left or right position when naming Start/End.

Source/Extensions/Blazorise.Snackbar/Snackbar.razor.cs
```diff
@@ -44,7 +46,12 @@ public partial class Snackbar : BaseComponent, IDisposable
     private CountdownTimer countdownTimer;
 
     /// <summary>
-    /// Flag that indicates if snackbar close action was delayed.
+    /// A flag that indicates if the snackbar is currently being closed with animation.
+    /// </summary>
+    private bool closeAnimation = false;
```
David-Moreira, line 51: Would `closingAnimation` be a better naming then? It would be consistent with `closingDelayed` and the documentation we've added also talks about "being closed" => closing

Source/Extensions/Blazorise.Snackbar/Utils/Extensions.cs
```diff
@@ -6,11 +6,21 @@ public static string GetName( this SnackbarLocation snackbarLocation )
     {
         switch ( snackbarLocation )
         {
+            case SnackbarLocation.Top:
+                return "top";
+            case SnackbarLocation.TopStart:
+                return "top-left";
+            case SnackbarLocation.TopEnd:
+                return "top-right";
```
David-Moreira, line 14: Can we have start/end support? Or does it not make sense here?

Source/Extensions/Blazorise.Snackbar/wwwroot/blazorise.snackbar.css
```diff
@@ -1,7 +1,7 @@
 ﻿.snackbar {
   align-items: center;
   background-color: var(--b-snackbar-background, #323232);
-  color: var(--b-snackbar-text-color, white);
+  color: var(--b-snackbar-text-color, rgb(255, 255, 255));
```
David-Moreira, line 4: curious, any special reason for this apparent redudant change?

David-Moreira (changes requested):

David-Moreira:
1. The **Top-Right-Aligned** takes a bit too long to show. Also if we want to show animations maybe it should be more explicit? A configuration like it's on the other pages? That just sets the animation for all the snackbars in the page? So the user or us can try it for every single one.
http://localhost:7713/tests/snackbar
2. Does the positioning also work correctly with stacked snackbars? That would be great.
3. Update enum doc:
image

Documentation/Blazorise.Docs/Pages/Docs/Extensions/Snackbar/SnackbarPage.razor
stsrki: This was wrong in the docs. It was always named `Interval`.

Source/Extensions/Blazorise.Snackbar/Enums/SnackbarLocation.cs
steal4life: So you think it should be Top-Start?

Source/Extensions/Blazorise.Snackbar/Snackbar.razor.cs
steal4life: Yes, definitely

Source/Extensions/Blazorise.Snackbar/Utils/Extensions.cs
steal4life: if i understand, you think that placing the snackbar in the middle of the screen on the left or right side is what can be added?... If so, i don't think it makes sense because the snackbar animations work by showing it from the bottom or top of the screen, possibly add an animation like offcanvas to slide on left or right side of the screen

steal4life:
> 1. The **Top-Right-Aligned** takes a bit too long to show. Also if we want to show animations maybe it should be more explicit? A configuration like it's on the other pages? That just sets the animation for all the snackbars in the page? So the user or us can try it for every single one.
>    http://localhost:7713/tests/snackbar
> 2. Does the positioning also work correctly with stacked snackbars? That would be great.
> 3. Update enum doc:
>    image

So the Top right appears longer due to the added parameter
AnimationDuration, which determines the length of the snackbar animation (how long it will last), with this we have enabled users to determine the length of the snackbar animation themselves in their future projects if they want so.

Yes, we added that the stack can also have the same positions as the snackbar.

Source/Extensions/Blazorise.Snackbar/Utils/Extensions.cs
stsrki: Agree, no need to make it at center left/right.

Source/Extensions/Blazorise.Snackbar/Enums/SnackbarLocation.cs
David-Moreira: I personally do. In the future we plan to have RTL support. So in that case the doc would be incorrect.

Source/Extensions/Blazorise.Snackbar/Utils/Extensions.cs
David-Moreira: I was asking about the css class naming. If it needs to be left right instead of start end

David-Moreira:
> > 1. The **Top-Right-Aligned** takes a bit too long to show. Also if we want to show animations maybe it should be more explicit? A configuration like it's on the other pages? That just sets the animation for all the snackbars in the page? So the user or us can try it for every single one.
> >    http://localhost:7713/tests/snackbar
> > 2. Does the positioning also work correctly with stacked snackbars? That would be great.
> > 3. Update enum doc:
> >    image
> 
> So the Top right appears longer due to the added parameter AnimationDuration, which determines the length of the snackbar animation (how long it will last), with this we have enabled users to determine the length of the snackbar animation themselves in their future projects if they want so.
> 
> Yes, we added that the stack can also have the same positions as the snackbar.

But does it make sense for 1 to be configurable in the page? Helps testing and seeing every use case.

steal4life:
I just added that parameter on one element to see the diference bettwen
that one and others, but we can do as you sad so it will make more sense.

steal4life:
Yes, tehnicaly in scss there is for instance snackbar-top-right etc... So
if we want to change the name here we should also do the same in scss.

Documentation/Blazorise.Docs/Pages/Docs/Helpers/Enums/SnackbarPage.razor
```diff
@@ -9,9 +9,15 @@
...
+            <UnorderedListItem><Code>Bottom</Code> Show the snackbar on the bottom side of the screen.</UnorderedListItem>
+            <UnorderedListItem><Code>BottomStart</Code> Show the snackbar on the bottom-left side of the screen.</UnorderedListItem>
+            <UnorderedListItem><Code>BottomEnd</Code> Show the snackbar on the bottom-right side of the screen.</UnorderedListItem>
+            <UnorderedListItem><Code>Start</Code> Show the snackbar on the left side of the screen (obsolete, use <Code>BottomStart</Code> instead).</UnorderedListItem>
+            <UnorderedListItem><Code>End</Code> Show the snackbar on the right side of the screen (obsolete, use <Code>BottomEnd</Code> instead).</UnorderedListItem>
+            <UnorderedListItem><Code>Top</Code> Show the snackbar on the top side of the screen.</UnorderedListItem>
+            <UnorderedListItem><Code>TopStart</Code> Show the snackbar on the top-left side of the screen.</UnorderedListItem>
+            <UnorderedListItem><Code>TopEnd</Code> Show the snackbar on the top-right side of the screen.</UnorderedListItem>
```
David-Moreira, line 20: left/right vs start/end

David-Moreira (changes requested):

David-Moreira:
In demo, Multiline snackbar and Info snackbar don't seem to be working with animation, either the parameter was forgotten or something's up

Source/Extensions/Blazorise.Snackbar/wwwroot/blazorise.snackbar.css
stsrki: @steal4life ?

Demos/Blazorise.Demo/Pages/Tests/SnackbarsPage.razor
stsrki: Tried it. Makes it complicated without reason.

steal4life:
Problably accidentaly, I did not notice it before.

David-Moreira (approved):

PR #4902, opened 2023-07-30
Grid utilities

stsrki:


Source/Blazorise.Bulma/BulmaClassProvider.cs
```diff
@@ -793,10 +793,20 @@ public override string RowColumns( RowColumnsSize rowColumnsSize, RowColumnsDefi
...
+    public override string Column( bool grid, bool hasSizes ) => hasSizes ? null : "column";
 
-    public override string Column( ColumnWidth columnWidth, Breakpoint breakpoint, bool offset )
+    public override string Column( bool grid, ColumnWidth columnWidth, Breakpoint breakpoint, bool offset )
     {
+        if ( grid )
+        {
+            if ( breakpoint != Blazorise.Breakpoint.None )
+            {
+                return $"is-grid-column-{ToColumnWidth( columnWidth )}-{ToBreakpoint( breakpoint )}";
```
David-Moreira, line 804: What's the difference between **is** and **are**?

Source/Blazorise/Base/BaseColumnComponent.cs
```diff
@@ -37,6 +37,21 @@ protected override void BuildClasses( ClassBuilder builder )
...
+    /// <summary>
+    /// Cascaded <see cref="Row"/> component in which this <see cref="Column"/> is placed.
+    /// </summary>
+    [CascadingParameter] public Row ParentRow { get; set; }
+
+    /// <summary>
+    /// Cascaded <see cref="Grid"/> component in which this <see cref="Column"/> is placed.
+    /// </summary>
+    [CascadingParameter] public Grid ParentGrid { get; set; }
+
```
David-Moreira, line 54: Could there be a collision? As far as I know cascading parameter cascades all the way.

David-Moreira:
On bulma last row does not seem consistent:
image

David-Moreira:
2. Is there a reason why the spacing between rows is bigger then the regular grid? I would expect both examples to be very similar/consistent with each other, despite the api being different?
image

David-Moreira:
3. Same thing as 2 but on the horizontal spacing between the columns
image

David-Moreira:
4. How is this different from regular row/column? Does it not already support that?
image

Source/Blazorise.Bulma/BulmaClassProvider.cs
stsrki: That's usually the Bulma naming convention. `is-` is meant for the single element, and `are-` is meant for multiple elements, usually child elements.

Source/Blazorise/Base/BaseColumnComponent.cs
stsrki: There could be, yes. But we can fix it later.

stsrki:
1 - 3 The spacing inconsistency comes from the default Margins on the Alert component. I could remove those for the examples or use a simple div as an alternative.

4. Yes, it supports. This new Grid also supports, but they are different things, Row/Column is one dimensional and uses flex. Grod/Column is two-dimensional and works with CSS grid.

David-Moreira:
> 1 - 3 The spacing inconsistency comes from the default Margins on the Alert component. I could remove those for the examples or use a simple div as an alternative.
> 
> 4. Yes, it supports. This new Grid also supports, but they are different things, Row/Column is one dimensional and uses flex. Grod/Column is two-dimensional and works with CSS grid.

3- I would think that keeping the whole page consistent would be best if the fault is ours for using a different component that generates a different spacing. And not the 2 features having different spacings.

David-Moreira:
The spacing still doesn't quite look the same?
image

Also why do you have to explicitly take off margin in alerts when using Grid, but standard Row/Col, you don't?

Other than that, lgtm.

stsrki:
> Also why do you have to explicitly take off margin in alerts when using Grid, but standard Row/Col, you don't?

Because Grid has a `gap` style. And when combined with the margins they have larger spacing.

David-Moreira (approved):

PR #4905, opened 2023-07-31
Fix Docs ITextLocalizerService changing the culture application wide

David-Moreira:
Fixes the recurring issue where our docs application would keep the language state application wide by changing our example to not set the default thread culture to the selected language/culture.

Added a note about this in the docs.

image


stsrki (approved):
Looking good.

PR #4912, opened 2023-08-06
TextSize utility

stsrki:
New utility with a predefined set of text sizes. This is going to help us with building standardized themes and blocks.


Source/Blazorise/Providers/ClassProvider.cs
```diff
@@ -1412,6 +1414,25 @@ public virtual string ToTextOverflow( TextOverflow textOverflow )
...
+            Blazorise.TextSize.Small => "sm",
+            Blazorise.TextSize.Medium => "md",
+            Blazorise.TextSize.Large => "lg",
+            Blazorise.TextSize.ExtraLarge => "xl",
+            Blazorise.TextSize.Heading1 => "1",
+            Blazorise.TextSize.Heading2 => "2",
+            Blazorise.TextSize.Heading3 => "3",
+            Blazorise.TextSize.Heading4 => "4",
+            Blazorise.TextSize.Heading5 => "5",
+            Blazorise.TextSize.Heading6 => "6",
```
David-Moreira, line 1431: is this correct? Just double checking

Source/Blazorise/Base/BaseComponent.cs
```diff
@@ -711,6 +716,21 @@ public TextOverflow TextOverflow
...
+    /// <summary>
+    /// Determines the font size of an element.
+    /// </summary>
+    [Parameter]
+    public TextSize TextSize
+    {
+        get => textSize;
+        set
+        {
+            textSize = value;
```
David-Moreira, line 728: I think i've seen you use the approach of checking whether the TextSize is already the value in most places? any reason why we're not doing that here?

Source/Blazorise.Tailwind/TailwindClassProvider.cs
```diff
@@ -1646,6 +1646,25 @@ public override string TextOverflow( TextOverflow textOverflow )
...
+    //        Blazorise.TextSize.Large => "",
+    //        Blazorise.TextSize.ExtraLarge => "",
+    //        Blazorise.TextSize.Heading1 => "",
+    //        Blazorise.TextSize.Heading2 => "",
+    //        Blazorise.TextSize.Heading3 => "",
+    //        Blazorise.TextSize.Heading4 => "",
+    //        Blazorise.TextSize.Heading5 => "",
+    //        Blazorise.TextSize.Heading6 => "",
+    //        _ => null,
+    //    };
```
David-Moreira, line 1665: commented code? Should it stay?

David-Moreira (changes requested):
The TextSize example is at the bottom. Should it be together with the other Text example?
Also should the examples just straight be alphabetically ordered?

Source/Blazorise/Providers/ClassProvider.cs
stsrki: It is correct.

Source/Blazorise/Base/BaseComponent.cs
stsrki: We need to revisit that. In the past with older Blazor, there were problems in some instances, and checking for the value change would not work. The component would not refresh when needed. I don't remember the reason, but I know that I had to be careful where to put the extra condition.

stsrki: Closing for now. I will open a new ticket for these optimizations.

David-Moreira: It would be good to move it to the OnParametersAsync, and try to develop a more concise api or helpers for these kind of stuff. 
I.E `DirtyIfParameterChanged(x=> x.MyParameter); `

David-Moreira (approved):
LGTM. Cna you give your input on this and just go ahead and merge if you feel like its fine,

> The TextSize example is at the bottom. Should it be together with the other Text example?
Also should the examples just straight be alphabetically ordered?

stsrki:
> LGTM. Cna you give your input on this and just go ahead and merge if you feel like its fine,
> 
> > The TextSize example is at the bottom. Should it be together with the other Text example?
> > Also should the examples just straight be alphabetically ordered?

No, it is a different component. The new TextSize utility is meant to be used by almost all components.

Issue #4374, opened 2022-12-07
Add option to change DataGridNumericColumns default column filter input box to NumericEdit or NumericPicker


asser-dk:
For Numeric Columns the default column filter input field is of type "text". Would it be possible to add a simple option to change that to "number" instead or even add the option to turn it into a default NumericPicker?

I know you can do that with a `FilterTemplate` but given that this column type is specifically for numbers it feels like an obvious addition :)

David-Moreira:
@stsrki Could we do the same as in this pending PR, where we add a dedicated filter component for that type of column? 
In this case it would be the `NumericPicker` since that's also what we use internally for editting.

![image](https://github.com/Megabit/Blazorise/assets/22283161/5155404b-d3bd-49e4-9869-70e1a71b576d)

stsrki:
> @stsrki Could we do the same as in this pending PR, where we add a dedicated filter component for that type of column? In this case it would be the `NumericPicker` since that's also what we use internally for editting.
> 
> ![image](https://user-images.githubusercontent.com/22283161/250115795-5155404b-d3bd-49e4-9869-70e1a71b576d.png)

Do you mean to have a new type of column? Like `DataGridNumericPickerColumn`? Or add a new parameter?

David-Moreira:
> > @stsrki Could we do the same as in this pending PR, where we add a dedicated filter component for that type of column? In this case it would be the `NumericPicker` since that's also what we use internally for editting.
> > ![image](https://user-images.githubusercontent.com/22283161/250115795-5155404b-d3bd-49e4-9869-70e1a71b576d.png)
> 
> Do you mean to have a new type of column? Like `DataGridNumericPickerColumn`? Or add a new parameter?

Well in this case I refactored the new celledit for select such that it was a more agnostic and reusable component, and reused for filter.
But bottom line is to make it so that if a numeric column shows up we use a numeric Picker and not a textedit like we're doing. I don't see why you are saying that would be a new column since the exiting numeric column already uses numeric Picker

stsrki:
@David-Moreira can you look into this if you think there is enough time before 1.3? Or we will move it to the next milestone.

David-Moreira:
I can try over the next days,

PR #4920, opened 2023-08-09
Datagrid : Filter support for specialized numeric column

David-Moreira:


Source/Extensions/Blazorise.DataGrid/_DataGridCellNumericEdit.razor
```diff
@@ -0,0 +1,59 @@
...
+               Type typeDecimalNull when typeDecimalNull == typeof( Decimal? ) => (Decimal?)( CellValue ?? (Decimal?)default ),
+               Type typeDouble when typeDouble == typeof( Double ) => (Double)( CellValue ?? (Double)default ),
+               Type typeDoubleNull when typeDoubleNull == typeof( Double? ) => (Double?)( CellValue ?? (Double?)default ),
+               Type typeFloat when typeFloat == typeof( Single ) => (Single)( CellValue ?? (Single)default ),
+               Type typeFloatNull when typeFloatNull == typeof( Single? ) => (Single?)( CellValue ?? (Single?)default ),
+               Type typeInt when typeInt == typeof( Int32 ) => (Int32)( CellValue ?? (Int32)default ),
+               Type typeIntNull when typeIntNull == typeof( Int32? ) => (Int32?)( CellValue ?? (Int32?)default ),
+               Type typeLong when typeLong == typeof( Int64 ) => (Int64)( CellValue ?? (Int64)default ),
+               Type typeLongNull when typeLongNull == typeof( Int64? ) => (Int64?)( CellValue ?? (Int64?)default ),
+               _ => throw new InvalidOperationException( $"Unsupported type {valueType}" )
```
stsrki, line 32: Do we want to throw an exception? It's not how we usually do it.

stsrki (commented):
Seems good to me. Just please remove some extra empty lines.

Source/Extensions/Blazorise.DataGrid/_DataGridCellNumericEdit.razor
David-Moreira: Kinda copied from existing implementations. I mean. It's one of those things, were it's not really expected to hit the exception. If it does, something is very wrong, and it's probably some change in development so it might be better to throw. 
image
image

What do you think?

David-Moreira:
> _DataGridCellNumericEdit.razor

Does your code cleanup automatically remove the empty lines? Or are you using something like code maid?

stsrki:
> > _DataGridCellNumericEdit.razor
> 
> Does your code cleanup automatically remove the empty lines? Or are you using something like code maid?

I do it manually.

Source/Extensions/Blazorise.DataGrid/_DataGridCellNumericEdit.razor
stsrki: I guess it can stay.

David-Moreira:
> > > _DataGridCellNumericEdit.razor
> > 
> > 
> > Does your code cleanup automatically remove the empty lines? Or are you using something like code maid?
> 
> I do it manually. Like a cave man.

Jesus christ, I can't even be bothered. If I may recommend CodeMaid then... I haven't used it for a while, but it has an option to "Clean up on save", and I believe you might be able to disable every other formating option you don't like.

David-Moreira:
Here's a commit "Auto format : Codemaid" where I've used it.

stsrki (commented):
The feature works. But in the original ticket #4374 it says that we need to add support for NumericEdit or NumericPicker. Should we have an enum API on `DataGridNumericColumn` that defines it?

stsrki:
I've tried CodeMaid in the past but it doesn't respect my formating. And it doesn't have as many options as I would like it to have.

stsrki:
> Here's a commit "Auto format : Codemaid" where I've used it on a signle file. `DataGrid.razor.cs`. See if you find it acceptable, and I'll start using it.

Nop. Please revert it 😅

David-Moreira:
> t my formating

That's not what I understand, the issue was that the filter input was not a number input and the user was just asking to make it a number input whether it was one or the other.
So keeping it consistent with what we already render for Edit input which is `NumericPicker` seems ideal.

We can additionally have the option, but I argue that might make it uneedlessly complex, specially if you use the same column and the parameters might not quite match between both. 

To me that would be best as a new column. But again I'm not sure we should do it.

David-Moreira:
> > Here's a commit "Auto format : Codemaid" where I've used it on a signle file. `DataGrid.razor.cs`. See if you find it acceptable, and I'll start using it.
> 
> Nop. Please revert it 😅

LOL

David-Moreira:
> > Here's a commit "Auto format : Codemaid" where I've used it on a signle file. `DataGrid.razor.cs`. See if you find it acceptable, and I'll start using it.
> 
> Nop. Please revert it 😅

Can you tell me what you didn't like about it? It's literally a 100x times superior to have something automated then manually doing it... I'm sure you miss alot of stuff, even though you're quite picky. :)

stsrki:
> > t my formating
> 
> That's not what I understand, the issue was that the filter input was not a number input and the user was just asking to make it a number input whether it was one or the other. So keeping it consistent with what we already render for Edit input which is `NumericPicker` seems ideal.
> 
> We can additionally have the option, but I argue that might make it uneedlessly complex, specially if you use the same column and the parameters might not quite match between both.
> 
> To me that would be best as a new column. But again I'm not sure we should do it.

Hm, in that case better to leave it as it is. We don't want to loos too much time before release.

stsrki:
Well, codemaid adds some unnecessary lines from what I can see in this one file.

image

image

stsrki (approved):

David-Moreira:
> Well, codemaid adds some unnecessary lines from what I can see in this one file.
> 
> image
> 
> image

I believe that's configurable. I'm not saying it comes as you'd like out of the box. haha. The point is that you configure it to your liking and then you start to have the peace of mind that everytime it saves a file, it keeps the spacing consistent.

image
image

David-Moreira:
Hell, I believe if you dig into it, it even has options to add the region gibberish by default you really like with the Properties Members, etc.

PR #4924, opened 2023-08-10
Autocomplete: Fix dropdown menu being anchored incorrectly

David-Moreira:
Well I had to introduce an anchorid to tell it to anchor off the input. What do you think?

### DIFF
--- a/Source/Blazorise/Components/Dropdown/Dropdown.razor.cs
+++ b/Source/Blazorise/Components/Dropdown/Dropdown.razor.cs
@@ -67,7 +67,7 @@ public partial class Dropdown : BaseComponent, IAsyncDisposable
         if ( firstRender )
         {
             JSModule.Initialize( ElementRef, ElementId,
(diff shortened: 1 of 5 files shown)


Source/Blazorise/Components/Dropdown/Dropdown.razor.cs
```diff
@@ -498,5 +498,11 @@ public Direction Direction
     /// </summary>
     [Parameter] public DropdownPositionStrategy PositionStrategy { get; set; } = DropdownPositionStrategy.Fixed;
 
+    /// <summary>
+    /// Provides a custom anchor element id for the dropdown menu.
+    /// This is useful when you want the dropdown menu to be anchored from a different element than the toggle.
+    /// </summary>
+    [Parameter] public string AnchorId { get; set; }
```
stsrki, line 505: The name is too generic. Something like `DropdownMenuAnchorId`, or `DropdownMenuTargetId` would be better. Or do you have a better name?

PR #4943, opened 2023-08-21
Prevent Server Thread Culture from being mutated in docs example

David-Moreira:
It seems like there was another place where users could mutate the server thread culture, no wonder it was in polish.
image


stsrki (approved):
LGTM.

stsrki:
PS. For future note. If we are fixing or improving documentation, we will do it on the current 1.3. Only new features are going into the master. That way we don't need to wait for publishing fixes on docs.

PR #4959, opened 2023-08-30
Docs : MarkdownShortcutsExample | Fix shortcut example

David-Moreira:


stsrki:
Any reason why targeting 1.2 branch and not 1.3?

David-Moreira:
Aren't we still releasing and updating v1.2 branch for docs?

David-Moreira:
Is that not the point of 
image
? Do you not compile the docs from 1.2 branch?

stsrki:
No. It is frozen now. It is only accessible in case that someone hasn't updated to 1.3 yet.

David-Moreira:
> No. It is frozen now. It is only accessible in case that someone hasn't updated to 1.3 yet.

I thought it'd make sense to fix any simple docs issue like this one in 1.2 and propagate through the code base. if it's frozen, let's retarget 1.3 then.

David-Moreira:
Ah jesus christ. Retarget made a mess. I'm just going to remake the PR. Closing...

PR #4963, opened 2023-08-31
Video: play Youtube videos

### LINKED ISSUE #4872: Play YouTube Video inside Blazorise Video component
### Discussed in https://github.com/Megabit/Blazorise/discussions/4871

<sup>Originally posted by **mdmontesinos** July 10, 2023</sup>
Blazorise Video component gets a CORB error when the source is set to a YouTube video URL.

```
Cross-Origin Read Blocking (CORB) blocked cross-origin response https://www.youtube.com/video_id with MIME type text/html. See https://www.chromestatus.com/feature/5629709824032768 for more details.
```

I know that the YouTube API requires an iframe to work properly, but I was wondering if there's any way to play it inside the video component, mainly to take advantage of the nice UI and controls it provides.

Also, the docs for the Video Component, in the Ratio attribute, reference YouTube, so it lead me to think it works somehow. 
_"As dimensions are not available from YouTube via SDK, 16:9 is forced as a sensible default."_  </div>

steal4life:
Closes #4872


steal4life:
So the idea is that the user just enters the id from youtube, the plan is to add an explanation to users where they can find that id and use it to add the desired video

steal4life:
so a youtube video can't be rendered using plyr without the /embed/ part of the link and what comes after the YoutubeVideoID, based on other cases I've found while working on this, I think it's the most optimal, because in this way user can play any video they want from YouTube

steal4life:
therefore, the user can reduce the width and height of the container in which YouTube is located, as for other parameters that are characteristic of the video, YouTube already covers it, I don't know if we should adjust that parameters when user is using youtube video and how can we do that?

Source/Extensions/Blazorise.Video/Video.razor.cs
```diff
@@ -15,6 +19,11 @@ public partial class Video : BaseComponent, IAsyncDisposable
 {
     #region Methods
 
+    ///<summary>
+    /// ID of youtube video link.
+    /// </summary>
+    public string YoutubeVideoID;
```
David-Moreira, line 25: Should this have the setter as public?  Seems like we set this ourselves and should not allow anyone to override it?

Source/Extensions/Blazorise.Video/Video.razor.cs
```diff
@@ -41,6 +51,7 @@ public override async Task SetParametersAsync( ParameterView parameters )
                     ProtectionHttpRequestHeaders = new { Changed = protectionHttpRequestHeadersChanged, Value = paramProtectionHttpRequestHeaders },
                     CurrentTime = new { Changed = currentTimeChanged, Value = paramCurrentTime },
                     Volume = new { Changed = volumeChanged, Value = paramVolume },
+                    YoutubeVideoSource = new { Changed = youtubeVideoSourceChanged, Value = paramYoutubeVideoSource }
```
David-Moreira, line 54: @stsrki @steal4life 
So this `YoutubeVideoSource` is an "option" of the underlying plyr library?

What if we instead added a `SourceType` enum instead?

`SourceType.Default`
`SourceType.Youtube`

and the user still defines the source in the same Source parameter, and we act accordingly to the defined SourceType internally?

Source/Extensions/Blazorise.Video/Video.razor.cs
David-Moreira: And should we really expose something specific to `Youtube` like this? Which we don't seem to require to do anything with it, and are adding specific regex logic that could perhaps change with time? i.e youtube changes the way they do ids or something?

stsrki:
We're in the process of moving to a new Video library that is supporting YouTube. Closing this.

PR #4969, opened 2023-09-04
Create bug-report-advanced

stsrki:


.github/ISSUE_TEMPLATE/bug-report-advanced
```diff
@@ -0,0 +1,74 @@
...
+body:
+  - type: markdown
+    attributes:
+      value: |
+        **Before submitting a bug report**
+
+        The issue list is reserved for bug reports and feature requests. If you have a usage question, you can:
+
+        - Read the [documentation](https://blazorise.com/docs)
+        - Ask questions on [Discord](https://discord.io/blazorise)
```
David-Moreira, line 14: Worth adding the GH Discussion tab for people who do not use Discord, but still want to write a custom question?

.github/ISSUE_TEMPLATE/bug-report-advanced
```diff
@@ -0,0 +1,74 @@
...
+      required: true
+  - type: input
+    id: reproduction-link
+    attributes:
+      label: Link to minimal reproduction
+      description: |
+        The easiest way to provide a reproduction is to provide a link to a public GitHub repository or a tool like [Telerik Repl](https://blazorrepl.telerik.com/).
+
+        The reproduction should be **minimal**. This means it should contain only the bare minimum amount of code needed
+        to show the bug.
```
David-Moreira, line 32: Maybe, Add some text about making it so the code provided is as complete as possible to get it running, i.e contain class definitions, etc...

.github/ISSUE_TEMPLATE/bug-report-advanced
```diff
@@ -0,0 +1,74 @@
...
+      placeholder: Reproduction Link
+    validations:
+      required: true
+  - type: textarea
+    id: steps-to-reproduce
+    attributes:
+      label: Steps to reproduce
+      description: |
+        What do we need to do after opening your repro to make the bug happen? Clear and concise reproduction instructions are important for us to be able to triage your issue in a timely manner. Note that you can use [Markdown](https://guides.github.com/features/mastering-markdown/) to format lists and code.
```
David-Moreira, line 41: Maybe link to the lists and code anchors, instead of full documention?
i.e https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#lists

.github/ISSUE_TEMPLATE/bug-report-advanced
```diff
@@ -0,0 +1,74 @@
...
+
+        - Read the [documentation](https://blazorise.com/docs)
+        - Ask questions on [Discord](https://discord.io/blazorise)
+
+        Also, try to search for your issue/feature - another user may have already requested something similar; Thanks!
+
+  - type: input
+    id: version
+    attributes:
+      label: Blazorise Version
```
David-Moreira, line 21: The Blazorise Provider might be good to add. Preferably if it's a dropdown of some sorts with our current providers.

.github/ISSUE_TEMPLATE/bug-report-advanced
```diff
@@ -0,0 +1,74 @@
+name: "\U0001F41E Bug report"
```
David-Moreira, line 1: Maybe add a mention somewhere that bug reports that do not follow the template and mandatory fields will be closed immediately. And if you want to explain why, a brief description along the lines of : this is to keep the developers focused on actually trobuleshooting & fixing bugs, then having to figure out with the user additional basic details everytime, as it currently happens, what blazorise version are you using, browser, code, etc...

.github/ISSUE_TEMPLATE/bug-report-advanced
```diff
@@ -0,0 +1,74 @@
...
+description: Create a report to help us improve
+title: "[Bug]: "
+labels: ["Triage"]
```
David-Moreira, line 4: Just an idea, not sure if worth it, a question like:
How urgent do you find your bug report to be?

MINOR
MODERATE
BREAKING
something along those lines? Could that immediately put some labels in the issues? That why we can quickly glance if some Triage issue is also reported as very urgent.

David-Moreira (changes requested):

.github/ISSUE_TEMPLATE/bug-report-advanced
stsrki: It is better to have an input so that users can write the exact version, eg 1.2.6.

.github/ISSUE_TEMPLATE/bug-report-advanced
stsrki: The form is already somewhat scary as it is to newcomers so don't want to scare them. Let's keep it like this for now.

.github/ISSUE_TEMPLATE/bug-report-advanced
stsrki: We already have that kind of labels. In any case, we are the ones to assign them, not the users who report an issue.

.github/ISSUE_TEMPLATE/bug-report-advanced
David-Moreira: Who said anything about changing the input?

> The **Blazorise Provider** might be good to add. Preferably if it's a **dropdown** of some sorts with our current **providers**.

I.E

Bootstrap
Bootstrap5
Material

etc...

.github/ISSUE_TEMPLATE/bug-report-advanced
David-Moreira: Right, it would still have the triage label and we would fix it, if we though the user incorrectly assigned the urgency, but it would give us an initial feeling for the priority of the issue with a quick glance.

.github/ISSUE_TEMPLATE/bug-report-advanced
stsrki: Ahhhh. Since you were linking "Blazorise Version" I understood you want to have an option for 1.1, 1.2, 1.3, etc. Yeah, we can have an additional field for providers.

.github/ISSUE_TEMPLATE/bug-report-advanced
stsrki: I've added more labels. We'll see how that works.

David-Moreira: I don't understand this. Doesn't this mean that the open issue will now have all these labels? The point is that the user would somehow select an option that would tie into a label? If that is not possible, then I'd just have the Bug & Triage labels set.
This way the issue created will just inherit all these labels that don't make sense, they should be mutually exclusive. Also Maybe "**Critical**" is a better naming then "**Breaking**", I just typed whatever I remembered at the time. And I thought you mentioned we already had those anyway, so just re-use?
labels: ["Type: Bug \U0001F41E", "Triage", "Minor", "Moderate", "Breaking"]

David-Moreira (approved):
LGTM. Review this comment first please.
image

stsrki:
I have realized with a new template, people would be forced to create a repro project on GitHub. Sometimes to reproduce then only need to paste a little code sample. Maybe we can also allow that? So, maybe we don't need to force validation on Link to reproducible??

David-Moreira:
> I have realized with a new template, people would be forced to create a repro project on GitHub. Sometimes to reproduce then only need to paste a little code sample. Maybe we can also allow that? So, maybe we don't need to force validation on Link to reproducible??

Yea. That's fine. As long as it comes as complete as possible, sometimes they come without a few class definitions and it's a bit of a bother to create the definitions.
Maybe you can have a dedicated box for the code? Could you make it mandatory at least link or code box?

stsrki:
i can make the field as markdown instead of simple input. That way they can paste a URL, or paste a code.

David-Moreira:
> i can make the field as markdown instead of simple input. That way they can paste a URL, or paste a code.

Fine by me. We can always further improve once its live.

PR #4983, opened 2023-09-06
DataGrid : Propagate Column FilterMethod to ReadData's DataGridColumnInfo

David-Moreira:


Source/Extensions/Blazorise.DataGrid/Models/DataGridColumnInfo.cs
```diff
@@ -57,4 +59,9 @@ public DataGridColumnInfo( string field, object searchValue, SortDirection sortD
     /// Gets the column type.
     /// </summary>
     public DataGridColumnType ColumnType { get; }
+
+    /// <summary>
+    /// Gets the column filter method.
+    /// </summary>
+    public DataGridFilterMethod? FilterMethod { get; }
```
stsrki, line 66: Why the new field is nullable?

David-Moreira: If `FilterMethod` is null. It Means that the filter has not been set for this column.

Parity with how filtering is done when it's static data:
image
image

And to what's defined in the column:
image

stsrki: Cotcha!

stsrki (approved):

PR #5003, opened 2023-09-14
Snackbar: fixed position on mobile

### LINKED ISSUE #4994: [Bug]: Snackbar: fixed position on mobile
### Blazorise Version

1.3.1

Bootstrap5

https://blazorise.com/docs/extensions/snackbar

Both my app and the Blazorise Demo of the SnackbarStack have the issue so I will just point back to the demo.

### Steps to reproduce

Set browser device mode to mobile device like iPhone 12 pro.

Go to the "Stacked snackbars" section of the docs toward bottom of page.

Press the Primary button.

### What is expected?

A snackbar/toast should appear completely on screen at the Bottom End of the screen saying:

"Current time is: " + DateTime.Now

It appears at Bottom but is cut off and placed at Bottom Start or there is just not enough room and it appears that way.

I've tried other positions and on mobile it doesn't seem to matter. The snackbar is cut in half with only a portion on screen aligned start/left.

Chrome, Microsoft Edge, Safari, Firefox

### Any additional comments?

_No response_

steal4life:
closes #4994

Reason behind:

In the previous version the snackbar stack didn't have keyframes in its scss that now allow opening and closing animations... among other things they caused only half of the snackbar to be visible on mobile, so I made a few changes with the animations to avoid that behavior


David-Moreira:
@steal4life 
Is this ready for review?

steal4life:
Yes it is @David-Moreira

Source/Extensions/Blazorise.Snackbar/Styles/blazorise.snackbar.scss
```diff
@@ -304,7 +300,7 @@ $snackbar-padding-y-multi: 1.25rem !default;
 .snackbar-stack-top-start .snackbar-hide,
 .snackbar-stack-top-end .snackbar-hide {
     transition: all var(--stack-transition-duration, 1000);
-    animation: hideTop var(--stack-animation-duration, 500) forwards;
+    animation: hideTop var(--stack-animation-duration, 500) backwards;
```
David-Moreira, line 303: Why is the animation changed to backwards? WAs it wrong before? Or will this now produce a different behaviour that users might not be expecting on their apps?

David-Moreira (approved):

Source/Extensions/Blazorise.Snackbar/Styles/blazorise.snackbar.scss
stsrki: Same question from me. Even if I revert it to `forwards` the animation for stacked snackbars still doent's work. When snackbar hides it should hide from bottom to top when it is placed on top of the screen. At the moment it closes from top to bottom, which seems strange.

steal4life:
@David-Moreira this solution works, but in case the snackbar is 50 percent of the screen height, then for some reason it will go down again, as long as it's smaller it should work

Source/Extensions/Blazorise.Snackbar/Styles/blazorise.snackbar.scss
```diff
@@ -297,93 +293,67 @@ $snackbar-padding-y-multi: 1.25rem !default;
...
     0% {
-        max-height: 0;
-        padding: 0;
-        transform: translate(-50%) translateY(-50px);
         opacity: 0;
     }
 
     100% {
-        max-height: 150px;
-        padding: $spacer-sm $spacer;
```
David-Moreira, line 337: Do we need to remove these?

David-Moreira (changes requested):
It does not seem to sit at bottom end, should it not?

image

Source/Extensions/Blazorise.Snackbar/Styles/blazorise.snackbar.scss
steal4life: I already removed that in last commit, do you see it?

stsrki (approved):

PR #5014, opened 2023-09-21
DropdownList: Add DropdownToggleSize parameter

### LINKED ISSUE #5010: Size Attribute for DropdownList Button
There is no size attribute for a DropdownList button. This would be nice to have since the dropdown component button has a size attribute.

I was told to post this here from the Blazorise Discord server.

steal4life:
closes #5010 

Reason behind:

I made it so that the user defines the size of the DropdonwToggle via the Size attribute, blazorise offers various options when that attribute is called and I think this might be the best solution, if anyone has a better idea feel free to share it with me.


Source/Extensions/Blazorise.Components/DropdownList.razor.cs
```diff
@@ -88,6 +88,11 @@ protected bool IsSelected( TValue value )
     /// </summary>
     [Parameter] public Color Color { get; set; }
 
+    /// <summary>
+    /// Defines the size of toggle button.
+    /// </summary>
+    [Parameter] public Size Size { get; set; }
```
David-Moreira, line 94: Would it be better to name it `DropdownToggleSize` or something along those lines? 

One could assume this to be the `Size` or the List or the item incorrectly?

Other than that I guess it's fine to make this available if it's a necessity. 

@stsrki Ideally in the future we can figure a way to pass a bag of parameters to inner components. (I think we have an issue open somewhere for that.)

David-Moreira (changes requested):
LGTM

Source/Extensions/Blazorise.Components/DropdownList.razor.cs
stsrki: I agree, that DropdownToggleSize seems more appropriate in this context.

> @stsrki Ideally in the future we can figure a way to pass a bag of parameters to inner components. (I think we have an issue open somewhere for that.)

We do have an issue. The main problem is that whatever route we go it feels like the inner API of the internal component structure will leak out. And in some cases, it is unique to certain providers.

Demos/Blazorise.Demo/Pages/Tests/DropdownListPage.razor
```diff
@@ -17,7 +17,8 @@
                                       ValueField="@((item) => item.Iso)"
                         @bind-SelectedValue="@selectedDropValue"
                                       Color="Color.Primary"
-                                      MaxMenuHeight="200px">
+                                      MaxMenuHeight="200px"
+                                      Size="Size.Large">
```
stsrki, line 21: You forgot to rename this parameter.

stsrki (approved):

PR #5020, opened 2023-09-25
RichTextEdit: Safe loading and invocation

njannink:
We see from time to time some RTE errors in or logging and I traced it back to JS disconnects. I changed the JS calls to safe invocations and also made the initialize routine safe


stsrki (approved):
LGTM. Thanks @njannink

PR #5048, opened 2023-10-09
DataGrid: Cell Edit feature (Rapid editing)

David-Moreira:
I believe we can do two separate features

- **Cell Edit (Rapid Editing)**
I believe this will be the most important to implement, able to click any cell and edit straight away with "instant" save. And a few shortcuts, Tab, Shift-Tab, Save on Enter, etc...

- **Batch/Bulk Edit** 
To me, this is the ability to accumulate several changes and only save upon strict Save Changes click. I believe this should be built upon the **Cell Edit** feature.
As such, this will be implemented on a best effort basis. Meaning, it will only be implemented, if we have time to do it as I believe it would be more complex. (Might do it on a seperate PR)

So the current PR is for **Cell Edit (Rapid Editing)**


David-Moreira:
@stsrki Would you take a look at this initial prototype and let me know your thoughts?

Source/Blazorise/Modules/JSUtilitiesModule.cs
```diff
@@ -53,7 +53,7 @@ public virtual ValueTask<bool> ParentHasClass( ElementReference elementRef, stri
 
     /// <inheritdoc/>
     public virtual ValueTask Focus( ElementReference elementRef, string elementId, bool scrollToElement )
-        => InvokeSafeVoidAsync( "focus", elementRef, elementId, scrollToElement );
+        => InvokeSafeVoidAsync( "focus", elementRef.Context is null ? null : elementRef, elementId, scrollToElement );
```
stsrki, line 56: What is the purpose of this null check? Can you explain?

Source/Extensions/Blazorise.DataGrid/_DataGridRow.razor
```diff
@@ -30,38 +30,38 @@
 
         @ParentDataGrid.groupIndentationFragment
 
-        @foreach ( var column in Columns )
+        @foreach (var column in Columns)
```
stsrki, line 33: Don't forget to reformat these lines.

Source/Extensions/Blazorise.DataGrid/Enums/DataGridEditMode.cs
```diff
@@ -19,4 +19,10 @@ public enum DataGridEditMode
     /// Values will be edited in the modal dialog.
     /// </summary>
     Popup,
+
+    /// <summary>
+    /// Any cell value can be edited directly allowing for rapid editing of data.
+    /// <para>It is of note that certain features may be disabled. e.g edit command column, etc...</para>
+    /// </summary>
+    Cell
```
stsrki, line 27: I guess `Cell` is an OK name in this case. But I'm already thinking about what will we name a batched cell edit mode?  🤔

Maybe BatchedCell?

stsrki (commented):
Overall it looks OK. There is not much code at the moment that seems off.

Source/Blazorise/Modules/JSUtilitiesModule.cs
David-Moreira: Because `ElementRef` is a struct. You need to explicitly send a null for `getRequiredElement` to use `elementId` instead.

image

Source/Extensions/Blazorise.DataGrid/Enums/DataGridEditMode.cs
David-Moreira: Despite batch mode being most used with cell type based edit (i.e excel like edit).

I was thinking that it could be a flag named `Batch` or `BatchEdit` that works with all edit modes, that should give some flexibility.

Instead of saving the data right away it will just keep in memory until **save or cancel** is hit. Whether I edit using modal, form, etc... the cells that were edited will have a visual indicator so the user can keep track of what's edited.

Source/Extensions/Blazorise.DataGrid/_DataGridRow.razor
David-Moreira: :(

Source/Extensions/Blazorise.DataGrid/Enums/DataGridEditMode.cs
stsrki: That makes sense. In fact I already see the benefit od doing it that way.

David-Moreira:
> Overall it looks OK. There is not much code at the moment that seems off.

Ok, but other then code review, did you try the feature? I wanted to know if it felt right for you.

Source/Extensions/Blazorise.DataGrid/Enums/DataGridEditMode.cs
David-Moreira: Right, yea, I think the flexibility will be a really good aspect of it.

I'm not sure if we should track edited values across page navigation, but I think we should. What do you think?

stsrki: > I'm not sure if we should track edited values across page navigation, but I think we should. What do you think?

We should track them across pages.

David-Moreira:
> > > Overall it looks OK. There is not much code at the moment that seems off.
> > 
> > 
> > Ok, but other then code review, did you try the feature? I wanted to know if it felt right for you.
> 
> I have tried it now. Overall it works but I have few suggestions.
> 
> 1. When I focus on a cell and edit field is shown the grid tends to resize the columns and it doesn't look really good to the user. We need to try to keep the table from not "jumping" around.
> 2. The edit value is only submitted when I hit ENTER. I think it makes sense to also submit when I click with a mouse outside of the cell, ie. lose focus.

Well, we have work cutout for us, not sure if 1 will be easy as the point is that the columns get resized to fit whatever is on them, except if the cell width is defined I gues.s... thanks for the feedback.

David-Moreira:
For number 1.
It does seem regular behavior. You will also find that behavior if you use regular Inline Edit and the Grid Resizes to accomodate the components.

I think if the user cares about it, he can adjust the buttons, or remove them all together. As he might want to edit without the Save / Cancel buttons.

David-Moreira:
@stsrki Validation is really weird... What do you think?

The user kinda can leave the field if validation fails.. Do you think we should lock him in the field somehow?
image

stsrki:
Hm, that's a though one. Maybe for cell editing, we can show validation on a tooltip inside of a cell. Also the user should still be able to leave field, but the submit will not happen until the validation passes.

image

David-Moreira:
> Hm, that's a though one. Maybe for cell editing, we can show validation on a tooltip inside of a cell. Also the user should still be able to leave field, but the submit will not happen until the validation passes.
> 
> image

Hey @stsrki 
there's apparently some bug with validation. Or maybe I'm doing something wrong. It's driving me nuts.

Since you're the expert in the validation, can you help me out?

Try editing the email here: https://localhost:5001/tests/datagrid/editing/rapid-editing
For example, you can manage to save it with an incorrect value ABC@

Then when you try to edit again, to ABC@gmail.com, it will always fail, and basically from debugging the validation.FieldIdentifier has a stale model that's used for validation.

**VALIDATION FAIL**
image

**On The Input Changed we can already see an incoherence?**
image

**VALIDATION RUNNING with correct internal value
(I have to admit, i don't understand why we already have the value and use additional abstractions to get it again?)**
image

Finally the **DataAnnotationValidationHandler** is running and it tries to grab the value from the field identifier, again, maybe we already had the value at hand and we're gonna recalculate it again? But there's probably a good reason.
image

David-Moreira:
@stsrki

stsrki:
Could it be that you're validating the wrong model? I think we have a different model for New and Edit and Validation editing modes. I will also try to debug it in the menatime.

David-Moreira:
> Could it be that you're validating the wrong model? I think we have a different model for New and Edit and Validation editing modes. I will also try to debug it in the menatime.

Please do. The idea with the new feature is that I'm trying to do everything as close to what already exists. I doubt I changed the model, it does not look like it. 
From my debug, the models get updated but the field keeps a stale Model. But since you're more accquainted with `Validation` you'll probably figure it out.

David-Moreira:
@stsrki bump

stsrki:
I tried to figure something out a few days ago but I didn't have much time. From what I could see, my suspect in ValidationItem. For some reason, it isn't updated properly, or it is reset once the validation fails.

David-Moreira:
> I tried to figure something out a few days ago but I didn't have much time. From what I could see, my suspect in ValidationItem. For some reason, it isn't updated properly, or it is reset once the validation fails.

I think I checked every item including ValidationItem and they had correct values. but the internal expression had a ref to something stale... Oh well.. I'll try to figure it out then, thanks!

David-Moreira:
I figured out the `Validation` problem. It was indeed a "stale" or incorrect ref for a `ValidationItem`, since Edit was being invoked twice in a row, but the component was being smart and being kept on memory and refreshing the Expression with the new `ValidationItem`.

Anyways long story short, it is fixed. I think the edit is okay like this validation wise. It keeps it more close to what & how it was already working.

The tabbing does not work properly on WASM, and it works partially well on Server, still need to figure that one out.
Probably will implement ESC Key press exists Edit? Do you think that makes sense? I'm thinking it could make sense as a standalone feature almost, ESC Key exiting edit, no matter what edit mode is on? @stsrki

stsrki:
I knew it haha. Great you could finally figure it out. Nice job.



> Probably will implement ESC Key press exits **Edit**?

Yeah, that makes perfect sense.

David-Moreira:
> I knew it haha. Great you could finally figure it out. Nice job.
> 
> > Probably will implement ESC Key press exits **Edit**?
> 
> Yeah, that makes perfect sense.

I believe we're getting there, can you take another look?

stsrki:
Tested and it working really nice. It's still "jumpy" but that is something that the user will have to figure for their usecases. Have a fixed size columns, hide command column, etc.

David-Moreira:
> Tested and it working really nice. It's still "jumpy" but that is something that the user will have to figure for their usecases. Have a fixed size columns, hide command column, etc.

Yep, In this case the "jumpyness"  you sometimes might see, is the re-rendering because the command column grows. We might make the example have it as a fixed width as an example.

I still found out that without the command column (and the submit button) pressing **Enter**, posts the entire page. Still have to figure out a workaround for that. Do you remember anything that can help with this specific issue?

I'll go ahead and try to fix this issue next and start working on docs, but to me this is mostly ready unless we find any issue.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -1866,6 +1887,9 @@ internal int GroupableColumnsCount
     internal bool IsFixedHeader
         => Virtualize || FixedHeader;
```
stsrki, line 1889: Add an XML comment explaining the property.

Source/Extensions/Blazorise.DataGrid/_DataGridRowEdit.razor.cs
```diff
@@ -64,6 +66,70 @@ internal protected async Task Save()
...
+                var currentIdx = OrderedColumnsForEditing.Index( x => x.IsEqual( column ) );
+
+                if ( args.ShiftKey )
+                {
+                    var previousColumn = OrderedColumnsForEditing.ElementAtOrDefault( currentIdx - 1 );
+                    if ( previousColumn is not null )
+                        await ParentDataGrid.HandleCellEdit( previousColumn, Item );
+                    else
+                    {
+                        var currentEditRowIdx = ParentDataGrid.Data.Index( x => x.IsEqual( Item ) );
```
stsrki, line 101: DO we need to check for null references here

- `ParentDataGrid.Data?.`
- ParentDataGrid.DisplayData?.
- etc

Source/Extensions/Blazorise.DataGrid/_DataGridRowEdit.razor.cs
```diff
@@ -64,6 +66,70 @@ internal protected async Task Save()
...
+                    return;
+
+
+                var currentIdx = OrderedColumnsForEditing.Index( x => x.IsEqual( column ) );
+
+                if ( args.ShiftKey )
+                {
+                    var previousColumn = OrderedColumnsForEditing.ElementAtOrDefault( currentIdx - 1 );
+                    if ( previousColumn is not null )
+                        await ParentDataGrid.HandleCellEdit( previousColumn, Item );
```
stsrki, line 98: Surround with braces.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
David-Moreira: Can you clarify where you put this?
I see an unrelated feature?

image

I am assuming I might have missed the new parameter? I'll take a look when I start writing the docs.

Source/Extensions/Blazorise.DataGrid/_DataGridRowEdit.razor.cs
David-Moreira: What the heck?
image

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
stsrki: No, just write the comment in the code. Not in the docs :)

David-Moreira: Yes, I understood, already done.

Source/Extensions/Blazorise.DataGrid/_DataGridRowCommand.razor
```diff
@@ -103,3 +104,11 @@ else if ( EditState == DataGridEditState.None && ParentDataGrid.IsCommandVisible
         </Div>
     </TableRowCell>
 }
+else if ( EditState == DataGridEditState.None && !ParentDataGrid.IsCommandVisible && ( ( ParentDataGrid.EditMode == DataGridEditMode.Inline || ParentDataGrid.EditMode == DataGridEditMode.Cell ) ) )
+{
+
+        <TableRowCell @onclick:stopPropagation="@(ParentDataGrid.CommandColumn?.PreventRowClick ?? false)" Class="@(Column.CellClass?.Invoke(Item))" Style="@BuildCellStyle()" TextAlignment="@Column.TextAlignment" VerticalAlignment="@Column.VerticalAlignment" Display="@Column.Display" Flex="@Column.Flex" Gap="@Column.Gap">
+            <Div Flex="@DefaultFlex" Gap="@DefaultGap">
```
stsrki, line 111: Why is there an empty Div here?

Source/Extensions/Blazorise.DataGrid/_DataGridCellSelectEdit.razor.cs
```diff
@@ -67,10 +73,36 @@ private void OnSelectedValueChanged( object value )
         }
     }
 
+    protected override async Task OnAfterRenderAsync( bool firstRender )
+    {
+        if ( firstRender )
+        {
+            if ( ParentDataGrid.IsCellEdit && Column.CellEditing )
+            {
+                await Task.Yield();
```
stsrki, line 82: Didn't an issue with calling `Focus` without `await Task.Yield();` already fixed?

Source/Extensions/Blazorise.DataGrid/DataGrid.razor
```diff
@@ -11,6 +11,13 @@
         @paginationFragment
     }
     <Form>
+
+        @if ( IsCellEdit && ( CommandColumn is null || !CommandColumn.SaveCommandAllowed ) )
+        {
+            //Disable the default form behaviour where pressing enter submits the form and posts the page which is not expected in typical SPA fashion.
+            <Button Type="ButtonType.Submit" Disabled Display="Display.None"></Button>
```
stsrki, line 18: I think you should add `PreventDefaultOnSubmit` on a Button here.

stsrki (commented):
Looks good so far,

Source/Extensions/Blazorise.DataGrid/DataGrid.razor
David-Moreira: Same behaviour, the button is not supposed to be triggered anyway.

Source/Extensions/Blazorise.DataGrid/_DataGridCellSelectEdit.razor.cs
David-Moreira: Sorry, I don't follow?

Due to the nature of single threaded WASM this was needed from what I can remember

Source/Extensions/Blazorise.DataGrid/_DataGridRowCommand.razor
David-Moreira: Use case where:
Inline or Cell mode, with **New Command** Disabled, but Save/Cancel commands are enabled. This means that an empty command column needs to be rendered in order for the columns to align properly. So basically an empty column.

David-Moreira: Actually when writing this... I figured this is even a problem in the current version with just **Inline** mode,

https://blazorise.com/docs/extensions/datagrid/features/editing
Change to **Inline** and click edit, and notice the column alignment.

Source/Extensions/Blazorise.DataGrid/_DataGridCellSelectEdit.razor.cs
stsrki: I remember we that that workaround in the past. 

Here is the diff

image

Source/Extensions/Blazorise.DataGrid/_DataGridRowCommand.razor
stsrki: Understand. In that case, place an `Span` instead. It is an `inline block` by default so it makes more sense in this context.

Source/Extensions/Blazorise.DataGrid/_DataGridCellSelectEdit.razor.cs
David-Moreira: Do you remember what led you to remove? Probably encountering same issue.
Anyway I can retest without it and I'll let you know. But I'm sure I'm not crazy and something was not working properly.

stsrki: It was fixed in .NET eventually, see https://github.com/dotnet/aspnetcore/pull/32017

David-Moreira: Just retried to remove it, does not work.

PR #5052, opened 2023-10-10
Fix employee data item mutation

David-Moreira:


Shared/Blazorise.Shared/Data/EmployeeData.cs
```diff
@@ -55,7 +55,24 @@ public EmployeeData( IMemoryCache memoryCache )
     };
 
     public async Task<List<Employee>> GetDataAsync()
-        => ( await cache.GetOrCreateAsync( employeesCacheKey, LoadData ) ).ToList(); //ToList is used so we make sure that we are not returning the same list reference avoiding an application wide "data corruption".
+        => ( await cache.GetOrCreateAsync( employeesCacheKey, LoadData ) )
+        .Select( x => new Employee()
```
stsrki, line 59: How about we add the new constructor overload that receivers `Employee other` that clones all the fields? The same as what we have here. That way we will not forget to update code in case we add another field.

David-Moreira: Yea good idea.

stsrki (approved):

PR #5059, opened 2023-10-12
Anchor: added parameter Unstyled to Link component

### LINKED ISSUE #5028: Add Unstyled utility to the Anchor component
Once enabled, the link should be unstyled to remove the default blue color. This is especially needed for `<a>` tags surrounding larger cards, or divs.

**API:**

`public bool Unstyled { get; set; }`

Enabled CSS should be:

```css
.css-class-name {
  text-decoration: none;
  color: inherit;
}
```

steal4life:
closes #5028 

while working on this i noticed that even though text-decoration is set to none the underline is still shown when we use bootstrap 4 and as I can see the reason is that _reboot.scss is called and gets its own link setting


Source/Blazorise.Bootstrap/Styles/_link.scss
```diff
@@ -0,0 +1,4 @@
+.link-unstyled {
+    text-decoration: none;
+    color: inherit !important;
```
stsrki, line 3: Prefer to have styles without `!important`. That way users can overwrite their own styles. Do the same on all other providers.

PS. you can have higher specificity with `a.link-unstyled`.

Source/Blazorise.Tailwind/TailwindClassProvider.cs
```diff
@@ -1797,10 +1797,23 @@ public override string ParagraphColor( TextColor textColor )
...
-    public override string Link() => "font-medium text-primary-600 dark:text-primary-500 hover:underline";
+    public override string Link() => "";
 
     public override string LinkActive( bool active ) => active ? "font-medium active" : null;
 
+    public override string LinkUnstyled( bool unstyled )
+    {
+        if ( unstyled )
+        {
+            return "no-underline text-inherit";
```
stsrki, line 1808: I believe we still want to keep `font-medium` and `hover:underline`.

Source/Extensions/Blazorise.Snackbar/Styles/blazorise.snackbar.scss
```diff
@@ -281,10 +281,6 @@ $snackbar-padding-y-multi: 1.25rem !default;
         position: relative;
         flex-direction: row;
         margin-bottom: 0;
-
```
stsrki, line 284: Why are the snackbar changes in this PR?

Source/Blazorise/Components/Link/Link.razor.cs
```diff
@@ -24,9 +24,20 @@ protected override void BuildClasses( ClassBuilder builder )
...
 
     #endregion
+
+    #region Properties
+
+    /// <summary>
+    /// Remove styles from Link component.
+    /// </summary>
+    [Parameter]
+    public bool Unstyled { get; set; }
```
stsrki, line 40: Move this parameter up to the BaseLinkComponent class.

Also mention it in the API section on the Link page in the documentation.

steal4life:
still working on ant-design @stsrki

Documentation/Blazorise.Docs/Pages/Docs/Components/Links/LinkPage.razor
```diff
@@ -95,6 +95,9 @@
     <DocsAttributesItem Name="Title" Type="string" Default="null">
         Defines the title of a link, which appears to the user as a tooltip.
     </DocsAttributesItem>
+    <DocsAttributesItem Name="Unstyled" Type="bool" Default="false">
+        Removes styles from link.
```
stsrki, line 99:
```suggestion
        Removes default color styles from the link.
```

stsrki (approved):

PR #5068, opened 2023-10-14
ValidationError: show content with higher priority

### LINKED ISSUE #4992: [Bug]: Validation : ValidationError priority for error message | Should ChildContent have an higher priority
### Blazorise Version

1.3

None

Something like:
```
<Validations @ref=_validationsRef HandlerType="ValidationHandlerType.DataAnnotation" Model="_model">
    <Validation>
        <Field>
            <FieldLabel>Phone Country Code</FieldLabel>
            <TextEdit @bind-Text="@_model.PhoneCountryCode">
                <Feedback>
                    <ValidationError>TEST</ValidationError>
                </Feedback>
            </TextEdit>
        </Field>
    </Validation>
</Validations>
<Button Clicked="Submit">Submit</Button>
@code {
    private ValidationLocalizationExample _model = new();
    private Validations _validationsRef;
    public class ValidationLocalizationExample
    {
        [RegularExpression( @"^(\+?\d{1,3}|\d{1,4})$" )]
        public string PhoneCountryCode { get; set; }
    }

    private async Task Submit()
    {
        if (await _validationsRef.ValidateAll())
        {
            Console.WriteLine( "Validation Success!" );
        }
    }
}
```

### Steps to reproduce

Validate

### What is expected?

We should consider to have an higher priority in the ValidationError ChildContent

The default error message shows, instead of what's configured **explicitly** in the **ValidationError**
This is a related issue in Blazorise Support : 
https://blazorise.com/support/issues/122/validation-error-content-being-ignored-and-replaced-with-error-message-from-dataannotation

_No response_

### Any additional comments?

_No response_

stsrki:
Closes #4992

### DIFF
--- a/Source/Blazorise/Components/Validation/ValidationError.razor
+++ b/Source/Blazorise/Components/Validation/ValidationError.razor
@@ -3,7 +3,7 @@
 @if ( ParentValidation?.Status == ValidationStatus.Error )
 {
     <div id="@ElementId" class="@ClassNames" style="@StyleNames" @attributes="@Attributes">
-        @if ( ErrorMessages != null && ErrorMessages.Any() )
+        @if ( ChildContent is null && ErrorMessages != null && ErrorMessages.Any() )
         {
             @if ( Multiline )
             {
--- a/Source/Blazorise/Components/Validation/ValidationError.razor
+++ b/Source/Blazorise/Components/Validation/ValidationError.razor
@@ -3,7 +3,11 @@
 @if ( ParentValidation?.Status == ValidationStatus.Error )
 {
     <div id="@ElementId" class="@ClassNames" style="@StyleNames" @attributes="@Attributes">
-        @if ( ChildContent is null && ErrorMessages != null && ErrorMessages.Any() )
+        @if ( ChildContent is not null )
+        {
+            @ChildContent
+        }
+        else if ( ErrorMessages != null && ErrorMessages.Any() )
         {
             @if ( Multiline )
             {
@@ -18,10 +22,6 @@
             {
                 @string.Join(";", ErrorMessages)
             }
-        }
-        else
-        {
-            @ChildContent
         }
     </div>
 }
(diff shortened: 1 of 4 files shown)


stsrki:
@David-Moreira can you look into tests? You're better at it then I am :)

Source/Blazorise/Components/Validation/ValidationError.razor
```diff
@@ -3,7 +3,7 @@
 @if ( ParentValidation?.Status == ValidationStatus.Error )
 {
     <div id="@ElementId" class="@ClassNames" style="@StyleNames" @attributes="@Attributes">
-        @if ( ErrorMessages != null && ErrorMessages.Any() )
+        @if ( ChildContent is null && ErrorMessages != null && ErrorMessages.Any() )
```
David-Moreira, line 6: Could it read better to do `ChildContent is not null` first and then else ifs? 

The reason being is that I think it reads better, it's the thing with priority, and it's more truthful that the `ChildContent` should be rendered when it is not null, and not the combination of those three logical results.

Tests/Blazorise.Tests/Components/ValidateAnnotationsComponentTest.cs
```diff
@@ -209,4 +210,27 @@ public void CanManuallyValidateName_InitiallyPopulated()
         btn.Click();
         Assert.Contains( "is-valid", edit.ClassList );
     }
+
+    [Fact]
+    public void CanAutoValidateName_InitiallyPopulated_ErrorOverride()
```
David-Moreira, line 215: I think the test is fine. 

I'd probably just add an additional one with an actual input other then **string.empty** for the invalid check?

Only thing of note is that if there is any race condition we need the usual `WaitForAssertion` helpers from bUnit. 
(Usually because the pipeline is more susceptible to it, but I think it's fine here.)

David-Moreira (approved):
LGTM

Tests/Blazorise.Tests/Components/ValidateAnnotationsComponentTest.cs
stsrki: Some tests are failing randomly, and I'm not sure if the problem is in the priority change or they didn't work before. Can you confirm?

David-Moreira: Heyia,

There's two tests failing because it can't locate the validationfeedback by text:
image

Is there a reason why you had removed it?

I reintroduced it.

Source/Blazorise/Components/Validation/ValidationError.razor
stsrki: Done

PR #4998, opened 2023-09-13
Migrating to .NET 8

stsrki:

PR #5091, opened 2023-10-20
Convert Blazorise.Docs to WASM client

stsrki:
This PR continues from #4998. I have separated it into a new PR so that we don't mess with the existing work.

It adds support for the new .NET 8 InteractiveAuto render mode. I'm still not sure if it's worth it. The problems that I see:

- all our secrets would be visible to anyone once the WASM mode is activated
- there are some bugs that prevent some JS scripts from loading. this might be fixed in the final .NET 8 release.

Overall, it works, but as I said, we need to consider carefully should we move forward with this mode.


Build/Blazorise.Docs.props
```diff
@@ -6,11 +6,14 @@
 		<IsPackable>true</IsPackable>
 		<LangVersion>10.0</LangVersion>
     	<AddRazorSupportForMvc>true</AddRazorSupportForMvc>
+		<NoDefaultLaunchSettingsFile>true</NoDefaultLaunchSettingsFile>
+		<StaticWebAssetProjectMode>Default</StaticWebAssetProjectMode>
```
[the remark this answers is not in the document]
stsrki: Because Blazorise.Docs now serve as a WASM client-side project this is needed. Or at least I've seen it used in the official Blazor project template.

Documentation/Blazorise.Docs/App.razor
```diff
@@ -58,8 +58,8 @@
     </script>
 
     <script src="_content/Blazorise.Animate/blazorise.animate.js?v=1.3.1.0"></script>
-    <script src="_content/Blazorise.Docs/blazorise.docs.min.js?v=1.3.1.0"></script>
+    <script src="blazorise.docs.min.js?v=1.3.1.0"></script>
 
-    <script src="_framework/blazor.web.js" suppress-error="BL9992"></script>
+    <script src="_framework/blazor.web.js"></script>
```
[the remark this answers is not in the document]
stsrki: I remember it was removed and mentioned on one of the .NET blog posts. I think it was used just as a temporary solution.

Documentation/Blazorise.Docs/Blazorise.Docs.csproj
```diff
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.NET.Sdk.Razor">
+﻿<Project Sdk="Microsoft.NET.Sdk.BlazorWebAssembly">
```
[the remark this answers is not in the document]
stsrki: As I mentioned. This project is now the client-side project.

David-Moreira:
> This PR continues from #4998. I have separated it into a new PR so that we don't mess with the existing work.
> 
> It adds support for the new .NET 8 InteractiveAuto render mode. I'm still not sure if it's worth it. The problems that I see:
> 
> * all our secrets would be visible to anyone once the WASM mode is activated
> * there are some bugs that prevent some JS scripts from loading. this might be fixed in the final .NET 8 release.
> 
> Overall, it works, but as I said, we need to consider carefully should we move forward with this mode.

I don't think our docs app warrants the need for this. It's a very simple app.
Thing is, if we want to use it just for the sake of being a way for us to get our feet wet on this feature, and also to have it deployed in production, see if we find any flaws.

As for the secrets problem, as any wasm app, they should be moved to server side if they are really sensitive, I don't think we have any, do we? Maybe we do for Blazorise Token? If we do, then that should be moved.

As for the js scripts being prvented of loading, does it still happen? We are in RC so very close to release, I really doubt a fix is coming unless it was reported recently...

Documentation/Blazorise.Docs/Program.cs
```diff
@@ -0,0 +1,14 @@
...
+namespace Blazorise.Docs;
+
+public class Program
+{
+    public static async Task Main( string[] args )
+    {
+        var builder = WebAssemblyHostBuilder.CreateDefault( args );
+
+        await builder.Build().RunAsync();
+    }
```
David-Moreira, line 13: Oh, I see what you mean with being a client project... you setup an actual wasm project.. that's strange does the feature not work without explicitly making it wasm by default?

David-Moreira: Because I don't understand how this entry point would have any effect in the actual entry point we use **Blazorise.Docs.Server** to setup everything.

stsrki: See this comment

image

David-Moreira: That's true for wasm projects, But you're not bootstraping a pure WASM project. You're going through a server ultimately which knows of what needs bootstraping.

David-Moreira: But maybe it's just my lack of knowledge in the new net8 changes. If you've tried it and it did not work, then ok.

PS: Your image shows difference between sdk.wasm and sdk.server.
You changed from sdk.razor to sdk.wasm. I still think sdk.razor is enough for the purpose of it being just a component library... but if you researched the feature and you found out that you absolutely need the sdk.wasm then ok.

Documentation/Blazorise.Docs/Blazorise.Docs.csproj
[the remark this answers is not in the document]
David-Moreira: From watching the day 1 from net conf, the reason for this is that when we are requesting for WASM components, this tells the application that this is a valid project to be downloaded as a DLL. 

This is because, as opposed to before you have an explicit server app, that is able to serve/orchestrate wasm simultaneously, so it needs to figure out what DLLs should be sent to the client. So apparently the `<Project Sdk="Microsoft.NET.Sdk.BlazorWebAssembly">` marker is what indicates this.

So this does make sense if you know your razor library is going to be consumed as WASM.

David-Moreira:
Actually on second though, I've changed my mind. Since it's a simple app, it's in fact easier to deploy this, and we can have a app working on auto mode, it's a way for us to have a production app on auto which allows us to have more confidence that auto is working properly. Which is expected to work properly anyway.

There might be less resource usage on our server also, since people would start loading WASM instead.

stsrki:
I have merged the latest 1.4, and trying to finish this PR but I get the error, see below

image

---

Any chance you could help and see what might be wrong? @David-Moreira

stsrki:
> As for the secrets problem, as any wasm app, they should be moved to server side if they are really sensitive, I don't think we have any, do we? Maybe we do for Blazorise Token? If we do, then that should be moved.

I don't think we will be able to overcome this. The token will still be visible on wasm mode once it is downloaded. That is something we will need to live with.

That leaves us with only needing to hide secrets for sending emails.

David-Moreira:
> I have merged the latest 1.4, and trying to finish this PR but I get the error, see below
> 
> image
> 
> Any chance you could help and see what might be wrong? @David-Moreira

That seems like some kind of weird checking, can you try deleting bin, obj, you know the drill... see if it helps.

David-Moreira:
> > As for the secrets problem, as any wasm app, they should be moved to server side if they are really sensitive, I don't think we have any, do we? Maybe we do for Blazorise Token? If we do, then that should be moved.
> 
> I don't think we will be able to overcome this. The token will still be visible on wasm mode once it is downloaded. That is something we will need to live with.
> 
> That leaves us with only needing to hide secrets for sending emails.

In the new mode, wasm mode only downloads the assemblies in the client project as far as I know. So anything from a dll perspective that should be obscured should remain in the server. Although I don't remember exactly what that Blazorise Token is? I'll have to take a look later.

stsrki:
Already tried deleting obj and bin folders. I figured the error happens after all wasm files are downloaded and I navigate to another page. It works only on first visit when the app still runs in interactive mode.

David-Moreira:
@stsrki Any reason why these do not compile?

David-Moreira:
@stsrki The issue should be fixed. But now you have to fix the DI for the WASM app. Let me know if you need additional help with this PR.

PR #5095, opened 2023-10-22
Fixed row headers on Table and DataGrid

### LINKED ISSUE #2339: Fixed row headers on Table and DataGrid
I hope Table and DataGrid can add the function of fixing row headers and columns

stsrki:
Closes #2339


stsrki:
@David-Moreira can you check this PR for some comments? Note that it's still WIP but I want to have second set of eyes on this.

Essentially this PR introduces the ability to freeze some columns. There are some rules to be met. 

1. All of the columns must have a defined `Width` in `px`
2. Fixed columns, row headers, and header cells must be defined with `Fixed="true"`

After that the Blazorise will figure the rest.

---

The current API is defined on Table and it's subcomponets. It's named `Fixed`. I'm still thinking of renaming it something else, maybe `Frozen`, or `Locked`. Thoughts?

The next step is to implement it on a DataGrid. That's why I wanted to cover the Table first.

David-Moreira:
I like Fixed, as our consumers should already be used to it from `FixedHeader`, or at least makes more sense to keep similar features named the same way.

1. Is this really mandatory? that's a shame? It'll also be a limitation for DataGrid, let's not forget that.
2. What happens if I just define one or the other? It will just partially work?

Demos/Blazorise.Demo/Pages/Tests/TablesPage.razor
```diff
@@ -474,6 +474,51 @@
...
+            <CardBody>
+                <CardText>Use <code>FixedColumns</code> to lock some columns.</CardText>
+            </CardBody>
+            <CardBody>
+                <Table Bordered FixedColumns>
+                     <TableHeader>
+                         <TableRow>
+                            <TableHeaderCell Width="@Width.Px(50)" Fixed>#</TableHeaderCell>
+                            <TableHeaderCell Width="@Width.Px(200)">Column heading</TableHeaderCell>
+                            <TableHeaderCell Width="@Width.Px(150)" Fixed>Column heading</TableHeaderCell>
```
David-Moreira, line 492: Can we change the content of the columns so it's easier to discern the feature?

Source/Blazorise/Components/Table/Table.razor.cs
```diff
@@ -30,12 +30,22 @@ public partial class Table : BaseDraggableComponent
...
 
+    private bool fixedColumns;
+
     private string fixedHeaderTableHeight = "300px";
 
     private string fixedHeaderTableMaxHeight = "300px";
 
     private bool resizable;
 
+    private TableRow lastTableRow;
```
David-Moreira, line 41: hmm a state for the `lastTableRow`, could this affect anyone trying to do dynamic stuff with the table?

Source/Blazorise/Components/Table/Table.razor.cs
```diff
@@ -350,6 +408,24 @@ public bool FixedHeader
         }
     }
 
+    /// <summary>
+    ///  Makes table have a fixed set of columns. This will make it so that the table columns could be fixed to the side of the table.
+    /// </summary>
+    [Parameter]
+    public bool FixedColumns
```
David-Moreira, line 415: Do we really need this? Isn't defining `Fixed` on a column enough to detect it?

Source/Blazorise/Components/Table/TableHeaderCell.razor.cs
```diff
@@ -17,19 +17,49 @@ public partial class TableHeaderCell : BaseDraggableComponent
...
 
         base.BuildClasses( builder );
     }
 
+    /// <inheritdoc/>
+    protected override void BuildStyles( StyleBuilder builder )
+    {
+        if ( Fixed && fixedLeftPosition != null )
+        {
+            builder.Append( $"left:{fixedLeftPosition:G29}px" );
```
David-Moreira, line 57: what the heck is G29? :D

Source/Blazorise/Components/Table/TableHeaderCell.razor.cs
```diff
@@ -54,6 +94,11 @@ protected Task OnClickHandler( MouseEventArgs eventArgs )
     /// </summary>
     [Parameter] public int? ColumnSpan { get; set; }
 
+    /// <summary>
+    /// Fixes the header cell to the start of the table.
```
David-Moreira, line 98: You mention start here? Will we not have a way to fix either at the start or the end?

Source/Blazorise/Fluent/FluentSizing.cs
```diff
@@ -234,16 +251,31 @@ public class FluentSizing :
     /// </summary>
     private readonly Dictionary<SizingSize, List<SizingDefinition>> rules = new();
 
+    /// <summary>
+    /// Holds the sizing rules for the style.
+    /// </summary>
+    private (string unit, double size) styleRule;
```
David-Moreira, line 257: Did we not dislike value tuples?

Source/Blazorise/Fluent/FluentSizing.cs
```diff
@@ -268,7 +300,7 @@ public string Class( IClassProvider classProvider )
         {
             void BuildClasses( ClassBuilder builder )
             {
-                if ( rules.Count > 0 )
+                if ( rules is not null && rules.Count > 0 )
```
David-Moreira, line 303: !NullOrEmpty?

David-Moreira (commented):
Looking good, I like the new Px, Rem, Em fluent utils. :D

Source/Blazorise/Components/Table/Table.razor.cs
stsrki: I wanted it to have it explicitly defined on a Table like this. The reason is if we do it from the TableCell up to the parent Table we would not know when to refresh the class names builder. This way it is _faster_ to render.

Source/Blazorise/Components/Table/TableHeaderCell.razor.cs
stsrki: This is to strip the trailing zeros, see https://stackoverflow.com/a/8675368/833106

Source/Blazorise/Components/Table/TableHeaderCell.razor.cs
stsrki: 🤔 In that case I would have to change the API `bool Fixed` to an enum `TableCellFixedPosition Fixed`(or similar), with

- Start
- End

values.

Source/Blazorise/Fluent/FluentSizing.cs
stsrki: `rules` is not a string, but a dictionary.

Source/Blazorise/Components/Table/TableHeaderCell.razor.cs
stsrki: So. I'm in the middle of refactoring Fixed to support Start and End but I realized it would not be an easy task. Let me explain. For the Start I just calculate offset from the left by iterating each table cell on initialize. But for the End, I cannot do that. Blazor renders cells in the order from left to right, or top to bottom, if you know what I mean. There is no such thing as reverse rendering. Any idea how to calculate the offset from the end/right position?

Source/Blazorise/Components/Table/TableHeaderCell.razor.cs
David-Moreira: hmm kinda unconventional, I'd go the extra mile and make a string constant with FORMAT_STRIP_TRAILING_ZEROS or something along those lines, we will easily forget this in the future I would say.

Although, to be fair it's easily googleable I guess...

Source/Blazorise/Components/Table/Table.razor.cs
David-Moreira: Why are you so quick to say it's faster to render, what's the reason?

Why would you not know? When the `TableCell` notifies the `Table` of initialization the parameters are already set and the `Table` can then decide whether Fixed Cols should apply or not.
Also it's not like that `Table` will be able to display any fixed column at all, until one actual shows up.

Source/Blazorise/Components/Table/TableHeaderCell.razor.cs
David-Moreira: Can't the end offset be calculated by the amount of existing row columns? So instead of being a fixed calc, it's a calculate as rendering happens, since columns can change, either dynamically, or like you said sequential init from blazor regular render.

Source/Blazorise/Fluent/FluentSizing.cs
David-Moreira: Do we not have a `IsNullOrEmpty` for collections?

Source/Blazorise/Components/Table/TableHeaderCell.razor.cs
stsrki: Possibly. But I don't know what performance hit that might take when there are many rows and columns.

Source/Blazorise/Components/Table/Table.razor.cs
stsrki: Because the parent element(Table) doesn't know how many child elements(Cells) there is. And on each cell initialize, we would have to request the parent table to refresh class, `Dirty()`.

Source/Blazorise/Fluent/FluentSizing.cs
stsrki: I forgot we have it. haha

Source/Blazorise/Components/Table/TableHeaderCell.razor.cs
David-Moreira: Could be a chance to investigate the "Defer" possibility if you have the patiente. :D

Source/Blazorise/Components/Table/Table.razor.cs
David-Moreira: Does that already not happen? It still needs to find whatever column to freeze, and each time to calculate a new offset or whatever? It needs to re-render anyway?

stsrki: Not at the moment. We just calculate the sum rf the offset from the start of the row. There is no rerendering involved.

Source/Blazorise/Components/Table/TableHeaderCell.razor.cs
stsrki: image

I don't think defer will help us. Maybe I didn't understand it.

David-Moreira: You should be able to collect everything like the comment implies.
So you have a `Table` with everything in it even before starting the rendering. At least that's my understanding, Although I haven't explored it either.
It should avoid those cases where
**Table** renders
**table row A** renders, does some calc and tells Table there's some additional re-rendering needed based on params, etc...
[...]
Or at least that's my understanding.

stsrki: Tried it. Not that easy. The problem is as before the order of rendering, from left to right. Also the bigger problem is that we really don't want to keep track of each individual row and cell in the memory. For smaller table it is not an issue but as soon as we try playing with virtualizations it becomes a problem.

I think going the JS route might be solution but I still haven't tried it. Virtualization also might be problematic.

All in all. I would rather make it like this and only have fixed cells to the left side, and don't bother with the right side. This is something that most of user would like to have anyways.

David-Moreira: I wouldn't be so sure, see: 
https://github.com/Megabit/Blazorise/issues/5101
User wants action buttons at the end.

Also we shouldn't forget that the naming should be `Start` and `End` instead of `Left` and `Right`. Right?

But I understand if it's complex and we don't have time yet, so better release as is then nothing at all?

stsrki: Do you want to tackle the problem yourself? Maybe I need another brain at this.

I will prepare some basic stuff from you, if yes.

David-Moreira:
@stsrki pushed a version based on events. This one might be better to implement actually.

David-Moreira:
TODO : 
Something seems to be broken here running DOCS:
image

stsrki:
Can you look again into fixed cells pr? For some reason it doesn't look on desktop screens when I apply it on DataGrid

David-Moreira:
> Can you look again into fixed cells pr? For some reason it doesn't look on desktop screens when I apply it on DataGrid

Can you clarify? It looks fine to me.
image

stsrki:
Not for me

image

David-Moreira:
> Not for me
> 
> image

I might be going blind, what's wrong in the picture?

stsrki:
> I might be going blind, what's wrong in the picture?

The horizontal scrollbar doesn't show. It should behave the same as it shows by default when I run the new columns demo under the `/tests/tables` page.

If I open chrome devtools and switch to responsive mode then the scrollbar shows.

David-Moreira:
> > I might be going blind, what's wrong in the picture?
> 
> The horizontal scrollbar doesn't show. It should behave the same as it shows by default when I run the new columns demo under the `/tests/tables` page.
> 
> If I open chrome devtools and switch to responsive mode then the scrollbar shows.

The scrollbar only shows when the content overflows horizontally is it not like that?

PR #5097, opened 2023-10-22
DataGridColumn: DataGridColumnFilterMethod with support for contextual filter | Numerc | Date | Greater or Less Than

David-Moreira:


David-Moreira:
@stsrki what do you think?

Source/Extensions/Blazorise.DataGrid/DataGridColumn.cs
```diff
@@ -759,7 +769,7 @@ public SortDirection CurrentSortDirection
     /// <para>Sets the filter method to be used for filtering the column.</para>
     /// <para>If null, uses the <see cref="DataGrid{TItem}.FilterMethod" /> </para>
     /// </summary>
-    [Parameter] public DataGridFilterMethod? FilterMethod { get; set; }
+    [Parameter] public DataGridColumnFilterMethod? FilterMethod { get; set; }
```
stsrki, line 772: Isn't this a breaking change API since we're renaming the enum?

Documentation/Blazorise.Docs/Models/Snippets.generated.cs
```diff
@@ -6853,19 +6853,19 @@ protected override async Task OnInitializedAsync()
...
     <FilterMenuTemplate>
         <Row>
             <Column ColumnSize=""ColumnSize.Is4"">
-                <Select TValue=""DataGridFilterMethod"" SelectedValue=""@context.GetFilterMethod()"" SelectedValueChanged=""e => { context.FilterMethodChanged.InvokeAsync(e); }"">
-                    <SelectItem TValue=""DataGridFilterMethod"" Value=""@DataGridFilterMethod.Contains"">Contains</SelectItem>
-                    <SelectItem TValue=""DataGridFilterMethod"" Value=""@DataGridFilterMethod.StartsWith"">Starts With</SelectItem>
-                    <SelectItem TValue=""DataGridFilterMethod"" Value=""@DataGridFilterMethod.EndsWith"">Ends With</SelectItem>
-                    <SelectItem TValue=""DataGridFilterMethod"" Value=""@DataGridFilterMethod.Equals"">Equals</SelectItem>
-                    <SelectItem TValue=""DataGridFilterMethod"" Value=""@DataGridFilterMethod.NotEquals"">Not Equals</SelectItem>
+                 <Select TValue=""DataGridColumnFilterMethod"" SelectedValue=""@context.GetFilterMethod()"" SelectedValueChanged=""e => { context.FilterMethodChanged.InvokeAsync(e); }"">
```
stsrki, line 6863: I bet the new VS formatter is automatically adding these new extra spaces. It's getting worse with every release.

Source/Extensions/Blazorise.DataGrid/DataGridColumn.cs
David-Moreira: Well yes. But the enum definitions are equivalent, so I was on the fence on this one. 
However I understand if we want to be rigid about that... 
We could just introduce an extra parameter and make that one obsolete?

Documentation/Blazorise.Docs/Models/Snippets.generated.cs
David-Moreira: Probably.. It's indeed a nightmare... sometimes I notice and fix manually...

Source/Extensions/Blazorise.DataGrid/DataGridColumn.cs
stsrki: We can just add new fields to the `DataGridFilterMethod`, and remove `DataGridColumnFilterMethod`.

David-Moreira: I don't like that as much because it's also used as part of the general DataGrid filter, which I don't feel like would make as much sense to have those.

David-Moreira: Although that would definitely be easier & less complex, from a code standpoint, of course.

stsrki: I like the idea of having a separate DataGridColumnFilterMethod, just fear that we will introduce a _breaking_ API.

David-Moreira: So either
we add the the additional enums to `DataGridFilterMethod`, I guess there could still be some utility.... Even though it will not make sense for text columns... maybe a `DataGrid` that just is number based...? 

Or we add a new `Parameter` with `ColumnFilterMethod` name and make the other one obsolete and with compatibility.

I'm fine with both honestly, but I think I like the second option better, you can go ahead and choose haha.

stsrki: Go with DataGridColumnFilterMethod. YOLO!

David-Moreira: What do you mean? The current version?

stsrki: Yes. 1.4.

stsrki (approved):

PR #5099, opened 2023-10-22
DataGridColumn | HelpText

David-Moreira:

PR #5103, opened 2023-10-23
DataGridSelectColumn | Rework internal setting

David-Moreira:
In the last PR, **#5080**  I just made a very wrong assumption by introducing the setter based on `ValueField` in the `DataGridSelectColumn`. It's just plain wrong, the value here, **should already been resolved by the inner Select**. This was just plain wrong, the fact that it worked properly, was just coincidence that the types matched...

----

Now... upon further ivnestigation... Well, DataGridSelectColumn is not correctly implemented. It just supports the destination to be a **string**. I guess since I made the developments based on the Gender Field which is a string property... I ended up not noticing it would not working with other types...

Since we cast everything down to object with the inner **Select** usage, the internals of `Select` make it so it's basically just a string representation, and anything else other then setting a string, it will not really work, as the object basically reflects a string and not the proper destination type, be it **enum** or an **int**, or whatever else...

We should revert the change I made in **#5080** in **v1.3** as although it gets enums working, it's really not a correct approach, and it breaks other field destination types anyway (altough only string should really be working... :s)...

Now I started exploring, how to make it work... Basically instead of the inner `Select` having a `TValue` of `object`, maybe having the actual correct `TValue`, so whatever is boxed into an object, is the correct `TValue`. So I believe the fix will be to add a new `Type` Parameter the users can set to hint at the proper `TValue` definition so the correct `Value` is then propagated properly.... Not very elegant, but it would get the job done.

Thing is we really should only introduce this new parameter in **v1.4**. Unless we really do introduce it in **v1.3** just because the column is really not working properly right now...

I just left the example in this PR so you can test and let me know your thoughts, or if you have any other ideas, at least to make this work in **v1.3** for the time being?


Source/Extensions/Blazorise.DataGrid/DataGridSelectColumn.cs
```diff
@@ -37,6 +35,12 @@ protected internal override void SetValue( TItem item, object value )
     /// </summary>
     [Parameter] public Func<object, object> ValueField { get; set; }
 
+    /// <summary>
+    /// Provides additional information about the underlying type. This is useful when the bound data is not easily castable to the desired type.
+    /// I.e: enums
+    /// </summary>
+    [Parameter] public Type Type { get; set; }
```
stsrki, line 42: Rename to `ValueType`?

stsrki:
Can it have an `object` by default? We should really add TValue in `1.4`, or even in `2.0` since it is a breaking change if we are required to use it.

Source/Extensions/Blazorise.DataGrid/DataGridSelectColumn.cs
David-Moreira: Yea that seems fine, I was just playing with it, making it work. I don't remember if there's a name collision for `ValueType` though.

David-Moreira:
Great news.

Managed to figure out a workaround. Which honestly I think it's pretty decent and we don't have to leave **v1.3** bugged.
Let me know your thoughts.

**Code to test:**
```
<DataGrid TItem="Employee" Data="@employeeList" PageSize="5" Responsive Editable Filterable EditMode="DataGridEditMode.Popup">
     <DataGridCommandColumn />

     <DataGridSelectColumn TItem="Employee" Field="@nameof( Employee.Role )" Caption="Role(User Sent)" Editable
                           Data="RolesNamesSource" ValueField="(x) => ((UsersRoles)Enum.Parse(typeof(UsersRoles), (string)x))" TextField="(x) => (string)x" />

     <DataGridSelectColumn TItem="Employee" Field="@nameof( Employee.RoleNullable )" Caption="Role With Null Support" Editable
                           Data="RolesNamesSourceWithNull" ValueField="(x) => ((UsersRoles?) ((x is null) ? null : Enum.Parse(typeof(UsersRoles), (string)x)))" TextField="(x) => (string)x" />

     <DataGridSelectColumn TItem="Employee" Field="@nameof( Employee.TestId )" Caption="Test Id" Editable
                           Data="tests" ValueField="(x) => ((Test)x).Id" TextField="(x) => ((Test)x).Name" />

     <DataGridSelectColumn TItem="Employee" Field="@nameof( Employee.TestIdNullable )" Caption="Test Id" Editable
                           Data="testsWithNull" ValueField="(x) => ((Test)x).Id" TextField="(x) => ((Test)x).Name" />

     <DataGridSelectColumn TItem="Employee" Field="@nameof( Employee.TestName )" Caption="Test Name" Editable
                           Data="tests" ValueField="(x) => ((Test)x).Name" TextField="(x) => ((Test)x).Name" />

 </DataGrid>

 @code {

    private List<Employee> employeeList = new List<Employee>() { new Employee( UsersRoles.Standard ) { TestId = 1 }, new Employee( UsersRoles.Premium ) { TestId = 1 }, new Employee( UsersRoles.Admin ) { TestId = 2 } };
    private List<string> RolesNamesSource = Enum.GetNames( typeof( UsersRoles ) ).ToList();
    private List<string> RolesNamesSourceWithNull;
    private List<Test> tests = new List<Test>()
    {
        new Test() { Id=1, Name = "Test1" },
        new Test() { Id=2, Name = "Test2" },
        new Test() { Id=3, Name = "Test3" },
    };

    private List<Test> testsWithNull = new List<Test>()
    {
        new Test() { Id=null, Name = "Test null" },
        new Test() { Id=1, Name = "Test1" },
        new Test() { Id=2, Name = "Test2" },
        new Test() { Id=3, Name = "Test3" },
    };

    {
        RolesNamesSourceWithNull = RolesNamesSource.Prepend( null ).ToList();
        base.OnInitialized();

    }

    class Employee
    {
        public Employee( UsersRoles role )
        {
            Role = role;
        }

        public UsersRoles Role { get; set; }

        public UsersRoles? RoleNullable { get; set; }

        public int TestId { get; set; }

        public int? TestIdNullable { get; set; }

        public string TestName { get; set; }
    }

    public class Test
    {
        public int? Id { get; set; }
    }



    enum UsersRoles
    {
        Standard,
        Premium,
        Admin
    }

}
```

Source/Extensions/Blazorise.DataGrid/_DataGridCellSelectEdit.razor.cs
```diff
@@ -26,4 +27,39 @@ public partial class _DataGridCellSelectEdit<TItem> : ComponentBase
     [Parameter] public EventCallback<object> CellValueChanged { get; set; }
 
     [Parameter] public bool ShowValidationFeedback { get; set; }
+
+    private System.Collections.Generic.List<(string text, object value, bool disabled)> selectItems;
```
stsrki, line 31: Better to have a designated model than the tupple.

David-Moreira: I'm fine with that. I will change after you take a look at approve the actual workaround.

Although to me I would be fine with value tuples in such a specific contained place where readability or navigation is unaffected in my opinion.

stsrki:
I don't see any problems with the fix. Apart from the tuple it is working as expected.

David-Moreira:
> I don't see any problems with the fix. Apart from the tuple it is working as expected.

Refactored Tuple. Please review & merge. And go ahead and review #5099 after unblocking it due to this bugfix.

stsrki (approved):

PR #5122, opened 2023-10-31
DataGrid: Batch Editing

David-Moreira:
Batch Edit Feature:

> - **Batch/Bulk Edit**
To me, this is the ability to accumulate several changes and only save upon strict Save Changes click. I believe this should be built upon the Cell Edit feature.
As such, this will be implemented on a best effort basis. Meaning, it will only be implemented, if we have time to do it as I believe it would be more complex. (Might do it on a seperate PR)


David-Moreira:
@stsrki This is what I'm currently thinking

- `BatchEdit` Flag
- New `SaveBatch` method
- Regular `Save` method (either by button or programatic) will accumulate changes in an internal collection and these will only be saved upon `SaveBatch` method invocation. It should still have a few events the user can react to
- I'm thinking of adding a new button to the Command Column that shows besides the `New` button which will call `SaveBatch`
- I'm not sure whether these should allow deletions and new items? If it does maybe we should have an enum by each entity stating whether it's new/edit/delete?
- Might be challenging to have stuff stay in the Grid temporarly, and if the changes are cancelled to be reverted back to previous state, since currently the bound collection is the "source of truth", I still need to investigate this one, which is the first thing I'll be trying out right away when starting development
- Ideally, any edited cell should have a distinct style. I.E `BatchEditCellChangedStyle`, if we also support new and delete, probably also 
`BatchEditNewRowStyle` and `BatchEditDeleteRowStyle`

Any suggestions, questions or other feedback?

stsrki:
Why we cannot have just a single Save() method?

David-Moreira:
> Why we cannot have just a single Save() method?

It will be less complex for us and to maintain if we clearly split the features in my opinion. 
And this way we take advantage of everything that calls `Save` method without too many changes, or at least that was my thinking, but now that you question it, this might not be quite clear to the user?

So was I overthinking it with the SaveBatch method? Maybe it isn't much work to make some refactoring so the internal `Save` does not call `DataGrid.Save()` but something like `SaveInternal` that knows whether to store the data ("batch data") or to "single save" right away.

And then calling `DataGrid.Save` will always "batch save" or "single save" according to configuration?

Thing is, if user is in `BatchEdit` mode and has a custom implementation where he calls programatically `DataGrid.Save` and wants it to store data instead of calling "batch save", then that's maybe a problem? In that case maybe the `SaveInternal` could be an additional public method (name can change of course).

David-Moreira:
@stsrki 
When saving the changes I don't think we should call the RowRemoved/Inserted/Updated events at all. 
We should stick to something like `BatchSaving` and `BatchSaved`? Would you agree?

stsrki:
Yeah, I think something BatchSaving and BatchSaved makes sense. Especially when the user works in ReadData mode where they use external store to persist the data.

Let me confirm. We're having an internal list of items that are being edited? And then on Save we submit it through BatchSaving and BatchSaved?

David-Moreira:
> Yeah, I think something BatchSaving and BatchSaved makes sense. Especially when the user works in ReadData mode where they use external store to persist the data.
> 
> Let me confirm. We're having an internal list of items that are being edited? And then on Save we submit it through BatchSaving and BatchSaved?

Yes pretty much. Already made most of the core implementation for it, if you want to take a look.

David-Moreira:
@stsrki please review.

stsrki:
1. The Save Changes button is always visible. It should be visible only when change happens.
2. It should also look like a `Link` button by default.
3. The Cancel button should also look like a link by default.
4. I don't think there is a need to show the number of changes, ie `(0)`
5. Maybe we can show Save Changes nd Cancel changes on top of the table instead of putting it inside of command column along with the New button. It takes too much space and the column grows too much.

image

---

When an item is deleted, remove it from the list. Don't mark it with a gray color. It is also strange that we can still Edit an item that is being deleted.

image

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -3129,5 +3388,41 @@ public int SelectedRowIndex
...
+
+    /// <summary>
+    /// Custom handler for the cell styling.
+    /// </summary>
+    [Parameter] public Action<TItem, DataGridColumn<TItem>, DataGridCellStyling> CellStyling { get; set; }
+
+    /// <summary>
+    /// Custom handler for the selected cell styling.
+    /// </summary>
+    [Parameter] public Action<TItem, DataGridColumn<TItem>, DataGridCellStyling> CellSelectedStyling { get; set; }
```
stsrki, line 3419: We already have similar API, `SelectedRowStyling`. So I would rename this to `SelectedCellStyling`.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -3129,5 +3388,41 @@ public int SelectedRowIndex
...
+
+
+    /// <summary>
+    /// Custom handler for the cell styling when the cell has batch edit changes.
+    /// </summary>
+    [Parameter] public Action<BatchEditItem<TItem>, DataGridColumn<TItem>, DataGridCellStyling> CellBatchEditStyling { get; set; }
```
stsrki, line 3425: Maybe rename it to `BatchEditCellStyling `?

Source/Extensions/Blazorise.DataGrid/DataGridCellStyling.cs
```diff
@@ -0,0 +1,22 @@
...
+
+    /// <summary>
+    /// Cell custom background color.
+    /// </summary>
+    public Background Background { get; set; }
+
+    /// <summary>
+    /// Cell custom text color.
+    /// </summary>
+    public TextColor TextColor { get; set; }
```
stsrki, line 21: Should we also include `Class` and `Style` for more control?

Source/Extensions/Blazorise.DataGrid/EventArguments/BatchChangeEventArgs.cs
```diff
@@ -0,0 +1,24 @@
...
+using System;
+#endregion
+
+namespace Blazorise.DataGrid;
+
+/// <summary>
+/// Represents the Batch Change Event Arguments.
+/// </summary>
+/// <typeparam name="TItem"></typeparam>
+public class BatchChangeEventArgs<TItem> : EventArgs
```
stsrki, line 11: Rename to `DataGridBatchChangeEventArgs`.

Source/Extensions/Blazorise.DataGrid/EventArguments/BatchSavedEventArgs.cs
```diff
@@ -0,0 +1,26 @@
...
+using System.ComponentModel;
+#endregion
+
+namespace Blazorise.DataGrid;
+
+/// <summary>
+/// Represents the Batch Saved Event Arguments.
+/// </summary>
+/// <typeparam name="TItem"></typeparam>
+public class BatchSavedEventArgs<TItem> : EventArgs
```
stsrki, line 13: Rename to `DataGridBatchSavedEventArgs`

Source/Extensions/Blazorise.DataGrid/EventArguments/BatchSavingEventArgs.cs
```diff
@@ -0,0 +1,25 @@
...
+using System.ComponentModel;
+#endregion
+
+namespace Blazorise.DataGrid;
+
+/// <summary>
+/// Represents the Batch Saving Event Arguments.
+/// </summary>
+/// <typeparam name="TItem"></typeparam>
+public class BatchSavingEventArgs<TItem> : CancelEventArgs
```
stsrki, line 12: Rename to `DataGridBatchSavingEventArgs`

Source/Extensions/Blazorise.DataGrid/Models/BatchEditItem.cs
```diff
@@ -0,0 +1,108 @@
...
+using Blazorise.Extensions;
+#endregion
+
+namespace Blazorise.DataGrid;
+
+/// <summary>
+/// Tracks the state of the batch edit item.
+/// </summary>
+/// <typeparam name="TItem"></typeparam>
+public class BatchEditItem<TItem>
```
stsrki, line 12: Rename to `DataGridBatchEditItem`

David-Moreira:
> 1. The Save Changes button is always visible. It should be visible only when change happens.
> 2. It should also look like a `Link` button by default.
> 3. The Cancel button should also look like a link by default.
> 4. I don't think there is a need to show the number of changes, ie `(0)`
> 5. Maybe we can show Save Changes nd Cancel changes on top of the table instead of putting it inside of command column along with the New button. It takes too much space and the column grows too much.
> 
> image
> 
> When an item is deleted, remove it from the list. Don't mark it with a gray color. It is also strange that we can still Edit an item that is being deleted.
> 
> image

1. Fine.
2. It looks like all the other commands by default. That one is customized.
3. It looks like all the other commands by default. That one is customized.
4. It's customized. It doesn't show by default.
5. I guess that could be fine. I'll try.

David-Moreira:
> > 1. The Save Changes button is always visible. It should be visible only when change happens.
> > 2. It should also look like a `Link` button by default.
> > 3. The Cancel button should also look like a link by default.
> > 4. I don't think there is a need to show the number of changes, ie `(0)`
> > 5. Maybe we can show Save Changes nd Cancel changes on top of the table instead of putting it inside of command column along with the New button. It takes too much space and the column grows too much.
> > 
> > image
> > When an item is deleted, remove it from the list. Don't mark it with a gray color. It is also strange that we can still Edit an item that is being deleted.
> > image
> 
> 1. Fine.
> 2. It looks like all the other commands by default. That one is customized.
> 3. It looks like all the other commands by default. That one is customized.
> 4. It's customized. It doesn't show by default.
> 5. I guess that could be fine. I'll try.



> > 1. The Save Changes button is always visible. It should be visible only when change happens.
> > 2. It should also look like a `Link` button by default.
> > 3. The Cancel button should also look like a link by default.
> > 4. I don't think there is a need to show the number of changes, ie `(0)`
> > 5. Maybe we can show Save Changes nd Cancel changes on top of the table instead of putting it inside of command column along with the New button. It takes too much space and the column grows too much.
> > 
> > image
> > When an item is deleted, remove it from the list. Don't mark it with a gray color. It is also strange that we can still Edit an item that is being deleted.
> > image
> 
> 1. Fine.
> 2. It looks like all the other commands by default. That one is customized.
> 3. It looks like all the other commands by default. That one is customized.
> 4. It's customized. It doesn't show by default.
> 5. I guess that could be fine. I'll try.

Do you think I should remove the customizations in the docs?

David-Moreira:
Do you think I should remove the customizations in the docs?

Source/Extensions/Blazorise.DataGrid/DataGridCellStyling.cs
David-Moreira: Well we already have a Func per column, didn't want to add even more redudancy that could make it confusing:
image

David-Moreira:
Also for **5.** do you think we should add a new API instead of being integrated in the command column?
Like `BatchEditBar` or something along those lines?

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
David-Moreira: I kinda would argue that this way is more discoverable. But I will change.

Source/Extensions/Blazorise.DataGrid/DataGridCellStyling.cs
stsrki: I was thinking the same thing. We could

1. expand DataGridCellStyling with Class and Style
2. introduce new CellStyling that is on DataGrid level
3. make obsolete current CellClass and CellStyle from DataGridColumn level

or

1. leave DataGridCellStyling as it is
2. introduce new CellStyling that is on DataGrid level
3. leave CellClass and CellStyle as they are on DataGridColumn level

I think the first choice is better in the long run and we will have the similar API as it is on the RowStyling

stsrki:
> Do you think I should remove the customizations in the docs?

Yes, keep it simple for the basic example. If needed we can have extra examples with the styling.

---

> Also for 5. do you think we should add a new API instead of being integrated in the command column?
Like BatchEditBar or something along those lines?

If we're going for the top bar, then I would say yes.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
stsrki: Possibly. But it is better to follow the same naming pattern to what we already have.

David-Moreira:
> 
> If we're going for the top bar, then I would say yes.

Actually I was thinking... We already have `ButtonRow` as an alternative to have top or bottom bar with contextual buttons.
I would leave the command column as is, and additionally introduce these new "commands" in the `ButtonRow`. Up to the user whether he uses one or another. Since he can control whether it's displayed with the `*Allowed` flags.

Source/Extensions/Blazorise.DataGrid/DataGridCellStyling.cs
David-Moreira: I'd prefer the first choice also. It was also my reasoning when introducing it. To be similar to `RowStyling` and defined in the `DataGrid` (so we don't have to define the same handler in all columns if it works the same for all cols).

I'll go ahead with the first one.

David-Moreira:
> > If we're going for the top bar, then I would say yes.
> 
> Actually I was thinking... We already have `ButtonRow` as an alternative to have top or bottom bar with contextual buttons. I would leave the command column as is, and additionally introduce these new "commands" in the `ButtonRow`. Up to the user whether he uses one or another. Since he can control whether it's displayed with the `*Allowed` flags.

PS: Please let me know your thoughts.

stsrki:
ButtonRow name is already quite generic, and I think it makes sense to also include Batch buttons in it.

David-Moreira:
> ButtonRow name is already quite generic, and I think it makes sense to also include Batch buttons in it.

Yep that seems like the best approach to me. Have it available on both Commands and ButtonRow. 
User chooses how to use it.

stsrki (commented):
I don't see any problem with the code. It works from what I could test.

stsrki:
Is there any more work you plan to do on this, or do you think it is done?

David-Moreira:
> Is there any more work you plan to do on this, or do you think it is done?

I'm confortable with how it's working. If you don't have any additional suggestions, I'd go ahead and merge.

It would be great if you can release a blazorise Preview or RC so we can try to get some feedback from the support user that really wanted this feature. 

(would be great to also release docs preview, so the feature can be seen in action)

We can also ask on the issue if anyone wants to take it for a spin.

PR #5125, opened 2023-11-01
Fix conversion of component values to InvariantCulture

### LINKED ISSUE #5105: [Bug]: Slider values in html are based on Culture
### Blazorise Version

1.3.1

Bootstrap5

 During developing on Windows with Polish language (for double period '.' is changed to comma ',' ). When i make slider with doubles it doesn't use InvariantCulture.

### Steps to reproduce

1. Add slider like that.
`<Slider TValue="double" Min="0.0" Max="1" Step="Step" @bind-Value="testKnow"></Slider>`
2. Change culture to culture where doubles are used with comma ',' (i have similar statement to use invariant in Program.cs as hotfix)
`CultureInfo.DefaultThreadCurrentCulture = new CultureInfo("pl-PL")`
3. The output is similar to `<input id="0HMUKK2MMLIIV" type="range" class="form-range" step="0,0001" min="0" max="1">`

### What is expected?

<input id="0HMUKK2MMLIIV" type="range" class="form-range" step="0.0001" min="0" max="1">

`<input id="0HMUKK2MMLIIV" type="range" class="form-range" step="0,0001" min="0" max="1">`

Chrome

### Any additional comments?

_No response_

stsrki:
Closes #5105

### DIFF
--- a/Source/Blazorise.AntDesign/Slider.razor
+++ b/Source/Blazorise.AntDesign/Slider.razor
@@ -4,7 +4,7 @@
     <div class="ant-slider-rail"></div>
     <div class="ant-slider-track" style="@TrackStyle"></div>
     <div class="ant-slider-step"></div>
-    <div tabindex="0" class="ant-slider-handle" role="slider" aria-valuemin="@Min" aria-valuemax="@Max" aria-valuenow="@Value" aria-disabled="@Disabled" style="@HandleStyle"></div>
+    <div tabindex="0" class="ant-slider-handle" role="slider" aria-valuemin="@MinString" aria-valuemax="@MaxString" aria-valuenow="@Value" aria-disabled="@Disabled" style="@HandleStyle"></div>
     <div class="ant-slider-mark"></div>
 </div>
 @ChildContent
--- a/Source/Blazorise.Bootstrap/NumericPicker.razor
+++ b/Source/Blazorise.Bootstrap/NumericPicker.razor
@@ -24,6 +24,6 @@ else
 {
     protected RenderFragment InputTemplate => __builder =>
     {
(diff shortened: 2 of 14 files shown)


David-Moreira:
Could this produce breaking behavior to users who are actually relying on the thread culture?
Would it be viable to introduce a Parameter that either 
- Let's you set a custom Culture
- Let's you specify whether you want it to be invariant or not (bool flag?)

David-Moreira (approved):

stsrki:
Good question. As far as I know, input elements should work with invariant values by default.

image

I might be wrong.

stsrki:
I think we're good. I've checked Blazor source code and it seems they also default to InvariantCulture in all cases

image

David-Moreira:
What's this code that I'm seeing their representation of html inputs? `<InputText>` `<InputNumber>` etc?
Making the assumption that they already follow or try to follow best industry practices, I'm guessing this is indeed a nice change for us to follow.

However I'd still have the same concern, that we didn't convert to Invariant before, so converting now, might lead to different behaviour, even if that's the supposed best practice?


On an off note I'm seeing they have some conditionals there you can base yourself off? They check for each of the value types instead of IFormatable

image

stsrki:
> On an off note I'm seeing they have some conditionals there you can base yourself off? They check for each of the value types instead of IFormatable

We also have those on NumericEdit, https://github.com/Megabit/Blazorise/blob/master/Source/Blazorise/Components/NumericEdit/NumericEdit.razor.cs#L129

But yeah, maybe you're right. Maybe I can remove the change for the Value as that is the most concerned change.

image

For Step, Min, and Max I don't think there is much harm as they are mostly to pass the number to the browser. Wo don't parse of convert it later, like we did for the Value.

PR #5137, opened 2023-11-08
MessageProvider: Change russian localization

denis-ivanov:
Change russian localization


stsrki (approved):
Looking good. Thank you for the contribution.

Issue #5116, opened 2023-10-28
DataGrid ReadData not handling pagination properly


David-Moreira:
### Discussed in https://github.com/Megabit/Blazorise/discussions/5115

<div type='discussions-op-text'>

<sup>Originally posted by **mdmontesinos** October 28, 2023</sup>
DataGrid with ReadData mode does not handle correctly pagination.

I have a PageSize of 1, 2 elements, and I am in page 2, everything's correct:

![image](https://github.com/Megabit/Blazorise/assets/90258222/5bf276ca-b9cd-4ad0-b1ee-da31d94d8b1e)

However, when changing to a PageSize of 5, the ReadData event is not triggered so the data source is not updated from the API, even if the DataGrid does change its internal value to 5 and shows that there's no previous page.

![image](https://github.com/Megabit/Blazorise/assets/90258222/354fd78f-62fd-499a-8698-6f06eaa9d1dd)

The last console.log I get is "PageSize: 1, CurrentPage: 2", when the ReadData event should be triggered with "PageSize: 5, CurrentPage: 1".

I'm using Tailwind provider, if it's any help.

Here's the code for it (the data source is just an API with simple pagination mecanism):

```razor
<DataGrid TItem="ClinicDto"
          Data="@clinics"
          ReadData="@OnReadData"
          TotalItems="@totalClinics"
          Responsive
          ShowPager
          ShowPageSizes
          PageSizes="new[] { 1, 5, 10, 15 }"
          PagerPosition="DataGridPagerPosition.Bottom">

     <DataGridColumn Field="@nameof(ClinicDto.Name)" Caption="Nombre" Sortable="true" />
     <DataGridColumn Field="@nameof(ClinicDto.Address)" Caption="Dirección" />
     <DataGridColumn Field="@nameof(ClinicDto.PhoneNumber)" Caption="Teléfono" />
     <DataGridColumn Field="@nameof(ClinicDto.Email)" Caption="Correo electrónico" />

 </DataGrid>

@code {
    PagedList<ClinicDto>? clinics;
    int totalClinics;

    private async Task OnReadData(DataGridReadDataEventArgs<ClinicDto> e)
    {
        if (e.CancellationToken.IsCancellationRequested) return;

        Console.WriteLine($"PageSize: {e.PageSize}, CurrentPage: {e.Page}");

        var parameters = new ClinicParameters()
            {
                PageSize = e.PageSize,
                PageNumber = e.Page,
            };

[...]

PR #5140, opened 2023-11-09
Fix ReadData handling between PageSizeChanged and PageChanged

David-Moreira:
**To reproduce:**

`DataGrid` with `ReadData`:

- Go to last page
- Change page size to allow more records per page
- Both PageSize and Page Changed events will trigger and internally it won't call `ReadData`.

**Problem**
- PageSizeChanged event triggers
- internal HandleReadData triggers
- IsLoading is set to true
- The task yields back control before calling `ReadData`
- PageChanged event triggers and cancels the current operation (CancellationToken)
- internal HandleReadData triggers
- IsLoading is true and jumps out
- The PageSizeChanged HandleReadData resumes and verifies the cancellation token that has been cancelled and jumps out
... 
with this, no `ReadData` was called and the data will not update.


stsrki:
Closes #5116

stsrki (approved):

PR #5160, opened 2023-11-25
ColorPicker: Add more configurability for the UI elements

### LINKED ISSUE #5156: Add more configurability to ColorPicker
I'm trying to use ColorPicker to choose colors with Hue and Alpha. Pickr has ability to do that, but current implementation falls short in configurability.

Add an option to address this line: https://github.com/Megabit/Blazorise/blob/master/Source/Blazorise/wwwroot/colorPicker.js#L30C24-L30C24 
Aka, add booleans to enable alpha, hue, etc. options.

Thanks for the absolutely brilliant work you've done in this repo.

stsrki:
Closes #5156

**Adds**

- ShowOpacitySlider
- ShowHueSlider
- ShowInputField

**Test**

```razor
<Field>
    <FieldLabel>@color1</FieldLabel>
    <FieldBody>
        <ColorPicker @bind-Color="@color1" />
    </FieldBody>
</Field>
<Field>
    <FieldLabel>@color2</FieldLabel>
    <FieldBody>
        <ColorPicker @bind-Color="@color2" ShowOpacitySlider="false" />
    </FieldBody>
</Field>
<Field>
    <FieldLabel>@color3</FieldLabel>
    <FieldBody>
        <ColorPicker @bind-Color="@color3" ShowHueSlider />
    </FieldBody>
</Field>
<Field>
    <FieldLabel>@color4</FieldLabel>
    <FieldBody>
        <ColorPicker @bind-Color="@color4" ShowOpacitySlider="false" ShowHueSlider ShowInputField="false" />
    </FieldBody>
</Field>
```


Source/Blazorise/wwwroot/colorPicker.js
```diff
@@ -26,8 +26,8 @@ export function initialize(dotnetAdapter, element, elementId, options) {
 
             // Main components
             preview: true,
-            opacity: true,
-            hue: false,
+            opacity: options.showOpacitySlider === true ? true : false,
```
David-Moreira, line 29: isn't this a truthful expression on itself?

Prefer  `options.showOpacitySlider === true` over `options.showOpacitySlider === true ? true : false`

David-Moreira: same for the ones below

Source/Blazorise/wwwroot/colorPicker.js
```diff
@@ -36,10 +36,10 @@ export function initialize(dotnetAdapter, element, elementId, options) {
                 hsla: false,
                 hsva: false,
                 cmyk: false,
-                input: true,
+                input: options.showInputField === true ? true : false,
                 save: false,
-                clear: options.showClearButton || true,
-                cancel: options.showCancelButton || true
+                clear: options.showClearButton === true ? true : false,
+                cancel: options.showCancelButton === true ? true : false
```
David-Moreira, line 42: Can we go over what the previous logic did and what this one does now.

Previously:
`options.showClearButton || true`
Which one of these is right for this truth expression?
1 - if null or false returns true?
2 - if null returns true?
I admit I don't remember how js checks it like this. I'm guessing it's number 2 as number 1 does not really make sense, right?

`options.showClearButton === true`
Basically only if true, and excludes **"true"**, right?

David-Moreira (approved):
LGTM, just passing down parameters to js. Just a two simple notes on the review

Source/Blazorise/wwwroot/colorPicker.js
stsrki: > isn't this a truthful expression on itself?
> 
> Prefer `options.showOpacitySlider === true` over `options.showOpacitySlider === true ? true : false`

Not necessarily always. In JS the truthful value can be any thing that is not false, non-empty string, number higher than 0, etc. This also ensures that the result is always bool.

Source/Blazorise/wwwroot/colorPicker.js
stsrki: Basically the same explanation as above.

PR #5165, opened 2023-11-28
Code cleanup for null checks

stsrki:


David-Moreira (approved):
I just glanced over it, LGTM, and I prefer this convention also.
As long as you used the find to make sure you've found the `!= null and == null` most, if not all must have been caught.
I'd go further ahead and make it a requite in editor config if you haven't yet. this will enforce it going forward.

PR #5171, opened 2023-12-02
License Limitations

David-Moreira:


Source/Blazorise/Licensing/BlazoriseLicenseProvider.cs
```diff
@@ -95,14 +127,182 @@ private async void BackgroundWorker_DoWork( object sender, DoWorkEventArgs e )
...
+                    return BlazoriseLicenseResult.Trial;
+                default:
+                    break;
+            }
+        }
+
+        return BlazoriseLicenseResult.Licensed;
+    }
+
+    internal int? GetDataGridRowsLimit()
```
stsrki, line 153: All the functions seems to do the sam thing. Can we make an additional function to resolve the with the same logic?

eg. `internal int? GetRowsLimit(string limitPropertyName, int communityMaxRows, int unlicensesMaxRows)`

stsrki (approved):

PR #5190, opened 2023-12-14
DataGrid: introduce CloneItemCreator parameter

### LINKED ISSUE #5184: DataGrid clone not working
When one of the fields on the model is a generic type, we get the error.

image

**To reproduce:**

```razor
<DataGrid TItem="Car" Data="@items" Editable EditMode="DataGridEditMode.Form">
    <DataGridColumns>
        <DataGridCommandColumn />
        <DataGridColumn Field="Model" Caption="Model" Editable />
        <DataGridColumn Field="Brand" Caption="Brand" Editable />
        <DataGridColumn Field="Extra" Caption="Extra" Editable />
    </DataGridColumns>
</DataGrid>

@code {
    List<Car> items = new()
    {
        new() { Brand = "VW", Model = "Golf" },
        new() { Brand = "Fiat", Model = "Punot" },
        new() { Brand = "Opel", Model = "Astra" },
        new() { Brand = "Porsche", Model = "Cayenne" },
    };

    class Car
    {
        public string Model { get; set; }
        public string Brand { get; set; }

        public SomeExtra<string> Extra { get; set; }
    }

    class SomeExtra<T>
    {
        public T Value { get; set; }
    }
}
```

Try to edit and click on Save.

stsrki:
Closes #5184

The culprit was not in #5184 example, but I'm just referencing it so that we can close the ticket.

As far as I was able to figure, the problem could be in the client possibly not having full-trust on their runtime objects. So DeepClone would fail. You can read about it on https://github.com/force-net/DeepCloner?tab=readme-ov-file#limitation

This PR introduces a new `CloneItemCreator` API. I am aware that we usually don't allow adding new APIs into existing versions but this time it was needed. I was able to test it on the client code and the change is working. On their side I just implement it as

```cs
MulticastAllocationEntry OnCloneItemCreator(MulticastAllocationEntry item)
{
    return item.ShallowClone();
}
```

### DIFF
--- a/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
+++ b/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
@@ -880,7 +880,7 @@ public partial class DataGrid<TItem> : BaseDataGridComponent
 
         var rowSavingHandler = editState == DataGridEditState.New ? RowInserting : RowUpdating;
 
-        var editItemClone = editItem.DeepClone();
+        var editItemClone = CloneItemCreator != null ? CloneItemCreator.Invoke( editItem ) : editItem.DeepClone();
         SetItemEditedValues( editItemClone );
 
         if ( await IsSafeToProceed( rowSavingHandler, editItem, editItemClone, editedCellValues ) )
@@ -2637,6 +2637,11 @@ public partial class DataGrid<TItem> : BaseDataGridComponent
     /// Function that, if set, is called to create a instance of the selected item to edit. If left null the selected item will be used.
     /// </summary>
     [Parameter] public Func<TItem, TItem> EditItemCreator { get; set; }
+
+    /// <summary>
+    /// Function that, if set, is called to clone an instance of the saving item. If left null the built-in DeepClone method will be used.
+    /// </summary>
+    [Parameter] public Func<TItem, TItem> CloneItemCreator { get; set; }
 
     /// <summary>
     /// Adds stripes to the table.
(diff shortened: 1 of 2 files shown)


Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -880,7 +880,7 @@ internal protected async Task SaveItem()
 
         var rowSavingHandler = editState == DataGridEditState.New ? RowInserting : RowUpdating;
 
-        var editItemClone = editItem.DeepClone();
+        var editItemClone = CloneItemCreator != null ? CloneItemCreator.Invoke( editItem ) : editItem.DeepClone();
```
David-Moreira, line 883: `is not `if you care to bother

David-Moreira (approved):
LGTM.

I just realized that we probably went overkill with DeepClone, we really only needed public read/set properties as that's the type of models that should be bound to the Grid, oh well, giving an option for the user to use their own implementation is nice.
image

David-Moreira:
**PS:** Did the client try to use a simpler DTO as an alternative?

stsrki:
> **PS:** Did the client try to use a simpler DTO as an alternative?

The object is quite complex with dozens of inherited classes and interfaces. Not to mention event handlers and callbacks. So having a simple DTO is not an option.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
stsrki: We will do it in 1.4, or 1.5.

David-Moreira:
> > **PS:** Did the client try to use a simpler DTO as an alternative?
> 
> The object is quite complex with dozens of inherited classes and interfaces. Not to mention event handlers and callbacks. So having a simple DTO is not an option.

Right my point exactly, I don't see how it is not an option, seems like an alternative other then introducing a new api in the middle of a Blazorise version. 
We are **conceding ourselves** on our software so the client does not do more work himself or waits for a new version. 

It's a somewhat harmless addition, but if we start conceding here and there... 

PS: Why are we adding in 1.3 which has already run through a few versions? I'd be much more pleased to have it 1.4 which just released.

Issue #5199, opened 2023-12-20
[Bug]: Ant Design Datagrid page size selection doesn't show list


Lasoty:
### Blazorise Version

1.4

### What Blazorise provider are you running on?

AntDesign

### Link to minimal reproduction, or a simple code snippet

Can be reporduced on https://antdesigndemo.blazorise.com/tests/datagrid/aggregates. 

![image](https://github.com/Megabit/Blazorise/assets/23433922/6e82b201-e1df-457a-ae54-2593bd83b355)

### Steps to reproduce

1. Open browser
2. Navigate to https://antdesigndemo.blazorise.com/tests/datagrid/aggregates
3. Try change datagrid page size. 

### What is expected?

List of page sizes is visible.

### What is actually happening?

The page size list is not visible. I can only see the scroll bar.

### What browsers are you seeing the problem on?

Chrome, Microsoft Edge

stsrki:
I assume the problem might be with the wrong `position: absolute` calculation. 

@steal4life Can you look into it?

PR #5207, opened 2023-12-28
DataGrid: fix position of Select component for AntDesign

steal4life:
cloeses #5199

If anybody has sugestion how can we modify elements.style withouth using !impotrant write it down.


Source/Blazorise.AntDesign/Select.razor.cs
```diff
@@ -233,6 +241,12 @@ protected string InputElementId
         set => inputElementId = value;
     }
 
+    protected string DropdownElementId
```
David-Moreira, line 244: Why are we adding this?
Why do we only assign a new Id, only if an attempt to get this Id is made? It seems like this could cause it so the html element start without an Id, and later if an access is made, suddently gets a random Id?

It should follow our current convention, the Id should be assigned when initialized and can be overriden
i.e:
image

Source/Blazorise.AntDesign/Select.razor.cs
```diff
@@ -315,7 +329,7 @@ string DropdownClassNames
     }
 
     string DropdownStyleNames =>
-        $"width: {(int)elementInfo.BoundingClientRect.Width}px; left: {(int)elementInfo.OffsetLeft}px; top: {(int)( elementInfo.OffsetTop + elementInfo.BoundingClientRect.Height )}px;";
+        $"width: {(int)dropdownElementInfo.BoundingClientRect.Width}px; left: {(int)dropdownElementInfo.OffsetLeft}px; top: {(int)( dropdownElementInfo.OffsetTop + dropdownElementInfo.BoundingClientRect.Height )}px;";
```
David-Moreira, line 332: I still don't get, how this is different then before? Maybe I'm not seeing or understanding something, can you explain?

David-Moreira (changes requested):

Source/Blazorise.AntDesign/Select.razor.cs
steal4life: Thing is I tried to make select list in antD to ref on dropdownElement not on element due to it sets wrong position for that list.

Source/Blazorise.AntDesign/Select.razor.cs
stsrki: I think setter is not needed in this case. But we still need to have `DropdownElementId`. AntDesign has a more complex HTML structure so we need to have the `id` placed on the places where we need to get the info.

Source/Blazorise.AntDesign/Select.razor.cs
steal4life: Basicalyit gets wrong info and sets wrong element.styles

Source/Blazorise.AntDesign/Select.razor.cs
David-Moreira: Can you clarify? Should it not follow the same approach of Auto generating an ID on Initialized, if we need it?

Also I can't seem to notice it being used? But maybe I'm not understanding.

Source/Blazorise.AntDesign/Select.razor.cs
David-Moreira: I still do not understand... :( 
Was this an attempt that did not work? Or is this working? Because personally, with the current code I do not see a difference between elementInfo and dropdownElementInfo

Is this PR working as expected? Or are we still figuring out an issue? We've marked this PR as review ready.

steal4life: I'm just trying to fix it, this doesn't work, but I'm trying to find a solution to fix the positioning of the select list, which for some reason appears in the wrong position and has the wrong width

steal4life: when I was fixing it, I noticed that when moving the position of the absolute from the class (ant-selecat dropdown), the list moves to the right position, but it breaks some other elements in blazorise that are directly connected to that same class, now I'm trying to find another way because it seems that this is not a good approach either

David-Moreira: Ah gotcha! No point in reviewing not ready code. Solution might change completly.
I'll see if I can take a look later today if you'd like.

steal4life: yes, why not, I'll try to find a solution, so if I change anything, I'll push the changes so you can take a look

David-Moreira:
Hello @steal4life 

I have taken a look at the issue.

I'm not sure if this is based on an old implementation of ant design, but it doesn't seem to match with the current implementation of ant design.

The problem is that currently you just have a div sitting besides the select input, and it relies on it rendering below this select input correctly positioned.

This makes it "vulnerable" to parent containers styling, which might make it does not render right below the select input as expected as displayed by this issue.
I believe modern solutions fix this with js by trying to be smart and dynamically allocating the dropdown according to coordinates and the current state of the document html elements. And I think the current ant design approach also does this, which is not the same as ours it seems.

@stsrki do you agree with my observations?
---
If my observations are correct, then proper fix would be to try and mimick the correct ant design behaviour which I'm not sure it's too easy at this stage... Another option is to try and use our **popperjs** system which should accomplish this.
----
Anyway, if we focus just on fixing this particular case I would fix it by doing something similar to what you did similarly, setting the parent container position to relative, in this case the pagination div container. In a way it's specific to this particular issue.

Then there's another problem, which seems like the values for the dimensions for the DropdownStyleNames are incorrect, the dimensions returned are incorrect and different from manually querying the element like so:
`document.getElementById("0HN0BKQQNVDB5").getBoundingClientRect()`

I would suggest taking a look at the `JSUtilitiesModule.GetElementInfo` see if there's something peculiar to how we grab the dimensions, and why they are different.
For example, maybe the following line, uses a stale element, and takes in the wrong dimensions? Maybe it should really requery the element again with `element = document.getElementById(elementId);`. If that's the problem maybe you can add an option to force it to requery the element by id.

David-Moreira:
Actually I was just testing by changing it to requery with 
`element = document.getElementById(elementId);` and it automatically worked properly... so figures... maybe adding an option to force it is appropriate, and force it for this particular case.

So whatever my observations were before, it seems to work alright, by just requerying the element correctly. Cheers.
image

I would just revert everything that you've done until this point and focus on this. 
Start with the forcing of  `element = document.getElementById(elementId);` and make sure nothing else is breaking, if it's not it should be alright to do with the optional forcing.

steal4life:
> element = document.getElementById(elementId);

the only way, because when I tried to change the position absolute to relative, it turned out fine, but in the forms it would disrupt the behavior of the multiselect form or the regular select form, so I gave up on that idea, thank you very much for your help, I will try this way

steal4life:
@David-Moreira sorry for the trouble, I'm trying to do what you said, but it's not entirely clear to me what exactly you did to get the result that it works

David-Moreira:
I just tried it again,

**1.** rel1.4 or rel1.4 based Fresh branch 

**2.** go to utilities.js and replace
```
    if (!element) {
        element = document.getElementById(elementId);
    }
```

with 

`    element = document.getElementById(elementId);`

**3.** run the app, and it should be working correctly (don't forget to clear browser cache if needed)

steal4life:
thanks @David-Moreira, it was problem in browser cache, now it's working corectly

stsrki (approved):

PR #5212, opened 2024-01-06
DataGrid: add State API

David-Moreira:
- Added `LoadState` - Loads the provided state into the `DataGrid`
- Added `GetState` - Gets the current `DataGrid` State
- Added demo page to test the feature
- Added docs page for the feature
- Added `Blazored.LocalStorage` to show examples of loading the DataGrid state from local storage in both the demo & docs.
  - This is because this might be the most sought use case for this feature

---


David-Moreira:
@stsrki Please review.
I believe the current state we are allowing to load is fine for now. 
But let me know if you have any suggestions on something we might have missed and might be good to load/get.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -1280,6 +1374,14 @@ private async Task ResetSorting()
         SortByColumns.Clear();
     }
 
+    private void ResetFiltering()
+    {
+        foreach ( var column in Columns )
```
stsrki, line 1379: Add a null check for `Columns`?

Source/Extensions/Blazorise.DataGrid/DataGridState.cs
```diff
@@ -0,0 +1,161 @@
...
+
+    /// <summary>
+    /// Gets or sets the filter state for the DataGrid columns.
+    /// </summary>
+    public List<DataGridColumnFilterState<TItem>> ColumnFilterStates { get; set; }
+
+    /// <summary>
+    /// Sets the DataGrid to the EditState.New.
+    /// </summary>
+    public void SetNew()
```
stsrki, line 56: Rename to `SetNewState()`?

Source/Extensions/Blazorise.DataGrid/DataGridState.cs
```diff
@@ -0,0 +1,161 @@
...
+    {
+        EditState = DataGridEditState.New;
+    }
+
+    /// <summary>
+    /// Sets the DataGrid to the EditState.Edit & provides the edit item.
+    /// </summary>
+    /// <param name="editItem"></param>
+    public void SetEdit( TItem editItem )
```
stsrki, line 65: Rename to ``SetEditState()`?`

Source/Extensions/Blazorise.DataGrid/DataGridState.cs
```diff
@@ -0,0 +1,161 @@
...
+    {
+        return ( fieldGetter.Body as MemberExpression ).Member.Name;
+    }
+}
+
+/// <summary>
+/// A DataGrid column sort state container.
+/// </summary>
+/// <typeparam name="TItem"></typeparam>
+public class DataGridColumnSortState<TItem>
```
stsrki, line 133: Move to a separate DataGridColumnSortState.cs file.

Source/Extensions/Blazorise.DataGrid/DataGridState.cs
```diff
@@ -0,0 +1,161 @@
...
+    public string FieldName { get; }
+
+    public SortDirection SortDirection { get; }
+}
+
+/// <summary>
+/// A DataGrid column filter state container.
+/// </summary>
+/// <typeparam name="TItem"></typeparam>
+public class DataGridColumnFilterState<TItem>
```
stsrki, line 150: Move to a separate DataGridColumnFilterState.cs file.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -1,4 +1,5 @@
 ﻿#region Using directives
+
```
stsrki, line 2: Remove extra line spaces. I can also see some bellow in the file.

stsrki (commented):
Overall it looks like a really simple and nice approach. I like it.

Source/Extensions/Blazorise.DataGrid/DataGridState.cs
David-Moreira: It's not a "new state" per say. it's the `EditState` that's set to new. `SetEditModeNew`? `SetEditStateNew`?
Or do you think the semantics wouldn't matter and better keep it short as you say : "`SetNewState`"?

stsrki: It all depends on the context. Here in `SetNew` we set the edit state to `New`. But then, in `SetEdit` method we set the state AND the `editItem`.

We could name them both `SetNewState` and `SetEditItem`, but it would be strange as they both have similar effects, just slight differences. Naming them `SetNewState`, and `SetEditState` would indicate what exactly they do.

David-Moreira: More or less agree with you, but no point in arguing further, I think it's fine, I'll just make the suggested changes.
Thanks.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
David-Moreira: I think it's okay, but if you could take the time to pull the branch and format using your IDE just to be sure? Thanks!

stsrki (approved):
Looks and works perfectly. Very good job on this one.

If you don't think of adding anything else I think I will proceed and merge it.

stsrki:
> If you don't think of adding anything else I think I will proceed and merge it.

@David-Moreira ?

David-Moreira:
> > If you don't think of adding anything else I think I will proceed and merge it.
> 
> @David-Moreira ?

Please do. Thanks.

Issue #5223, opened 2024-01-13
[Bug]: Bulma Carousel Slide Transition Animation is broken


David-Moreira:
### Blazorise Version

1.4.0

### What Blazorise provider are you running on?

None

### Link to minimal reproduction, or a simple code snippet

https://bulmademo.blazorise.com/tests/carousel

### Steps to reproduce

Visit the carousel page. 
Select Previous or Next Slide

### What is expected?

.

### What is actually happening?

 No animation when slides transition.

David-Moreira:
@stsrki Can you confirm whether this is indeed a bug? I'm guessing we've had animation in bulma in the past? Or was it always like this?

stsrki:
I can see the transition animation in Bulma CSS, so I would assume it worked in the past.

David-Moreira:
You said something about next bulma version. So let's leave it for now?

stsrki:
Closing now as we plan to update to Bulma 1.0.

Issue #5222, opened 2024-01-13
[Bug]: AntDesign is currently broken


David-Moreira:
### Blazorise Version

1.4.0

### What Blazorise provider are you running on?

AntDesign

### Link to minimal reproduction, or a simple code snippet

https://antdesigndemo.blazorise.com/

### Steps to reproduce

Just visit the page. Error should be in the console.

### What is actually happening?

![image](https://github.com/Megabit/Blazorise/assets/22283161/e9131fa3-fd94-418f-a0c6-3fd36c515b2f)

### What browsers are you seeing the problem on?

Microsoft Edge

PR #5225, opened 2024-01-13
Carousel | Improve previous/next slide transition responsiveness

David-Moreira:
From my tests, this greatly improves responsiveness when switching between slides. The problem was that we were only setting the Animation to being ended after 2s. I'm not sure why we decided on that, seems like an oversight.

Notes:
- **AntDesign** is blocked from testing due to https://github.com/Megabit/Blazorise/issues/5222
- **Bulma** seems to have no animations, I'm not sure if it's intended or not, but it doesn't affect the viability of this PR. For now we made it so it considers bulma as having no animation, so it is very snappy when switching between slides with no animation.
  - https://github.com/Megabit/Blazorise/issues/5223

### DIFF
--- a/Source/Blazorise.Bulma/CarouselSlide.cs
+++ b/Source/Blazorise.Bulma/CarouselSlide.cs
@@ -1,0 +1,12 @@
+﻿#region Using directives
+#endregion
+
+namespace Blazorise.Bulma;
+
+public class CarouselSlide : Blazorise.CarouselSlide
+{
+    /// <summary>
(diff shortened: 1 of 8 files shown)


stsrki:
That's a strange error for AntD, indeed.

Regarding the Bulma animation. Don't bother with it for now. Bulma is going to release v1.0 in the coming months, and hopefully, we will be part of the beta testers. So, the plan is to adjust everything for that version.

stsrki (approved):

PR #5232, opened 2024-01-14
DataGrid: Auto Generate Columns

David-Moreira:


David-Moreira:
For now we are generating the columns for **value** type & **string** properties.
- `Caption` based on the `Display` or `DisplayName` attributes
- `Field` based on the property name. 
 
I believe this is a good starting point for the most part.

I'm not seeing any great mappings to do out of the **DataAnnotations** model, I was thinking... in order for us to give more flexibility, we would introduce our own `DataGridColumnAttribute` that would be used with all the relevant properties, this would give way more flexibility in my opinion and we could probably expose most of the `DataGridColumn` Parameters on it easily, order, editable, filter mode, etc...

What do you think @stsrki ?

Shared/Blazorise.Shared/Models/Employee.cs
```diff
@@ -28,24 +28,41 @@ public Employee( Employee other )
         Zip = other.Zip;
     }
 
+    [Display( Name = "Id" )]
     public int Id { get; set; }
 
     [Required]
+    [Display( Name = "First Name" )]
```
stsrki, line 35: What do you think we generate `Caption` from the field name if `DisplayAttribute` is not defined?

Example:

- `FirstName` to `"First Name"`
- `DateOfBirth` to `"DateOfBirth"`
- `City` to `"City"`

Source/Blazorise/Utilities/ReflectionHelper.cs
```diff
@@ -0,0 +1,44 @@
...
+            return displayAttribute.GetName();
+        }
+
+        var displayNameAttribute = propertyInfo.GetCustomAttribute<DisplayNameAttribute>();
+        if ( displayNameAttribute is not null )
+        {
+            return displayNameAttribute.DisplayName;
+        }
+
+        return propertyInfo.Name;
```
stsrki, line 42: Oh now I see that you return property name as a default. That leaves it only to properly format Title cased name to include spaces as a separator.

stsrki (commented):
It's hard to see what exactly you have changed in the DataGrid.razor.cs. GitHub diff have problems with comparing changes. Did you included some kind of file-wide format or is there really so many changes in this PR?

Shared/Blazorise.Shared/Models/Employee.cs
David-Moreira: We take in the property name already. But I'm guessing you gave the example for the Date Of Birth incorrectly?
It should be `DateOfBirth` to "`DateOfBirth`" 
Was that what you mean? Well, us trying to guess will only work correctly if people use Properties with PascalCase... 
Then what if they don't want our smartass generation? Then they are stuck with names they don't want.

Source/Blazorise/Utilities/ReflectionHelper.cs
David-Moreira: Right, see related comment above. Us trying to be smart about it, relies on people using the PascalCase convention and not being able to opt out if we make it default.

David-Moreira:
> It's hard to see what exactly you have changed in the DataGrid.razor.cs. GitHub diff have problems with comparing changes. Did you included some kind of file-wide format or is there really so many changes in this PR?

I usually do diffs before comitting, I don't think I saw a bunch of them... Must have been to do with end of line stuff... so annoying...

David-Moreira:
> > It's hard to see what exactly you have changed in the DataGrid.razor.cs. GitHub diff have problems with comparing changes. Did you included some kind of file-wide format or is there really so many changes in this PR?
> 
> I usually do diffs before comitting, I don't think I saw a bunch of them... Must have been to do with end of line stuff... so annoying...

I'll try to fix it later...

David-Moreira:
> > > It's hard to see what exactly you have changed in the DataGrid.razor.cs. GitHub diff have problems with comparing changes. Did you included some kind of file-wide format or is there really so many changes in this PR?
> > 
> > 
> > I usually do diffs before comitting, I don't think I saw a bunch of them... Must have been to do with end of line stuff... so annoying...
> 
> I'll try to fix it later...

@stsrki  Fixed.

Source/Blazorise/Utilities/ReflectionHelper.cs
stsrki: The default naming convention in C# is PascalCase, according to the [docs](https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/identifier-names#naming-conventions). And people can always use `DisplayAttribute` if they don't like it.

Source/Extensions/Blazorise.DataGrid/DataGridColumn.cs
```diff
@@ -39,26 +41,39 @@ public DataGridColumn()
...
-    protected override void OnInitialized()
+    /// <summary>
+    /// Provides a way to generate column specific dependencies outside the Blazor engine.
+    /// </summary>
+    /// <param name="parentDataGrid"></param>
+    /// <param name="serviceProvider"></param>
+    internal void InitializeGeneratedColumn( DataGrid<TItem> parentDataGrid, IServiceProvider serviceProvider )
     {
-        base.OnInitialized();
+        this.ParentDataGrid = parentDataGrid;
```
stsrki, line 63: I guess this is the reason for `SuppressMessage` settings? Can we make it on a method level instead of making it a global setting? Easier to see it in the future if we directly set properties somewhere else.

Source/Extensions/Blazorise.DataGrid/DataGridColumn.cs
```diff
@@ -39,26 +41,39 @@ public DataGridColumn()
...
 
     #region Methods
 
-    protected override void OnInitialized()
+    /// <summary>
+    /// Provides a way to generate column specific dependencies outside the Blazor engine.
+    /// </summary>
+    /// <param name="parentDataGrid"></param>
+    /// <param name="serviceProvider"></param>
+    internal void InitializeGeneratedColumn( DataGrid<TItem> parentDataGrid, IServiceProvider serviceProvider )
```
stsrki, line 61: Name it `InitializeAutoGeneratedColumn`?

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -384,6 +395,27 @@ protected override async Task OnAfterRenderAsync( bool firstRender )
...
+
+            var column = new DataGridColumn<TItem>
+            {
+                Caption = ReflectionHelper.ResolveCaption( property ),
+                Field = property.Name,
+            };
+            this.AddColumn( column );
+        }
+
+        StateHasChanged();
```
stsrki, line 416: `InvokeAsync(StateHasChanged)`

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -384,6 +395,27 @@ protected override async Task OnAfterRenderAsync( bool firstRender )
...
+        {
+            if ( !( property.PropertyType.IsValueType || property.PropertyType == typeof( string ) ) )
+            {
+                continue;
+            }
+
+            var column = new DataGridColumn<TItem>
+            {
+                Caption = ReflectionHelper.ResolveCaption( property ),
+                Field = property.Name,
```
stsrki, line 411: What else might we get from the property?

-` [Required]` attribute?
- Enum that generates the Select with options?

Source/Extensions/Blazorise.DataGrid/DataGridColumn.cs
David-Moreira: I like the supression file alot. It keeps all the supression in a single file and easily referable to with justification. You should not need to be "poluting" code with it.

There is no issue with the line you are linking. It is being set in the component itself.

The reason is for the setters on `Parameters` when newing up a Blazor component in the `Datagrid.AutoGenerateColumns`.
image

But let me know if you still prefer to inline it.

Source/Extensions/Blazorise.DataGrid/DataGridColumn.cs
David-Moreira: hmm I didn't name it "Auto" because it can be a column added by user even. Now users should be able to add the columns programatically. They could not before, there were initializers missing.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
David-Moreira: It's redudant the thread is synced up here. Specially since it isn't async.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
David-Moreira: >  [Required] attribute? 
I'm blanking out here, what can we set in the Column with the Required attribute?

> Enum that generates the Select with options?
What's your idea with Enum? If an Enum is found we figure out the Enum members and generate the Select, that's it right? No need for Attributes here, right?

Source/Extensions/Blazorise.DataGrid/DataGridColumn.cs
stsrki: What do you mean? Isn't the method marked as `internal`?

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
stsrki: > I'm blanking out here, what can we set in the Column with the Required attribute?
> 
> >

Basically, validation.

Source/Extensions/Blazorise.DataGrid/DataGridColumn.cs
David-Moreira: I mean, the method should be invoked whenever a generated column is detected. Here, we are refering to generated column as a column that was created imperatively through code instead of declaretevly through the razor code.

So whenever a user calls `DataGrid.AddColumn` it should end up calling this method also, that's why I wanted it to be a more generic name not tied directly to our `AutoGenerateColumns` feature.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
David-Moreira: > > I'm blanking out here, what can we set in the Column with the Required attribute?
> > >
> 
> Basically, validation.

I still don't understand what you want to do here, there's no validation atributes that we set specifically on the column is there? The `DataAnnotations` validation should work automatically if the `DataGrid` is set to validate, we've talked about this.

stsrki: You're right. I completely forgot that we mentioned it last time.

Source/Extensions/Blazorise.DataGrid/DataGridColumn.cs
stsrki: OK, we can leave at as it is.

stsrki (commented):
Looks good to me.

What do you think we add custom attributes for auto generated fields? Or do we wait until people request them?

David-Moreira:
- What do you think we add custom attributes for auto generated fields?

I already gave my suggestion for that. At least these attributes we know of, and we don't have to imagine user scenarios. 
image

Otherwise I'm fine leaving it like this, and wait for user requests.

stsrki:
It would be good to introduce some new attributes. But we need to take more time to consider which ones are needed. And we can also make them in the core project to reuse them in some other places if required. Dynamic Forms comes to mind.

OK, we leave it as it is for now. I will open a new planning ticket for the future.

Issue #5248, opened 2024-01-20
[2.0] Strip all logic that uses Collapse within the Accordion


stsrki:
## Reason

Using the Collapse within accordion is leaking too much of logic between components that are not that directly connected. Since we have added dedicated AccordionItem, AccordionHeader, and AccordionBody there is no need for using Collapse in the Accordion anymore.

tesar-tech:
So this means removing the `Collapse` component from the codebase entirely, right?

It seems there is no other usage for it apart from being part of the `Accordion`, which will no longer be the case in 2.0.

Or what's the reason for keeping `Collapse` component?

stsrki:
We want to leave Collapse component.It still has it's use cases. 

But any usage WITHIN the accordion needs to be removed.

PR #5249, opened 2024-01-20
Refactor Accordion components

### LINKED ISSUE #5247: New Accordion subcomponents
Instead of using the Collapse component for the accordion structure, we need to introduce new Accordion specific components

- AccordionItem
- AccordionHeader
- AccordionBody

stsrki:
Closes #5247

Introduced new Accordion subcomponets

- `AccordionItem`, and
- `AccordionHeader`, and
- `AccordionBody`

Reason is that using the Collapse was leaking too much of logic between accordion and collapse, which are not that much directly connected. This PR should solve this because we now have dedicated component just for accordion. In the future we should strip collapse from accordion completely #5248.


Source/Blazorise/Components/Accordion/AccordionToggle.razor.cs
```diff
@@ -41,6 +43,11 @@ protected async Task ClickHandler( MouseEventArgs eventArgs )
             await ParentCollapse.Toggle();
         }
 
+        if ( ParentAccordionItem is not null )
```
David-Moreira, line 46: If might be a little more correct to do an else if. Although it's one of those conditions where you already expect only one of them is to be executed.

Source/Blazorise.FluentUI2/Components/AccordionToggle.razor
```diff
@@ -6,5 +6,5 @@
     @ChildContent
 </button>
 @code {
-    string AriaExpandedString => CollapseVisible ? "true" : "false";
+    string AriaExpandedString => AccordionItemVisible ? "true" : "false";
```
David-Moreira, line 9: Do we not need to check for Collapse?

Source/Blazorise/Components/Accordion/AccordionToggle.razor.cs
```diff
@@ -76,6 +88,24 @@ public bool CollapseVisible
         }
     }
 
+    /// <summary>
+    /// Gets or sets the content visibility.
+    /// </summary>
+    [CascadingParameter( Name = "AccordionItemVisible" )]
+    public bool AccordionItemVisible
```
David-Moreira, line 95: Didn't we use a state mechanism for this things before?

Source/Blazorise/Components/Accordion/Accordion.razor.cs
```diff
@@ -56,6 +58,35 @@ internal bool IsLastInAccordion( Collapse collapse )
...
+    {
+        if ( accordionItems is not null && accordionItems.IndexOf( accordionItem ) == 0 )
+            return true;
+
+        return false;
+    }
+
+    internal bool IsLastInAccordion( AccordionItem accordionItem )
+    {
+        if ( accordionItems is not null && accordionItems.IndexOf( accordionItem ) == accordionItems.Count - 1 )
```
David-Moreira, line 84: I doubt it, but could this have a dumb bug like:
accordionItems being empty, 
which means  
**accordionItems.IndexOf( accordionItem )** => **-1**
**accordionItems.Count - 1** would also lead to **-1**

Source/Blazorise.Material/MaterialClassProvider.cs
```diff
@@ -37,5 +37,9 @@ public class MaterialClassProvider : Bootstrap.BootstrapClassProvider
 
     public override string BarItemHasDropdown( BarMode mode ) => mode == Blazorise.BarMode.Horizontal ? "dropdown" : null;
 
+    public override string AccordionToggle() => "btn btn-link btn-block text-left";
+
+    public override string AccordionToggleCollapsed( bool collapsed ) => collapsed ? null : "collapsed";
```
David-Moreira, line 42: is this right? If collapsed null otherwise "collapsed"?

David-Moreira (approved):
LGTM, more coherent api.

Source/Blazorise/Components/Accordion/Accordion.razor.cs
stsrki: If the accordionItems are empty, then it is already invalid. I would leave it like this. I don't want to introduce another bug by accident. If anyone complains in the future, we will change it.

Source/Blazorise.Material/MaterialClassProvider.cs
stsrki: It is. We send it a `Visible`  parameter, and in here we name it `collapsed`. It's inverted so there's a confusion.

Source/Blazorise/Components/Accordion/AccordionToggle.razor.cs
stsrki: For Accordion no. We keep it simple until we don't :)

PR #5277, opened 2024-01-31
DataGrid : Column Preventrowclick does not work anymore

David-Moreira:
It seems we have removed mistakenly the stopPropagation attribute before when implementing a new feature.

image

Razor components do not support having both `@onlick`, `@onlick:stopPropagation` so you need to use an actual **Parameter** and set it on the html element.

We would be introducing two new Parameters to `TableRowCell` in a patch. Not ideal, Should we move to 1.5? Or take the added change? @stsrki ?


stsrki (commented):
Better to move it to v1.5 since we're introducing new APIs.

PR #5278, opened 2024-02-01
DataGrid : Column Preventrowclick does not work anymore 

David-Moreira:
It seems we have removed mistakenly the stopPropagation attribute before when implementing a new feature.

Razor components do not support having both @onlick, @onlick:stopPropagation so you need to use an actual Parameter and set it on the html element.

### DIFF
--- a/Source/Blazorise/Components/Table/TableRowCell.razor.cs
+++ b/Source/Blazorise/Components/Table/TableRowCell.razor.cs
@@ -146,5 +146,15 @@ public partial class TableRowCell : BaseDraggableComponent
     /// </summary>
     [Parameter] public RenderFragment ChildContent { get; set; }
 
+    /// <summary>
+    /// Used to prevent the default action for an <see cref="OnClickHandler"/> event.
+    /// </summary>
+    [Parameter] public bool ClickPreventDefault { get; set; }
+
+    /// <summary>
+    /// Used to stop progation of the click action event.
+    /// </summary>
+    [Parameter] public bool ClickStopPropagation { get; set; }
+
     #endregion
 }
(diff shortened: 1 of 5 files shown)


Source/Blazorise/Components/Table/TableRowCell.razor.cs
```diff
@@ -146,5 +146,15 @@ public Color Color
     /// </summary>
     [Parameter] public RenderFragment ChildContent { get; set; }
 
+    /// <summary>
+    /// Used to prevent the default action for an <see cref="OnClickHandler"/> event.
+    /// </summary>
+    [Parameter] public bool ClickPreventDefault { get; set; }
```
stsrki, line 152: Naming seems somewhat weird. But I can't find a better alternative.

Can you just add the new API to the documentation?

David-Moreira: I personally don't think it does.
Maybe `OnClickPreventDefault`. But I'm fine with either.

David-Moreira: Oh I guess your concern is that this is kind of a js nomeanclature, the prevent default? I'd just keep it anyway, most front end devs will most likelly interpret it correctly, if we try to come up with a new one it might not be a good idea.

David-Moreira: Whoever does not understand it, probably has no need for it anyway. hehe

stsrki (commented):
Please just update the docs before I merge it. Thanks

David-Moreira:
> Please just update the docs before I merge it. Thanks

Done

stsrki:
> > Please just update the docs before I merge it. Thanks
> 
> Done

Release notes aro greater, but I meant the mention in the table page API section 😂😅

David-Moreira:
> > > Please just update the docs before I merge it. Thanks
> > 
> > 
> > Done
> 
> Release notes are greate, but I meant the mention in the table page API section 😂😅

ah, done.

stsrki (approved):

PR #5289, opened 2024-02-04
Blazorise Captcha component

David-Moreira:
Added support for **google reCAPTCHA** based on the v2 version which is the one I think we are most accustomed to. There's a newer **v3** version, which seems to be made to be more hidden on the page? I'm not sure what exactly is the difference, I have never used that one, I'll explore it.

The v2 version we implemented should support both the **checkbox & invisible** versions from the grecaptcha docs by setting the appropriate recaptcha Size option (normal,compact,invisible).
https://developers.google.com/recaptcha/docs/display
https://developers.google.com/recaptcha/docs/invisible
image

If the captcha is set as checkbox, the user has to click a checkbox.
If the captcha is set as invisible, it's up to the dev to call the Submit operation which will show a captcha puzzle at times (click all cars, buses, etc...) if it's not sure its a real user (less intrusive)


stsrki:
Just a note on the usage API.

We need to have `Captcha.razor` as that one will be used by the user. Users should not use the implementation component directly.

**DO:**

```razor
<Captcha />
```

**DONT:**

```razor
<ReCaptcha />
```

With the Dependency Injection, every provider will register their own implementation. In the case of reCAPTCHA this will be done inside of `.AddGoogleReCAPTCHA()`.

```cs
serviceCollection.AddTransient( typeof( Blazorise.Captcha.Captcha ), typeof( Blazorise.Captcha.reCAPTCHA.reCAPTCHA );
```

David-Moreira:
**PS:** we just got a notification by GH about the exposed server side secret key. It's just a test key of mine that I will delete, it's fine.

Demos/Blazorise.Demo.Bootstrap.Server/Pages/_Host.cshtml
```diff
@@ -41,6 +43,7 @@
 
     <script src="_content/Blazorise.Animate/blazorise.animate.js?v=1.4.2.0"></script>
     <script src="_content/Blazorise.Demo/demo.js?v=1.4.2.0"></script>
+    <script src="_content/Blazorise.Captcha.reCAPTCHA/recaptcha.js?v=1.4.2.0"></script>
```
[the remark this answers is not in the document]
David-Moreira: It's WIP I'm still developing the surface api and doing tests, that was just faster for me. I'll start updating to blazorise conventions soon.

On another note, what are your thoughts on the naming? reCAPTCHA is the actual name of the product but that isn't PascalCase like we usually name projects, namespaces, etc..

Source/Extensions/Blazorise.Captcha.reCAPTCHA/ReCAPTCHA.razor
stsrki, line 1: Don't forget to also rename files, folders, and project names.

David-Moreira: Gotcha, I'll check

David-Moreira:
@stsrki You can do some reviewing now please.

**The PR description has been updated.**

---
**Notes:**
We have an `OnExpired` event which is kind of based on the grecaptcha expired event, I'm not entirely sure it makes sense to generically have it, but maybe it does since I guess solving a captcha should ideally always be valid for a limited time.

We have a `OnValidate` event which we could ommit, and the user would do any validations on the `OnSolved` event instead? What do you think?

---
**TODO:**
- Docs & Examples
- Please create a new blazorise v2 checkbox key in advance so we can showcase on an example on live docs
  - You can create a separate one in advance for our captcha solvers in the blazorise site.
  https://www.google.com/recaptcha/admin/create (You can create a classic key, should be enough for our needs) any doubts, contact me on discord.

stsrki:
If we're implementing v2 and the new v3 is on the way maybe we should rename our current to include the version number, eg `Blazorise.Captcha.ReCaptcha2`?

Source/Extensions/Blazorise.Captcha.ReCaptcha/Enums/ReCaptchaBadge.cs
```diff
@@ -0,0 +1,24 @@
+﻿namespace Blazorise.Captcha.ReCaptcha;
+
+/// <summary>
+/// Reposition the reCAPTCHA badge. 'inline' lets you position it with CSS.
+/// <para>This option is only available for invisible reCAPTCHA.</para>
+/// </summary>
+public enum ReCaptchaBadge
```
stsrki, line 7: Rename to `ReCaptchaBadgePosition`.

Source/Extensions/Blazorise.Captcha.ReCaptcha/Enums/ReCaptchaBadge.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    /// <summary>
+    /// Repositions the reCAPTCHA badge to the bottom right.
+    /// </summary>
+    BottomRight,
```
stsrki, line 12: Rename to `BottomStart`.

Source/Extensions/Blazorise.Captcha.ReCaptcha/Enums/ReCaptchaBadge.cs
```diff
@@ -0,0 +1,24 @@
...
+
+    /// <summary>
+    /// Repositions the reCAPTCHA badge to the bottom left.
+    /// </summary>
+    BottomLeft,
```
stsrki, line 17: Rename to `BottomEnd`.

Source/Extensions/Blazorise.Captcha/Captcha.razor.cs
```diff
@@ -0,0 +1,107 @@
...
+    /// The Captcha expired event.
+    /// </summary>
+    [Parameter] public EventCallback OnExpired { get; set; }
+
+
+    protected override async Task OnAfterRenderAsync( bool firstRender )
+    {
+        if ( firstRender )
+        {
+            await Render();
```
stsrki, line 35: I think this should be called `InitializeCaptcha` or similar. Not `Render`.

Source/Extensions/Blazorise.Captcha/Captcha.razor.cs
```diff
@@ -0,0 +1,107 @@
...
+    /// <summary>
+    /// The current state of the Captcha.
+    /// </summary>
+    public CaptchaState State { get; private set; } = new();
+
+    /// <summary>
+    /// A Captcha solved event. 
+    /// Provides contextual information about the Captcha state after the user has solved.
+    /// </summary>
+    [Parameter] public EventCallback<CaptchaState> OnSolved { get; set; }
```
stsrki, line 18: We usually don't have `On` prefixes. Just name them in past tense, `Validated`, `Expired`.

stsrki (commented):
Code-wise it looks really good.

Source/Extensions/Blazorise.Captcha/Captcha.razor.cs
David-Moreira: Well, it does renders the captcha in practice... hehe

Source/Extensions/Blazorise.Captcha/Captcha.razor.cs
David-Moreira: `Validated`? Maybe `Validating`? `Validated` makes me think it's done validating. And the purpose is for the user to apply validation. Anyway refer to : 
> We have a `OnValidate` event which we could ommit, and the user would do any validations on the OnSolved event instead? What do you think?

Maybe we should just remove the validate event, and if the user needs to do validation he does it on the `Solved` event?

stsrki: If it's an ongoing validation, then I would name it `Validate`.

David-Moreira:
@stsrki 
I have just tried v3. It's pretty similar to how v2 works.
So by just using a v3 key, I was able to reuse our implementation successfully. 


**What changes with v3**:
- instead of a puzzle, the recaptcha evaluates the user interactions and gives a score representing how likely the user is a bot, it's up to the developer to interpret the score and proceed accordingly... i.e low score, ask for 2fa, etc...

This score is sent back when the developer does the backend validation, which isn't our concern. Up to the developer to implement the backend validation using our Validate event.
Example server validation (again I'll invalidate the keys once I'm done) :  
image

- Example response for v3:
image
image

- v3 also has a new action parameter which just serves as an identifier, which then is returned back on the server validation callback. We've used the captcha Id for this action. So the user can just use the Id to whatever he wants if he really wants to propagate the action. I'd guess most cases he won't need to though.

So I think we don't need to bother with making a separate project.

For our purposes I'm thinking we'll be using the v2 checkbox for our "bot validations".

stsrki:
That's great. We can then leave it as it is now. If we want to fully support v3, then we can introduce a few APIs where needed. And we can mention in the documentation the differences.

David-Moreira:
@stsrki I'm finished with the feature. Please review.

stsrki (approved):
Works great, and the API is easy to follow.

I only made some small changes to formating and optimized docs a bit.

If you don't need to add anything else I think it is ready for merge.

David-Moreira:
Oh yea, It's fine by me.
Thanks for taking the time to cleanup & format.

stsrki:
Thank you for taking time to create such a wonderful peace of work :)

PR #5293, opened 2024-02-07
Rating: add tooltip for the items

### LINKED ISSUE #5292: Rating: add tooltip for the items
This will make it easy for the user to indicate the right score

**Example**

image

stsrki:
Closes #5292


Source/Blazorise/Models/RatingItemTooltipInfo.cs
```diff
@@ -0,0 +1,63 @@
+﻿namespace Blazorise;
+
+/// <summary>
+/// Defines the tooltip information for the <see cref="Rating"/> component.
+/// </summary>
+public class RatingItemTooltipInfo
```
David-Moreira, line 6: What about **RatingItemTooltip** You are defining how the actual tooltip gets constructed not just the "info".

Source/Blazorise/Models/RatingItemTooltipInfo.cs
```diff
@@ -0,0 +1,63 @@
+﻿namespace Blazorise;
+
+/// <summary>
+/// Defines the tooltip information for the <see cref="Rating"/> component.
```
David-Moreira, line 4: Is it for the `Rating` component or the `RatingItem` component?

Source/Blazorise/Components/Rating/RatingItem.razor
```diff
@@ -1,14 +1,33 @@
 ﻿@namespace Blazorise
 @inherits BaseComponent
-@if ( Rating.ReadOnly )
+@{
+    var tooltipInfo = Rating?.GetItemTooltipInfo?.Invoke( Value );
+}
+
+@if ( tooltipInfo is null || string.IsNullOrEmpty( tooltipInfo.Text ) )
```
David-Moreira, line 7: can probably do `string.IsNullOrEmpty( tooltipInfo?.Text )`. Not a bother.

Source/Blazorise/Components/Rating/Rating.razor.cs
```diff
@@ -225,5 +226,10 @@ public int SelectedValue
     /// </summary>
     [Parameter] public EventCallback<int?> HoveredValueChanged { get; set; }
 
+    /// <summary>
+    /// Expression that allows the rating item to have additional user-friendly information.
+    /// </summary>
+    [Parameter] public Func<int?, RatingItemTooltipInfo> GetItemTooltipInfo { get; set; }
```
David-Moreira, line 232: Maybe just GetItemTooltip?

David-Moreira: Is there a Rating Item that's represented by null? That's why we have nullable int? Would we have an actual tooltip render when the value is null?

David-Moreira: Also is it ok to use `Get` prefix for `Func` from a coding style perspective? I'm not sure we use that approach, but we do seem pretty lax about it across the code base. 
Looking at some of them, could it perhaps be `ItemTooltipCreator`?

Source/Blazorise/Models/RatingItemTooltipInfo.cs
stsrki: Renamed it to `RatingTooltip`. We already don't actively expose `RatingItem`, only the `Rating` component. So this makes the most sense.

Source/Blazorise/Components/Rating/Rating.razor.cs
stsrki: I don't think `Creator` makes sense. This is just a delegate. So from a delegate perspective the `Get` prefix makes the most sense. And we already have that naming in other places, TreeView, for example.

Source/Blazorise/Components/Rating/Rating.razor.cs
```diff
@@ -225,5 +226,13 @@ public int SelectedValue
     /// </summary>
     [Parameter] public EventCallback<int?> HoveredValueChanged { get; set; }
 
+    /// <summary>
+    /// Expression that provides the tooltip for the rating item to have additional user-friendly information.
+    /// </summary>
+    /// <remarks>
+    /// The expression receives the rating item <Code>Value</Code> to determine the result.
+    /// </remarks>
+    [Parameter] public Func<int?, RatingTooltip> GetTooltip { get; set; }
```
David-Moreira, line 235: Is there a Rating Item that's represented by null? That's why we have nullable int? Would we have an actual tooltip render when the value is null? What happens?

David-Moreira (approved):
LGTM

Source/Blazorise/Components/Rating/Rating.razor.cs
David-Moreira: Made*? :P

PR #5297, opened 2024-02-09
Vulnarable references | DeepCloner dependency imbue in Blazorise

David-Moreira:
Like we've talked. We went ahead and made a new `DeepCloner` folder and copied the `DeepCloner` implementation.

Code should remain mostly intact for now and as it was implemented by the original authors. I have to say I don't understand half of the code hehe.

**Changes we've made specifically to their original files:**
- Removed any files or code that had predecessor directives to point to non dotnet core versions
- File Scoped namespaces
- Added a ReadMe with proper explanations of the reasoning behind the copy and proper info about where the implementation came from
- Added a mention on the **public** class `DeepClonerExtensions` mentioning that it is originally based on 
https://github.com/force-net/DeepCloner

Every test is still passing, including the dedicated tests we had for `DeepCloner`.
Also briefly tested DataGrid Edit. Anyway no reason why it would not work unless we've made a mistake copying or removing code. :)

### DIFF
--- a/Source/Blazorise/Blazorise.csproj
+++ b/Source/Blazorise/Blazorise.csproj
@@ -42,7 +42,6 @@
 
   <ItemGroup>
     <PackageReference Include="Blazorise.Licensing" Version="1.2.0" />
-    <PackageReference Include="DeepCloner" Version="0.10.4" />
   </ItemGroup>
 
   <PropertyGroup>
--- a/Source/Blazorise/DeepCloner/DeepClonerExtensions.cs
+++ b/Source/Blazorise/DeepCloner/DeepClonerExtensions.cs
@@ -1,0 +1,75 @@
+﻿using System;
+using System.Security;
+using Blazorise.DeepCloner.Helpers;
+
+namespace Blazorise;
+
+/// <summary>
+/// Extensions for object cloning
+/// </summary>
+public static class DeepClonerExtensions
+{
+    /// <summary>
+    /// Performs deep (full) copy of object and related graph
+    /// </summary>
+    public static T DeepClone<T>( this T obj )
+    {
+        return DeepClonerGenerator.CloneObject( obj );
+    }
+
+    /// <summary>
+    /// Performs deep (full) copy of object and related graph to existing object
+    /// </summary>
+    /// <returns>existing filled object</returns>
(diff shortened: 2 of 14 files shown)


David-Moreira:
@stsrki are we fine with having this new extensions available under `Blazorise` or should we move them specifically under `DeepCloner`?
image

I know we allocate most of our stuff under `Blazorise`, but maybe I'd keep these under `DeepCloner` since they are not very usual to use, what are your thoughts?

stsrki:
Now that I'm looking through the code, I realized several issues. 

1. First, you already mentioned. What namespace to choose? I think since this is a patch, it should leave the original name, and we can change it in the 1.5 to Blazorise.DeepCloner
2. Their codebase doesn't follow our guidelines for naming, formatting, etc. There is probably a warning unless we suppress them or change namings and formatting.
3. Should we move the DeepCloent folder under the Vendors folder? Feels strange to keep at the root level,

Or we can go back to our original plan, fork the repo, and release it under our own Nuget package.

David-Moreira:
> Or we can go back to our original plan, fork the repo, and release it under our own Nuget package.

It's still the same. We get a new `Blazorise.DeepCloner` namespace. Since it's just a few files and don't seem to add complexity / time to the build process I'm fine with having it live in a folder.

As for your concerns, I will be moving this to master if you don't want to add a new namespace `Blazorise.DeepCloner` in a patch. Which is effectly true, that's introducing a new api for all intents and purposes.

As for your concerns regarding the formatting, I did not see any warning actually? Does your IDE throw off any warnings? Can you check? Anyway it's just formatting we can reformat if we really want to. But from the brief look I had in the files I didn't see any code that wasn't following regular C# code conventions.

David-Moreira:
> As for your concerns, I will be moving this to master if you don't want to add a new namespace Blazorise.DeepCloner in a patch. Which is effectly true, that's introducing a new api for all intents and purposes.

I was thinking about this, we could argue that it's not introducing a new api per say, because it already existed under the previous package namespace. The api is switching namespaces.
But if for some reason anyone used the underlying deepcloner through our package then it is a breaking change, because the original package will no longer exist. 
I'm also up for moving this to master definitely.

stsrki:
I think keeping it under the `Blazorise.DeepCloner` and targeting master is the best approach.

PR #5304, opened 2024-02-10
ModalProvider | Stateful feature

David-Moreira:
Initially I was basing myself on guessing whether a modal would be kept in memory by using the `RenderMode`.

If **RenderMode** was **Default** Or **LazyLoad**, the modal would be kept in memory when closed.

This would make it so users currently using `ModalProvider` on their projects could unknowingly have alot of instances in memory. 

As such I decided to introduce a new `Stateful` parameter, this makes it so users looking for this feature will opt in explicitly to it and be more mindful that instances are kept in memory and they might need to manage them if they plan to have alot of modals across the application.

In order for a user to correctly open the same instance as before, either he sets the `ElementId`, or uses the reference returned by the `ModalService.Show` to reopen the same instance. This has been explained in the docs. And there is a new example in the Demo and in the Docs to showcase the feature in action.

Since now instances can remain in memory, new methods have been provided that might give further flexibility for users to get creative with it:
- GetInstances
- Remove
- Reset


stsrki (approved):
Looking good.

PR #5309, opened 2024-02-13
Re-enable build-nuget.yml | Nightly builds

David-Moreira:
For us to enable the automatic build. 

Let's use this PR to make decisions on how to proceed & test.


David-Moreira:
@stsrki 
Adapted the proceed based on our current CI pipeline.

In this case the existing nuget build does not yet support automatic versioning & periodic run.

**Next Steps:**
1. Can we do test runs with some dummy tag for now? What dummy flag can we use?
2. Decide upon the periodic cadence. If we have problems with space, I'm thinking every Week on Friday or Sunday? Or Fortnightly?
3. Decide on Version suffix: 1.5.0-preview-{gh_build_number} or 1.5.0-dev-{gh_build_number} or 1.5.0-{gh_build_number} or whatever else...

David-Moreira:
> We also have a problem with versioning query strings. If we raise assembly version to 1.5.0 we also need raise the version that is used for loading JS modules. E.g. `utilities.js?v=1.5.0.0`, If we leave it to an older query version in runtime, users will have different files loaded.
> 
> Is there any way to override those during the build somehow? Can files be changed dynamically?

I guess that's a different problem, right? I mean they are files, you can probably add a build step that runs some kind of replacement? I will need to do some research, but that would be a separate issue no? Should we open another issue?

Currently when we raise the assembly version, we do it manually, same for those files.

David-Moreira:
@stsrki 
For the JS files, Something like this I'm thinking, it should work I think:
https://itecnote.com/tecnote/r-how-to-replace-string-in-file-using-msbuild/

Anyway separate issue?

For the purposes of this particular task, we'll start incrementing the assembly version right away on master?

stsrki:
It's not a problem to increment the version number. It's a problem when I merge the patch branch back to the master. I got to solve hundreds of conflicts manually. That's why I usually raise the number at the last possible time before release.

stsrki:
> @stsrki For the JS files, Something like this I'm thinking, it should work I think: https://itecnote.com/tecnote/r-how-to-replace-string-in-file-using-msbuild/ (No idea how up to date the post is... But I think csprojs still support this kind of stuff.)
> 
> Anyway separate issue?
> 
> For the purposes of this particular task, we'll start incrementing the assembly version right away on master?

Seems promising. When file is modified is it going to be shown in Git changes? Or can we make it so that only the published files are replaced?

David-Moreira:
> > It's not a problem to increment the version number. It's a problem when I merge the patch branch back to the master. I got to solve hundreds of conflicts manually. That's why I usually raise the number at the last possible time before release.
> 
> What merge strategy are you using? It should not conflict on files that have not been changed. It's probably comparing alot of stuff that it should not.
> 
> I.e: You patch the file DataGrid.razor.cs. when you merge to master, it should not look into csproj version or js files version as these have not changed?

AHHH I guess you are merging in the commits that change the patch numbers? Any way these could not be merged at all?

David-Moreira:
> > @stsrki For the JS files, Something like this I'm thinking, it should work I think: https://itecnote.com/tecnote/r-how-to-replace-string-in-file-using-msbuild/ (No idea how up to date the post is... But I think csprojs still support this kind of stuff.)
> > Anyway separate issue?
> > For the purposes of this particular task, we'll start incrementing the assembly version right away on master?
> 
> Seems promising. When file is modified is it going to be shown in Git changes? Or can we make it so that only the published files are replaced?

I'm not sure, I'm thinking that it would affect the output files, not the source files.
So the source files would look like 
**js?v=placeholder** ALWAYS
The ouput files (bin, etc...) would have the placeholder replaced with the version number. I am thinking that is it?

stsrki:
> AHHH I guess you are merging in the commits that change the patch numbers? Any way these could not be merged at all?

Yes, I change the patch numbers. Well I can not merge them but I would prefer to have them in the history.

stsrki:
> **js?v=placeholder** ALWAYS
> The ouput files (bin, etc...) would have the placeholder replaced with the version number. I am thinking that is it?

If that is true then I would like it very much. Please investigate.

David-Moreira:
> > AHHH I guess you are merging in the commits that change the patch numbers? Any way these could not be merged at all?
> 
> Yes, I change the patch numbers. Well I can not merge them but I would prefer to have them in the history.

hmm I see... Not sure how I feel about that. So that's the reason why you don't have master on latest version right away... 
So if you patch an older version, i.e : 1.3 and you bring in the fixes, you bring in the 1.3 patch version commit, also & for that one you fix the conflicts maintaining whatever master version? 

If you don't want to change how you merge, then I can only think of setting an env variable with the current master version. But then there's the js files problem indeed... 
Maybe we do need to automate that right away then... Let me think about it.

Source/Blazorise.Bootstrap5/wwwroot/modal.js
```diff
@@ -1,4 +1,4 @@
-import { addClassToBody, removeClassFromBody } from "../Blazorise/utilities.js?v=1.4.2.0";
+import { addClassToBody, removeClassFromBody } from "../Blazorise/utilities.js?v=1.4.2";
```
stsrki, line 1: The version must contain 4 values, ending in 0 in this case.

David-Moreira: Any special reason? I might change the approach anyway, I'm still testing. I just found out that MSBuild supports executing C# code, that might be better because it will be cross platform build-able more easily.

stsrki: Because the version that is sent from Blazorise at runtime must match the version that is used by the JS to load additional JS modules. 


```js
public override string ModuleFileName => $"./_content/Blazorise/closable.js?v={VersionProvider.Version}";
```

And making it always explicitly have 4 values make it easier to handle without introducing additional rules like trimming the end. And find and replace is much easier.

David-Moreira: I see. Just the way I got the version from : 
image
Does not yield the 0. But I can figure out some other symbol with it. Surely there's a way to get the assembly version. 
I'll probably just use C# to get the current assembly version. We'll see.
Thanks.

Build/Blazorise.props
```diff
@@ -61,4 +61,56 @@
...
+
+	<Target Name="BeforeBuildStep" BeforeTargets="Build">
+		<ItemGroup>
+			<JsFiles Include='wwwroot\*.js' />
+		</ItemGroup>
+		
+		<ReplaceVersionPlaceholderInJsFiles
+		   SourceFiles="@(JsFiles)"
+		   MatchExpression="\?v=([\d.]+)"
+		   ReplacementText="?v=$(Version).0" />
```
stsrki, line 115: What will happen when we have a version that `1.4.3.1`? This will convert it to `1.4.3.1.0`,

David-Moreira: This is temporary. I need to get proper assembly version like I've already told before.
image

David-Moreira: Actually it might be best to correctly set the version with 4 digit groups in the project if that is expected.
Although it's naive to assume we'll ever do **1.4.3.1** as If I remember correctly we've defined that we would stick to the semantic versioning which just uses the first 3 groups?

Now with automatic versioning. Maybe you could change the JS files to be 3 digit groups instead? Because the original issue where it would be easier for you to replace no longer applies, since it's automatic.

David-Moreira:
@stsrki 
It should now be working:
Build currently uses an hard coded version in the yml:
image
We should be able to add an additional 0 if we want the four digits thing:
image

The pack then gets built with the dev- suffix we talked about, since it builds again I believe both the assembly and js will have the same dev version:
image
image

May we go ahead and enable nuget publish and publish once so I can test the resulting packages?

Build/Blazorise.props
stsrki: We had invalid patches, so I had to use the fourth number to fix it.

image

David-Moreira: I see. Well if everything works as expected, you just need to properly set the Version. And it will propagate properly. See my latest comment for results.

Source/Blazorise.Bootstrap/wwwroot/modal.js
```diff
@@ -116,4 +116,4 @@ export function resetAdjustments(element) {
 export function getScrollBarWidth() {
     const documentWidth = document.documentElement.clientWidth;
     return Math.abs(window.innerWidth - documentWidth);
-}
\ No newline at end of file
+}
```
stsrki, line 119: There are a bunch of changes in JS files that seem to be because of line endings. Can we revert them to the previous version?

David-Moreira: Lol, honestly no idea we just replace with `File.ReadAllText(fileName)` I'll try to figure it out.

David-Moreira:
@stsrki Seems to work decently.

image
image
image
image

Although I'm seeing some files load `1.5.0.0` instead of `1.5.0-dev-20` I'm not really sure why the difference. Any idea?

David-Moreira:
Ahh dumbass, I figured it out. 
Some modules are requested by c# which only has knowledge of the assembly version being 1.5.0.0.
The modules requested by javascript, have the replaced version.

Probably not a big deal for a dev version to be published like this. Do you think we should try to normalize, or just leave it?

stsrki:
Yes we need to match the version from assembly being 1.5.0.0, and the same be in JS.

David-Moreira:
Reminder to self :  
Add workflow dispatch option

David-Moreira:
> Yes we need to match the version from assembly being 1.5.0.0, and the same be in JS.

Meh. What a bother I'll see what we can do.

Source/Blazorise.Bootstrap/wwwroot/modal.js
David-Moreira: added **.trim** Seems to have fixed it.

David-Moreira:
@stsrki Seems fixed & working as expected:
image
image
Please review.

Source/Blazorise.Bulma/wwwroot/modal.js
stsrki, line 1: Bulma js file content got completely deleted.

David-Moreira: That's strange, let me check. Maybe I commited a bad attempt.

David-Moreira: Seems like it was on the PShell version. so probably commited wrongly.
image

David-Moreira: Rebuilt and no changes. Should be fixed. 
image

David-Moreira:
I think I'm going to do an improvement where it does not overwrite the file if there's no match for the version regex. This will avoid overwriting files that have no version update.

So please review, but do not merge yet.

David-Moreira:
@stsrki Done, please review again. Thanks!

stsrki:
Can you make it so that when we have 

```
<Version>1.5.0</Version>
<PackageVersion>1.5.0</PackageVersion>
```

The generated query will still be

```
import { Behave } from "./vendors/Behave.js?v=1.5.0.0";
import { getRequiredElement } from "./utilities.js?v=1.5.0.0";
```

and not

```
import { Behave } from "./vendors/Behave.js?v=1.5.0";
import { getRequiredElement } from "./utilities.js?v=1.5.0";
```

David-Moreira:
> 1.5.0.0

Done, it now tries to make it 4 "version groups" at least.

stsrki:
Still not good :(

If I do the

```
<Version>1.5.0-preview1</Version>
<PackageVersion>1.5.0-preview1</PackageVersion>
```

it generates

```
import "./vendors/Pickr.js?v=1.5.0-preview1.0.0.0.0";
import * as utilities from "./utilities.js?v=1.5.0-preview1.0.0.0.
```
And it should generate

```
import { Behave } from "./vendors/Behave.js?v=1.5.0.0";
import { getRequiredElement } from "./utilities.js?v=1.5.0.0";
```

it also adds extra zeros at the and.

---

To explain why we always need to have 4 values, with zeros included at the end. During runtime, the framework will give us 4 value version numbers. At runtime, it also doesn't know the version names. The `1.5.0-preview1`, `1.5.0-dev1` will always end in 1.5.0.0. That's why we must match it in JS also.

David-Moreira:
> Still not good :(
> 
> If I do the
> 
> ```
> <Version>1.5.0-preview1</Version>
> <PackageVersion>1.5.0-preview1</PackageVersion>
> ```
> 
> it generates
> 
> ```
> import "./vendors/Pickr.js?v=1.5.0-preview1.0.0.0.0";
> import * as utilities from "./utilities.js?v=1.5.0-preview1.0.0.0.
> ```
> 
> And it should generate
> 
> ```
> import { Behave } from "./vendors/Behave.js?v=1.5.0.0";
> import { getRequiredElement } from "./utilities.js?v=1.5.0.0";
> ```
> 
> it also adds extra zeros at the and.
> 
> To explain why we always need to have 4 values, with zeros included at the end. During runtime, the framework will give us 4 value version numbers. At runtime, it also doesn't know the version names. The `1.5.0-preview1`, `1.5.0-dev1` will always end in 1.5.0.0. That's why we must match it in JS also.

You are not supposed to set the version to 1.5.0-preview1. 
The version is still 1.5.0. You set the package version to 1.5.0-preview1 and generate the 1.5.0-preview1 packagage.

stsrki:
Ahhh. Gotcha. I always did change both in the past.

David-Moreira:
> 1.5.0



> Ahhh. Gotcha. I always did change both in the past.

I mean you are not wrong. 
The version marks explicitly the dll version. (1.5.0)... 1.5.0.0
The PackageVersion is what packs it as a preview version.

But you could indeed arguee the dll version should be marked with the preview. 
This way is easier I guess without any downside as far as I know. Because you can correctly get the same version for the JS without too much custom code. Otherwise we'd have to do some guessing with the version if now we gotta be expecting dashes and chars. Doable still if you're really keen on it.

stsrki:
From I can see the end result is still valid

image

stsrki:
Well, I test manually(localy) at the moment. 

With

```
<Version>1.5.0-</Version>
<PackageVersion>1.5.0-preview1</PackageVersion>
```

- The JS files get the right query number
- The NuGet generated get the right version
- Nugets are properly named

All seems good.

David-Moreira:
> 1.5.0-

In your comment the version has a dash. Did you try it like that? Because it would probably do 1.0.**5-**.0

stsrki (commented):
Build works now.

Can you update https://github.com/Megabit/Blazorise/blob/master/NuGet/help.md and include the steps that we will need to to in the future. For example what we write in Version and what in PackageVersion. So that we document all steps.

Also, can you revert JS files to preview version? They are all listed in this PR with changes, and nothing seems different. I think it might be line ending but I'm unable to compare,

David-Moreira:
> Build works now.
> 
> Can you update https://github.com/Megabit/Blazorise/blob/master/NuGet/help.md and include the steps that we will need to to in the future. For example what we write in Version and what in PackageVersion. So that we document all steps.
> 
> Also, can you revert JS files to preview version? They are all listed in this PR with changes, and nothing seems different. I think it might be line ending but I'm unable to compare,

Done.

Build/Blazorise.props
```diff
@@ -1,64 +1,152 @@
...
+	  AssemblyFile="$(MSBuildToolsPath)\Microsoft.Build.Tasks.Core.dll">
+    <ParameterGroup>
+      <SourceFiles ParameterType="Microsoft.Build.Framework.ITaskItem[]" Required="true" />
+      <Version ParameterType="System.String" Required="true" />
+    </ParameterGroup>
+    <Task>
+      <Using Namespace="System"/>
+      <Using Namespace="System.IO"/>
+      <Using Namespace="System.Text.RegularExpressions" />
+      <Code Type="Fragment" Language="cs">
```
stsrki, line 77: Do you think it makes sense to create a console app that do the same as this and we execute it instead of having the code build into the project configuration? We can keep the `JsTrim.exe`(of however we call it) in the NuGet subfolder.

David-Moreira: No. This is much more elegant in my opinion. It's an integral part of the build process and makes sure the versions are always correct, does not depend on human error.
Also this is used to publish master 1.5.0 weekly correctly version wise.

stsrki: OK, we can always revisit it later.

stsrki:
Do you think it is done? Ready for merge?

David-Moreira:
> Do you think it is done? Ready for merge?

Yes. We can always improve as we go.

stsrki:
OK, I think I will merge it after the Fluent UI is merged. Last time i have released preview 1 from that branch. So I don't want to have a new nightly build without it.

David-Moreira:
> OK, I think I will merge it after the Fluent UI is merged. Last time i have released preview 1 from that branch. So I don't want to have a new nightly build without it.

Well we already do. I think I oublished 3-4 times to test. you can delete those packages if u want

stsrki:
No. I meant the `1.5.0-preview1` that we have on Nuget is published from `dev/fluentui` branch.

And, if we start releasing nightly builds now those builds will be behind the `1.5.0-preview1`

stsrki:
Those will be removed :)

stsrki (approved):

PR #5315, opened 2024-02-16
Remove package.json from all projects

### LINKED ISSUE #5306: Remove all package.json from solution projects

stsrki:
Closes #5306

We used it in the past, and it was causing so many problems that we disabled the actions from GitHub. Since then, it has just been standing unused in the solution.


David-Moreira (approved):
Well, LGTM. As long as it doesn't affect package restore in any way.

I still found stuff doing a search : 
Do we care about any of these?
image

stsrki:
I will need to remove those as well if they are part of git commits.

PR #5319, opened 2024-02-17
Autocomplete: Refactor to use DropdownItem ShowCheckbox feature

David-Moreira:
No idea why check handler is not working properly when in AutoComplete. Gotta figure it out.


David-Moreira:
**Note to self:** Investigate whether it's to do with TextFocused

stsrki:
Can it be because of closable.js that is implemented in dropdown?

David-Moreira:
> Can it be because of closable.js that is implemented in dropdown?


> Can it be because of closable.js that is implemented in dropdown?

Nah, that's does not seem like it.
To re-explain the problem.
- Clicking the anchor structure is fine.
- Clicking the structure that makes up the checkbox makes it so it tries to focus the checkbox which in turn will trigger the TextBlur and TextFocused = false from the Autocomplete, which will make the dropdown not visible.

I investigated a while on how to make an element no focusable in html, people say, disabled it the way to go. I just tried to disable the checkbox and it indeed worked... But as you know disabling it, is not really a great option. I suspect I had similar problems before when implementing checkboxes that's why we implemented an icon that acts as a checkbox.

Do you have any suggestions?
I'm thinking in hacking away at it, propagate the Checkbox Focused event, and when the TextBlur event triggers, if it was after the checkbox focused event, we'll ignore it...
But I'm taking a break, I'll think a bit more about it.

David-Moreira:
The Pr's for the checkbox implementation are hard to navigate, but it looks like I found the same limitations at the time.
It seems that at the time, the option we were taking was disabled + styling... hmmm

image

David-Moreira:
**Note to self:**
- Fix tests
- Fix FluentUI look

David-Moreira:
@stsrki I'm not able to compile scss at all. Can you try it?

Also could you try fixing fluent ui checkbox look?
image

stsrki (commented):
Did you notice that when you clock on a BS checkbox instead of a dropdown item, that the menu is closed? Can we keep it open because user will expect to keep it while selecting.

David-Moreira:
> Did you notice that when you clock on a BS checkbox instead of a dropdown item, that the menu is closed? Can we keep it open because user will expect to keep it while selecting.

Can you give me the steps to reproduce.
It seems to be working correctly.
It is of note, however, that when you unselect the menu will close. But that's something that already happens. You could indeed argue this is a bug, and I would fix it in a separate task.

stsrki:
> It is of note, however, that when you unselect the menu will close but this is regardless whether its a checkbox or dropdown click.

Yep, this basically. If you could fix this it would be great.

David-Moreira:
> > It is of note, however, that when you unselect the menu will close but this is regardless whether its a checkbox or dropdown click.
> 
> Yep, this basically. If you could fix this it would be great.

Great. Since it's a production bug, I'd proactively fix it as part of v1.4. Going ahead and opening a separate issue. Please review this feature regardless of this particular bug or we can leave it on hold until we merge the fix in.

Demos/Blazorise.Demo/Pages/Dashboard.razor
stsrki, line 1: Don't forget to rollback the dashboard.

stsrki (approved):

PR #5322, opened 2024-02-20
Modalprovider loses parameter reference when closing popup

David-Moreira:
More info is found in the github issue.

Added where `TComponent : notnull, IComponent` to the `TComponent` version. While it is in theory a breaking change. If users use a non `IComponent` it will just throw an exception as it's not valid for blazor rendering.

What's notable is that we specifically capture the parameters setter provided by the user:
image


stsrki (approved):
Not a fan of this kind of changes in the middle of the production version. But I think it is not going to hurt since it works with IComponent.

LGTM

David-Moreira:
I don't see an issue. But I wouldn't mind moving the constraint to 1.5 if you'd like.

PR #5324, opened 2024-02-20
DataGrid: Column chooser 

David-Moreira:


stsrki:
> @stsrki I'm thinking this will be allocated in the `Pager`, just like the `ButtonRow` with the same kind of positioning settings. Agree?

I think so, yes; at least, that is how I see it being done when I investigate some other libs.

Source/Extensions/Blazorise.DataGrid/_DataGridPagination.razor
```diff
@@ -50,53 +63,89 @@
...
-                        },
-                        SaveBatchCommand = new ()
-                        {
-                            Clicked = SaveBatchClick,
-                            LocalizationString = saveBatchButtonString,
-                        },
-                        CancelBatchCommand = new()
+                            @foreach ( var column in ParentDataGrid.GetColumns().Where( x => x.IsRegularColumn ) )
+                            {
+                                <Row Margin="Blazorise.Margin.Is0">
```
stsrki, line 134: We should use the Dropdown built-in checkbox feature instead of adding the Check box here. That way the UI will be more consistent across all providers.

Source/Extensions/Blazorise.DataGrid/_DataGridPagination.razor.cs
```diff
@@ -10,14 +10,26 @@ namespace Blazorise.DataGrid;
...
+    private bool dropdownColumnChooserVisible;
+
+    #endregion
+
     #region Methods
 
     private bool ShowButtonRow()
         => ButtonRowTemplate is not null && ParentDataGrid.IsButtonRowVisible;
 
+    private bool ShowColumnChooser()
```
stsrki, line 24: I think it makes more sense for this to be a getter property instead of a method. Agree?

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -3611,5 +3616,10 @@ public int SelectedRowIndex
     /// </summary>
     [Parameter] public Action<DataGridBatchEditItem<TItem>, DataGridColumn<TItem>, DataGridCellStyling> BatchEditCellStyling { get; set; }
 
+    /// <summary>
+    /// Gets or sets whether the column chooser is visible.
+    /// </summary>
+    [Parameter] public bool ShowColumnChooser { get; set; }
```
stsrki, line 3622: Do we also want to have an option to place it at the Start or End of DataGrid?

Source/Extensions/Blazorise.DataGrid/Contexts/ColumnChooserContext.cs
```diff
@@ -0,0 +1,21 @@
+﻿#region Using directives
+#endregion
+
+using System.Collections.Generic;
```
stsrki, line 4: Wrap inside of region.

Source/Extensions/Blazorise.DataGrid/Contexts/ColumnChooserContext.cs
```diff
@@ -0,0 +1,21 @@
...
+using Blazorise.DataGrid.EventArguments;
+using Microsoft.AspNetCore.Components;
+
+namespace Blazorise.DataGrid;
+
+public class ColumnChooserContext<TItem>
```
stsrki, line 10: Add XML comments.

Documentation/Blazorise.Docs/Pages/Docs/Extensions/DataGrid/Examples/DataGridColumnChooserExample.razor
```diff
@@ -0,0 +1,47 @@
+﻿@namespace Blazorise.Docs.Docs.Examples
+
+<DataGrid TItem="Employee"
+          Data="inMemoryData"
+          Responsive
+          ShowColumnChooser
+          PagerPosition="DataGridPagerPosition.Top"
+          ShowPager
+          ShowPageSizes>
+    <DataGridColumns>
```
stsrki, line 10: What do you think we simplify Examples by not defining the columns manually? We now have autogenerate columns from the Model.

stsrki (commented):
Feature wise. it works like a charm.

Source/Extensions/Blazorise.DataGrid/_DataGridPagination.razor.cs
David-Moreira: I personally don't mind either way. It's just not consistent with what already exists.
If I change this one to be getter, I would change ShowButtonRow also?

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
David-Moreira: Hmm It's in the pager, the pager options should work to configure it.
Also there's the PagerPosition configuration.
Should be flexible enough.

Documentation/Blazorise.Docs/Pages/Docs/Extensions/DataGrid/Examples/DataGridColumnChooserExample.razor
David-Moreira: I like it alot for simplifying the examples.
What I get a little bit worried is not showing how columns get built on the examples.

Source/Extensions/Blazorise.DataGrid/_DataGridPagination.razor
David-Moreira: I think I had a limitation, but I'll try again, I can't remember.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
stsrki: I meant to show it here if users prefer it 

image

But we can always do that feature later.

David-Moreira: And I just told you it works with the PagerPosition feature like the others.

Documentation/Blazorise.Docs/Pages/Docs/Extensions/DataGrid/Examples/DataGridColumnChooserExample.razor
stsrki: Well, doesn't need to on all examples. Just some that we want to focus on specific feature. Like this column chooser. having too many columns in the code samples add a noise IMO.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
David-Moreira: **This:**
**PagerElementPosition**
image
image
image

Documentation/Blazorise.Docs/Pages/Docs/Extensions/DataGrid/Examples/DataGridColumnChooserExample.razor
David-Moreira: Right, but most examples, **DO** focus on a specific feature. So they'll be missing the columns. 
You know how less experienced users are sometimes... 
Anyways can't hurt trying. Do you think we should only apply to this one, or apply to all we think its worth it?

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
stsrki: Ahh, I found the option now. 

I still fear the user might want to have it at the end of the table.

image

David-Moreira: hmm Strange I thought that was how it worked actually.
Maybe we did not really implement the feature like that.
I'd open a separate issue to review how the `PagerElementPosition` is working. I would expect that if I set to end, that it tries to actually put it at the end just like you mentioned.

Source/Extensions/Blazorise.DataGrid/_DataGridPagination.razor
David-Moreira: Works perfectly. ;Maybe I was just dumb...

David-Moreira:
@stsrki The Column chooser is allocated at the end after all...
Can you verify how you tested it?
image

Additionally I added Margin from top 2 like the total items have for consistency. Although it's a bit weird it needs the specific margin applied?

Go ahead and close #5326 if no issue is verified

stsrki (approved):
Working good. Please check once again on your side. I have made some changes in how pagination columns are spaced.

David-Moreira:
> Working good. Please check once again on your side. I have made some changes in how pagination columns are spaced.

I can't seem to see any visual difference. I am okay with it.
However why would we prefer to use margin over flex. If we are already using flex for positioning we should be consistent? End result seems to be the same eitherway?

stsrki:
From my observation of Bootstrap, AntDesign, and other frameworks, it is the preferred way to align the items to the right with margins. I really rarely see the flex reverse, if ever.

David-Moreira:
> From my observation of Bootstrap, AntDesign, and other frameworks, it is the preferred way to align the items to the right with margins. I really rarely see the flex reverse, if ever.

My point is that if we chose to use the flex system we should stick to it, unless there was a technical limitation.

For example, C# standard convention for private members is a prefix underscore and camelcase and that does not necessarily mean you need to follow it, neither do you or the blazorise code base... :)

So unless there's a technical limitation, to me it makes more sense to be consistent with it.

stsrki:
True, but we always tend to follow the native behavior of the frameworks that we use. And for them margin auto from start is the preferred method.

> We're changing pagination classes with no reason also. Imagine it affects someone's styling.

Also true, but since this is kinda the internal change, not the public facings API, if anyone styling breaks they can easily override it with a small change to their CSS selectors.

David-Moreira:
> True, but we always tend to follow the native behavior of the frameworks that we use. And for them margin auto from start is the preferred method.
> 
> > We're changing pagination classes with no reason also. Imagine it affects someone's styling.
> 
> Also true, but since this is kinda the internal change, not the public facings API, if anyone styling breaks they can easily override it with a small change to their CSS selectors.

Ok, so we are taking the approach of:
- this was incorrectly implemented. 

Would the providers mix flex with margin like we have or some other way? Since we are fixing it, might as well make sure to fix it all.

---

It's not always noticeable until a customer complains, if he does, or someone finds it out some time later hehe

stsrki:
**AntD**

image

---

**BS4**

image

---

**BS5**

image


---

**Bulma**

image

---

**Fluent**

image

---

**Material**

image

----

**Tailwind**

image

---

It seems to work now on all.

David-Moreira:
Can we just try the center option since we have modified it also.

David-Moreira:
> Center
> 
> image

LGTM then.

stsrki (approved):
LGTM

PR #5340, opened 2024-03-02
Expose internal validation APIs

### LINKED ISSUE #5210: Validation: expose internal APIs as public
Proposal

- `Task InitializeInput( IValidationInput inputComponent )`
- `Task InitializeInputPattern<T>( string patternString, T value )`
- `Task InitializeInputExpression<T>( Expression<Func<T>> expression )`

and

- `Task NotifyInputChanged<T>( T newExpressionValue, bool overrideNewValue = false )`

stsrki:
Closes #5210


Source/Blazorise/Components/Validation/Validation.razor.cs
```diff
@@ -110,7 +119,18 @@ internal async Task InitializeInput( IValidationInput inputComponent )
...
+    /// a new regex pattern is created, and the input is re-validated if applicable.
+    /// </summary>
+    /// <typeparam name="T">The type of the value to validate against the pattern.</typeparam>
+    /// <param name="patternString">The regex pattern string for validation.</param>
+    /// <param name="value">The current value of the input to validate against the new pattern.</param>
+    /// <remarks>
+    /// This method optimizes performance by avoiding re-instantiation of the regex pattern if it has not changed.
+    /// It ensures that validation is only re-triggered if the component has been initialized and validation conditions are met.
+    /// </remarks>
+    public async Task InitializeInputPattern<T>( string patternString, T value )
```
David-Moreira, line 133: I don't know much about our validation works internally. But do we need to expose this and `InitializeInputExpression`? Can't we just expose a single Initialization that handles accordingly to whatever is defined?

Since both methods will only really do something if the `Validation` is configured accordingly, seems like an internal implementation detail that should be part of a more generic **Initialization** method?

Source/Blazorise/Components/Validation/Validations.razor.cs
```diff
@@ -102,23 +102,41 @@ private async Task<bool> TryValidateAll()
...
+    /// Notifies the validation system that the status of a validation component has changed. This method handles the logic for updating the overall validation status based on the mode (Auto or Manual) and the current validation results.
+    /// </summary>
+    /// <param name="validation">The validation component whose status has changed.</param>
+    /// <remarks>
+    /// In Auto mode, this method triggers the aggregation of validation results and potentially raises a status changed event.
+    /// It is designed to minimize the number of status changed events by aggregating validation results.
+    /// Special consideration is needed to ensure that the status changed event is raised only once per validation cycle,
+    /// even if multiple validations fail.
+    /// </remarks>
+    public void NotifyValidationStatusChanged( IValidation validation )
```
David-Moreira, line 139: Is there a reason why this takes in an `IValidation`? It doesn't seem to be used?

David-Moreira (approved):

Source/Blazorise/Components/Validation/Validation.razor.cs
stsrki: I don't remember TBH. I think some components have different initialization flow so I made them as separate function to be more flexible. Maybe we can indeed optimize it by having only one function.

David-Moreira: Yea otherwise you're making it so the consumer has to study how he's going to initialize the Validation.
Try to put yourself on his shoes. How would you use this api?
Call every Initialize in a row? Why do I explicitly need to initialize the other ones? Does it need to happen somewhere else?
```
InitializeInput();
InitializeInputPattern();
InitializeInputExpression();
```
Will it work? Do I need additional checks? The moment you make a user have to think too much or investigate the internal api, then the exposed api might not be clear.

stsrki: Its might be difficult. See this

**TimeEdit**;

```cs
if ( parameters.TryGetValue<Expression<Func<TValue>>>( nameof( TimeExpression ), out var expression ) )
    await ParentValidation.InitializeInputExpression( expression );

if ( parameters.TryGetValue<string>( nameof( Pattern ), out var pattern ) )
{
    // make sure we get the newest value
    var value = parameters.TryGetValue<TValue>( nameof( Time ), out var inTime )
        ? inTime
        : InternalValue;

    await ParentValidation.InitializeInputPattern( pattern, value );
}

await InitializeValidation();
```

**TextEdit**:

```cs
if ( parameters.TryGetValue<Expression<Func<string>>>( nameof( TextExpression ), out var expression ) )
    await ParentValidation.InitializeInputExpression( expression );

if ( parameters.TryGetValue<string>( nameof( Pattern ), out var pattern ) )
{
    // make sure we get the newest value
    var value = parameters.TryGetValue<string>( nameof( Text ), out var paramText )
        ? paramText
        : InternalValue;

    await ParentValidation.InitializeInputPattern( pattern, value );
}

await InitializeValidation();
```

We have different parameter names for parameters and different data types. Some are generic(TValue), some could be known, like `string`.

Source/Blazorise/Components/Validation/Validations.razor.cs
stsrki: I was used at some point but it got not-used over time. I would leave it just in case we need it and that we don't brake the API if we do need it.

Source/Blazorise/Components/Validation/Validation.razor.cs
David-Moreira: Right. I don't like it at all.
Anyway, could **parameters** be wrapped in a dedicated method and we abstract away this bit?
Like 
`OnParametersSetValidation(parameters);`

Otherwise we can just take the approach of properly documenting the process of creating a validatable blazorise component with these nuances and refer anyone to it.

stsrki: Also, not all components have Pattern validation. So that's that.

Source/Blazorise/Components/Validation/Validations.razor.cs
David-Moreira: The problem is that it is now a public api... again, the consumer might be looking at this puzzled trying to pass in the validation and he does not know it's not required at all.

Does this add unnecessary extra effort for the consumer?

stsrki: I've propagated the argument to be used as an extra information for the ValidationsStatusChangedEventArgs. In case anyone needs it, and we don't need to remove it now. Agree?

Source/Blazorise/Components/Validation/Validation.razor.cs
stsrki: What do you think of this approach?

```cs
public async Task Initialize<TValue>( IValidationInput inputComponent, ParameterView parameters, string valueParameterName, string expressionParameterName )
{
    if ( parameters.TryGetValue<Expression<Func<TValue>>>( expressionParameterName, out var paramExpression ) )
        await InitializeInputExpression( paramExpression );

    if ( parameters.TryGetValue<string>( nameof( Pattern ), out var paramPattern ) )
    {
        // make sure we get the newest value
        var value = parameters.TryGetValue<TValue>( valueParameterName, out var paramValue )
            ? paramValue
            : inputComponent.ValidationValue;

        await InitializeInputPattern( paramPattern, value );
    }

    await InitializeInput( inputComponent );
}
```

stsrki: Check it here: https://github.com/Megabit/Blazorise/commit/be5a6e51687dce015004dc86bf1dce8d0c3b7f91

If you think it is worth it I will merge it here in this PR. At least we have less code to intitialise.

David-Moreira: That seems way better to me. The fact that the parameters argument is in there already hints at the consumer this should be in the ParametersSet. But maybe the method name should be more descriptive somehow? Because altought it's called Initialize, it will be called multiple times.
Initialize generally is something that is called once.

David-Moreira: Oh there's still quite a bit of ceremony in fact.... hmm

David-Moreira: If it's only called once = > validationInitialized why do we need it to be on the parameters set again? Because we need to access the parameters collection?

stsrki: Yes, we need to get the latest value from the parameters. Which I now realized the new code might be wrong. Previously I checked Expression and Pattern parameters on each component change. And if I leave that code placed in the InitializeValidation it will not work. Possibly breaking users code if they change those parameters dynamically.

Back to the drawing board.

stsrki: OK I fixed it. Now it should work as expected.

And renamed `Initialize` into `InitializeValidationParameters`.

stsrki: https://github.com/Megabit/Blazorise/commit/0ca72d7f5ddc07dbb5efee426540bf118566296f

stsrki: In the end, I think it is best to have separate APIs for each validation process. While it works to have them all under the one API I think it is not that flexible for us and for users. Going to merge this PR now.

PR #5341, opened 2024-03-03
Introduce new XXL and Full2K breakpoint size

### LINKED ISSUE #5215: Built-in Bootstrap 5 XXL breakpoint support
I'm trying to use Bootstrap's 5 `xxl` breakpoint using Blazorise, but it's not available.
Blazorise breakpoints go up to `FullHD`, which corresponds to Bootstrap's `xl` breakpoint.

A Blazorise breakpoint that goes up to `≥1400px`.

stsrki:
Closes #5215

In this PR, I have focused mainly on BS4, Material, and Bulma. Other providers have already native support for `xxl` size.

Bulma could potentially not be fully covered, but this is intentional because in the next Blazorise v1.6 we are migrating to new Bulma v1.0 anyways. We're already part of their closed beta group so I'm already investigation it.


Source/Blazorise.Bootstrap/Styles/utilities/_display.scss
```diff
@@ -0,0 +1,14 @@
+﻿@each $breakpoint in map-keys($grid-breakpoints) {
+    // only generate xxl that is not available in native Bootstrap 4
+    @if $breakpoint == "xxl" {
```
David-Moreira, line 3: does it even make sense to iterate?

David-Moreira (approved):
Looks good it seems.
2k seems like an exageration. hehe. Are front end frameworks even supporting 4k or something?

Source/Blazorise.Bootstrap/Styles/utilities/_display.scss
stsrki: Maybe not. But I want to make it future proof if we introduce another breakpoint.

stsrki:
> Looks good it seems. 2k seems like an exageration. hehe. Are front end frameworks even supporting 4k or something?

I really didn't find any better way than 2k. We already have FullHD, which is close to HD but not quite. So it seemed natural to have a higher one as Full2K. 

Also I think only Fluent has 4K breakpoint natively.

Maybe I could rename Full2K to QuadHD? Google say it like this:

image

Source/Blazorise.Bootstrap/Styles/utilities/_display.scss
David-Moreira: Wouldn't future proof be an iteration **over only the keys that you need** instead of doing **ifs**?

David-Moreira:
> > Looks good it seems. 2k seems like an exageration. hehe. Are front end frameworks even supporting 4k or something?
> 
> I really didn't find any better way than 2k. We already have FullHD, which is close to HD but not quite. So it seemed natural to have a higher one as Full2K.
> 
> Also I think only Fluent has 4K breakpoint natively.
> 
> Maybe I could rename Full2K to QuadHD? Google say it like this:
> 
> image

Unfortunately I think 2k and 4k are more widely used at least when marketing devices. So people are probably more acquainted with that?

However for consistency sake, using `QHD` and `UHD` does seem like the better option?

stsrki:
> However for consistency sake, using `QHD` and `UHD` does seem like the better option?

I like 

- `QuadHD`, and
- `UltraHD`

more than the shorter `QHD` and `UHD` versions.

But I fear users might get confused about which HD is the highest. So, 2K and 4K might be more obvious at first.

David-Moreira:
> > However for consistency sake, using `QHD` and `UHD` does seem like the better option?
> 
> I like
> 
> * `QuadHD`, and
> * `UltraHD`
> 
> more than the shorter `QHD` and `UHD` versions.
> 
> But I fear users might get confused about which HD is the highest. So, 2K and 4K might be more obvious at first.

Yes. The blame is on marketing and how they mostly dropped QHD and UHD in favor of 2k and 4k, because these two do look more explciit and bigger to the consumers...
So I also expect that the general person is more acquainted with 2k and 4k.

But then again, programmers are usually more acquainted with these nuances so it might be fine to use the `QuadHD` and `UltraHD`, again for consistency. We still have docs and examples to support it.

Source/Blazorise.Bootstrap/Styles/utilities/_display.scss
stsrki: It has to be an `if` because `breakpoint-infix` has special rules where it checks if there are smaller sizes than are currently running.

David-Moreira:
> I went with QuadHD in the end, as it is a more standardized name than Full2K.

I agree, Full2K seems like something made up. Either it was 2K or QuadHD in my opinion. But since we can't really do 2K...

PR #5344, opened 2024-03-06
Bar: Toggle behavior feature

David-Moreira:
I'm not sure ToggleSingle is the best naming for this feature, any ideas?

Please review the implementation before I go ahead with the docs.


Source/Blazorise/Components/Bar/Bar.razor.cs
```diff
@@ -174,10 +175,50 @@ private async void OnLocationChanged( object sender, LocationChangedEventArgs ar
             await Toggle();
     }
 
+
+    internal async Task HideOtherBarItems( BarItem barItem )
```
stsrki, line 179: Rename it to `HideAllExcept`? And make it public?

Source/Blazorise/Components/Bar/Bar.razor.cs
```diff
@@ -174,10 +175,50 @@ private async void OnLocationChanged( object sender, LocationChangedEventArgs ar
...
+        }
+    }
+
+    /// <summary>
+    /// Notifies the <see cref="Bar"/> of a BarItem to be removed.
+    /// </summary>
+    /// <param name="barItem">Reference to the <see cref="BarItem"/> that is placed inside of this <see cref="Bar"/>.</param>
+    internal void NotifyBarItemRemoved( BarItem barItem )
+    {
+        BarItems.Remove( barItem );
```
stsrki, line 209: Do we need to check for null?

Source/Blazorise/Components/Bar/BarDropdown.razor.cs
```diff
@@ -48,7 +48,10 @@ protected override void BuildClasses( ClassBuilder builder )
     /// <inheritdoc/>
     protected override Task OnInitializedAsync()
     {
-        ParentBarItem?.NotifyBarDropdownInitialized( this );
+        if ( ParentBarDropdown is null ) //BarItem tracks the first direct menu. Ignore submenus
```
stsrki, line 51: For clarity, maybe have `ParentBarDropdown is null` under the getter property named `IsUnderFirstMenu,` or similar.

Source/Blazorise/Components/Bar/BarDropdown.razor.cs
```diff
@@ -265,6 +268,13 @@ public bool Visible
 
             state = state with { Visible = value };
 
+            if ( value )
+            {
+#pragma warning disable CS4014 // Because this call is not awaited, execution of the current method continues before the call is completed
```
stsrki, line 273: Hm, not liking this code smell. Can we use `ExecuteAfterRender`?

Or make is as async discard https://stackoverflow.com/questions/42838362/ignoring-the-return-value-from-an-async-task-method?

Source/Blazorise/Components/Bar/Bar.razor.cs
```diff
@@ -353,5 +394,10 @@ public BarCollapseMode CollapseMode
     /// </summary>
     [CascadingParameter] protected LayoutHeader LayoutHeader { get; set; }
 
+    /// <summary>
+    /// Keeps a single bar item open at a time.
+    /// </summary>
+    [Parameter] public bool ToggleSingle { get; set; }
```
stsrki, line 400: How about we have an enum? `BarMenuToggleBehavior`?

- Single
- Multiple

Source/Blazorise/Components/Bar/BarDropdown.razor.cs
David-Moreira: The code smell is already because of the architecture of the component in my opinion that makes use of the setter like this. Also we already have another async call there, VisibleChanged.InvokeAsync, that should also not be appropriate.

I'd prefer to do ParametersSet and check for visible changed as we've seen recommended before, if you really want to change.

Source/Blazorise/Components/Bar/Bar.razor.cs
David-Moreira: No idea, I tought about it, but I wasn't sure it'd make sense? Do you think it'd make more sense?

Source/Blazorise/Components/Bar/Bar.razor.cs
David-Moreira: Not, sure, seems like a very niche helper method. Do you think it's worth being public?

Source/Blazorise/Components/Bar/Bar.razor.cs
stsrki: Yeah, I think so, yes.

🤔 maybe better names would be

- AllowSingleMenu
- AllowMultipleMenus

in case we have more behaviors in the future.

Source/Blazorise/Components/Bar/BarDropdown.razor.cs
David-Moreira: > Or make is as async discard https://stackoverflow.com/questions/42838362/ignoring-the-return-value-from-an-async-task-method?

I don't see the point of this? It is equivalent to what we have. 
Let's not mistake a compiler warning not showing for something, making it right suddently.

stsrki: Ok, change it to ParametersSet. We can start slowly with that change whenever we work on some component.

Source/Blazorise/Components/Bar/Bar.razor.cs
stsrki: Not sure haha

Source/Blazorise/Components/Bar/BarDropdown.razor.cs
David-Moreira: The point is to be consistent, in my opinion I would leave it for consistency in order to not have mixed ways to do things. 
If we chose this "architecture" for this component, we stick to it in my opinion. If you still prefer that I move it, I'm ok with moving it.

stsrki: I agree to move it. We need to do it anyway. I'm just saying that by doing it peace by peace, we're making it easier for ourselves in the future, instead of going all-in on all components at once.

stsrki: Hopefully done. Please retest on your side.

David-Moreira: That does seem like it would work more as I personally would expect, with a private variable tracking the visible state and listening to the `Visible` Parameter. But I need to take a look better at it, because I'm not entirely convinced it's synced correctly. Also I believe you forgot to trigger `VisibleChanged`.

Also I'm still puzzled why we keep creating a new object reference for the state each time. Why not just use a class if we are not using the record semantics apparently? 
So basically just doing : `state.Visible = trueOrFalse`?

stsrki: The record is used because we want the component children to keep track of the state changes. The class is checked by reference, and the record is checked by the values.

David-Moreira: Ok gotcha, so it's for the automatic change detection. Could it no be a struct then?

David-Moreira: I honestly think it should. Let me find docs

David-Moreira: Actually, What if it's being triggered because you're newing up an object instead of the equality factor?

David-Moreira: You could argue that you could implement some kind of abstraction where you'd have a State Base Class where you have equality as a mandatory override. 
And you could have the best of both worlds, since you don't need the immutability feature and wouldn't need to pay the cost to new up an object every single time.

David-Moreira: **PS:** There's also **record struct**

stsrki: It might indeed be because of the newing of the record object. But there is another neat record feature. The `with` keywoard. It is a lot easier to clone an object that way than to have a separate `Clone()` method. It's all handled by the compiler.

David-Moreira: You didn't read everything... :) It's more ceremony indeed. But consumers of your library shouldn't have to pay the additional memory cost because you want to write less code if you can avoid it.

- Either a source generated class **OR**
- a record struct
could possibly at least remove the price we're paying with all the **With**. Which I honestly can't tell you if that price is huge or not in the grand scheme of things.

> You could argue that you could implement some kind of abstraction where you'd have a State Base Class where you have equality as a mandatory override.
And you could have the best of both worlds, since you don't need the immutability feature and wouldn't need to pay the cost to newing up an object every single time. Just more cerimony of course. But with source code generator that could be mitigated.

stsrki: I will need to investigate more about `record struct`.

But isn't regular regular struct always copied when assigned to another variable? Which potentially means cascaded struct could be copied on each rerender. 🤔 Which would have even more impact on the performance. With the current record, we at least try to clone it only when needed.

David-Moreira: As far as my understanding goes, record struct is a struct with record semantics. 
This means that you should get the benefits of a struct. 
Which means it should act as any value type, like an int for instance. So it should work as any other instance value type member of that class as far as I understand and should remove the allocated heap memory out of the equation and all the objects being created each time and having to be garbage collected.
So I don't understand your reasoning on it having more impact.

David-Moreira: But again, we can just open a separate issue and investigate that as part of the separate issue. And use some memory profile tooling to actually take concrete evidence.

David-Moreira: https://learn.microsoft.com/en-us/dotnet/csharp/advanced-topics/performance/
image

stsrki: Yeah, I'm aware of the difference between class and struct.

---

I quickly updated and converted `record` to `record struct`. Here are the results:

image

image

So a lot of our internal logic would break with value based types.

David-Moreira: I see. We could still do some tricks to check whether the state really exists instead of null checks, seems like it would be an easy fix. But anyway if you don't want to investigate this at all it's fine. Resolving.

Source/Blazorise/Components/Bar/BarDropdown.razor.cs
```diff
@@ -131,14 +126,34 @@ public async Task Toggle( string dropdownToggleElementId )
...
 
-        await InvokeAsync( StateHasChanged );
+        await RaiseEvents( visible );
+    }
+
+    /// <summary>
+    /// Fires all the events for this dropdown.
+    /// </summary>
+    /// <param name="visible"></param>
+    protected virtual Task RaiseEvents( bool visible )
```
David-Moreira, line 154: Why is this virtual? Should it be?

David-Moreira:
Visible seems to be more understandable and working correctly. Thank you for your  help.

stsrki:
> Visible seems to be more understandable and working correctly. Thank you for your help.

You started it. Thank you :)

stsrki (approved):

PR #5350, opened 2024-03-08
DataGridColumn: Add AggregateTemplate

David-Moreira:
Although it's redundant for most columns since `DisplayTemplate` already exists in the `AggregateColumn`.  
This best allows to configure each column individually including the Command and multi select columns and any additional special column we might introduce.

Maybe in the future we could consider in removing the `Aggregate` Column `DisplayTemplate` instead? 
What do you think? 

Even the AggregateColumns definition could possibly be moved to the DataGridColumn, making the definition more centralized. 
However I have to say I personally do like the separation.

### DIFF
--- a/Demos/Blazorise.Demo/Pages/Tests/DataGrid/DataGridPage.razor
+++ b/Demos/Blazorise.Demo/Pages/Tests/DataGrid/DataGridPage.razor
@@ -245,11 +245,7 @@
                         </Span>
                     </EmptyFilterTemplate>
                     <DataGridAggregates>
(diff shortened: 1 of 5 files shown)


stsrki (approved):

stsrki:
> Maybe in the future we could consider in removing the `Aggregate` Column `DisplayTemplate` instead?
> What do you think?

That would only work if the user defines all Aggregates for all columns displayed in the DataGrid. Because the aggregate need to be placed in an explicitly defined column to show properly.

PR #5356, opened 2024-03-12
RichTextEdit: Use module system for loading JS references

njannink:


njannink:
@stsrki I have update the RTE to use the modern way for loading the external references and included the stylesheet and js in the vendors repo

Source/Blazorise/Modules/BaseJSModule.cs
```diff
@@ -159,7 +159,18 @@ protected async ValueTask<TValue> InvokeSafeAsync<TValue>( string identifier, pa
...
     {
-        return moduleTask ??= jsRuntime.InvokeAsync<IJSObjectReference>( "import", ModuleFileName ).AsTask();
+        return moduleTask ??= InitializeModule();
+
+        async Task<IJSObjectReference> InitializeModule()
+        {
+            var @ref = await jsRuntime.InvokeAsync<IJSObjectReference>( "import", ModuleFileName );
+            if ( ModuleLoaded is not null )
+            {
+                _ = Task.Run( () => ModuleLoaded( @ref ) );
```
stsrki, line 169: Can this just be awaited without using `Task.Run`?

Source/Extensions/Blazorise.RichTextEdit/JSRichTextEditModule.cs
```diff
@@ -34,25 +29,32 @@ public JSRichTextEditModule( IJSRuntime jsRuntime, IVersionProvider versionProvi
...
+    {
+        List<string> styles = new();
+        if ( options.UseBubbleTheme )
+            styles.Add( "bubble" );
+        if ( options.UseBubbleTheme )
+            styles.Add( "snow" );
+
+        if ( styles.Count > 0 )
+        {
+            await @ref.InvokeVoidAsync( "loadStylesheets", string.Join( ',', styles ), VersionProvider.Version );
```
stsrki, line 48: Instead of csv we can have an array of strings.

njannink: ok I will change that

njannink:
Im testing the RTE and see something doesnt work anymore. The update of the content from code inside the editor. Im looking at that now aswell

njannink:
I resolved the issue with content not updating. It is using a possible dangerous call so I added a remark. I don't know if there is a html sanatize function available in Blazorise.

Source/Extensions/Blazorise.RichTextEdit/wwwroot/richtextedit.js
```diff
@@ -66,8 +82,7 @@ export function initialize(dotnetAdapter, element, elementId, readOnly, placehol
     function setContent() {
         if (contentUpdating) return;
 
-        const content = quill.clipboard.convert(contentRef.innerHTML);
```
njannink, line 69: this old convert method is no longer functional and has been replaced 
https://quilljs.com/docs/modules/clipboard#dangerouslypastehtml

stsrki: We use some libraries on our internal projects to sanitize HTML. But I think, in this case, it would be better not to import additional libraries into RTE. Instead, we can add a callback API to the RTE so that users could use for their own sanitation implementation.

eg. `[Parameter] public Func<string, string> SanitizeHtml { get; set; }`

njannink: we could leave the responsibility with the caller of `SetHtmlAsync( string html )` to not bload the API of the RTE component. As long as it's clearly mentioned in the documentation that the caller is responsible for sending sanatized html.

stsrki: Makes sense. Having a note in the docs would be enough.

Source/Blazorise/Modules/BaseJSModule.cs
```diff
@@ -192,5 +204,10 @@ private Task<IJSObjectReference> GetModule()
     /// </summary>
     protected IVersionProvider VersionProvider => versionProvider;
 
+    /// <summary>
+    /// Runs once after the module has been loaded.
+    /// </summary>
+    protected Func<IJSObjectReference, ValueTask> ModuleLoaded { get; set; }
```
stsrki, line 210: How about we convert this to virtual method `protected virtual ValueTask OnModuleLoaded`?

Source/Extensions/Blazorise.RichTextEdit/JSRichTextEditModule.cs
```diff
@@ -34,25 +29,35 @@ public JSRichTextEditModule( IJSRuntime jsRuntime, IVersionProvider versionProvi
         : base( jsRuntime, versionProvider )
     {
         this.options = options;
+
+        ModuleLoaded = OnModuleLoaded;
```
stsrki, line 33: And then we override OnModuleLoaded instead of assigning it?

Source/Blazorise/Modules/BaseJSModule.cs
njannink: fine by me

njannink:
Nice teamwork @stsrki. Good to have this updated / resolved. Still want to look further at the microsoft RTE, but good to have the QuillJS also still functional

Source/Extensions/Blazorise.RichTextEdit/wwwroot/richtextedit.js
```diff
@@ -1,6 +1,9 @@
 ﻿import "./vendors/quill.js?v=1.4.2.0";
 import { getRequiredElement } from "../Blazorise/utilities.js?v=1.4.2.0";
 
+document.getElementsByTagName("head")[0]
```
stsrki, line 4: I don't see this `blazorise.richtextedit.bundle.scp.css` file in the wwwroot. Did you forget to include it?

stsrki (commented):
We don't have this file in wwwroot. Did you for

stsrki:
> Nice teamwork @stsrki. Good to have this updated / resolved. Still want to look further at the microsoft RTE, but good to have the QuillJS also still functional

You did a great job.

stsrki (approved):
LGTM. Big thanks for your help.

PR #5359, opened 2024-03-13
Project cleaning: restructure providers and components

stsrki:


David-Moreira (approved):
So If I got this right, you've namespaced components & the styling providers for each provider. I don't see anything wrong, LGTM.

stsrki:
> So If I got this right, you've namespaced components & the styling providers for each provider. I don't see anything wrong, LGTM.

Exactly. It also helps to fight those users who accidentally define namespace like Blazorise.Bootstrap in their _Import.razor, and then they have conflicts.

PR #5093, opened 2023-10-21
DatePicker: Add an option to control input mask

stsrki:
Closes #5034

I still have trouble with long-month names when I define InputFormat as `MMMM.dd.yyyy`. Reported it on [#2751](https://github.com/RobinHerbots/Inputmask/issues/2751)

Test with

```
<CardBody>
    <DatePicker TValue="DateTime?" InputFormat="dd/MM/yyyy" DisplayFormat="dd/MM/yyyy" />
</CardBody>
<CardBody>
    <DatePicker TValue="DateTime?" InputFormat="MMM/dd/yyyy" DisplayFormat="MMM/dd/yyyy" />
</CardBody>

<CardBody>
    <DatePicker TValue="DateTime?" InputFormat="MMMM/dd/yyyy" DisplayFormat="MMMM/dd/yyyy" />
</CardBody>
```

Edit* 

> I still have trouble with long-month names when I define InputFormat as `MMMM.dd.yyyy`. Reported it on [#2751](https://github.com/RobinHerbots/Inputmask/issues/2751)

It seems this is an unfinished feature in inputmask. Hopefully, it will be fixed soon. In any case for the most part this implementation should work, long-month-names excluded. For that I will make a not in the docs.

PR #5363, opened 2024-03-13
DatePicker: Fix flatpickrWrapper is not defined error

panmona:
First of all thanks for maintaining this library!

When updating from Blazorise 1.3.3 to the newest 1.4.3, we started to notice the following error logs in the console:

```
datePicker.js?v=1.4.3.0:118 Uncaught ReferenceError: flatpickrWrapper is not defined
    at ClassWatcher.successClassAddHandler [as classAddedCallback] (datePicker.js?v=1.4.3.0:118:17)
    at MutationObserver.mutationCallback (observer.js?v=1.4.3.0:90:30)
```

The component behavior seems to still be as excepted though.

I suspect that this is because in #5093 the successClassAddHandler functions were moved outside of the scope where the flatpickrWrapper is available. In the PR I unfortunately wasn't able to see why this was changed. The PR seems to me a bit unrelated to those functions.
I'm very open to change the PR if my analysis was incorrect!

### DIFF
--- a/Source/Blazorise/wwwroot/datePicker.js
+++ b/Source/Blazorise/wwwroot/datePicker.js
@@ -110,19 +110,19 @@ export function initialize(dotnetAdapter, element, elementId, options) {
 
                     picker.errorClassWatcher = new ClassWatcher(picker.altInput, options.validationStatus.errorClass, errorClassAddHandler, errorClassRemoveHandler);
                 }
-            }
-        }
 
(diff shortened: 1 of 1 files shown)


panmona:

panmona:
recheck

stsrki (approved):
LGTM. 

Not sure how we missed that as it seems pretty obvious not to be placed there. 

Thank you for the contribution!

Issue #5376, opened 2024-03-19
[Bug]: Richtext edit no longer works in 1.5


blenke:
### Blazorise Version

1.5

### What Blazorise provider are you running on?

Bootstrap5

### Link to minimal reproduction or a simple code snippet

After upgrading to 1.5 my RichtTextEdit renders incorrect, also on your the blazorise example page , see:
https://blazorise.com/docs/extensions/richtextedit

### Steps to reproduce

https://blazorise.com/docs/extensions/richtextedit

### What is expected?

layout as in previous versions

### What is actually happening?

layout distorted and icons are huge

### What browsers do you see the problem on?

Chrome, Microsoft Edge

panmona:
Additionally to that our quill modules seem to have stopped working correctly.
Do you have any guidance on how we need to update those? We use modules/blotFormatter, modules/better-table, modules/imageCompress and modules/mention.

stsrki:
@njannink Can you help?

panmona:
Thanks.
We already used Quill 2.0 before this update (but in a previous version: v2.0.0-dev.4, which is now also already quite old) so that could be an issue. Unfortunately the modules used don't seem to have been gotten an update.
I'll check again whether it works once the theme fix is there 👍 

It would be nice if changes like that would be highlighted more prominently in the release notes in the future: "No breaking changes" is not true in this case unfortunately.

PR #5380, opened 2024-03-19
#5376 Fix snow theme css not loading

njannink:
Fix for #5376


stsrki (approved):
LGTM. Thanks for the help @njannink

PR #5395, opened 2024-03-21
Add Blazorise Google Captcha to HelpUsImprove & Contact Pages

David-Moreira:


David-Moreira:
@stsrki Let's use our own implementation to handle this.

When implementing it, I kinda wish we had a better integration with the validation system, what do you think?

stsrki (commented):
Wouldn't it be better to create something like a `CaptchaInput` component and bind the result to a variable. eg.

```
<CaptchaInput @bind-Value="captchaValid" />
```

It is a lot easier to handle it then inheriting from a base page class.

David-Moreira:
> Wouldn't it be better to create something like a `CaptchaInput` component and bind the result to a variable. eg.
> 
> ```
> <CaptchaInput @bind-Value="captchaValid" />
> ```
> 
> It is a lot easier to handle it then inheriting from a base page class.

I guess you're right, it'll be better. I'll make that change.

David-Moreira:
@stsrki Well... "alot easier" I guess the correct word is "more elegant", because it did take me substiantly more time to implement it properly haha. 
I made use of `BaseInputComponent` abstraction and some of the code from the `Check` so we can have proper blazorize validation support. Please check whether it's well implemented as I'm not well versed in these apis.

stsrki (approved):
LGTM

stsrki:
You did such a great work that it is worth making a blog #5403

🤣

PR #5468, opened 2024-04-19
DataGrid: Fix Localization of Clear Button

panmona:
## Description

We noticed that the clear button text is currently hardcoded, even though a custom _ClearFilterButtonLocalizer_ and translations exist.
Now it uses the localizer.

### Questions / Unclear points
The filter button is also not localized, but for this one **no** translations currently exist. Therefore I've not added it yet. Is it possible for you to provide those? 

Additionally I'm not sure whether it is intended that the _less than_, _less than or equal_, _greater than_, _greater than or equal_ methods all refer to the _FilterMethodContainsLocalizer_. Translations already exist for these so I could easily add the new custom localizers if this is a mistake.

## Types of changes

## Checklist:

### DIFF
--- a/Source/Extensions/Blazorise.DataGrid/_DataGridMenuFilter.razor
+++ b/Source/Extensions/Blazorise.DataGrid/_DataGridMenuFilter.razor
@@ -40,7 +40,7 @@
                 </Column>
                 <Column ColumnSize="ColumnSize.Is12" Flex="Flex.JustifyContent.End.AlignItems.Center" Gap="Gap.Is2">
                     <Button Clicked="@ParentDataGrid.FilterData" Color="Color.Primary">Filter</Button>
-                    <Button Clicked="@(() => ParentDataGrid.ClearFilter(Column.Field))" Color="Color.Light">Clear</Button>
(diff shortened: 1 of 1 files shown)


panmona:
Yes I can submit it to rel-1.5.

And yes, there are more missing.
As written in the description, I would need translations for the "Filter" for all languages. How can I get those?

What about the greater than/less than filter methods: Should these really refer to the Contains Filter method localization? Seems wrong to me.

David-Moreira:
> Yes I can submit it to rel-1.5.
> 
> And yes, there are more missing. As written in the description, I would need translations for the "Filter" for all languages. How can I get those?
> 
> What about the greater than/less than filter methods: Should these really refer to the Contains Filter method localization? Seems wrong to me.

If they refer to the same localization string key then it's wrong as they will not load the correct localization when someone tries to override, if you'd be so kind to fix it also, we'd appreciate it, thanks!!

panmona:
@David-Moreira Okay! I will fix this then aswell.

Can you provide translations for the "Filter"? This translation is not yet in the json file.

David-Moreira:
> @David-Moreira Okay! I will fix this then aswell.
> 
> Can you provide translations for the "Filter"? This translation is not yet in the json files.

We know it's not the most accurate approach, but we use google translator to make these translations. For simple translations it should be accurate enough.
If a community member notices a badly translated string we'll gladly take the correct translation.

If you don't want to bother with it, submit the PR and we'll complete the json translations later.

panmona:
Okay, I'll re-open the PR when the

Issue #5383, opened 2024-03-19
Allow navigating DataGrid with keyboard when in EditMode="DataGridEditMode.Cell"


mtbayley:
I made a comment about this a few years ago. I wanted to create an official feature request for it.

> It would be really nice if I could batch inline edit in a data grid. I.e.
> 
> * If I'm editing a cell and hit **Tab** it will start editing the cell to the **Right**.
> * If I'm editing a cell and hit **Shift+Tab** it will start editing the cell to the **Left**.
> * If I'm editing a cell and hit **Enter** it will start editing the cell **Below**.
> * If I'm editing a cell and hit **Shift+Enter** it will start editing the cell **Above**.

https://github.com/Megabit/Blazorise/issues/855#issuecomment-985887652

I'd like to be able to navigate with the keyboard within a DataGrid when EditMode="DataGridEditMode.Cell"

Similar to this demo:
https://blazor.syncfusion.com/demos/datagrid/batch-editing

David-Moreira:
`Tab` and `Shift tab` should already move between cells.
I'm not sure `Enter` and `Shift+Enter` editing cell below and above is intuitive, I personally never seen that? 
`Enter` is usually a commit key. And it commits the value presently.

So would you not agree with the current shortcuts and implementation?

stsrki:
The `Enter` key is the remnant of the DOS era, where it was used for navigation of the "forms." I'm not sure if we would want to have that in Blazorise, as it is not standard behavior on the web. It would break any AODA compatibility that we plan to have later.

mtbayley:
What my users are looking for is "Excel like" editing. Currently, it is very difficult to input a column of values because there is no way to navigate vertically with the keyboard. You have to type your value, then use the mouse to move to the next row in the same column.

In Excel and that Syncfusion demo, cell navigation works as described above.

You can also use the arrow keys to move between cells. Maybe that could be implemented as an opt-in feature? Or allow the user to customize the keys which are used for navigation?

mtbayley:
Syncfusion has a whole page dedicated to keyboard navigation. There are extra things in there such as inserting or deleting a row, but the cell navigation is a big one for us.
https://blazor.syncfusion.com/demos/datagrid/keyboard-navigation

I think the current rapid editing feature in Blazorise is very close to be able to used for entering in lots of data for our ERP application

stsrki:
[...]

PR #5506, opened 2024-05-11
DataGrid: Cell Navigation support | CellEdit Mode options

David-Moreira:
Ok so while improving the cell edit feature I realized the following:

**Cell navigation** can just be a standalone feature. By activating the feature, both single clicking, using arrow keys & tabbing will navigate through the cells.
I was initially doing it as part of the **cell edit mode**, but if it's a standalone feature, the user can just activate whenever he wants. We can recommend to activate it in the docs for cell edit mode.

Introduced a new `DataGridEditModeOptions`. Here we'll have two new options: 
`CellEditOnSingleClick`, `CellEditOnDoubleClick `that are self explanatory.
Here I would take the breaking change where we default to double click. Users can still enable the old behaviour by setting the `CellEditOnSingleClick` to true.

Just by adding the extra cell navigation and the single/double click the user has enough flexibility from what I've experienced.

All in all by just adding `CellNavigable` to a **CellEdit enabled Grid** the experience is much better!

Let me know your thoughts.

**Notes:**
You will see that we require a new bit of javascript for this feature, I did try to make it work in C#, but as you know, when  you have alot of dynamic elements and you have to try to figure out stuff in DOM it gets really hard and cumbersome. I think it's best to do it in js.


David-Moreira:
Friendly reminder

Source/Blazorise/wwwroot/blazorise.css
```diff
@@ -1060,6 +1060,10 @@ thead tr th {
   position: relative;
 }
 
+td:focus {
```
stsrki, line 1063: I would like it more to have any extra CSS for each provider individually instead of in core Blazorise.

Documentation/Blazorise.Docs/Pages/Docs/Extensions/DataGrid/DataGridApi.razor
```diff
@@ -329,6 +329,12 @@
         Gets or sets whether the DataGrid should automatically generate columns.
         Columns will only be automatically generated if no columns have been provided.
     </DocsAttributesItem>
+    <DocsAttributesItem Name="DataGridEditModeOptions" Type="DataGridEditModeOptions">
+        Gets or sets DataGridEditMode options.
+    </DocsAttributesItem>
+    <DocsAttributesItem Name="CellNavigable" Type="bool" Default="true">
```
stsrki, line 335: Should say `Default="false"`

Source/Extensions/Blazorise.DataGrid/wwwroot/datagrid.js
```diff
@@ -1,18 +1,28 @@
 import { getRequiredElement } from "../Blazorise/utilities.js?v=1.5.2.0";
 
+const QUERYSELECTOR_ALL_COLUMNS = "tbody tr td";
+const QUERYSELECTOR_ALL_TABLE_HEAD_INPUT = "tbody tr td";
```
stsrki, line 4: Does this selector work for all providers?

David-Moreira: Actually I forgot to test. But I'd expect that they all obbey that basic structure for a table.

David-Moreira: Working on all providers.

mtbayley:
Hi @David-Moreira
I've reviewed this branch and I do like the added navigation features.

However, I am wondering if you are still considering adding these two features I mentioned in my feature request. https://github.com/Megabit/Blazorise/issues/5383#issuecomment-2015238362. These features would be extremely valuable for my users.

>- Typing any value on the keyboard would put the cell in editing mode.

>- Entering editing mode, selects all text of the current value, allowing a new value to be easily updated

David-Moreira:
Hello @mtbayley 
Thanks for reviewing and helping us deliver a better feature.

Both your suggestions seem fine,
- Typing any value on the keyboard would put the cell in editing mode.
- Entering editing mode, selects all text of the current value, allowing a new value to be easily updated

Would you agree @stsrki ? 
Do you guys think it's worth keeping these behind an optional flag or it's fine if the grid starts working like this?

Also any of you have any suggestions for this feature?
-> "selects all text of the current value" 
I don't think I personally ever tried something like this, does blazor support this in some way? I'm guessing not? You guys know the js for this?

stsrki:
> Typing any value on the keyboard would put the cell in editing mode.

This seems fine to me. We can do this.

> Entering editing mode, selects all text of the current value, allowing a new value to be easily updated

I also don't remember any time doing this. So I would skip it.

---

The question is how to opt-in. I would add an additional flag to the `DataGridEditModeOptions`.

mtbayley:
> The question is how to opt-in. I would add an additional flag to the `DataGridEditModeOptions`.

I agree. Options are good. What might work for me, may not work for someone else.

> Also any of you have any suggestions for this feature?
> -> "selects all text of the current value"
> I don't think I personally ever tried something like this, does blazor support this in some way? I'm guessing not? You guys know the js for this?

I see Blazorise uses the autoNumeric library to select all on focus for the Numeric Picker component.
https://github.com/Megabit/Blazorise/blob/95d0ecda056b515dffc8ebbed07e99ffe17f83a2/Source/Blazorise/wwwroot/numericPicker.js#L116-L118

Can you add something similar to your other input components?

https://github.com/autoNumeric/autoNumeric/blob/ea4a977c230c616fdcda6a5441c6009ce850d2c8/src/AutoNumeric.js#L6421-L6432

    _onFocusIn(e) {
        if (this.settings.selectOnFocus) {
            // The whole input content is selected on focus (following the `selectOnFocus` and `selectNumberOnly` options)
            //XXX Firefox <47 does not respect this selection...Oh well.
            this.select();
        } else {
            // Or we decide where to put the caret using the `caretPositionOnFocus` option
            if (!AutoNumericHelper.isNull(this.settings.caretPositionOnFocus)) {
                AutoNumericHelper.setElementSelection(e.target, this._initialCaretPosition(AutoNumericHelper.getElementValue(this.domElement)));
            }
        }
    }

https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/select

This is how AG Grid works by default.

msedge_Mb5Ttz4B9L

David-Moreira:
@stsrki Friendly reminder

David-Moreira:
@stsrki Could you evaluate @mtbayley suggestion about the select all focus? 
Would probably be best to do it on a separate PR and merge it in after.

stsrki:
We have an API on our input that could be used to select all text. But I would leave that for another PR.

stsrki:
Do it in another PR. For 1.6.

mtbayley:
I appreciate your efforts to get this in 1.6. This will help me out a lot 🙂

David-Moreira:
@stsrki I just realized you said we already support it? Why do we need another PR then?

"We have an API on our input that could be used to select all text."

David-Moreira:
Done.
There was indeed already an api for selecting the text. No point in doing anything in a separate PR I guess?

@mtbayley Would you be so kind to take it for a spin if you have time?

mtbayley:
@David-Moreira Looks great! Everything is working how I would expect it to.

Maybe some more clarification in the documentation is what I would add.

image
- Any key to to enter editing mode
- Document CellEditSelectTextOnEdit, I think this should be highlighted in the docs since it is such a great feature, and you would only know about it if you stumbled upon it.

David-Moreira:
@stsrki Friendly reminder

stsrki:
I'm now testing and it works mostly. But it still need some work. For example, it doesn't work as best with the focus feature.

1. Focus the cell

image

2, Hit `a` key

image

3. Hit `b` key

image


Notice how only the `b` is now written. Should it work like that or should it continue writing from `a`? Maybe we need to add an additional option to control the behavior?

- select all on focus
- move caret at the end on focus

David-Moreira:
> I'm now testing and it works mostly. But it still need some work. For example, it doesn't work as best with the focus feature.
> 
> 1. Focus the cell
> 
> image
> 
> 2, Hit `a` key
> 
> image
> 
> 3. Hit `b` key
> 
> image
> 
> Notice how only the `b` is now written. Should it work like that or should it continue writing from `a`? Maybe we need to add an additional option to control the behavior?
> 
> * select all on focus
> * move caret at the end on focus

I'm not sure actually. But you can always turn off the Select feature. But maybe the select feature should only select the text if the key pressed was a non text key, like enter or regular click

stsrki:
Actually, from a UX perspective, it makes sense to select all when we double-click to edit the cell. But when we start typing we need to type all characters. Including the first that is now lost.

mtbayley:
I had missed what @stsrki found. I agree, it should not select all on the first character entered. Instead the cursor should be at the end of the character entered.

> it makes sense to select all when we double-click to edit the cell

Agree

> select the text if the key pressed was a non text key, like enter or regular click

Agree, this how AG grid works.

I also noticed that the `Backspace` key will clear the cell in AG Grid.

David-Moreira:
Done, thanks for the continuous feedback guys.

- It should now only select the text under the same conditions AND if the text hasn't changed.
- Backspace now clears
- I found some bugs while testing, apparently our `UpdateCellEditValue` requires the values to be already in the destination type, so we had to try and do some conversions when trying to set strings, ints, decimals, dates, etc..
  - Not super happy with this, but when we're already boxing stuff, oh well... It works, that's what it matters. hehe
- There was also another bug where apparently the **js select function** does not exist in a select html element, so it would throw an exception, I added a function check

@stsrki please review again.

stsrki:
I've tested the new feature and it works fantastic. Still need to go review the code.

mtbayley:
Agree. It works great now. Excellent work @David-Moreira

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -3791,5 +3844,15 @@ public int SelectedRowIndex
...
 
+    /// <summary>
+    /// Gets or sets DataGridEditMode options.
+    /// </summary>
+    [Parameter] public DataGridEditModeOptions DataGridEditModeOptions { get; set; }
+
+    /// <summary>
+    /// Gets or sets whether the Datagrid is Cell Navigable, users will be able to navigate the Grid cells by pressing the Keyboard's ArrowLeft, ArrowUp, ArrowRight and ArrowDown keys.
+    /// </summary>
+    [Parameter] public bool CellNavigable { get; set; }
```
stsrki, line 3855: We now have two APIs, CellNavigable and Navigable. Both are bool. I think we need to make them more coherent.

I propose to have `DataGridNavigationMode NavigationMode` with

- `Cell`
- `Row`

The reason is that having both might be confusing and are too similar by name. What happens if user have bth enabled?

So we need to obsolete `Navigable`, and introduce this new API.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -3791,5 +3844,15 @@ public int SelectedRowIndex
     /// </summary>
     [Parameter] public bool AutoGenerateColumns { get; set; } = true;
 
+    /// <summary>
+    /// Gets or sets DataGridEditMode options.
+    /// </summary>
+    [Parameter] public DataGridEditModeOptions DataGridEditModeOptions { get; set; }
```
stsrki, line 3850: I don't like the DataGridEditModeOptions parameter name. Rename it to `EditModeOptions`. 

PS. the parameter, not the type.

David-Moreira: Don't like it either, an oversight I guess. :)

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
David-Moreira: For the row you mean up, down? :)
Your suggestion seems like a nice idea.

stsrki: Yes, Row would behave the same as existing Navigable. Pressing Up/Down will focus rows.

stsrki (approved):
LGTM.

Made some cleaning since I have installed SonarLint installed and it was complaining with some warnings. There are a bunch more I didn't wanted to do them as I don't want to make unwanted bugs. We can do the rest later.

Please check and if you agree I will merge

David-Moreira:
> LGTM.
> 
> Made some cleaning since I have installed SonarLint installed and it was complaining with some warnings. There are a bunch more I didn't wanted to do them as I don't want to make unwanted bugs. We can do the rest later.
> 
> Please check and if you agree I will merge

Yea I'm okay with that, I'd stick to this PR changes as much as possible and do those changes in a separate PR, so we can properly identify them if anything goes wrong.

stsrki:
> > LGTM.
> > Made some cleaning since I have installed SonarLint installed and it was complaining with some warnings. There are a bunch more I didn't wanted to do them as I don't want to make unwanted bugs. We can do the rest later.
> > Please check and if you agree I will merge
> 
> Yea I'm okay with that, I'd stick to this PR changes as much as possible and do those changes in a separate PR, so we can properly identify them if anything goes wrong.

If you want you can also install SonarLint yourself. It's a free extension for VS.

David-Moreira:
> > > LGTM.
> > > Made some cleaning since I have installed SonarLint installed and it was complaining with some warnings. There are a bunch more I didn't wanted to do them as I don't want to make unwanted bugs. We can do the rest later.
> > > Please check and if you agree I will merge
> > 
> > 
> > Yea I'm okay with that, I'd stick to this PR changes as much as possible and do those changes in a separate PR, so we can properly identify them if anything goes wrong.
> 
> If you want you can also install SonarLint yourself. It's a free extension for VS.

Oh it's free? I might give it a go. :) Thanks for the suggestion.

stsrki:
Brace yourself for a lot of warnings :) But I guess it is a good thing to be as comply as possible with the standards.

David-Moreira:
> Brace yourself for a lot of warnings :) But I guess it is a good thing to be as comply as possible with the standards.

Depends on who defines those standards, and if they make sense to follow. As always, let's not just blindly follow some tool. haha

PR #5507, opened 2024-05-12
DataGrid: Dynamic data support through ExpandoObject

David-Moreira:
While it wasn't that hard to integrate the loading of data. 
I found it might be a bit limited regarding the edit operations. This is because due to the dynamic nature, we can't really figure out the Type of a particular field without having an actual structure.

**This means that when editing:**
We need to infer from the **actual item** field type in order to know what input to render (numeric, checkbox, etc), so this means a nullable won't be easily figured out, if the value is null, there's no way you can tell which type would be in this particular field for example.
So the TypeGetter method was updated to include the Item as a parameter to look into it.
image

**When creating:** 
We need to override the `NewItemCreator` and provide the structure so the Grid can figure out the inputs to render, checkboxes, numeric, etc...
image

What are your thoughts?


Source/Extensions/Blazorise.DataGrid/DataGridColumn.cs
```diff
@@ -35,19 +36,62 @@ public partial class DataGridColumn<TItem> : BaseDataGridColumn<TItem>
...
     {
-        // TODO: move this to cached FunctionCompiler so it doesn't get compiled every time
-        valueTypeGetter = new( () => FunctionCompiler.CreateValueTypeGetter<TItem>( Field ) );
-        defaultValueByType = new( () => FunctionCompiler.CreateDefaultValueByType<TItem>( Field ) );
-        valueGetter = new( () => FunctionCompiler.CreateValueGetter<TItem>( Field ) );
-        valueSetter = new( () => FunctionCompiler.CreateValueSetter<TItem>( Field ) );
-        sortFieldGetter = new( () => FunctionCompiler.CreateValueGetter<TItem>( SortField ) );
-    }
 
+        if ( typeof( TItem ) == typeof( ExpandoObject ) )
```
stsrki, line 40: I think it would be worth it to have a non-static `FunctionCompiler` class that could be abstracted and then have the implementation for each TItem object type. Instead of having `if`s like this for each type.

stsrki:
> While it wasn't that hard to integrate the loading of data. I found it might be a bit limited regarding the edit operations. This is because due to the dynamic nature, we can't really figure out the Type of a particular field without having an actual structure.
> 
> **This means that when editing:** We need to infer from the **actual item** field type in order to know what input to render (numeric, checkbox, etc), so this means a nullable won't be easily figured out, if the value is null, there's no way you can tell which type would be in this particular field for example. So the TypeGetter method was updated to include the Item as a parameter to look into it. image
> 
> **When creating:** We need to override the `NewItemCreator` and provide the structure so the Grid can figure out the inputs to render, checkboxes, numeric, etc... image
> 
> What are your thoughts?

If there is no other option I don't see any objections regarding this. As long as it doesn't affect the current work.

Source/Extensions/Blazorise.DataGrid/DataGridColumn.cs
David-Moreira: I'm not sure I follow, independent of whatever abstraction you always need to check for type of ExpandoObject.

David-Moreira:
Well, I believe, this is it. What do you think?

David-Moreira:
Friendly reminder

Demos/Blazorise.Demo/Pages/Dashboard.razor
```diff
@@ -1,185 +1,109 @@
...
+                <DataGridAggregate Field="@nameof( Employee.Salary )" Aggregate="DataGridAggregateType.Sum" DisplayFormat="{0:C}" DisplayFormatProvider="@System.Globalization.CultureInfo.GetCultureInfo("fr-FR")" />
+                <DataGridAggregate Field="@nameof( Employee.IsActive )" Aggregate="DataGridAggregateType.TrueCount" />
+                <DataGridAggregate Field="@nameof( Employee.Childrens )" Aggregate="DataGridAggregateType.Sum" />
+            </DataGridAggregates>
+            <DataGridColumns>
+                <DataGridCommandColumn></DataGridCommandColumn>
+                <DataGridColumn Editable TextAlignment="TextAlignment.Center" Field="@nameof( Employee.Id )" Caption="#" Width="60px" />
+                <DataGridColumn Editable Field="@nameof( Employee.FirstName )" Caption="First Name">
+                </DataGridColumn>
+                <DataGridColumn Editable Field="@nameof( Employee.LastName )" Caption="Last Name" />
```
stsrki, line 29: Should we define RAW field names instead of using Employee?

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -518,19 +520,65 @@ protected override async Task OnAfterRenderAsync( bool firstRender )
...
+                    DataGridColumn<TItem> column = GetColumnByType( type );
+
+                    column.Editable = true;
+                    column.Caption = expandoKeyValue.Key;
+                    column.Field = expandoKeyValue.Key;
+                    this.AddColumn( column );
+                } 
+            }
+
+            StateHasChanged();
```
stsrki, line 550: Use `InvokeAsync(StateHasChanged)`

stsrki (commented):
Try to create a separate test page in the Demo project. And also add unit test(s).

Demos/Blazorise.Demo/Pages/Dashboard.razor
David-Moreira: Yea... probably better so some devs aren't confused.

Source/Extensions/Blazorise.DataGrid/DataGridColumn.cs
David-Moreira: friendly reminder

stsrki: Meh, leave it. If we ever have more TItem types, we will optimize it.

stsrki:
There are conflicts now.

David-Moreira:
> There are conflicts now.

Done.

David-Moreira:
@stsrki Friendly reminder

stsrki (approved):
LGTM. If you don't plan to add anything else I will merge it.

David-Moreira:
> LGTM. If you don't plan to add anything else I will merge it.

I am ok with the feature for now. Better have it introduced and get feedback from users using it then to overdo it.

PR #5514, opened 2024-05-18
DataGrid: Add more auto-generator attributes

David-Moreira:
- IgnoreFieldAttribute
- OrderAttribute
- NumericAttribute
- SelectAttribute
- DateAttribute


Source/Blazorise/Attributes/IgnoreFieldAttribute.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Blazorise;
+
+/// <summary>
+/// Ignores the field from being included.
```
stsrki, line 10: Should better define in what the field is not included.

Source/Blazorise/Attributes/NumericAttribute.cs
```diff
@@ -0,0 +1,44 @@
+﻿using System;
+using Microsoft.AspNetCore.Components;
+
+namespace Blazorise;
+
+/// <summary>
+/// Configures a numeric field.
```
stsrki, line 7: More descriptive comment.

eg. `Represents an attribute that can be applied to numeric properties or fields to specify additional metadata.`

Source/Blazorise/Attributes/OrderAttribute.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
+
+namespace Blazorise;
+
+/// <summary>
+/// Configures the field order.
```
stsrki, line 6: Also more descriptions.

PS. Chat GPT works well for this.

stsrki (commented):
Other than the comments, looks good to me.

David-Moreira:
Added also Select and Date Attributes, I think this attributes are fine for now as they complement the possible DataGrid columns.

Source/Blazorise/Attributes/SelectAttribute.cs
```diff
@@ -0,0 +1,35 @@
...
+    /// The name of the function that will be used to get the data source.
+    /// By default it's set to "GetSelectData".
+    /// </summary>
+    public string GetDataFunction { get; set; } = "GetSelectData";
+
+    /// <summary>
+    /// Used to get the display field from the supplied data source.
+    /// Defaults to "Description".
+    /// </summary>
+    public string TextField { get; set; } = "Description";
```
stsrki, line 21: Should we even have default names for the fields? The combination of `Description` and `Id` seems very rare in the real world.

Maybe name them differently?

- `Value`
- `Text`

David-Moreira: Well rare for you...? :D I'm thinking in the perspective of someone binding a DTO.
So at least, I have worked with Id commonly. So wanting the **Id** as the **Value** which corresponds to the database entity. That seems like a common case? Even if its a DTO the dto generally also points to the Id. 

As for the **Text** I think it's generally Description or Text,..?

David-Moreira: I also don't mind not assuming any defaults.
The idea was that if someone were to have the same type of binding, he would have to define less metadata in the attribute.

David-Moreira: Here's an example on our own code base, for support forum, that if we used a `SelectList` the `ValueField` would be the `Id`
image

David-Moreira: Done.

stsrki:
Can you add release notes? I have already create an empty 1.6 page and merged it here.

stsrki (approved):

PR #5535, opened 2024-05-27
DatePicker: fixed validation rules for Dates parameter

### LINKED ISSUE #5508: [Bug]: DatePicker validaton not working for Dates
### Blazorise Version

1.5.2

None

-

### Steps to reproduce

-

### What is expected?

Validation should work.

When using `DatePicker` with `@bind-Dates` (selection mode Range or Multiple) the validation does not work and I get error:
```
Unable to determine the validator type.
```

After looking in the source it seems like only the `Date` input is initialized for validation in `SetParametersAsync`  (DatePicker.razor.cs), thus `Dates` input validation will not work. Probably need this:
```
if ( parameters.TryGetValue<Expression<Func<TValue>>>( nameof( DatesExpression ), out var datesExpression ) )
    await ParentValidation.InitializeInputExpression( datesExpression );
```

Workaround: inherit `DatePicker` and override `SetParametersAsync`:
```
    public override async Task SetParametersAsync(ParameterView parameters)
    {
        if (parameters.TryGetValue<Validation>(nameof(ParentValidation), out var parentValidation) && parentValidation is not null)
        {
            if (parameters.TryGetValue<Expression<Func<IReadOnlyList<TValue>>>>(nameof(DatesExpression), out var expression))
            {
                await parentValidation.InitializeInputExpression(expression);
            }
        }

        await base.SetParametersAsync(parameters);
    }
``` 

_No response_

### Any additional comments?

_No response_

stsrki:
Closes #5508

- Fixed the Dates validation.
- Add built-in validation rules for dates
- updated docs

Test:

```razor
<Validation Validator="ValidationRule.IsDateSelected<DateTime?>">
    <DatePicker TValue="DateTime?" @bind-Date="selectedDate" InputMode="DateInputMode.Date">
        <Feedback>
            <ValidationSuccess>Date is good.</ValidationSuccess>
            <ValidationError>Pick a date!</ValidationError>
        </Feedback>
    </DatePicker>
</Validation>
<Validation Validator="ValidationRule.AreDatesSelected<DateTime?>">
    <DatePicker TValue="DateTime?" @bind-Dates="selectedDates" InputMode="DateInputMode.Date" SelectionMode="DateInputSelectionMode.Multiple">
        <Feedback>
            <ValidationSuccess>Dates are good.</ValidationSuccess>
            <ValidationError>Pick dates!</ValidationError>
        </Feedback>
    </DatePicker>
</Validation>
@code {
    DateTime? selectedDate;
    IReadOnlyList<DateTime?> selectedDates;
}
```


Source/Blazorise/Components/DatePicker/DatePicker.razor.cs
```diff
@@ -33,10 +33,14 @@ public partial class DatePicker<TValue> : BaseTextInput<IReadOnlyList<TValue>>,
     /// <inheritdoc/>
     public override async Task SetParametersAsync( ParameterView parameters )
     {
+        var datesUsed = parameters.TryGetValue( nameof( Dates ), out IEnumerable<TValue> paramDates );
```
David-Moreira, line 36: Isn't this always true?

Source/Blazorise/Components/DatePicker/DatePicker.razor.cs
```diff
@@ -33,10 +33,14 @@ public partial class DatePicker<TValue> : BaseTextInput<IReadOnlyList<TValue>>,
...
+
         if ( Rendered )
         {
-            var dateChanged = parameters.TryGetValue<TValue>( nameof( Date ), out var paramDate ) && !Date.Equals( paramDate );
-            var datesChanged = parameters.TryGetValue( nameof( Dates ), out IEnumerable<TValue> paramDates ) && !Dates.AreEqual( paramDates );
+            var dateUsed = parameters.TryGetValue<TValue>( nameof( Date ), out var paramDate );
```
David-Moreira, line 40: Isn't this always true?

Source/Blazorise/Components/Validation/ValidationRule.cs
```diff
@@ -182,5 +184,29 @@ public static void IsFileSelected( ValidatorEventArgs e )
...
+    /// <summary>
+    /// Checks if the date is selected.
+    /// </summary>
+    /// <typeparam name="TValue">Data-type used be binded by the Date property.</typeparam>
+    /// <param name="e"></param>
+    public static void IsDateSelected<TValue>( ValidatorEventArgs e )
+    {
+        var dates = e.Value as TValue[];
+
+        e.Status = dates?.Count( x => !x.IsEqual( default ) ) >= 1 ? ValidationStatus.Success : ValidationStatus.Error;
```
David-Moreira, line 196: `Any?` might shortcut sooner? Also conveys meaning a little bit better?

Source/Blazorise/Components/Validation/ValidationRule.cs
```diff
@@ -182,5 +184,29 @@ public static void IsFileSelected( ValidatorEventArgs e )
...
+    /// <summary>
+    /// Checks if the date range is selected.
+    /// </summary>
+    /// <typeparam name="TValue">Data-type used be binded by the Dates property.</typeparam>
+    /// <param name="e"></param>
+    public static void AreDatesSelected<TValue>( ValidatorEventArgs e )
+    {
+        var dates = e.Value as IReadOnlyList<TValue>;
+
+        e.Status = dates?.Count( x => !x.IsEqual( default ) ) >= 2 ? ValidationStatus.Success : ValidationStatus.Error;
```
David-Moreira, line 208: Ah, I see why you used Count up above...
I have to say I don't exactly get what this is doing, it has a bunch of dates that represent date ranges, that it? So whenever there's a data range it's valid?

Could for example the count be 3 and still be valid?

David-Moreira (changes requested):

Source/Blazorise/Components/DatePicker/DatePicker.razor.cs
stsrki: It is not. Blazor will only pass what is been explicitly defined as a parameter,

Source/Blazorise/Components/DatePicker/DatePicker.razor.cs
stsrki: Same as above.

Source/Blazorise/Components/Validation/ValidationRule.cs
stsrki: Any selection that is above or equal to 2 is valid. This is only a default behavior that is true for most use-cases. For any custom validation users would have to write their own logic for their business.

Source/Blazorise/Components/Validation/ValidationRule.cs
stsrki: Any might be cleaner, I agree.

Source/Blazorise/Components/DatePicker/DatePicker.razor.cs
David-Moreira: I honestly didn't know about that. I thought every parameter was always populated with defaults in that collection. Guess I was wrong ^^

Source/Blazorise/Components/Validation/ValidationRule.cs
```diff
@@ -182,5 +184,36 @@ public static void IsFileSelected( ValidatorEventArgs e )
...
+    /// <summary>
+    /// Checks if the date is selected.
+    /// </summary>
+    /// <typeparam name="TValue">Data-type used be binded by the Date property.</typeparam>
+    /// <param name="e"></param>
+    public static void IsDateSelected<TValue>( ValidatorEventArgs e )
+    {
+        var dates = e.Value as IEnumerable<TValue>;
+
+        if ( dates is not null )
```
David-Moreira, line 196: What happens if eventually null?

In the `AreDatesSelected` the `ValidationStatus.Error` is explicitly set.

Source/Blazorise/Components/DatePicker/DatePicker.razor.cs
```diff
@@ -106,17 +110,35 @@ public override async Task SetParametersAsync( ParameterView parameters )
...
-                await ParentValidation.InitializeInputExpression( expression );
+            if ( datesUsed )
+            {
+                if ( parameters.TryGetValue<Expression<Func<IReadOnlyList<TValue>>>>( nameof( DatesExpression ), out var datesExpression ) )
+                    await ParentValidation.InitializeInputExpression( datesExpression );
+
+                if ( parameters.TryGetValue<string>( nameof( Pattern ), out var pattern ) )
+                {
+                    // make sure we get the newest value
+                    var value = parameters.TryGetValue<IReadOnlyList<TValue>>( nameof( Dates ), out var inDates )
```
David-Moreira, line 121: Don't we already do this in the begging of the method? Also probably can remove the comment in favor of renaming the variable to `newestValue`. 
As usual, if we feel like there's a need to add a comment, then maybe the code can be made more explicit by renaming variables or using methods.

David-Moreira: I know you know this, but it's a good reminder sometimes... :)

David-Moreira (changes requested):

PR #5549, opened 2024-06-01
DataGrid: add SetLoading API

David-Moreira:
I tried something like this, but that didn't seem to yield good results or at least seemed quirky. even the current version I'm not too sure about calling StateHasChanged, but that might be better for the user?

```
if ( isLoading && !IsLoading )
{
    IsLoading = true;
    await Task.Yield();
}

if ( !isLoading && IsLoading )
{
    IsLoading = false;
}
```


Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -942,6 +942,17 @@ private async void OnPageChanged( int currentPage )
...
     #region Commands
 
+    /// <summary>
+    /// Sets the DataGrid into the loading state. 
+    /// <para>Makes sure to invoke the StateHasChanged method.</para>
+    /// </summary>
+    /// <param name="isLoading">Whether the grid is loading or not</param>
+    public void SetLoading(bool isLoading)
+    {
+        IsLoading = isLoading;
```
stsrki, line 952: Should be also make IsLoading public? But only the `get` part.

David-Moreira: Done.

stsrki (approved):

PR #5576, opened 2024-06-19
Link: automatically add rel tag

### LINKED ISSUE #5565: [Link] Define rel attribute for Blank links
When `Target` is set to `Blank` we should automatically add a `rel` attribute:

- `rel="nofollow noreferrer"`, or
- `rel="noopener noreferrer nofollow"`, or
- `rel="noopener noreferrer"`

---

Resources https://stackoverflow.com/questions/74633505/security-of-html5-target-blank-attribute

stsrki:
Closes #5565

ChatGPT explained it better

When setting the Target attribute of a link to _blank in the Blazorise Link component, it is crucial to include the rel attribute to prevent potential security risks and improve SEO. The options provided are:

- rel="nofollow noreferrer"
- rel="noopener noreferrer nofollow"
- rel="noopener noreferrer"

Here's an analysis of each option:

### Option 1: rel="nofollow noreferrer"

- nofollow: Tells search engines not to follow the link, preventing the transfer of SEO value to the linked page.
- noreferrer: Prevents the browser from sending the referrer information to the target site.

### Option 2: rel="noopener noreferrer nofollow"

- noopener: Ensures that the new page runs in a separate process, which prevents the new page from being able to access the window.opener property and potentially gain control over the original page.
- noreferrer: As above, prevents the referrer information from being sent.
- nofollow: As above, prevents search engines from following the link.

### Option 3: rel="noopener noreferrer"

- noopener: As above, prevents the new page from being able to access the window.opener property.
- noreferrer: As above, prevents the referrer information from being sent.

Recommendation

The best option to use in the Blazorise Link component when the Target is set to _blank is:

- rel="noopener noreferrer"

Reasoning:

- Security: Using noopener is crucial for security as it prevents the new page from accessing the original page through the window.opener property, mitigating the risk of malicious activities such as phishing.
- Privacy: noreferrer ensures that no referrer information is sent to the new page, preserving user privacy.
- SEO: Including nofollow in the rel attribute can impact SEO by instructing search engines not to pass on link equity. While this might be desirable in some specific cases, it generally reduces the SEO benefit of linking to other pages. Unless you have a specific need to prevent search engines from following these links, it is typically better to omit nofollow.

By using rel="noopener noreferrer", you ensure a secure and privacy-conscious implementation without unnecessarily impacting SEO. If there is a specific requirement for nofollow, it can be added as needed on a case-by-case basis.

### DIFF
--- a/Documentation/Blazorise.Docs/Pages/News/2024-07-15-release-notes-160.razor
+++ b/Documentation/Blazorise.Docs/Pages/News/2024-07-15-release-notes-160.razor
@@ -107,6 +107,14 @@
     Added a new <Code>SetLoading</Code> so you're able to control the loading state of the DataGrid.
 </Paragraph>
 
+<Heading Size="HeadingSize.Is4">
+    Automatic <Code>rel</Code> Attribute for Blazorise Link Component
+</Heading>
+
+<Paragraph>
+    Links with <Code>Target="Target.Blank"</Code> now automatically include <Code>rel="noopener noreferrer"</Code> for enhanced security and privacy. This update prevents new pages from accessing the original page, reducing phishing risks, and ensures referrer information is not sent to the target site, thereby enhancing the security and privacy of external links by default.
+</Paragraph>
+
 <Heading>
     Wrap Up
 </Heading>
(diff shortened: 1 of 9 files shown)


David-Moreira (approved):
LGTM, could this produce a breaking change in any way?

stsrki:
> LGTM, could this produce a breaking change in any way?

Nothing breaking. This will prevent any security risks by giving the browser some extra hints. The user is not affected by it.

PR #5577, opened 2024-06-19
InputMask: add new MaskPlaceholder parameter

### LINKED ISSUE #5560: [Bug]: InputMask and placeholder
### Blazorise Version

latest

None

N/a

### Steps to reproduce

`<InputMask ShowMaskOnFocus="false" ShowMaskOnHover="false" Mask="****-****-****" Background="Background.White" Placeholder="Enter HUID (e.g. hdc1-tst1-hcu1)" />`

### What is expected?

Shouldn't use placeholder as mask

The placeholder become a mask it's totally wtf

_No response_

### Any additional comments?

_No response_

stsrki:
Closes #5560

### DIFF
--- a/Documentation/Blazorise.Docs/Pages/Docs/Components/InputMasks/InputMaskPage.razor
+++ b/Documentation/Blazorise.Docs/Pages/Docs/Components/InputMasks/InputMaskPage.razor
@@ -26,8 +26,12 @@
     </UnorderedListItem>
 </UnorderedList>
 
+<DocsPageSubtitle>
+    Examples
+</DocsPageSubtitle>
+
 <DocsPageSection>
-    <DocsPageSectionHeader Title="Basic Example">
+    <DocsPageSectionHeader Title="Basic Mask">
         To start you only need to define the <Code>Mask</Code> parameter.
     </DocsPageSectionHeader>
     <DocsPageSectionContent Outlined FullWidth>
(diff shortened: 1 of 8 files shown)


Documentation/Blazorise.Docs/Pages/Docs/Components/InputMasks/InputMaskPage.razor
```diff
@@ -26,8 +26,12 @@
     </UnorderedListItem>
 </UnorderedList>
 
+<DocsPageSubtitle>
+    Examples
+</DocsPageSubtitle>
+
 <DocsPageSection>
-    <DocsPageSectionHeader Title="Basic Example">
```
David-Moreira, line 30: We do use the Example suffix in more places, if we now don't like this, should it be changed widely?

Source/Blazorise/Components/InputMask/InputMask.razor.cs
```diff
@@ -71,7 +71,7 @@ protected override async Task OnFirstAfterRenderAsync()
             Alias,
             InputFormat,
             OutputFormat,
-            Placeholder,
```
David-Moreira, line 74: Strange, I see this Placeholder being removed here, but I don't see a property reference being removed? Am I missing something?

David-Moreira (approved):

Source/Blazorise/Components/InputMask/InputMask.razor.cs
David-Moreira: Or is it staying with a different purpose?

stsrki: MaskPlaceholder is now passed to JS lib.

Documentation/Blazorise.Docs/Pages/Docs/Components/InputMasks/InputMaskPage.razor
David-Moreira: My bad, I was under the impression we used Example alot more.

Source/Blazorise/Components/InputMask/InputMask.razor.cs
David-Moreira: I see that. I meant that `Placeholder` was removed, but I guess it now serves a diffrent purpose, which is not the same as the maskplaceholder.

stsrki: Yes. Now, it is only applied directly to the input.

PR #5578, opened 2024-06-19
DatePicker: add DisabledDays parameter

### LINKED ISSUE #5543: Disabled Days
I'm looking for a Blazor date-time picker to use for selecting delivery date on an eCommerce site. The problem is that as well as disabling particular _**dates**_ we also need to the ability to disable _**days**_ of the week (no post on Sundays).

Ideally I'd imagine this would work almost exactly the same as the existing `DisabledDates` attribute, but instead of an array of `DateTime` values it could take an array of `DayOfWeek` values.

stsrki:
Closes #5543

### DIFF
--- a/Source/Blazorise/wwwroot/datePicker.js
+++ b/Source/Blazorise/wwwroot/datePicker.js
@@ -42,6 +42,11 @@ export function initialize(dotnetAdapter, element, elementId, options) {
     const mutationObserver = new MutationObserver(mutationObserverCallback);
     mutationObserver.observe(document.getElementById(elementId), { attributes: true });
 
+    const disableDatesOptions = options.disabledDates || [];
+    const disableDaysOptions = options.disabledDays ? [function (date) {
+        return options.disabledDays && options.disabledDays.length > 0 && options.disabledDays.includes(date.getDay());
+    }] : [];
+
     const defaultOptions = {
         enableTime: options.inputMode === 1,
         dateFormat: options.inputMode === 1 ? 'Y-m-d H:i' : 'Y-m-d',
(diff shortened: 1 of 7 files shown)


Source/Blazorise/wwwroot/datePicker.js
```diff
@@ -42,6 +42,11 @@ export function initialize(dotnetAdapter, element, elementId, options) {
     const mutationObserver = new MutationObserver(mutationObserverCallback);
     mutationObserver.observe(document.getElementById(elementId), { attributes: true });
 
+    const disableDatesOptions = options.disabledDates || [];
+    const disableDaysOptions = options.disabledDays ? [function (date) {
```
David-Moreira, line 46: Just to make sure, does this trully take in an array of functions?

David-Moreira: Think it's safe to assume that the date they pass in is valid? If we usually null check everything that's ours, then something from a 3rd party lib which we might not be sure, it's probably safer if we do instead of trusting it?

Source/Blazorise/wwwroot/datePicker.js
```diff
@@ -56,7 +61,7 @@ export function initialize(dotnetAdapter, element, elementId, options) {
         },
         time_24hr: options.timeAs24hr ? options.timeAs24hr : false,
         clickOpens: !(options.readOnly || false),
-        disable: options.disabledDates || [],
+        disable: disableDatesOptions.concat(disableDaysOptions),
```
David-Moreira, line 64: have we tried with 2 functions, and it works correctly?

David-Moreira (approved):

Source/Blazorise/wwwroot/datePicker.js
stsrki: Yes, the lib work with date values and with functions. Any they can be mixed.

Source/Blazorise/wwwroot/datePicker.js
stsrki: I've tried it locally.

PR #5579, opened 2024-06-19
BarLink: add missing CustomMatch parameter

### LINKED ISSUE #5567: [Bug]: CustomMatch is not settable on BarLink
### Blazorise Version

1.3

None

```
<BarItem>
      <BarLink To="MyUrl"
               Match="Match.Custom"
               CustomMatch="(url) => true"
               Padding="Padding.Is2.OnX">
          <BarIcon Class="notranslate"
                   IconName="Blazorise.Icons.Material.MaterialIcons.People" />
          MyUrl
      </BarLink>
  </BarItem>
```

### Steps to reproduce

Compile the code :)

### What is expected?

As the BarLink allows to use Match.Custom, it should also expose the CustomMatch Func, just like the Link and Anchor components.

Unfortunately this method is not exposed.

_No response_

### Any additional comments?

Additionally I noticed that the type shown for "Match" in the "BarLink" documentation is not correct. It is of type Match and not of type bool: https://blazorise.com/docs/components/bar

For now we are using Anchor with the class "b-bar-link" as a workaround.

stsrki:
Closes #5567

I also added the same on SidebarLink.

### DIFF
--- a/Documentation/Blazorise.Docs/Pages/Docs/Components/Links/LinkPage.razor
+++ b/Documentation/Blazorise.Docs/Pages/Docs/Components/Links/LinkPage.razor
@@ -122,6 +122,9 @@
         URL matching behavior for a link.
     </DocsAttributesItem>
     <DocsAttributesItem Name="CustomMatch" Type="Func<string, bool>" Default="null">
+        A callback function that is used to compare current uri with the user defined uri. Must enable <Code>Match="Match.Custom"</Code> to be used.
+    </DocsAttributesItem>
+    <DocsAttributesItem Name="CustomMatch" Type="Func<string, bool>" Default="null">
         A callback function that is used to compare current uri with the user defined uri. Must enable <Code>Match.Custom</Code> to be used.
     </DocsAttributesItem>
     <DocsAttributesItem Name="Title" Type="string" Default="null">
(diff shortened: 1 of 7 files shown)


Documentation/Blazorise.Docs/Pages/Docs/Components/Links/LinkPage.razor
```diff
@@ -121,6 +121,9 @@
     <DocsAttributesItem Name="Match" Type="Match" Default="Prefix">
         URL matching behavior for a link.
     </DocsAttributesItem>
+    <DocsAttributesItem Name="CustomMatch" Type="Func<string, bool>" Default="null">
+        A callback function that is used to compare current uri with the user defined uri. Must enable <Code>Match="Match.Custom"</Code> to be used.
```
David-Moreira, line 125: Why do we need to enable `Match.Custom`? Just by providing the function, should be enough on itself?
We're getting on a place where `CustomMatch` can be defined and not work at all, because `Match.Custom` was not defined, and I'm not sure there's a benefit to that, it's just needlessly more verbose for the user?

David-Moreira (approved):

David-Moreira:
**PS:** Missing release notes

Documentation/Blazorise.Docs/Pages/Docs/Components/Links/LinkPage.razor
stsrki: We could change it but that is a breaking change then. Maybe for 2.0.

stsrki:
> **PS:** Missing release notes

It's a small change so I will skip release notes and just list it in the changelog.

Documentation/Blazorise.Docs/Pages/Docs/Components/Links/LinkPage.razor
David-Moreira: Just open a 2.0 task then and keep it like this for now I guess.

stsrki: done

PR #5582, opened 2024-06-20
Datagrid add toodatastring api to the readdata

David-Moreira:


David-Moreira:
@stsrki Any customer that would be willing to take this for a spin using 1.6 preview?

stsrki:
> @stsrki Any customer that would be willing to take this for a spin using 1.6 preview?

I have someone in mind.

mrpmorris (commented):
1. Edit the page at tests/datagrid
2. In OnReadData add `string oDataQuery = e.ToODataString( "https://ibm.com" );`
3. Run the app, visit the page
4. Enter `gmail` in the email filter

Expected: `oDataQuery` should have a `$filter` entry
Actual: It doesn't

David-Moreira:
Fixed @mrpmorris Would you be so kind to retest?

@stsrki The problem here, was that the column filter method is null on read data, which is somewhat correct since these columns have no FilterMethod specifically defined. 
However, I think it should be fine, that if it's not defined at column level, we propagate the current DataGrid filter method as I've just changed. Please let me know your thoughts.

stsrki:
> @stsrki The problem here, was that the column filter method is null on read data, which is somewhat correct since these columns have no FilterMethod specifically defined.
> However, I think it should be fine, that if it's not defined at column level, we propagate the current DataGrid filter method as I've just changed. Please let me know your thoughts.

Yeah, I think that should be fine.

stsrki:
@mrpmorris mentioned a way to test the filter quickly. 

```
...

public Task<List<Employee>> FilterData( IEnumerable<DataGridColumnInfo> dataGridColumns, string oDataQuery ) <-- Added string parameter
...

        var values =
            new Uri( oDataQuery )
            .Query
            .ToKeyValuePairs()
            .ToDictionary( x => x.Key, x => HttpUtility.UrlDecode( (string)x.Value ) );

        var source = filteredData.AsQueryable().OData();
        if (values.TryGetValue("$filter", out string f))
            source = source.Filter( (string)f );


        //foreach ( var column in dataGridColumns.Where( x => !string.IsNullOrWhiteSpace( x.SearchValue?.ToString() ) ) )
        //{
        //    var valueGetter = FunctionCompiler.CreateValueGetter<Employee>( column.Field );
        //    filteredData = filteredData.Where( x => valueGetter( x )?.ToString().IndexOf( column.SearchValue.ToString(), StringComparison.OrdinalIgnoreCase ) >= 0 ).ToList();
        //}
        return Task.FromResult( source.ToList() );
    }
```

and

`<PackageReference Include="Community.OData.Linq" Version="2.1.24-rc" />`

David-Moreira:
> LGTM.
> 
> @David-Moreira If you think you will not need anything else I will merge it.

Yep, fine by me.

PR #5585, opened 2024-06-22
DataGrid : Add Check | Date cell edit components | Filter by contextual components

David-Moreira:
Further improving upon contextual Filtering

Finally centralized the filter components in a `_DataGridCellFilter` component that's reused by both standard filter and menu filter.

It now renders, **numeric**, **select**, **check**, **date** and **text** edit components according to the user's provided `DataGrid` column.


Source/Extensions/Blazorise.DataGrid/_DataGridCellDateEdit.razor
```diff
@@ -0,0 +1,41 @@
...
+
+@code {
+    private RenderFragment dateFragment => builder =>
+    {
+        var type = typeof( DateEdit<> ).MakeGenericType( new[] { valueType } );
+
+        builder.OpenComponent( 0, type );
+        builder.AddAttribute( 1, nameof( DateEdit<object>.Date ), valueType switch
+        {
+            Type typeDateTime when typeDateTime == typeof( DateTime ) => (DateTime)( CellValue ?? (DateTime)default ),
```
stsrki, line 18: I haven't checked. But how performant is all this checking for a type? Can we do the check during OnInitialized, store the information in local varaobles, and then just render based on that?

stsrki: I don't suppose it will be a problem, though, since this is only rendered in a table header, so not a lot of render cycles should be made.

stsrki (approved):

PR #5593, opened 2024-06-26
Calculate input name attribute in SSR mode

### LINKED ISSUE #5592: [Bug]: Form inputs do not have a name
### Blazorise Version

1.5.2

Bootstrap5

N/A

### Steps to reproduce

Create a page with form binding to a model. Use the Blazor <InputText> and also <TextEdit>

### What is expected?

The generated html should have a name, just as InputText does.

Blazor InputText adds a name attribute to InputText etc, this allows Blazor SSR <form> binding.

Blazorise TextEdit does not do this, so the built in form binding does not work

**<InputText @bind-Value=@Model.Account/>**
<input aria-invalid="true" **name="Model.Account"** class="invalid" value="asd">

**<TextEdit @bind-Text=@Model.Account>**
<input id="0HN4L55VES4O2" type="text" class="form-control is-invalid" aria-readonly="false" value="asd">

Microsoft Edge

### Any additional comments?

N/A

stsrki:
Closes #5592 

At first, I tried using `FieldIdentifier`, but we had several problems.

1. FieldIdentifier is only available on Validations. I tried to bring it down to input level, but it was too complex and it broke the validation. We can reconsider this for 2.0 as I have realized we had to break some APIs also.
2. The FieldIdentifier was generating invalid names, eg. `FieldName` instead of `Model.FieldName`.

So, I went and saw how they do it in .NET core. I have copied their helpers, who are internal only and brought them to Blazorise. Then, I used some of the code to generate the name attribute. simplified it, and made it to work with our components. It works as expected.

One problem, though. On our `Radio` component, we already have the `Name` parameter. I still need to see what to do with it.

I still want to see if there is only room for improvement. This is only a first run.

---

@mrpmorris Any chance you could test it?

### DIFF
--- a/Source/Blazorise/Base/BaseInputComponent.razor.cs
+++ b/Source/Blazorise/Base/BaseInputComponent.razor.cs
@@ -65,12 +65,7 @@ public abstract class BaseInputComponent<TValue> : BaseComponent, IValidationInp
     /// <inheritdoc/>
     public override async Task SetParametersAsync( ParameterView parameters )
     {
-        if ( !hasInitializedParameters )
-        {
-            shouldGenerateFieldNames = !OperatingSystem.IsBrowser();
-
-            hasInitializedParameters = true;
-        }
+        InitializeParameters();
 
         await base.SetParametersAsync( parameters );
 
@@ -152,7 +147,7 @@ public abstract class BaseInputComponent<TValue> : BaseComponent, IValidationInp
     /// <summary>
     /// Initializes parameters that are not going to be changed again.
     /// </summary>
-    protected void IntializeParameters()
+    protected void InitializeParameters()
     {
         if ( hasInitializedParameters )
             return;
--- a/Source/Blazorise/Base/BaseInputComponent.razor.cs
+++ b/Source/Blazorise/Base/BaseInputComponent.razor.cs
@@ -5,7 +5,6 @@ using Blazorise.Extensions;
 using Blazorise.Modules;
 using Blazorise.Utilities;
 using Microsoft.AspNetCore.Components;
-using Microsoft.AspNetCore.Components.Forms;
 using Microsoft.AspNetCore.Components.Web;
 #endregion
 
@@ -70,11 +69,11 @@ public abstract class BaseInputComponent<TValue> : BaseComponent, IValidationInp
         await base.SetParametersAsync( parameters );
 
         // For modals we need to make sure that autofocus is applied every time the modal is opened.
-        if ( parameters.TryGetValue<bool>( nameof( Autofocus ), out var autofocus ) && this.autofocus != autofocus )
+        if ( parameters.TryGetValue<bool>( nameof( Autofocus ), out var paramAutofocus ) && autofocus != paramAutofocus )
         {
-            this.autofocus = autofocus;
+            autofocus = paramAutofocus;
 
-            if ( autofocus )
+            if ( paramAutofocus )
             {
                 if ( ParentFocusableContainer is not null )
                 {
--- a/Source/Blazorise/Base/BaseInputComponent.razor.cs
+++ b/Source/Blazorise/Base/BaseInputComponent.razor.cs
@@ -3,7 +3,9 @@ using System;
 using System.Threading.Tasks;
 using Blazorise.Extensions;
 using Blazorise.Modules;
+using Blazorise.Utilities;
 using Microsoft.AspNetCore.Components;
+using Microsoft.AspNetCore.Components.Forms;
 using Microsoft.AspNetCore.Components.Web;
 #endregion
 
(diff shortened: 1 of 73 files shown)


Source/Blazorise/Base/BaseInputComponent.razor.cs
```diff
@@ -125,6 +149,20 @@ protected virtual void ReleaseResources()
         }
     }
 
+    /// <summary>
+    /// Initializes parameters that are not going to be changed again.
+    /// </summary>
+    protected void IntializeParameters()
```
David-Moreira, line 155: Did you end up not using this method? Can you check?

Source/Blazorise/Base/BaseInputComponent.razor.cs
```diff
@@ -125,6 +149,20 @@ protected virtual void ReleaseResources()
...
+    {
+        if ( hasInitializedParameters )
+            return;
+
+        // we only need to generate field names on the server side
+        shouldGenerateFieldNames = !OperatingSystem.IsBrowser();
+
+        hasInitializedParameters = true;
```
David-Moreira, line 163: Might prefer to move this up, just to be sure there's no way this code can execute twice? Although I think it probably never will.

Source/Blazorise/Base/BaseInputComponent.razor.cs
```diff
@@ -125,6 +149,20 @@ protected virtual void ReleaseResources()
...
+    /// <summary>
+    /// Initializes parameters that are not going to be changed again.
+    /// </summary>
+    protected void IntializeParameters()
+    {
+        if ( hasInitializedParameters )
+            return;
+
+        // we only need to generate field names on the server side
+        shouldGenerateFieldNames = !OperatingSystem.IsBrowser();
```
David-Moreira, line 161: Should we try/catch instead of trusting the underlying code? Nowadays it seems like Microsoft introduces too many modes/variables... Maybe just to be sure, this doesn't break in some mode for some reason...

Source/Blazorise/Components/DatePicker/DatePicker.razor.cs
```diff
@@ -477,6 +477,17 @@ private object GetLocalizationObject()
     /// <inheritdoc/>
     protected override bool IsSameAsInternalValue( IReadOnlyList<TValue> value ) => value.AreEqual( InternalValue );
 
+    /// <inheritdoc/>
+    protected override string GetFormatedValueExpression()
+    {
+        if ( DateExpression is null )
```
David-Moreira, line 483: You created an issue to rename stuff to **Value**. **Expression** should be included I guess. If Expression is somehow centralized like Expression<Func<TValue>> something along these lines, then we might not need to override this per component, and just do it on base?

David-Moreira (approved):
LGTM, I didn't bother going through the microsoft classes.

Source/Blazorise/Base/BaseInputComponent.razor.cs
stsrki: No, I left it on purpose. I plan to use it more in the future. It will be part of the refactoring `Value`.

Source/Blazorise/Base/BaseInputComponent.razor.cs
stsrki: I doubt. The logic is practically the same as in .NET project. If they think it will not be executed twice I guess it is fine.

Source/Blazorise/Base/BaseInputComponent.razor.cs
stsrki: Again, the same as in .NET project. The `OperatingSystem.IsBrowser()` basically just checks that we are in WASM browser mode. Should be safe. I don't see any exception in their API docs https://learn.microsoft.com/en-us/dotnet/api/system.operatingsystem.isbrowser?view=net-8.0

Source/Blazorise/Components/DatePicker/DatePicker.razor.cs
stsrki: Exactly. That is the plan with `ValueExpression`. All these overrides are needed because we have have too many different APIs. With `Value`, we will centralize it in the base input component.

Source/Blazorise/Base/BaseInputComponent.razor.cs
stsrki: You were right. It was a duplicate. Thanks.

Source/Blazorise/Base/BaseInputComponent.razor.cs
David-Moreira: Don't forget that Microsoft employees are still humans, and they can also make mistakes. Anyway I also doubt, I just took a look inside IsBrowser and its returning super immediately, so yes.
The "correct way" for this case, could be as stated.

PR #5602, opened 2024-06-29
Chart: Zoom plugin component

David-Moreira:
https://www.chartjs.org/chartjs-plugin-zoom/latest/


Source/Extensions/Blazorise.Charts/Options/ChartZoom/ChartZoom.cs
stsrki, line 1: All Zoom options must be inside of the new `Blazorise.Charts.Zoom` project. And must be used the same way as other "plugins". To configure the `plugin` options you can use code similar as `Blazorise.Charts.Annotation`, or `Blazorise.Charts.DataLabels`

Example from annotations:

```
import { getChart } from "../Blazorise.Charts/charts.js?v=1.5.3.0";

export function addAnnotation(canvasId, options) {
    const chart = getChart(canvasId);

    if (chart) {
        if (options) {
            if (!chart.options.plugins) {
                chart.options.plugins = [];
            }

            chart.options.plugins.annotation = adjustOptions(options);
        }

        chart.update();
    }
}

function adjustOptions(options) {
    return { annotations: options };
}
```

David-Moreira: I see, my bad, the plugin seemed so simple that a dedicated project/package wasn't warranted. Thanks for the heads up and directions, I will make the recommended changes.

David-Moreira:
It is of note that the zoom plugin has an inner zoom options so I ended up naming the outer one **ChartZoomPluginOptions** please let me know if you find a better naming scheme.

stsrki (approved):
Made some small modifications with naming and formating mostly. The chart works great. Great job.

PR #5608, opened 2024-07-03
Tables & DataGrid: Mobile Mode

David-Moreira:


David-Moreira:
Here's an initial implementation of this based on https://benjamin-keller.github.io/bootstrap-mobile-tables/

image

What are your thoughts?

Also the way we show the caption in the row is by making use of the data-content, so I introduced a `MobileModeTitle` in order to test it, but I'm not sure it's a good api to have to define these? The alternative would be to have to somehow resolve the correct table header column and extract the ChildContent from there? also probably would not work correctly since it's a renderfragment and can have anything.
Do you have any suggestions on how to tackle this?

stsrki:
I haven't tried it yet. But, how good, or bad, will this feature work with

- virtualization
- fixed headers
- fixed columns

Should we say that once mobile mode is enabled, all of the other features are disabled?

David-Moreira:
> * fixed headers
> * fixed columns

Most likely, there will have to be concessions you can't really expect those to work and some might not even make sense? And there's probably more stuff that won't work correctly.

stsrki:
> > * fixed headers
> > * fixed columns
> 
> Most likely, there will have to be concessions you can't really expect those to work and some might not even make sense? And there's probably more stuff that won't work correctly.

Exactly. And we need to make it clear for everyone so that they know what to expect. Also, need to work for all providers.

David-Moreira:
> > > * fixed headers
> > > * fixed columns
> > 
> > 
> > Most likely, there will have to be concessions you can't really expect those to work and some might not even make sense? And there's probably more stuff that won't work correctly.
> 
> Exactly. And we need to make it clear for everyone so that they know what to expect. Also, need to work for all providers.

Well I'm more worried about getting something working decently first. Like I said, I worry about how should we set the header captions in these cases, I'd like your help, see if you have any suggestions.

Source/Blazorise/Styles/_table.scss
```diff
@@ -27,3 +27,68 @@ thead {
         position: relative;
     }
 }
+
+
+/*-- ==============================================================
+ Screen smaller than 760px and iPads.
+ ============================================================== */
+
+@media only screen and (max-width: 760px), (min-device-width: 768px) and (max-device-width: 1024px) {
```
stsrki, line 36: We might want to define media breakpoints per CSS provider since all of them have a different mobile size.

Source/Blazorise/Components/Table/Table.razor.cs
```diff
@@ -70,6 +71,7 @@ protected override async Task OnAfterRenderAsync( bool firstRender )
     protected override void BuildClasses( ClassBuilder builder )
     {
         builder.Append( "b-table" );
+        builder.Append( "b-table-mobile", ResponsiveMode == TableResponsiveMode.Mobile );
```
stsrki, line 74: Make "b-table-mobile" as part of `IClassProvider`. And maybe name it accordoing to each provider naming scheme.

Also make it receive the TableResponsiveMode enum in case we want to resolve different modes in the future. Default will return null.

stsrki:
Regarding the MobileModeTitle. I would try to use ChildContent. I know it can be anything besides text but I would communicate the limitations of mobile mode once again. No need to introduce a new API at this time.

David-Moreira:
> Regarding the MobileModeTitle. I would try to use ChildContent. I know it can be anything besides text but I would communicate the limitations of mobile mode once again. No need to introduce a new API at this time.

Ok but my problem is that, it's buried under different components. 
You have to do quite some gynmnastics to get it.
So we'd need to track every component under table, and then match respectively assuming they are under the same indexes, does this make sense?

```
<Table>
<TableHeader>
<TableHeaderCell>
I have to get this caption to allocate on the TableRowCell
</TableHeaderCell>
<TableBody>
<TableRow>
<TableRowCell>
Content
</TableRowCell>
</TableRow>

</Table>

```

David-Moreira:
Also I haven't tried doing `data-content="@ChildContent"`. I am assuming it will work, but it might also not. 
Thoughts?

David-Moreira:
> Also I haven't tried doing `data-content="@ChildContent"`. I am assuming it will work, but it might also not. Thoughts?

Seems like it's out of the question. it doesn't get rendered at all.
image
image
image

Thoughts? :(

Source/Blazorise/Styles/_table.scss
```diff
@@ -27,3 +27,68 @@ thead {
...
+
+    /* Default layout */
+    .b-table-mobile td {
+        /* Behave  like a "row" */
+        border: none;
+        position: relative;
+        padding-top: 1.5rem !important;
+    }
+
+    .b-table-mobile td:before {
```
stsrki, line 79: Can we just have a CSS rule for `td` placed inside of `thead`? Would that solve the problem of not need the `data-content`?

David-Moreira: Can you elaborate?

stsrki: I mean to have `b-table-mobile thead td:before`.

David-Moreira: The structure is quite different, you have to somehow extract an header value into some other place. Or change the css in a way that somehow each table header cell would be followed by the corresponding table body row cell, which I don't think it's possible with just css.

What I am thinking now, is that doing JS and extracting whatever value is on the header will be more resilient then trying to do it on Blazor/css.

David-Moreira: > I mean to have `b-table-mobile thead td`.

What does that accomplish differently? Now you would have the rows at head level, and you need to grab the corresponding values from tbody columns somehow.
That's what the data-content trick does basically...

stsrki: I understand the problem now. I agree we have two options.

- Introduce a new API, as you have already done
- Use JS to do the heavylifint

The new API is the best way forward. The user must enable the mobile mode, and adding one more attribute to the table cell should not be much of a problem. Not to mention that it is much easier for us also.

stsrki: Yeah, fot DataGrid it is simple as it will be done automatically. It is only a problem if they construct the table from scratch.

David-Moreira: Right, but my point was that the DataGrid is going to inherit limitations, also, one of them the Caption Template. :)

stsrki: I forgot the CaptionTemplate 😱

David-Moreira: Well, should we try the js based approach instead? I'm still conflicted. I feel like the js based approach might allow less limitations regarding the caption at least.

stsrki:
i would ignore the CapionTemplate for now. We want to release the feature. In the release notes, we will mention that it is still experimental and that we want to see how it is used before adding more features.

David-Moreira:
@stsrki It ended up looking quite decent I believe.
Please review.
image

Source/Blazorise.Bootstrap5/Styles/components/_table.scss
```diff
@@ -99,4 +99,41 @@ tr {
 
 td:focus {
     outline: -webkit-focus-ring-color auto 1px;
+}
+
+@media only screen and (max-width: 760px), (min-device-width: 768px) and (max-device-width: 1024px) {
+
+    .b-table-mobile {
```
stsrki, line 106: The classnames should follow the CSS provider naming, as close as it can. For BS it should be `table-mobile`.

Source/Blazorise.Bulma/Styles/components/_table.scss
```diff
@@ -89,4 +89,41 @@ tr {
 
 td:focus {
     outline: -webkit-focus-ring-color auto 1px;
+}
+
+@media only screen and (max-width: 760px), (min-device-width: 768px) and (max-device-width: 1024px) {
+
+    .b-table-mobile {
```
stsrki, line 96: Name it `is-table-mobile`.

Source/Blazorise.FluentUI2/Styles/components/_table.scss
```diff
@@ -340,4 +340,41 @@
 
 td:focus {
     outline: -webkit-focus-ring-color auto 1px;
+}
+
+@media only screen and (max-width: 760px), (min-device-width: 768px) and (max-device-width: 1024px) {
+
+    .b-table-mobile {
```
stsrki, line 347: Name it `fui-Table-mobile`.

Source/Blazorise.Tailwind/Styles/components/_table.scss
```diff
@@ -52,4 +52,41 @@ tr.b-table-row {
 
 td:focus {
     outline: -webkit-focus-ring-color auto 1px;
+}
+
+@media only screen and (max-width: 760px), (min-device-width: 768px) and (max-device-width: 1024px) {
+
+    .b-table-mobile {
```
stsrki, line 59: Name it `b-table-mobile`.

Documentation/Blazorise.Docs/Pages/Docs/Components/Tables/TablePage.razor
```diff
@@ -130,6 +130,21 @@
     <DocsPageSectionSource Code="TableDarkHeaderExample" />
 </DocsPageSection>
 
+<DocsPageSection>
```
stsrki, line 133: It might be good to implement desktop/mobile mode in the examples so that user can switch the view. This can be done by using the `iframe` and setting the width. 

Check how it is done on Blocks https://github.com/Megabit/BlazoriseSupport/blob/master/WebApps/BlazoriseSupport.Blocks/Components/BlockPageSectionContent.razor

stsrki (approved):
Looking good. If you think you're not going to add additional changes I will merge it.

David-Moreira:
> Looking good. If you think you're not going to add additional changes I will merge it.

Yea I'm ok with it.

PR #5609, opened 2024-07-04
Video: Update to VidstackPlayer library

### LINKED ISSUE #4914: Update Video player to new library
The Plyr.js is going to be obsolete by the new library at some point https://github.com/sampotts/plyr/issues/2408

We must plan for the upgrade once it is stabilized and exits beta.

New Library: https://www.vidstack.io/

### LINKED ISSUE #5479: Change video captions
Is there a way of changing captions of the video with custom buttons. What I mean by that is, I want to make my own button that displays available captions and by selecting them, they get updated in the video player.

I want this so that i can position and style my own buttons and menus with captions and make my application more customizable.

stsrki:
Closes #4914
Closes #5479

Updated Video to use VidstackPlayer library. The Plyr library was archived in favor of the new library https://github.com/sampotts/plyr/issues/2737

In this version, I have made it work as closely as before, and from my tests, it should not have any breaking changes.


Source/Extensions/Blazorise.Video/Utilities/VideoParsers.cs
```diff
@@ -0,0 +1,33 @@
...
+
+namespace Blazorise.Video.Utilities;
+
+internal static class VideoParsers
+{
+    public static double? ParseAspectRatio( string aspectRatio )
+    {
+        if ( string.IsNullOrEmpty( aspectRatio ) || !aspectRatio.Contains( ':' ) )
+        {
+            return null;
```
David-Moreira, line 11: Why does this one not warrant a special message? Specially the` !aspectRatio.Contains( ':' )`

David-Moreira (changes requested):
- I can't tell if changing the quality does anything
- Changing captions or speed does not seem to have any effect, I don't see any subtitles
- Sometimes when clicking changing video, nothing happens, page seems broken.

Source/Extensions/Blazorise.Video/Utilities/VideoParsers.cs
stsrki: Because in Plyr.js, Ratio is defined as a string, e.g., `"16:9"`, while in the new library, it needs to be described as a number `16 / 9`. So to not add a breaking change I just parse and convert it to a number.

David-Moreira: Also it is of note that we usually don't typify/return exceptions and just silently return null. Not sure if we want to be consistent with that or not.

stsrki: I am more open to going the silent route than raising an exception.

stsrki:
> * I can't tell if changing the quality does anything

This should not be changed, hence the DefaultQuality parameter name. I removed the change button.

> * Changing captions or speed does not seem to have any effect, I don't see any subtitles

Yeah, this seems to not work with Plyr layout while it works with default layout. I might report the issue on their repo.

> * Sometimes when clicking changing video, nothing happens, page seems broken.

For me it works. Cannot reproduce.

stsrki:
The menu should work now.

David-Moreira:
**DEMO:**
- Go To 20 seconds only works once. Maybe the `CurrentTime` needs to be bound?
- It would be nice if clicking change video would alternate between both videos
- Menu seems to work perfectly right now, but I still don't get any captions at all, even though the options are there? Is this intended?
- When you click stop, you can only scrub the video, you can no longer play/pause the video, the video just stays paused?

**DOCS:**
- Is a video missing here
image
- The Widevine DRM instantiation doesn't seem to play at all?
- The Multiple files example, displays 3 files being bound, but you can't really use the other files? Would it not be great to see it, being able to select any of the 3 files?

Demos/Blazorise.Demo/Pages/Tests/VideoPage.razor
```diff
@@ -16,31 +16,50 @@
...
             <CardBody>
-                <Video @ref="@videoRef" Source="@videoSource" CurrentTime="@currentTime" DefaultQuality="@quality" />
+                <Video @ref="@videoRef" Source="@videoSource" Thumbnails="https://files.vidstack.io/sprite-fight/thumbnails.vtt" CurrentTime="@currentTime" DefaultQuality="@defaultQuality" />
             </CardBody>
         </Card>
     </Column>
 </Row>
 @code {
     Video videoRef;
+    // VideoSource videoSource = new VideoSource()
```
David-Moreira, line 34: Do we want this old code?

Documentation/Blazorise.Docs/Pages/News/2024-07-15-release-notes-160.razor
```diff
@@ -71,6 +71,17 @@
     Visit our docs for more information, <Anchor To="docs/extensions/chart-zoom">Chart Zoom Plugin</Anchor>
 </Paragraph>
 
+<Heading Size="HeadingSize.Is3">
+    Video update
+</Heading>
+
+<Paragraph>
+    We have updatet our Video extension component to work with the new <Anchor To="https://www.vidstack.io/">VidStack</Anchor> library. Until now we have used a Plyr library, but it was slowly been replaced by a neweset and more modern VidStack library. So,  the obvious reason was for us also to make an update as it provides more features, and is much more stable. Not to mention that it also has many more features.
```
David-Moreira, line 79: - updatet => updated.
- we have used **the** Plyr
- it was slowly been replaced => it has slowly been replaced 
- neweset?
- So,  the obvious reason was for us also to make an update => Since the newer library is much more stable and provides more feature, we've went ahead and updated, implementing the new features, and making the transition as seamless as possible.
- Not to mention that it also has many more features. => Again?

I guess you'll review the notes when you release anyway.

Documentation/Blazorise.Docs/Pages/News/2024-07-15-release-notes-160.razor
```diff
@@ -71,6 +71,17 @@
...
+</Paragraph>
+
+<Paragraph>
+    In this update we have mostly focused to make it work without any breaking changes, or introduction of new features. We have tried our best to polish it but as always there might be some edgecases that we have missed. In case you find any issue please report it to us and we will try to fix them as soon as possible.
```
David-Moreira, line 83: Despite XXXX bringing many new features, we've focused on migrating to this new library without any breaking changes. We have tried... bla bla...

Be on the lookout for new features on the next version (like... ) => tease about whatever features might be relevant

Source/Extensions/Blazorise.Video/Video.razor.cs
```diff
@@ -77,22 +81,24 @@ protected override async Task OnAfterRenderAsync( bool firstRender )
...
                 Volume,
                 ClickToPlay,
                 DisableContextMenu,
                 ResetOnEnd,
-                Ratio,
+                AspectRatio = VideoParsers.ParseAspectRatio( Ratio ),
                 InvertTime,
-                DefaultQuality,
-                AvailableQualities,
+                DefaultQuality = new { Height = DefaultQuality },
```
David-Moreira, line 94: Only Requires Height?

Source/Extensions/Blazorise.Video/Video.razor.cs
```diff
@@ -77,22 +81,24 @@ protected override async Task OnAfterRenderAsync( bool firstRender )
...
+                AvailableQualities = AvailableQualities?.Select( x => new { Height = x } ),
```
David-Moreira, line 95: Only Requires Height?

David-Moreira (changes requested):

Documentation/Blazorise.Docs/Pages/News/2024-07-15-release-notes-160.razor
David-Moreira: PS:
We've added ProtectionServerCertificateUrl and thumbnails? Not sure if relevant to mention.

Demos/Blazorise.Demo/Pages/Tests/VideoPage.razor
stsrki: Yes, for testing purpsoses.

Source/Extensions/Blazorise.Video/Video.razor.cs
stsrki: It is enough, yes.

Source/Extensions/Blazorise.Video/Video.razor.cs
stsrki: Yep.

stsrki:
I have fixed the problems you mentioned. Some other things were already behaving as before, so I will skip them to avoid postponing the release.

Proceeding with the merge 🤞

PR #5627, opened 2024-07-20
Theme: add FontOptions on ThemeBody

### LINKED ISSUE #5526: Theme: add an option to define a font-family
**Proposal:** Add it to the `ThemeBodyOptions`

```diff
/// <summary>
/// Defines the body theme options.
/// </summary>
public record ThemeBodyOptions
{
    /// <summary>
    /// Gets or sets the body background color.
    /// </summary>
    /// <remarks>
    /// If defined it will also override any <see cref="Background.Body"/>.
    /// </remarks>
    public string BackgroundColor { get; set; }

    /// <summary>
    /// Gets or sets the body text color.
    /// </summary>
    /// <remarks>
    /// If defined it will also override any <see cref="TextColor.Body"/>.
    /// </remarks>
    public string TextColor { get; set; }

+    public string FontFamily { get; set; }
}
```

stsrki:
Closes #5526

### DIFF
--- a/Source/Blazorise/Themes/Options/ThemeFontOptions.cs
+++ b/Source/Blazorise/Themes/Options/ThemeFontOptions.cs
@@ -1,0 +1,27 @@
+﻿namespace Blazorise;
+
+/// <summary>
+/// Defines the font theme options.
+/// </summary>
+public record ThemeFontOptions
+{
+    /// <summary>
+    /// Gets or sets the font family.
+    /// </summary>
+    public string Family { get; set; }
+
+    /// <summary>
+    /// Gets or sets the font style.
+    /// </summary>
+    public string Style { get; set; }
+
+    /// <summary>
+    /// Gets or sets the font size.
+    /// </summary>
+    public string Size { get; set; }
+
+    /// <summary>
+    /// Gets or sets the font weight.
+    /// </summary>
+    public string Weight { get; set; }
+}
--- a/Source/Blazorise/Themes/ThemeGenerator.cs
+++ b/Source/Blazorise/Themes/ThemeGenerator.cs
@@ -109,6 +109,23 @@ public abstract class ThemeGenerator : IThemeGenerator
         {
             Variables[ThemeVariables.BodyTextColor] = ToHex( ParseColor( theme.BodyOptions.TextColor ) );
         }
+
+        if ( theme.BodyOptions.FontOptions is not null )
+        {
+            var fontOptions = theme.BodyOptions.FontOptions;
+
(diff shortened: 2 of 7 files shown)


Source/Blazorise/Themes/Options/ThemeFontOptions.cs
```diff
@@ -0,0 +1,27 @@
+﻿namespace Blazorise;
+
+/// <summary>
+/// Defines the font theme options.
+/// </summary>
+public record ThemeFontOptions
```
David-Moreira, line 6: Do we want to have some kind of smart types to help the user set these? or are we fine with "free edit"?

Source/Blazorise/Themes/ThemeGenerator.cs
```diff
@@ -109,6 +109,23 @@ protected virtual void GenerateBodyVariables( Theme theme )
         {
             Variables[ThemeVariables.BodyTextColor] = ToHex( ParseColor( theme.BodyOptions.TextColor ) );
         }
+
+        if ( theme.BodyOptions.FontOptions is not null )
+        {
+            var fontOptions = theme.BodyOptions.FontOptions;
+
+            if ( !string.IsNullOrEmpty( fontOptions.Family ) )
```
David-Moreira, line 117: I generally like to do whitespace just to be a bit more "safe". Otherwise it' fine anyway.

David-Moreira (approved):
LGGTM.

Source/Blazorise/Themes/ThemeGenerator.cs
stsrki: Kinda prefer `IsNullOrEmpty`.

Source/Blazorise/Themes/Options/ThemeFontOptions.cs
stsrki: What do you mean by smart types?

David-Moreira: Like the ones we have for other stuff but still allow strings.

Font.Family.Helvetiva, etc....
Font.Size.1px rem, etc...?
so on...

stsrki: Ah, that. No. It would be impossible to have every type. The user should be able to define any custom font family or size.

David-Moreira: Like I said, it should still allow introduction of custom strings. I think we have that somewhere? Icons?
You're right that it's ALOT of types, but we could have the most common.
Just an idea, of coursse it's way more work to make it work.

PR #5648, opened 2024-07-25
BlazoriseLicenseChecker: Null checks and apply defaults for unlicensed limits

### LINKED ISSUE #5647: [Bug]: NullReferenceException thrown in BaseChart::LimitDataSets method
### Blazorise Version

1.6

Bootstrap5

Unfortunately, creating a code snippet to reproduce this would take quite a lot of time. Suffice it to say, during the execution of a Blazor page `OnInitializedAsync` method, it is possible for the `LicenseChecker` to be null.

The offending code in our app is this:
```
await _lineChart.AddDatasetsAndUpdate(chartData);
```

The exception gets thrown in `Blazorise.Charts.BaseChart`:
```
106 private void LimitDataSets(params TDataSet[] datasets)
107 {
108  	if (!datasets.IsNullOrEmpty() && LicenseChecker.GetChartsRowsLimit().HasValue)
109  	{
110  		foreach (TDataSet val in datasets)
111  		{
112  			val.Data = LimitData(val.Data);
113  		}
114  	}
115 }
```

At line 108, `LicenseChecker` is `null`.

### Steps to reproduce

Execute `Chart.AddDatasetsAndUpdate` while in the `OnInitializedAsync` phase of a page load.

### What is expected?

The expected behavior is that no exception will be thrown in the `BaseChart` class.

A `NullReferenceException` is thrown from the `BaseChart` class.

Chrome, Microsoft Edge

### Any additional comments?

There's a very simple fix. Change line 108 in `BaseChart` either to do null propagation or check for a null `LicenseChecker`. Or even better, because the `LimitData` method winds up calling `GetChartRowsLimit` in a loop, it would make more sense to check it once get rid of `LimitData` altogether:

```
private void LimitDataSets(params TDataSet[] datasets)
{
	if (datasets.IsNullOrEmpty() || LicenseChecker is null) return;
	
	var rowsLimit = LicenseChecker.GetChartsRowsLimit();
	if (!rowsLimit.HasValue) return;
		
	foreach (TDataSet val in datasets)
	{
		val.Data = val.Data.Take(rowsLimit).ToList();
	}
}
```

There is another code smell, of course, in that the code executes a method and uses it like a property. The method `BlazoriseLicenseProvider::GetDataGridRowsLimit` (ultimately the source of `LicenseChecker`'s data) could be refactored.

David-Moreira:
Well it is not the first time we hear about someone complaining about a null in `BlazoriseLicenseChecker`.
Although typically this should not be null as per our standard **Blazorise** registration.

This should make sure that users can't try to game the system by removing the service from the DI Container, by assuming the user as unlicensed if `BlazoriseLicenseChecker` is null.
Also gives the flexibility for users to initialize the services themselves, and if they happen to not initialize the `BlazoriseLicenseChecker`, then we still don't throw an exception, but still apply our limits.

The naming of the class can change if you wish, I went back and forth with a few but ultimately, stating that the purpose of it is related to `BlazoriseLicenseLimits` seems fine, the suffix can change if you prefer something like `Manager`, `Utils`, or whatever else.

**PS:** Also it could probably be extensions in the `LicenseChecker` that check whether it is null itself? But I think this is clearer? Let me know your thoughts.

closes #5647

### DIFF
--- a/Source/Blazorise/Licensing/BlazoriseLicenseChecker.cs
+++ b/Source/Blazorise/Licensing/BlazoriseLicenseChecker.cs
@@ -87,7 +87,7 @@ public sealed class BlazoriseLicenseChecker
     /// Null if no limit is set.
     /// </summary>
     /// <returns></returns>
-    public int? GetAutoCompleteRowsLimit()
+    public int? GetAutocompleteRowsLimit()
     {
         return blazoriseLicenseProvider.GetAutocompleteRowsLimit();
     }
--- a/Source/Blazorise/Licensing/BlazoriseLicenseLimitsHelper.cs
+++ b/Source/Blazorise/Licensing/BlazoriseLicenseLimitsHelper.cs
@@ -1,0 +1,57 @@
+﻿namespace Blazorise.Licensing;
+
+/// <summary>
+/// Ultimately provides sane defaults for Blazorise licensing, in case required scoped services are not provided
+/// </summary>
+public static class BlazoriseLicenseLimitsHelper
+{
+    /// <summary>
+    /// Returns the maximum number of rows that can be displayed.
+    /// Null if no limit is set.
+    /// </summary>
+    /// <returns></returns>
+    public static int? GetDataGridRowsLimit( BlazoriseLicenseChecker blazoriseLicenseChecker )
+    {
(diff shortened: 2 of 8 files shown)


stsrki (approved):
Looks good to me. As you have mentioned, if anyone tries to start the app without registering license services in DI than we properly fallback to default limits. This is fine. Good job.

PR #5671, opened 2024-08-04
Optimizations: refactor class provider

### LINKED ISSUE #5294: Modify IClassBuilder to make it provider focused
We need to make all methods in the `IClassBuilder` in a way that the provider that implements it will handle it in their own way.

**Current:**

```cs
string DropdownShow()
```

**Proposed:**

```cs
string DropdownShow( bool visible )
```

Use case is:

```cs
builder.Append( ClassProvider.DropdownShow( Visible ) );
```

So, what does this give us? In some cases, certain providers tend to flip the visibility class. Instead of `.show` it can be `.hide`. By moving the classname builder login to the provider we can be certain that all cases are handled.

stsrki:
Closes #5294 

This PR makes all IClassProvider methods accept extra parameters so that every CSS provider can decide how to handle and return the class name.

### DIFF
--- a/Source/Blazorise/Extensions/ColorExtensions.cs
+++ b/Source/Blazorise/Extensions/ColorExtensions.cs
@@ -14,4 +14,14 @@ public static class ColorExtensions
     {
         return color == null || color == Color.Default;
     }
+
+    /// <summary>
+    /// Determines if the supplied color is defined, i.e. not null or empty.
+    /// </summary>
+    /// <param name="color">Color value to check.</param>
+    /// <returns>True if color is not <c>null</c> or <see cref="Color.Default"/>.</returns>
+    public static bool IsNotNullOrDefault( this Color color )
+    {
+        return color != null && color != Color.Default;
+    }
 }
--- a/Source/Blazorise/Extensions/ColorExtensions.cs
+++ b/Source/Blazorise/Extensions/ColorExtensions.cs
@@ -34,4 +34,14 @@ public static class ColorExtensions
     {
         return background == null || background == Background.Default;
     }
+
+    /// <summary>
+    /// Determines if the supplied background color is defined, i.e. not null or empty.
+    /// </summary>
+    /// <param name="background">Color value to check.</param>
+    /// <returns>True if background color is not <c>null</c> or <see cref="Background.Default"/>.</returns>
+    public static bool IsNotNullOrDefault( this Background background )
+    {
+        return background != null && background != Background.Default;
+    }
 }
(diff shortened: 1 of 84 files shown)


Source/Blazorise/Extensions/ColorExtensions.cs
```diff
@@ -14,4 +14,34 @@ public static bool IsNullOrDefault( this Color color )
...
+    }
+
+    /// <summary>
+    /// Determines if the supplied background color is defined, i.e. not null or empty.
+    /// </summary>
+    /// <param name="background">Color value to check.</param>
+    /// <returns>True if background color is not <c>null</c> or <see cref="Background.Default"/>.</returns>
+    public static bool IsNotNullOrDefault( this Background background )
+    {
+        return background != null && background != Background.Default;
```
David-Moreira, line 45: `is not null` if you care

Source/Blazorise/Extensions/ColorExtensions.cs
```diff
@@ -14,4 +14,34 @@ public static bool IsNullOrDefault( this Color color )
...
+    }
+
+    /// <summary>
+    /// Determines if the supplied background color is null or empty, i.e. not containing any color.
+    /// </summary>
+    /// <param name="background">Color value to check.</param>
+    /// <returns>True if background color is <c>null</c> or <see cref="Background.Default"/>.</returns>
+    public static bool IsNullOrDefault( this Background background )
+    {
+        return background == null || background == Background.Default;
```
David-Moreira, line 35: Probably can just do `background == Background.Default` since null is never equal to `Background.Default`. 
Otherwise if you want to read exactly,  `is null` if you care.

Source/Blazorise/Extensions/ColorExtensions.cs
```diff
@@ -14,4 +14,34 @@ public static bool IsNullOrDefault( this Color color )
...
     }
+
+    /// <summary>
+    /// Determines if the supplied color is defined, i.e. not null or empty.
+    /// </summary>
+    /// <param name="color">Color value to check.</param>
+    /// <returns>True if color is not <c>null</c> or <see cref="Color.Default"/>.</returns>
+    public static bool IsNotNullOrDefault( this Color color )
+    {
+        return color != null && color != Color.Default;
```
David-Moreira, line 25: `is not null` if you care

David-Moreira (approved):

Source/Blazorise/Extensions/ColorExtensions.cs
stsrki: I would leave it explicit just in case we do something with Default in the future.

PR #5673, opened 2024-08-05
TextDecoration: add new utilities

### LINKED ISSUE #5667: TextDecoration utilities
https://getbootstrap.com/docs/5.3/utilities/text/#text-decoration

stsrki:
Closes #5667


Documentation/Blazorise.Docs/Pages/Docs/Components/Typographies/TypographyPage.razor
```diff
@@ -132,6 +132,16 @@
     <DocsPageSectionSource Code="TypographyTextSizeExample" />
 </DocsPageSection>
 
+<DocsPageSection>
+    <DocsPageSectionHeader Title="Text Decoration">
+        Blazorise allows you to style text with various decorations. You can use underline to place a line beneath the text, line-through to strike through it, or overline to add a line above. To remove any decorations, use the none option. These styles help enhance and highlight text as needed.
```
David-Moreira, line 137: I actually like the bullet point approach with the description for each one.
Thoughts?

Source/Blazorise/Base/BaseComponent.cs
```diff
@@ -690,6 +695,21 @@ public TextTransform TextTransform
...
+    /// <summary>
+    /// Gets or sets the text decoration.
+    /// </summary>
+    [Parameter]
+    public TextDecoration TextDecoration
+    {
+        get => textDecoration;
+        set
+        {
+            textDecoration = value;
```
David-Moreira, line 707: You usually have the approach of checking if the value == existingvalue. Is it applicable here?

I was thinking, does it make sense to have an helper on the base component that does this? Not sure if it makes sense, just throwing the idea out there.
setProperty(field, newvalue){
if (field==newValue)

....
DirtyClasses()
}

Source/Blazorise/Enums/TextDecoration.cs
```diff
@@ -0,0 +1,37 @@
...
+
+/// <summary>
+/// Defines the text decoration.
+/// </summary>
+public enum TextDecoration
+{
+    /// <summary>
+    /// No decoration. The text renders as it is. This is default.
+    /// </summary>
+    Default,
```
David-Moreira, line 11: I'm not sure how to better describe this.
But it can still inherit decoration from somewhere else can it not?

The thing is that default, basically does nothing, it does not apply any special css class, is this not right?

Saying that it **renders the text as it is** might be misleading I'm not sure.

David-Moreira (approved):

Source/Blazorise/Enums/TextDecoration.cs
stsrki: I can add the comment saying that it will use rule from CSS framework, or inherited from the parent element.

Source/Blazorise/Base/BaseComponent.cs
stsrki: Done.

Not sure how it would that work unless we use something like reflection. Better to stick with the manual approach.

Documentation/Blazorise.Docs/Pages/Docs/Components/Typographies/TypographyPage.razor
stsrki: The example already shows it good IMO.

Source/Blazorise/Base/BaseComponent.cs
David-Moreira: Hmm probably can make the method generic or overloads for the common cases. Don't think reflection is needed.

Documentation/Blazorise.Docs/Pages/Docs/Components/Typographies/TypographyPage.razor
David-Moreira: I meant the text not the code example

PR #5689, opened 2024-08-14
TreeView: Add virtualization support

### LINKED ISSUE #5649: TreeView: add virtualization
We need to consider adding virtualization support for big trees, which can decrease performance on client systems. Virtualization might help in this case.

ddjerqq:
## Added virtualization support to the TreeView component

Closes #5649

## How Has This Been Tested?
I ran the demo project and checked that the TreeView worked.

## Types of changes

## Checklist:

- [x] I've added relevant tests.


Documentation/Blazorise.Docs/Pages/Docs/Extensions/TreeView/TreeViewPage.razor
```diff
@@ -192,6 +192,9 @@
     <DocsAttributesItem Name="AutoExpandAll" Type="bool" Default="false">
         Defines if the treenode should be automatically expanded. Note that it can happen only once when the tree is first loaded.
     </DocsAttributesItem>
+    <DocsAttributesItem Name="Virtualize" Type="bool">
+        Defines if hidden treenodes should be visible. See <a href="https://learn.microsoft.com/en-us/aspnet/core/blazor/components/virtualization">docs for Virtualization</a>.
```
stsrki, line 196: Use our `Anchor` component for links. 

Example: `<Anchor To="https://learn.microsoft.com/en-us/aspnet/core/blazor/components/virtualization">docs for Virtualization</Anchor>`

Source/Extensions/Blazorise.TreeView/Internal/_TreeViewNode.razor.cs
```diff
@@ -305,6 +305,14 @@ protected virtual Task OnContextMenuHandler( TreeViewNodeState<TNode> nodeState,
     /// </summary>
     [Parameter] public bool AutoExpandAll { get; set; }
 
+    /// <summary>
+    /// Defines if hidden treenodes should be visible
+    /// </summary>
+    /// <see href="https://learn.microsoft.com/en-us/aspnet/core/blazor/components/virtualization">
```
stsrki, line 311: Prefer to have `<see` as inline and not span more lines.

Source/Extensions/Blazorise.TreeView/Internal/_TreeViewNode.razor.cs
```diff
@@ -305,6 +305,14 @@ protected virtual Task OnContextMenuHandler( TreeViewNodeState<TNode> nodeState,
     /// </summary>
     [Parameter] public bool AutoExpandAll { get; set; }
 
+    /// <summary>
+    /// Defines if hidden treenodes should be visible
```
stsrki, line 309: The description could be more descriptive of what Virtualize does.

Source/Extensions/Blazorise.TreeView/TreeView.razor.cs
```diff
@@ -341,6 +341,14 @@ public TreeViewSelectionMode SelectionMode
     /// </summary>
     [Parameter] public bool AutoExpandAll { get; set; }
 
+    /// <summary>
+    /// Defines if hidden treenodes should be visible
```
stsrki, line 345: The description could be more descriptive of what Virtualize does.

Source/Extensions/Blazorise.TreeView/TreeView.razor.cs
```diff
@@ -341,6 +341,14 @@ public TreeViewSelectionMode SelectionMode
     /// </summary>
     [Parameter] public bool AutoExpandAll { get; set; }
 
+    /// <summary>
+    /// Defines if hidden treenodes should be visible
+    /// </summary>
+    /// <see href="https://learn.microsoft.com/en-us/aspnet/core/blazor/components/virtualization">
```
stsrki, line 347: Prefer to have `<see` as inline and not span more lines.

David-Moreira:
**DOCS:**
1. Why are all the overflow bars visible by default? I'd use a more nicer example, where only the required vertical bar is shown. Also Do you need the outside one also?

image

Example from DataGrid : 
image

2. By setting **Virtualize** to true should we not have sane defaults where it would just work? Why are we making our users have to manually set additional settings? In my opinion it should automatically work, but users can further customize it if they want to, changing max height, etc...

3. Can we see more nesting? What will happen? Will it work ok? Will it be a bunch of vertical bars?

4. How is dynamic data handled? Imagine I had data to the treeview will it break anything? would be nice to adap the DEMO version to work with it, since you can already do alot of interactions with the treeview, and you can check how the Virtualize affects these.

Documentation/Blazorise.Docs/Pages/News/2024-10-15-release-notes-170.razor
```diff
@@ -98,6 +98,10 @@
     Optimizations
 </Heading>
 
+<Paragraph>
+    We added options for Virtualized rendering to the TreeView component. This should help with large TreeViews.
```
David-Moreira, line 102: Very generic. What does Virtualize actually do? Maybe just extract a brief paragraph out of Microsoft docs.

Documentation/Blazorise.Docs/Pages/Docs/Extensions/TreeView/TreeViewPage.razor
```diff
@@ -192,6 +219,9 @@
     <DocsAttributesItem Name="AutoExpandAll" Type="bool" Default="false">
         Defines if the treenode should be automatically expanded. Note that it can happen only once when the tree is first loaded.
     </DocsAttributesItem>
+    <DocsAttributesItem Name="Virtualize" Type="bool" Default="false">
+        Defines if hidden treenodes should be visible. See <Anchor To="https://learn.microsoft.com/en-us/aspnet/core/blazor/components/virtualization">docs for Virtualization</Anchor>.
```
David-Moreira, line 223: > Defines if hidden treenodes should be visible. 

Are we sure this conveys the exact meaning of what `Virtualize` does? I'd say it's something along the lines of rendering the nodes as you scroll through the data limiting the number of html elements that are rendered at a time which in turns helps page performance when handling huge sources of data?

stsrki:
> **1.** Why are all the overflow bars visible by default? I'd use a more nicer example, where only the required vertical bar is shown. Also Do you need the outside one also?

It seems this is the problem of the bigger picture. We pass all the parameters that were assigned to the root treeview elements down to the child tree nodes. I think we should not pass the down, only the necessary that are needed for the child nodes to get the data and states.

image

David-Moreira:
> > **1.** Why are all the overflow bars visible by default? I'd use a more nicer example, where only the required vertical bar is shown. Also Do you need the outside one also?
> 
> It seems this is the problem of the bigger picture. We pass all the parameters that were assigned to the root treeview elements down to the child tree nodes. I think we should not pass the down, only the necessary that are needed for the child nodes to get the data and states.
> 
> image

Probably not sure? Could these be acting as a king of global way of styling all nodes? So if you remove these, some TreeViews that might rely on them, might get a different UI result?

stsrki:
> > > **1.** Why are all the overflow bars visible by default? I'd use a more nicer example, where only the required vertical bar is shown. Also Do you need the outside one also?
> > 
> > 
> > It seems this is the problem of the bigger picture. We pass all the parameters that were assigned to the root treeview elements down to the child tree nodes. I think we should not pass the down, only the necessary that are needed for the child nodes to get the data and states.
> > image
> 
> Probably not sure? Could these be acting as a king of global way of styling all nodes? So if you remove these, some TreeViews that might rely on them, might get a different UI result?

In this case yes, they act as a semi-global styles. Which is also much different than we do in any other component. We almost never pass styles from the parent down to the child component. The right approach would be to have a callback like `GetSomeParameter` that we already have for node elements.

I think we should revise all styles on this and only pass the ones that are not going to break the TreeVIew component. Text styling is OK. Overflow and sizing are not OK to be passed.

stsrki:
I have removed styling parameters from not being passed down and tested all examples. And it seems to be working as expected.

David-Moreira:
@stsrki Already asking for review?

Already gone through all the points? **1, 2,3, 3.1, 3.2, 5?**

stsrki:
I will do the 2. point. For others, there should not be any problems.

David-Moreira:
> I will do the 2. point. For others, there should not be any problems.

So you're telling me you don't see any problems with this?
> 3. Can we see more nesting? What will happen? Will it work ok? Will it be a bunch of vertical bars?
EDIT : 3.1 I suspect it might not work correctly not sure? Since Virtualize really likes fixed dimensions...
3.2: I'd like to see the Virtualize working also on the root items.

stsrki:
> So you're telling me you don't see any problems with this?
> 
> > 3. Can we see more nesting? What will happen? Will it work ok? Will it be a bunch of vertical bars?
> >    EDIT : 3.1 I suspect it might not work correctly not sure? Since Virtualize really likes fixed dimensions...
> >    3.2: I'd like to see the Virtualize working also on the root items.

It works on all levels. Root level is the one where height and overflow is defined. I tested it on multiple levels, and it works as it should.

David-Moreira:
I still get an horizontal scroll bar which I don't think it's needed.
The vertical bar shows even though there's nothing to scroll yet, I think there's a setting to make it show **only** there's scrollable content. (I think it's auto, but I don't remember)
website-capture_2024-8-31

Also if I disable virtualize the TreeView does not go back to the original styling.

David-Moreira:
I am getting a huge list, can you take a look?
image


I'll commit my demo code

stsrki:
> I am getting a huge list, can you take a look? image
> 
> I'll commit my demo code
> 
> **PS:** Use the AddNode button to increase the nodes

I have fixed it. The problem was that we generated TreeViewNodeState each time from the original list. So, in turn, the Key would also be new, making the duplicate between renderings. I've added a simple check for reference on the `.Node` itself.

stsrki:
@David-Moreira please re-review.

stsrki (approved):

PR #5691, opened 2024-08-14
Tooltip Delay option

### LINKED ISSUE #5686: Tooltip delay
Add an option to delay the showing and hiding of tooltips.

https://atomiks.github.io/tippyjs/v6/customization/

stsrki:
Closes #5686

Also made some other improvements.

1. Make a serialization object with the first letter in lowercase. This ensures that we get appropriately named JSON objects in case users have different JSON serialization settings. Unlikely but you never know.
2. Added extra checks for Parameter properties.
3. I made tippy.js and popper.js as separate JS files and imported them.


Documentation/Blazorise.Docs/Pages/Docs/Components/Tooltips/TooltipPage.razor
```diff
@@ -125,4 +135,7 @@
     <DocsAttributesItem Name="AppendTo" Type="string" Default="body">
         The element to append the tooltip to. If <Code>Interactive = true</Code>, the default behavior is appendTo: "parent".
     </DocsAttributesItem>
+    <DocsAttributesItem Name="Delay" Type="(int Show, int Hide)" Default="(0, 0)">
```
David-Moreira, line 138: You don't usually like these at all, not even for internal things, what changed?

David-Moreira (approved):

David-Moreira:
> Make a serialization object with the first letter in lowercase. This ensures that we get appropriately named JSON objects in case users have different JSON serialization settings. Unlikely but you never know.

You touch on an important point here. We should not rely on anything external or conventions. We should precisely define how our own stuff gets serialized so it can't fail due to external conditions. Is this not possible?

Source/Blazorise/Components/Tooltip/Tooltip.razor.cs
```diff
@@ -73,20 +73,21 @@ protected override void OnInitialized()
...
+                placement = ClassProvider.ToTooltipPlacement( Placement ),
+                multiline = Multiline,
+                alwaysActive = AlwaysActive,
+                showArrow = ShowArrow,
+                fade = Fade,
+                fadeDuration = FadeDuration,
+                trigger = ToTippyTrigger( Trigger ),
+                triggerTargetId = TriggerTargetId,
+                maxWidth = Theme?.TooltipOptions?.MaxWidth,
+                autodetectInline,
```
David-Moreira, line 86: Is this one missing the assignment?

stsrki: It's already lowercase, so no.

stsrki:
> > Make a serialization object with the first letter in lowercase. This ensures that we get appropriately named JSON objects in case users have different JSON serialization settings. Unlikely but you never know.
> 
> You touch on an important point here. We should not rely on anything external or conventions. We should precisely define how our own stuff gets serialized so it can't fail due to external conditions. Is this not possible?

Well, we could introduce internal serialization class objects that make use of `JsonPropertyNameAttribute` to strongly type the name of the field. That way, it would be 100% right.

Documentation/Blazorise.Docs/Pages/Docs/Components/Tooltips/TooltipPage.razor
stsrki: Nothing much. I still prefer more complex objects over tuples. However, since this API will hardly change in the future, I find it appropriate in this case.

Source/Blazorise/Components/Tooltip/Tooltip.razor.cs
David-Moreira: I said assignment. It's just autodetectInline. Nothing assigned is it intended?

David-Moreira:
> > > Make a serialization object with the first letter in lowercase. This ensures that we get appropriately named JSON objects in case users have different JSON serialization settings. Unlikely but you never know.
> > 
> > 
> > You touch on an important point here. We should not rely on anything external or conventions. We should precisely define how our own stuff gets serialized so it can't fail due to external conditions. Is this not possible?
> 
> Well, we could introduce internal serialization class objects that make use of `JsonPropertyNameAttribute` to strongly type the name of the field. That way, it would be 100% right.

Yes. If we care about it. That's a way to go I guess, not doing guess work and hoping nothing external happens in my opinion.

Source/Blazorise/Components/Tooltip/Tooltip.razor.cs
stsrki: C# will use the variable name if its unassigned. Try it and you will see little warning that assignment is not needed.

stsrki:
> > > > Make a serialization object with the first letter in lowercase. This ensures that we get appropriately named JSON objects in case users have different JSON serialization settings. Unlikely but you never know.
> > > 
> > > 
> > > You touch on an important point here. We should not rely on anything external or conventions. We should precisely define how our own stuff gets serialized so it can't fail due to external conditions. Is this not possible?
> > 
> > 
> > Well, we could introduce internal serialization class objects that make use of `JsonPropertyNameAttribute` to strongly type the name of the field. That way, it would be 100% right.
> 
> Yes. If we care about it. That's a way to go I guess, not doing guess work and hoping nothing external happens in my opinion.

I've created an issue so we can plan for it.

Source/Blazorise/Components/Tooltip/Tooltip.razor.cs
David-Moreira: LOL, i'm not speaking about the anonymous type.
I'm speaking about actual assignment var a = 1.
The variable autodetectInline has no assigned value at all. That's my question? Will it be null? That's the point?

stsrki: Ahhh that. No, it is going to have a value assigned. See

image

David-Moreira: Ahhh, I missed that. Thanks.

PR #5692, opened 2024-08-14
Remove validator APIs that are not supported

### LINKED ISSUE #5690: [Bug]: Validator dateTime
### Blazorise Version

1.6

FluentUI2

### Steps to reproduce

```
<Field>
   <FieldLabel RequiredIndicator>Date</FieldLabel>
   <FieldBody>
     <Validation Validator="@ValidationRule.IsDateSelected<DateTime?>">
       <DatePicker TValue="DateTime?" DisplayFormat="dd/MM/yyyy" @bind-Date="@Date">
       </DatePicker>
     </Validation>
   </FieldBody>
 </Field>
```

### What is expected?

No errors

RZ9986	Component attributes do not support complex content (mixed C# and markup). Attribute: 'Validator', text: 'ValidationRule.IsDateSelected<DateTime?>'	

_No response_

### Any additional comments?

_No response_

stsrki:
Closes #5690

I have, by mistake, added some validation helper APIs that are not supported.

### DIFF
--- a/Documentation/Blazorise.Docs/Pages/Docs/Components/Validations/ValidationPage.razor
+++ b/Documentation/Blazorise.Docs/Pages/Docs/Components/Validations/ValidationPage.razor
@@ -264,12 +264,6 @@
     <DocsMethodsItem Name="IsFileSelected">
         Checks if the file is selected.
     </DocsMethodsItem>
-    <DocsMethodsItem Name="IsDateSelected<TValue>">
-        Checks if the date is selected.
-    </DocsMethodsItem>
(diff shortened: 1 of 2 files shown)


David-Moreira (approved):
Why are these not supported? Feels weird to remove something that was already introduced?

Edit: I understand you added by mistake, but how hard would it be to just make them work?

stsrki:
> Edit: I understand you added by mistake, but how hard would it be to just make them work?

We would have to introduce a new `Validator` API that uses a generic `T` parameter. But since that is possible only in the next major version, I would rather remove it for now.

PR #5696, opened 2024-08-16
add api verify

SimonCropp:
This adds snapshot testing to the repo using Verify

As an initial example i added a test for the API. This means any time the public API changes you need to accept that change. this secondary (in addition to the code that changes the api) acceptance serves as a confirmation of any change and the api change to be visualized in different way

to accept any change there are a variety of approaches https://github.com/VerifyTests/Verify?tab=readme-ov-file#snapshot-management

if u want to see the workflwo for manageing snapshots:

 * Initial Verification https://github.com/VerifyTests/Verify?tab=readme-ov-file#initial-verification
 * Subsequent Verification https://github.com/VerifyTests/Verify?tab=readme-ov-file#subsequent-verification

if u want to see some more content on how verify works https://github.com/VerifyTests/Verify?tab=readme-ov-file#media

let me know if u want to jump on a call to discuss any of the above

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,6 @@
 * text=auto
 
 *.json      text eol=crlf working-tree-encoding=UTF-8
+
+
+*.verified.txt text eol=lf working-tree-encoding=UTF-8
--- a/.gitignore
+++ b/.gitignore
@@ -273,3 +273,6 @@ Documentation/Blazorise.Docs.Server/wwwroot/img/blog/.DS_Store
 
 # Source Generation directory
 __SOURCEGENERATED__/
+
+
+*.received.*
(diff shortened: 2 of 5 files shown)


stsrki:
Thank you for the submission.

If I'm not mistaken, this will only check if the public API should be allowed to be publisdhed. It will not check if the new API can be used? 

For example, in https://github.com/Megabit/Blazorise/pull/5692, we never supported generic methods to be used as a validator on a non-generic delegate.

GeertvanHorrik:
Thanks @SimonCropp for creating the PR, it was on my to-do list so thanks!

@stsrki : this will indeed check every public API change you make, and you have to manually approve the api change for the unit test to succeed. This way you are made double aware of any changes you make and can verify it in a different context.

About the usability: that's something that can be covered by regular unit tests?

PR #5700, opened 2024-08-17
DataGrid: Add SelectedCell API

David-Moreira:


stsrki (approved):
The feature is working great. The only thing that might be interested for the users is that by hitting ENTER key we can focus to the cell in the next row. I think that is how it is usually handled in similar UIs. If you think it is not worth it I will merge as it is.

David-Moreira:
> The feature is working great. The only thing that might be interested for the users is that by hitting ENTER key we can focus to the cell in the next row. I think that is how it is usually handled in similar UIs. If you think it is not worth it I will merge as it is.

I don't think it's worth, as we're already implemented it like this. 
I'd take in feedback first from the people using **Rapid Editing** and this and act if necessary.

PR #5703, opened 2024-08-19
Cropper: update to v2-rc2

### LINKED ISSUE #5677: [Bug]: Autocomplete issues when Cropper is used
### Blazorise Version

1.5.4 (Also verified on 1.6.0)

Bootstrap5

https://github.com/DaveAPerry/WASMHyphenTest

### Steps to reproduce

1. Navigate to the Counter Page
2. Type in the autocomplete component - unable to use +/- characters

### What is expected?

Autocomplete component should function normally and accept valid inputs

Autocomplete does not accept + or - characters from the keyboard

Microsoft Edge

### Any additional comments?

We recently used the Cropper component on one of our internal applications - a couple weeks after deployment we were notified that the user could not use - in the autocomplete when looking for a specific value containing a "-" character.

stsrki:
Closes #5677


David-Moreira (approved):
AS long as you've tested and it's working. What can I say? :D 
LGTM.

PR #5704, opened 2024-08-20
Tooltip: global options

### LINKED ISSUE #5686: Tooltip delay

stsrki:
Closes #5686

Added an option to define global tooltip settings.


Source/Blazorise/BlazoriseOptions.cs
```diff
@@ -124,5 +124,23 @@ public string LicenseKey
...
+    /// </summary>
+    public BlazoriseTooltipOptions TooltipOptions { get; set; }
+
     #endregion
+}
+
+/// <summary>
+/// Represents the options available for configuring the behavior of a Tooltip in Blazorise.
+/// </summary>
+public class BlazoriseTooltipOptions
```
David-Moreira, line 138: I was more of a separate file for every class in the past and I believe I've seen similar reviews from you in the past.
Personally, nowadays I don't mind keeping related stuff in the same file. What about you? Are you fine with keeping both of these in the same file?

Source/Blazorise/Components/Tooltip/Tooltip.razor.cs
```diff
@@ -305,7 +314,7 @@ public TooltipTrigger Trigger
     /// <summary>
     /// Specifies the delay in ms once a trigger event is fired before a Tooltip shows or hides.
     /// </summary>
-    [Parameter] public (int Show, int Hide) Delay { get; set; } = (0, 0);
+    [Parameter] public (int Show, int Hide)? Delay { get; set; }
```
David-Moreira, line 317: What about `ShowDelay` and `HideDelay` as two separate parameters? 
I still don't like how we changed our mind about using values tuples, I would almost certainly say we don't have any other api with value tuples. I like them more for internal stuff and less for public stuff.
I'd just keep it consistent and use the same as we've been doing.

I honestly think the two separate params are more clear as a public api. I don't mind either anyway.

David-Moreira (approved):

Source/Blazorise/Components/Tooltip/Tooltip.razor.cs
stsrki: The more I'm thinking, you're right. It is better not to use a tuple and to have a separate parameters instead. Going to change it now.

Source/Blazorise/BlazoriseOptions.cs
stsrki: Done

David-Moreira (approved):

PR #5706, opened 2024-08-21
Video: Update events API

### LINKED ISSUE #5458: Video: add events for rewind and fast forward
Is there a feature to skip/rewind 5 or 10 seconds of the video, like in youtube for example?

I saw that there is a rewind and fast-forward option, but is there any event that triggers when the buttons are triggered?

I think it would be a valuable feature to have :)

stsrki:
Closes #5458

This PR changes the API of `Seeking` and `Seeked` events. I know we usually don't do this, but I find it appropriate in this case as in the previous version; having these events without the value parameter didn't make sense. And we didn't have that information because the previous video library didn't have that information. 

I have updated the migration notes. I don't think it will affect many people, but still, we want to have as seamless migration as possible.

---

Also, update documentation and split the API into a separate component/page.


Source/Extensions/Blazorise.Video/Video.razor.cs
```diff
@@ -728,12 +730,12 @@ public Task NotifyQualityChange( int? quality )
     /// <summary>
     /// Sent when a seek operation begins.
     /// </summary>
-    [Parameter] public Func<Task> Seeking { get; set; }
+    [Parameter] public Func<double, Task> Seeking { get; set; }
```
David-Moreira, line 733: Since we are touching the apis, why don't these just use regular event callbacks? Any special reason?

Source/Extensions/Blazorise.Video/wwwroot/video.js
```diff
@@ -478,11 +478,11 @@ function registerToEvents(dotNetAdapter, player) {
     });
 
     player.addEventListener('seeking', (event) => {
-        invokeDotNetMethodAsync(dotNetAdapter, "NotifySeeking");
+        invokeDotNetMethodAsync(dotNetAdapter, "NotifySeeking", event.detail || 0);
```
David-Moreira, line 481: `event.detail || 0`
So in js if we use double pipe it coalesces the value? so if event.detail is null, 0 will be sent?

David-Moreira (approved):
A part of me wants to say that we should just have args based api on all event like apis.
Allows way more flexibility when making changes. 

Thoughts?

Source/Extensions/Blazorise.Video/Video.razor.cs
stsrki: Regular event callbacks refreshes the page, and func is not. I want to avoid refreshing whenever necessary. Especially with heavy components like video.

Source/Extensions/Blazorise.Video/wwwroot/video.js
stsrki: Yeah. I don't think null is valid value in vidstack but this is just a precaution. 0 can be associated with zero time in the player.

Source/Extensions/Blazorise.Video/Video.razor.cs
David-Moreira: Gotcha, nice thinking. Although I'm not entirely sure about the veracity of "Regular event callbacks refreshes the page".
As far as I know, It calls `StateHasChanged` which not necessarily means that the page will refresh. The component specifically will be checked for changes and then refreshed, is this not the case?

stsrki: Yes. It will call StateHasChanged but it will still refresh everything on the same level of rendering tree, and bellow.

David-Moreira: Gotcha, my point. So I would say that it would really be doing almost nothing at all since Video won't inherit children?  Anyway this approach is indeed better nevertheless, just wanted to be sure that we were on the same page about the statement :
> "Regular event callbacks refreshes the page"

PR #5719, opened 2024-08-31
Datagrid | DataGridDateEditColumn |  Prefer DatePicker | Add public api

David-Moreira:


stsrki (commented):
Looking good to me.

What do you think if we add API, like `bool NativeMode`, that will keep the `DateEdit<>` if enabled?

David-Moreira:
> Looking good to me.
> 
> What do you think if we add API, like `bool NativeMode`, that will keep the `DateEdit<>` if enabled?

Then I'd do the same thing for the numeric column. Would you agree? Also altough **NativeMode** is a clear name for us, it might not be for our users? Do they understand exactly what **NativeMode** means? Altough I'm not finding a better name.

stsrki:
> Then I'd do the same thing for the numeric column. Would you agree?

Since we are going with new options I think it makes sense to make it for numeric column also.

> Also altough NativeMode is a clear name for us, it might not be for our users? Do they understand exactly what NativeMode means? Altough I'm not finding a better name.

Maybe `NativeInputMode`?

David-Moreira:
> > Then I'd do the same thing for the numeric column. Would you agree?
> 
> Since we are going with new options I think it makes sense to make it for numeric column also.
> 
> > Also altough NativeMode is a clear name for us, it might not be for our users? Do they understand exactly what NativeMode means? Altough I'm not finding a better name.
> 
> Maybe `NativeInputMode`?

So I was starting to work on this, but I noticed that there really isn't feature parity. The native one might have different parameters, features...
Do you think it makes sense to have Parameters that will only have effect when in a certain Mode? That might be a little bit confusing and might mislead users to try and use incompatible combinations? 
i.e 
`NativeInputMode=true+ DisabledDays`
`NativeInputMode=false + Step`, etc...

image
image

stsrki:
> > > Then I'd do the same thing for the numeric column. Would you agree?
> > 
> > 
> > Since we are going with new options I think it makes sense to make it for numeric column also.
> > > Also altough NativeMode is a clear name for us, it might not be for our users? Do they understand exactly what NativeMode means? Altough I'm not finding a better name.
> > 
> > 
> > Maybe `NativeInputMode`?
> 
> So I was starting to work on this, but I noticed that there really isn't feature parity. The native one might have different parameters, features... Do you think it makes sense to have Parameters that will only have effect when in a certain Mode? That might be a little bit confusing and might mislead users to try and use incompatible combinations? i.e `NativeInputMode=true+ DisabledDays` `NativeInputMode=false + Step`, etc...
> 
> image image

We just need to document it and say that in native mode those parameters will be ignored.

David-Moreira:
> > > > Then I'd do the same thing for the numeric column. Would you agree?
> > > 
> > > 
> > > Since we are going with new options I think it makes sense to make it for numeric column also.
> > > > Also altough NativeMode is a clear name for us, it might not be for our users? Do they understand exactly what NativeMode means? Altough I'm not finding a better name.
> > > 
> > > 
> > > Maybe `NativeInputMode`?
> > 
> > 
> > So I was starting to work on this, but I noticed that there really isn't feature parity. The native one might have different parameters, features... Do you think it makes sense to have Parameters that will only have effect when in a certain Mode? That might be a little bit confusing and might mislead users to try and use incompatible combinations? i.e `NativeInputMode=true+ DisabledDays` `NativeInputMode=false + Step`, etc...
> > image image
> 
> We just need to document it and say that in native mode those parameters will be ignored.

It's both modes. Both modes might have some parameter that's incompatible with the other.

stsrki:
Yeah, that makes it more complicated. 

If there are overlapping parameters in terms of features then prefer the higher ones. Introduce `DisabledDays` but don't introduce `Step`.

Source/Extensions/Blazorise.DataGrid/DataGridDateColumn.cs
```diff
@@ -10,4 +13,64 @@ public class DataGridDateColumn<TItem> : DataGridColumn<TItem>
     /// Hints at the type of data that might be entered by the user while editing the element or its contents.
     /// </summary>
     [Parameter] public DateInputMode InputMode { get; set; }
+
+    /// <summary>
+    /// Gets the date format based on the current <see cref="InputMode"/> settings.
+    /// </summary>
+    [Parameter] public string DateDisplayFormat { get; set; }
```
stsrki, line 20: Can we reuse the `DisplayFormat` from `BaseDataGridColumn`? It is the same thing.

Source/Extensions/Blazorise.DataGrid/_DataGridCellDatePicker.razor
```diff
@@ -0,0 +1,91 @@
+@using Blazorise.Utilities
+@typeparam TItem
+@inherits ComponentBase
+
+@if ( TypeHelper.IsDate( valueType ) )
+{
+    @if ( Column.NativeInputMode )
+    {
+        @dateEditFragment
```
stsrki, line 9: Just a small note. Make it uppercase so it matches the naming for numeric component, `(NumericEditFragment)`.

stsrki (commented):
It would be good to have an example in the docs that re-enables native mode.

Source/Extensions/Blazorise.DataGrid/DataGridDateColumn.cs
David-Moreira: Makes sense, although this would have the advantage that you could have a different display format (when viewing the table) and edit format (when editing the records)

stsrki (approved):
LGTM

PR #5758, opened 2024-09-30
Chart: add callbacks to tooltip options

### LINKED ISSUE #5741: Chart: add more support for callbacks
https://blazorise.com/support/issues/218/chart-percentage-tooltip

stsrki:
Closes #5741

Added all callbacks from https://www.chartjs.org/docs/3.7.1/configuration/tooltip.html#tooltip-callbacks

As in other parts of the code I have used `LambdaConverter` that converts C# to JS code. For simple tasks it seems to work as expected. 

```cs
ChartOptions chartOptions = new()
{
    AspectRatio = 1.5,
    Plugins = new ChartPlugins()
    {
        Tooltip = new ChartTooltip()
        {
            Enabled = true,
            UsePointStyle = true,
            Callbacks = new ChartTooltipCallbacks
            {
                Title = ( items ) => "Custom title: " + items[0].Parsed,
                Label = ( item ) => "Custom label: " + item.Parsed,
            }
        }
    }
};
```

The only problem is that the `Parsed` of the `ChartTooltipItemContext` can only be defined as `object` type. We cannot do something like `Title = ( items ) => "Custom title: " + items[0].Parsed.x,` because `object` is unknown in C#. I tried the `dynamic` type, but C# complained that it could not be used in lambda expression.

I could introduce a known type for the `Parsed` field, but I don't know how many different objects can be there. Some charts have it as a simple number. Some are `{ x, y }`. 

So I think I will leave it as `object` for now.

### DIFF
--- a/Demos/Blazorise.Demo/Pages/Tests/ChartsPage.razor.cs
+++ b/Demos/Blazorise.Demo/Pages/Tests/ChartsPage.razor.cs
@@ -20,7 +20,20 @@ public partial class ChartsPage
 
     ChartOptions chartOptions = new()
     {
-        AspectRatio = 1.5
+        AspectRatio = 1.5,
+        Plugins = new ChartPlugins()
+        {
+            Tooltip = new ChartTooltip()
+            {
+                Enabled = true,
+                UsePointStyle = true,
+                Callbacks = new ChartTooltipCallbacks
+                {
+                    Title = ( items ) => "Custom title: " + items[0].Parsed,
+                    Label = ( item ) => "Custom label: " + item.Parsed,
+                }
+            }
+        }
     };
 
     LineChartOptions lineChartOptions = new()
--- a/Documentation/Blazorise.Docs/Pages/News/2024-10-15-release-notes-170.razor
+++ b/Documentation/Blazorise.Docs/Pages/News/2024-10-15-release-notes-170.razor
@@ -46,6 +46,11 @@
             <Strong>RichTextEdit:</Strong> Table support
         </Paragraph>
     </UnorderedListItem>
+    <UnorderedListItem>
+        <Paragraph>
+            <Strong>Chart:</Strong> Tooltip callbacks
+        </Paragraph>
+    </UnorderedListItem>
 </UnorderedList>
 
 <Paragraph>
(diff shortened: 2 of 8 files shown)


David-Moreira (approved):
LGTM. Just curious how the heck does this even work? Does this run actual c# code or tries to parse to equivalent js code and execute in js? O.o

stsrki:
> LGTM. Just curious how the heck does this even work? Does this run actual c# code or tries to parse to equivalent js code and execute in js? O.o

It converts C# code into JS equivalent. And then runs it in JS. Unfortunately, there is no way to run C# code. All JS callbacks are synchronous by nature, so they cannot execute asynchronous JS interop code.

David-Moreira:
> > LGTM. Just curious how the heck does this even work? Does this run actual c# code or tries to parse to equivalent js code and execute in js? O.o
> 
> It converts C# code into JS equivalent. And then runs it in JS. Unfortunately, there is no way to run C# code. All JS callbacks are synchronous by nature, so they cannot execute asynchronous JS interop code.

Certainly there are other limitations, I don't think it will be able to convert everything?

I was thinking could we setup our own engine that would be able to callback into C#? I'm not sure if that would be useful over running on JS.

stsrki:
> I was thinking could we setup our own engine that would be able to callback into C#?

I think we would get the Nobel prize if we would solve this 🤯

> Certainly there are other limitations, I don't think it will be able to convert everything?

Yes, that are many limitations. But for simple titles and labels, it is just enough to get things done.

David-Moreira:
> I think we would get the Nobel prize if we would solve this 🤯

What do you mean? the dotnetobjectreference can call in to methods through a string based representation. As long as we setup something that knows that when the js callback happens it goes through the dotnetobjectreference and calls in the respective c# handler, it should work?
Of course there may be limitations or it might be harder to handle things like specific arguments.
But it seems possible thinking about it?

stsrki:
> > I think we would get the Nobel prize if we would solve this 🤯
> 
> What do you mean? the dotnetobjectreference can call in to methods through a string based representation. As long as we setup something that knows that when the js callback happens it goes through the dotnetobjectreference and calls in the respective c# handler, it should work? Of course there may be limitations or it might be harder to handle things like specific arguments. But it seems possible thinking about it?

Yes, we can call it through dotnetobjectreference. But. ChartJS callbacks will not wait for the result. ChartJS callbacks are synchronous, and any call to asynchronous method will be ignored, or the result of such call will not be used.

stsrki:
See this https://github.com/mariusmuntean/ChartJs.Blazor/issues/90

David-Moreira:
> ChartJS callbacks are synchronous, and any call to asynchronous method will be ignored

What do you mean? 
Isn't there are a way to make a sync js method wait for the async method synchronously or some other trick? Anyway, I'm going to assume that guy has already investigated like mad himself.

Let's just leave it, I was just throwing out the idea, clearly much more involved then what I was thinking.

stsrki:
> > ChartJS callbacks are synchronous, and any call to asynchronous method will be ignored
> 
> What do you mean? Isn't there are a way to make a sync js method wait for the async method synchronously or some other trick? Anyway, I'm going to assume that guy has already investigated like mad himself.
> 
> Let's just leave it, I was just throwing out the idea, clearly much more involved then what I was thinking.

I tried it many times in the past. And there really is not solution. Which is a shame.

David-Moreira:
If there's something I learned in software dev, is to not say something like : " And there really is not solution."
You never know... :P 

Anyway shame indeed.

PR #5759, opened 2024-10-01
Radio: Add Color parameter

### LINKED ISSUE #5728: Add Mixed Styles With RadioGroup Buttons
Currently, all colors are the same with the Color parameter on RadioGroup. I'm requesting an easy way to customize the colors per button. See bootstrap docs.

https://getbootstrap.com/docs/5.3/components/button-group/#mixed-styles

image

stsrki:
Closes #5728


Documentation/Blazorise.Docs/Pages/Docs/Components/Radios/Examples/RadioGroupButtonColorsExample.razor
```diff
@@ -0,0 +1,7 @@
+﻿@namespace Blazorise.Docs.Docs.Examples
+
+<RadioGroup TValue="string" Name="side" Buttons>
+    <Radio Value="@("left")" Color="Color.Danger">Left</Radio>
```
David-Moreira, line 4: is there any reason for @("left") instead of "left". What's the limitation?

David-Moreira (approved):
LGTM

Documentation/Blazorise.Docs/Pages/Docs/Components/Radios/Examples/RadioGroupButtonColorsExample.razor
stsrki: Blazor has difficulty figuring out the `string` value type and probably tries to bind to a variable with the same name. If we would to add `TValue="string"` then it would work as you said.

David-Moreira: strange, isn't this a case where it inherits TValue from the parent?

stsrki: I think that only works on runtime as a cascaded parameter.

David-Moreira: So it's not a cascaded parameter? What do you mean runtime? That's also compile time checked, that's the point of it, make the compiler aware of the TValue automatically?

stsrki: I agree it should work. But it is not.

image

Maybe it's a bug in the razor engine?

David-Moreira: No idea. But the middle value definitely seems like it should work.

David-Moreira: I just took it for a spin, I was curious. For example an int, works as expected with cascading type param support.
While if you put a Value="2" it's smart about it to convert to int automatically. 

Since the type of Value is generic, TValue
then as soon as you put a reference type, such as string, then the Value="" assignment tries to look for a symbol (type) instead of taking in a string by default. So it's actually looking for variables, etc... that's why `Value="left"` is invalid, it's looking for 
```
@code{

private string left="example";
}
```
I mean, makes sense when you think about it, but I'd expect it'd be smart and work more similar to a 
`[Parameter] public string Value {get;set,}` which in this case is smart enough to look for a string by default instead of symbol.

Anyway it's working as expected I think, but probably not what Blazor developers would expect haha.

David-Moreira: Which also leads to these weird off cases like the example above, where you have to explicitly tell it, it's a string.

In my opinion, this isn't consistent with the rest of the product, where ` [Parameter] public string Value {get;set,} ` by default it knows how to bind the string immediately, and if you want to bind a symbol, you use the **@**. Here it works the reverse way. It tries to look for the symbol immediately, I think they should have put an if case for strings to keep it consistent haha

PR #5761, opened 2024-10-01
ToastProvider: Add Closing and Opening events

### LINKED ISSUE #5736: ToastProvider  missing arguments Closing/Opening while toast does have it
please add Closing/Opening

stsrki:
Closes #5736


Documentation/Blazorise.Docs/Pages/News/2024-10-15-release-notes-170.razor
```diff
@@ -229,4 +229,12 @@
     However, since this module is not officially supported, we are marking this feature as experimental. This designation will remain until an official table module is released by the Quill.js team.
 </Paragraph>
 
+<Heading Size="HeadingSize.Is3">
+    ToastProvider Improvements
+</Heading>
+
+<Paragraph>
+    We have made some improvements to the ToastProvider component. The ToastProvider is a powerful helper component built on top of the Toast component and is used for showing simple alerts and notifications. The ToastProvider now includes a new option to to handle Opening and Closing events. This can be useful if you want to perform some actions before the toast is shown or hidden.
```
David-Moreira, line 237: Why are we re-explaining what ToastProvider is? This is release notes, the component already exists, I'd just go straight to the point?

Source/Blazorise/Components/Toast/Toast.razor.cs
```diff
@@ -70,12 +70,15 @@ public override async Task SetParametersAsync( ParameterView parameters )
...
         {
-            if ( paramVisible && await IsSafeToOpen() )
+            parameters.TryGetValue<Func<ToastOpeningEventArgs, Task>>( nameof( Opening ), out var paramOpening );
+            parameters.TryGetValue<Func<ToastClosingEventArgs, Task>>( nameof( Closing ), out var paramClosing );
+
+            if ( paramVisible && await IsSafeToOpen( paramOpening ) )
             {
                 await base.SetParametersAsync( parameters );
                 await SetVisibleState( true );
             }
```
David-Moreira, line 80: This makes it so `Opening` and `Closing` is called before  `base.SetParametersAsync( parameters ); `could this have any unintended side effects?

David-Moreira: What if it's not evaluated to true?
`await base.SetParametersAsync( parameters );` will not be called? parameters will not be updated?

David-Moreira (changes requested):

Documentation/Blazorise.Docs/Pages/News/2024-10-15-release-notes-170.razor
stsrki: Because there are always other features going into release notes, I always refine the release notes before the final version. So that the reading feels more nature. This is just so I do not forget it until then.

Source/Blazorise/Components/Toast/Toast.razor.cs
stsrki: > What if it's not evaluated to true? `await base.SetParametersAsync( parameters );` will not be called? parameters will not be updated?

Hmm, we're calling `await base.SetParametersAsync( parameters );` in other `if` statements. But, yeah there might be a situation when it would not called. Let me refine it,

```cs
if ( parameters.TryGetValue<bool>( nameof( Visible ), out var paramVisible ) && state.Visible != paramVisible )
{
    parameters.TryGetValue<Func<ToastOpeningEventArgs, Task>>( nameof( Opening ), out var paramOpening );
    parameters.TryGetValue<Func<ToastClosingEventArgs, Task>>( nameof( Closing ), out var paramClosing );

    if ( paramVisible && await IsSafeToOpen( paramOpening ) )
    {
        await base.SetParametersAsync( parameters );
        await SetVisibleState( true );
    }
    else if ( !paramVisible && await IsSafeToClose( paramClosing ) )
    {
        await base.SetParametersAsync( parameters );
        await SetVisibleState( false );
    }

    if ( Rendered )
}
else
{
    await base.SetParametersAsync( parameters );
}
```

David-Moreira: Ok, right, now, what about the order? The callbacks being called before ? Any concerns?

stsrki: > This makes it so `Opening` and `Closing` is called before `base.SetParametersAsync( parameters ); `could this have any unintended side effects?

I'm not sure, TBH. But considering that we always had the same logic in other places (like Modal), I would assume that it is fine unless we get it reported.

David-Moreira (approved):

PR #5763, opened 2024-10-02
Skeleton component

### LINKED ISSUE #1086: Skeleton components
**Loading "Skeleton" feature.**

New component to provide a generic loading templates.
Similar to this [Ant Design implentation](https://ant.design/components/skeleton/).

This is provides a visually appealing way for the user to wait on data loading from the server.

Example:
image

stsrki:
Closes #1086


Documentation/Blazorise.Docs/Pages/Docs/Components/Skeletons/SkeletonPage.razor
```diff
@@ -0,0 +1,88 @@
+﻿@page "/docs/components/skeleton"
+
+<Seo Canonical="/docs/components/skeleton" Title="Blazorise Skeleton system" Description="Learn to use and work with the Blazorise Skeleton component, loading placeholders for your components or pages to indicate something may still be loading." />
+
+<DocsPageTitle Path="Components/Skeleton">
+    Blazorise Skeleton component
+</DocsPageTitle>
+
+<DocsPageLead>
+    Use loading placeholders for your components or pages to indicate something may still be loading.
```
David-Moreira, line 10: And just serving as a placeholder is that a viable case?

Source/Blazorise/Components/Skeleton/Skeleton.razor.cs
```diff
@@ -0,0 +1,49 @@
...
+
+    #region Properties
+
+    /// <summary>
+    /// Gets or sets the animation style applied to the skeleton.
+    /// </summary>
+    /// <value>
+    /// A <see cref="SkeletonAnimation"/> value that determines the animation style.
+    /// </value>
+    [Parameter] public SkeletonAnimation Animation { get; set; }
```
David-Moreira, line 38: Do we need the dirty we usually have?

David-Moreira (approved):
Code LGTM. I haven't taken it for a spin. I'll do it later if you're able to hold on to the PR.

Source/Blazorise/Components/Skeleton/Skeleton.razor.cs
stsrki: Done.

Documentation/Blazorise.Docs/Pages/Docs/Components/Skeletons/SkeletonPage.razor
stsrki: That's why they are used, yes.

stsrki:
> Code LGTM. I haven't taken it for a spin. I'll do it later if you're able to hold on to the PR.

No problem. I can wait.

David-Moreira:
Looking good to me, just a few points:

1. Do we need the first Card example? I come to the skeleton page expecting skeleton examples, at first I'm trying to figure out how Skeleton affects the first example, only to check the code to see it's just a regular card example
image

2. I'd expect the `SkeletonTable` to at least have the `Animation` Parameter. I guess it might be common to make it seem more like it's loading a table, instead of just being a static like table?

3. The examples are all just simple lines. Isn't it common to also use circles or rectangles, etc? Do we support that easily? Can we add examples?

stsrki:
> Looking good to me, just a few points:
> 
> 1. Do we need the first Card example? I come to the skeleton page expecting skeleton examples, at first I'm trying to figure out how Skeleton affects the first example, only to check the code to see it's just a regular card example
>    image
> 2. I'd expect the `SkeletonTable` to at least have the `Animation` Parameter. I guess it might be common to make it seem more like it's loading a table, instead of just being a static like table?
> 3. The examples are all just simple lines. Isn't it common to also use circles or rectangles, etc? Do we support that easily? Can we add examples?

1. It's a copy of an example from BS documentation where it shows side-by-side how it would represent a final UI.

image

2. You're right. I forgot to add it.

3. For now, we only support lines.

David-Moreira:
1. Well they make sure to document why the first card example is there. I would do the same.

David-Moreira:
2. Only other thing I can think off, is if you had also SkeletonTableRow / Column, for additional flexibility. But I think it's probably too much for this feature. We can do it if a user requests I guess?

David-Moreira:
> 2. Only other thing I can think off, is if you had also SkeletonTableRow / Column, for additional flexibility. But I think it's probably too much for this feature. We can do it if a user requests I guess? User can always build his skeleton anyway using the base parts.

PR #5768, opened 2024-10-04
DataGrid: add text styling parameters

### LINKED ISSUE #5656: Add TextOverflow Properties to Datagrid Columns
Currently you cannot Truncate the Text fields inside DataGridColumns.

It would be nice if you could set it on table level like the Text Component has.

On Tables with much data you could set a "easy scroll" function so that every row is equal in size and rows with much data won't expand.

stsrki:
Closes #5656

The original ticket only mentioned TextOverflow. But, I have also added all other parameters related to text stylings. I have covered display cell, filter, header, and aggregate cells.

### DIFF
--- a/Documentation/Blazorise.Docs/Pages/Docs/Extensions/DataGrid/DataGridApi.razor
+++ b/Documentation/Blazorise.Docs/Pages/Docs/Extensions/DataGrid/DataGridApi.razor
@@ -480,17 +480,77 @@
     <DocsAttributesItem Name="ReverseSorting" Type="bool" Default="false">
         Whether the sort direction will be reversed.
     </DocsAttributesItem>
-    <DocsAttributesItem Name="TextAlignment" Type="TextAlignment?" Default="Default">
+    <DocsAttributesItem Name="TextAlignment" Type="TextAlignment" Default="Default">
         Defines the alignment for display cell.
+    </DocsAttributesItem>
+    <DocsAttributesItem Name="TextTransform" Type="TextTransform" Default="Default">
+        Gets or sets the text transformation for display cell.
+    </DocsAttributesItem>
+    <DocsAttributesItem Name="TextDecoration" Type="TextDecoration" Default="Default">
+        Gets or sets the text decoration for display cell.
+    </DocsAttributesItem>
+    <DocsAttributesItem Name="TextWeight" Type="TextWeight" Default="Default">
+        Gets or sets the text weight for display cell.
+    </DocsAttributesItem>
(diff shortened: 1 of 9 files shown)


Documentation/Blazorise.Docs/Pages/Docs/Extensions/DataGrid/DataGridApi.razor
```diff
@@ -480,18 +480,78 @@
...
+    <DocsAttributesItem Name="TextOverflow" Type="TextOverflow" Default="Default">
+        Determines how the text will behave when it is larger than a parent container for display cell.
+    </DocsAttributesItem>
+    <DocsAttributesItem Name="TextSize" Type="IFluentTextSize" Default="null">
+        Determines the font size of an element for display cell.
+    </DocsAttributesItem>
     <DocsAttributesItem Name="HeaderTextAlignment" Type="TextAlignment?" Default="null">
         Defines the alignment for column header cell.
     </DocsAttributesItem>
+    <DocsAttributesItem Name="HeaderTextTransform" Type="TextTransform?" Default="null">
```
David-Moreira, line 504: Have we always been using nullable types in the docs?

Source/Extensions/Blazorise.DataGrid/DataGridColumn.cs
```diff
@@ -647,6 +762,31 @@ public SortDirection CurrentSortDirection
     /// </summary>
     [Parameter] public TextAlignment TextAlignment { get; set; }
 
+    /// <summary>
+    /// Gets or sets the text transformation for display cell.
```
David-Moreira, line 766: The doc mentions this is for **display cell.** However, both Filter and Aggregate parameters will fallback to these regular parameters. Is this clear to the user? I think it might be misleading as it says **"for display cell"**

David-Moreira (approved):
LGTM

Documentation/Blazorise.Docs/Pages/Docs/Extensions/DataGrid/DataGridApi.razor
stsrki: Do you mean the `?` sign? I believe so, we did where it was valid to be mentioned.

Source/Extensions/Blazorise.DataGrid/DataGridColumn.cs
stsrki: You're right. I will change a description a bit.

PR #5776, opened 2024-10-14
Video: add ClickToFullscreen API

### LINKED ISSUE #5775: Video: Disable Fullscreen
Hi, 

is there any way to disable the video to enter fullscreen if a user double taps on a video.

stsrki:
Closes #5775


Documentation/Blazorise.Docs/Pages/Docs/Extensions/Video/VideoApi.razor
```diff
@@ -146,4 +146,7 @@
     <DocsAttributesItem Name="SettingsList" Type="VideoSettingsType[]" Default="Captions, Quality, Speed, Loop">
         If the default controls are used, you can specify which settings to show in the menu.
     </DocsAttributesItem>
+    <DocsAttributesItem Name="ClickToFullscreen" Type="bool" Default="true">
+        Double-press the video container to toggle fullscreen.
```
David-Moreira, line 150: Is it double press or single click? I mean it should be fairly obvious that a regular click will not full screen it, however to make sure we're explicit and don't leave room for misinterpretations, shouldn't the name of the parameter better reflect this?

"`DoubleClickToFullScreen`?"

Documentation/Blazorise.Docs/Pages/News/2024-10-15-release-notes-170.razor
```diff
@@ -312,4 +312,12 @@
     We have improved the way how our Fluent Flex utilities work. Previously it was not possible to define a breakpoint on per-flex rule, which made it hard to define different flex rules for different breakpoints. Now you can define the flex rules for each breakpoint separately, and you can also define the order of the flex rules.
 </Paragraph>
 
+<Heading Size="HeadingSize.Is3">
+    Disable Video Fullscreen
+</Heading>
+
+<Paragraph>
+    Added <Code>ClickToFullscreen</Code> parameter to the <Code Tag>Video</Code> component. This can be useful if you want to prevent the video from going fullscreen when the user double clicks on the video player.
```
David-Moreira, line 320: Seems like "Disable Video Fullscreen" is a byproduct of disabling the **ClickToFullscreen**.  So the title really should be "Double Click To FullScreen" or something along those lines? And then on the text we can inform, that's useful to disable it in order to prevent fullscreen.

Source/Extensions/Blazorise.Video/Video.razor.cs
```diff
@@ -70,37 +70,41 @@ protected override async Task OnAfterRenderAsync( bool firstRender )
...
+                Controls = Controls,
+                ControlsList = ControlsList,
+                SettingsList = SettingsList,
+                AutomaticallyHideControls = AutomaticallyHideControls,
+                AutoPlay = AutoPlay,
+                AutoPause = AutoPause,
+                Muted = Muted,
+                Source = Source,
+                Poster = Poster,
+                Thumbnails = Thumbnails,
```
David-Moreira, line 84: If we want to, we can make a dedicated constructor, so it's more close to the previous implementation and also enforces the number of arguments.

I do like seeing the properties explicitly defined however.

David-Moreira (changes requested):

Documentation/Blazorise.Docs/Pages/News/2024-10-15-release-notes-170.razor
David-Moreira: i.e: "You can now control whether double clicking the video player will enter full screen. By disabling this parameter, the user will not be able to enter full screen in this way."

(Is there a full-screen button or programatic way to enter fullscreen? If there is then we need to be explicit that this prevents full screen only on double click, but full screen can still be achieved through the button or programatically.)

stsrki: Yes, there is still an option to define a button by defining it as part of the `ControlsList` parameter.

Documentation/Blazorise.Docs/Pages/News/2024-10-15-release-notes-170.razor
```diff
@@ -312,4 +312,12 @@
     We have improved the way how our Fluent Flex utilities work. Previously it was not possible to define a breakpoint on per-flex rule, which made it hard to define different flex rules for different breakpoints. Now you can define the flex rules for each breakpoint separately, and you can also define the order of the flex rules.
 </Paragraph>
 
+<Heading Size="HeadingSize.Is3">
+    Disable Video Fullscreen
```
David-Moreira, line 316: I'd still add "Disable Video Fullscreen on Double Click". But I guess it's fine to keep the Heading simpler if you prefer.

David-Moreira (approved):

PR #5779, opened 2024-10-15
DataGridColumn: Undeprecate CellClass and CellStyle

### LINKED ISSUE #5773: Undeprecate CellStyle / CellClass
Deprecating CellStyle and CellClass at the DataGridColumn level in favor of CellStyling at the DataGrid level seems pretty inconvenient for simple cases like just disabling line wrapping for a particular column.  The CellStyling lambda provides a DataGrid column which I guess theoretically could be used to find the column that I want to apply the style/class to, but how?  Some columns have a Field property that I guess you could use, but it's not required, and my command columns will definitely not have it.  If there were an ordinal property it would help, but we don't currently get that.  Worse if I have 2 or more columns that I want to apply a CellStyle or a CellClass to now I have to write a switch statement or something which will be really ugly.  I can imagine CellStyling could be very useful and powerful in certain circumstances, but please consider making it's use optional for simpler cases by undeprecating CellStyle and CellClass.

stsrki:
Closes #5773

### DIFF
--- a/Documentation/Blazorise.Docs/Pages/News/2024-10-15-release-notes-170.razor
+++ b/Documentation/Blazorise.Docs/Pages/News/2024-10-15-release-notes-170.razor
@@ -188,6 +188,14 @@
     All the new text styling parameters are also available for headers, filter, and aggregate cells of the DataGrid.
 </Paragraph>
 
+<Heading Size="HeadingSize.Is4">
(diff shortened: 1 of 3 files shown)


David-Moreira (approved):
I guess it's for the best. Code won't be as elegant to handle all the multiple cases, but the surface api flexibility is what the devs really care about.

PR #5787, opened 2024-10-17
RouterTabs component

David-Moreira:
Based on Ant-design's implementation, we've based ourselves on their implementation to make the core of it work (tabs based on navigation, etc...). This should now serve as a base for us to expand however we want.

- By default it uses the Type name as the tab title.
- The title and other properties can be defined by using a specialized attribute:
- The url is synced when switching tabs happens
- The tabs don't really have the concept of closing, so I added the `CloseButton` and a custom handler, but it might be good to add `CloseButton` support for tabs on a separate ticket?
image
image

I'm not sure If I would add too much functionality around this component yet. I believe this might be a fine starting point for us to collect user feedback on its usage.

**Notes:**
- Don't forget to add proper Service Injection
- Don't forget Docs/examples/how to 
- @stsrki are you fine with current naming?

Please let me know your thoughts.


Source/Extensions/Blazorise.Components/RouterTabsPageAttribute.cs
```diff
@@ -0,0 +1,33 @@
...
+{
+    /// <summary>
+    /// Sets the title of the router tab.
+    /// </summary>
+    public string Title { get; set; }
+
+    /// <summary>
+    /// Sets the css class of the router tab.
+    /// </summary>
+    public string TabCssClass { get; set; }
```
stsrki, line 21: Why are `CssClass`es needed?

Demos/Blazorise.Demo/Pages/Dashboard.razor
```diff
@@ -1,5 +1,6 @@
 ﻿@page "/"
 @inject IVersionProvider VersionProvider
+@attribute [RouterTabsPageAttribute( Title = "Dashboard", TabCssClass = "font-weight-bold" )]
```
stsrki, line 3: The class names should be CSS provider agnostic. So in this case `"font-weight-bold"` would only work for BS.

Source/Extensions/Blazorise.Components/RouterTabs.razor.cs
```diff
@@ -0,0 +1,66 @@
...
+
+    [Inject] public NavigationManager NavigationManager { get; set; }
+    [Inject] public RouterTabsService RouterTabsService { get; set; }
+
+    protected override void OnInitialized()
+    {
+        base.OnInitialized();
+
+        NavigationManager.LocationChanged += OnLocationChanged;
+        RouterTabsService.OnStateHasChanged += OnStateHasChanged;
```
stsrki, line 26: Should we check for nulls?

Source/Extensions/Blazorise.Components/RouterTabsService.cs
```diff
@@ -0,0 +1,140 @@
...
+    public string TabCssClass { get; set; }
+    public string TabPanelCssClass { get; set; }
+    public bool Closeable { get; set; } = true;
+
+}
+
+public class RouterTabsService
+{
+
+    internal event Action OnStateHasChanged;
```
stsrki, line 24: We usually don't name events with `On` prefixes. The handlers have it usually.

Source/Extensions/Blazorise.Components/RouterTabsPageAttribute.cs
David-Moreira: Why not? It's an attribute, user sets whatever he wants in there?

Demos/Blazorise.Demo/Pages/Dashboard.razor
David-Moreira: Uhm, it's whatever the user wants, this is just an example.

Source/Extensions/Blazorise.Components/RouterTabs.razor.cs
David-Moreira: We can. But the component would already be in a bad state if these were nulls.

Source/Extensions/Blazorise.Components/RouterTabsService.cs
David-Moreira: `StateHasChangedHandler`?

David-Moreira:
Please review the actual component and how it works, see if there's any concerns before we polish / docs.

Source/Extensions/Blazorise.Components/RouterTabsService.cs
stsrki: `StateHasChanged`.

stsrki:
> Please review the actual component and how it works, see if there's any concerns before we polish / docs.

Not much to says here. It works.

David-Moreira:
@stsrki 
1. The tabs don't really have the concept of closing, so I added the CloseButton and a custom handler, but it might be good to add CloseButton support for tabs on a separate ticket?
2. Are you fine with current naming?

stsrki:
> 1. The tabs don't really have the concept of closing, so I added the CloseButton and a custom handler, but it might be good to add CloseButton support for tabs on a separate ticket?

Not sure about this one. As you have said regular tabs have no concept of closing. We don't need to make it part of `CloseButton`.

> 2. Are you fine with current naming?

I would lose the `Css` in the middle of attribute names. ie `TabClass`. 

For `Title`, maybe name it a `Name`?

stsrki:
I think something is wrong with the DocsRouterTabsLayout.razor and how we show the router example. It messes the rendering of the page, and I noticed that it makes a refresh every time I navigate to the `docs/extensions/routertabs` page.

It also doesn't detect all headers for the TOC on the side:

image

---

Can we make router examples some other ways maybe? Could the `iframe` help it to make it isolated?

stsrki (commented):
Code seems good to me. Just the example in the docs seem "problematic" to me-

David-Moreira:
**1. It also doesn't detect all headers for the TOC on the side:**
Not an issue with the component. It should render the Page Body as any other blazor use-case. The installation was set as a PageSubtitle. It was probably copied from somewhere like that, by the way.

**2. Makes a refresh every time**
The issue does not seem to be related to the component. I tried removing the component and just using the Layout. By moving between two different layouts, it seems like it gets "refreshed". Makes sense if you think about it, it's a new set of layout components if you switch between layouts.
In order to fix this "issue", I removed the custom layout and introduced a field in the regular Layout to track whether the RouterTabs should render according to the location. I think this is fine for the purposes of the docs / example.

stsrki (approved):

PR #5788, opened 2024-10-18
Web: theme selector

stsrki:
I was a little bored, so I created a dark theme selector for our website. 

Currently, all it does is use a default dark theme that comes from Bootstrap 5 by default. I have implemented a simple ThemeService that just raises a `changed` event when the theme is selected, and then in various places, I listen for the event. Then, just apply light or dark colors where needed.


David-Moreira:
> I was a little bored, so I created a dark theme selector for our website.

Little bored? There's plenty of tickets to go around!! (j/k j/k :D)

stsrki:
> > I was a little bored, so I created a dark theme selector for our website.
> 
> Little bored? There's plenty of tickets to go around!! (j/k j/k :D)

I'm leaving those as your last dance 😜

Documentation/Blazorise.Docs/App.razor
```diff
@@ -117,3 +118,20 @@
...
 </body>
 </html>
+@code {
+    RenderFragment customIcon =@<Image Source="/img/logos/blazorise-small.svg" Text="Small Blazorise logo" Style="width:32px; height: 32px" />;
+
+    [Inject] private ThemeService ThemeService { get; set; }
+
+    protected override void OnInitialized()
+    {
+        ThemeService.ThemeChanged += OnThemeChanged;
```
David-Moreira, line 128: Not sure whether dispose matters or not here?

Documentation/Blazorise.Docs/Components/Commercial/PlanSwitch.razor.cs
```diff
@@ -0,0 +1,43 @@
+﻿using System;
+using System.Threading.Tasks;
+using Blazorise.Docs.Services;
+using Microsoft.AspNetCore.Components;
+
+namespace Blazorise.Docs.Components.Commercial;
+
+public partial class PlanSwitch : IDisposable
+{
+    private Background GetSelectedBackground( string value ) => SelectedValue == value ? Background.Primary : Background.Default;
```
David-Moreira, line 10: I don't understand these getters, if the **SelectedValue** is already equal to the value that's in the argument it chooses Background.Primary, otherwise Default?

Documentation/Blazorise.Docs/Components/Commercial/PlanSwitch.razor.cs
```diff
@@ -0,0 +1,43 @@
...
+
+    private TextColor GetTextSelectedColor( string value ) => SelectedValue == value
```
David-Moreira, line 12: Same thing, don't understand why SelectedValue == value turns into TextColor.White?

Documentation/Blazorise.Docs/Components/Footer.razor
```diff
@@ -103,4 +104,26 @@
...
+
+        base.OnInitialized();
+    }
+
+    public void Dispose()
+    {
+        ThemeService.ThemeChanged -= OnThemeChanged;
+    }
+
+    async void OnThemeChanged( object sender, string theme )
```
David-Moreira, line 125: If you do StateHasChanged in app.razor, is that not enough to update all the underlying components?

Documentation/Blazorise.Docs/Routes.razor.cs
```diff
@@ -0,0 +1,56 @@
...
+    [Inject] ILocalStorageService LocalStorage { get; set; }
+
+    protected override void OnInitialized()
+    {
+        ThemeService.ThemeChanged += OnThemeChanged;
+
+        base.OnInitialized();
+    }
+
+    protected override async void OnAfterRender( bool firstRender )
```
David-Moreira, line 21: Am I missing something? Isn't there a trully async version?

Documentation/Blazorise.Docs/Components/Footer.razor
```diff
@@ -103,4 +104,26 @@
             </UnorderedList>
         </Column>
     </Row>
-</Container>
\ No newline at end of file
+</Container>
+@code{
+    [Inject] private ThemeService ThemeService { get; set; }
+
+    private TextColor IconTextColor => ThemeService.CurrentTheme == "Dark" ? TextColor.Light : TextColor.Dark;
```
David-Moreira, line 111: `ThemeService.IsDark` helper?

Documentation/Blazorise.Docs/Routes.razor.cs
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    public void Dispose()
+    {
+        ThemeService.ThemeChanged -= OnThemeChanged;
+    }
+
+    async void OnThemeChanged( object sender, string theme )
+    {
+        if ( theme == "Dark" )
```
David-Moreira, line 43: Check some centralized constant on `ThemeService`?

Documentation/Blazorise.Docs/Services/ThemeService.cs
```diff
@@ -0,0 +1,33 @@
...
+public class ThemeService
+{
+    public string CurrentTheme = "Light";
+
+    public EventHandler<string> ThemeChanged;
+
+    public Background BarBackground = Background.Light;
+    public ThemeContrast BarThemeContrast = ThemeContrast.Light;
+
+    public bool IsDark => CurrentTheme == "Dark";
```
David-Moreira, line 14: "Dark" as constant?

Source/Blazorise/wwwroot/utilities.js
```diff
@@ -24,6 +24,14 @@ export function toggleClass(element, classname) {
     }
 }
 
+export function addAttribute(element, attribute, value) {
+    element.setAttribute(attribute, value);
```
David-Moreira, line 28: You usually have the habit of null checking? Should null check?

David-Moreira (approved):

Documentation/Blazorise.Docs/Routes.razor.cs
```diff
@@ -0,0 +1,56 @@
...
+        {
+            await JSUtilitiesModule.AddAttributeToBody( "data-bs-theme", "dark" );
+        }
+        else
+        {
+            await JSUtilitiesModule.RemoveAttributeFromBody( "data-bs-theme" );
+        }
+
+        await LocalStorage.SetItemAsync( "theme", theme );
```
tesar-tech, line 52: Using local storage with Blazor Server results in unintended behavior. 
Since the first "render" happens on the server, where local storage is not accessible, it will always serve the light (default) mode. Then, it checks for the "dark" mode and switches accordingly. This results in a quick "light theme blink" on refresh, which is exactly what's happening. The "quickness" of the blink depends on internet speed. You can throttle in devtools for a more noticeable effect. 

To my knowledge, the only solution is to store the "dark" mode value inside a cookie, which will be sent together with the first request. This ensures the first render happens in the correct dark/light theme.

Documentation/Blazorise.Docs/Components/DocsAttributes.razor.cs
```diff
@@ -1,16 +1,30 @@
...
 namespace Blazorise.Docs.Components;
 
-public partial class DocsAttributes
+public partial class DocsAttributes : IDisposable
 {
     #region Methods
 
+    protected override void OnInitialized()
+    {
+        ThemeService.ThemeChanged += OnThemeChanged;
```
tesar-tech, line 18: This is here because we want to make sure the page re-render with the correct theme (`@ThemeContrast`)? 
I think it's enough if the re-render happens in some parent component, resulting in re-render also in here, thus the whole `ThemeService.ThemeChanged`, `StateHasChanged` code has no use in here (and in other child components). It just adds another re-render.

Documentation/Blazorise.Docs/Routes.razor.cs
```diff
@@ -0,0 +1,56 @@
...
+        base.OnInitialized();
+    }
+
+    protected override async void OnAfterRender( bool firstRender )
+    {
+        if ( firstRender )
+        {
+            var theme = await LocalStorage.GetItemAsync<string>( "theme" );
+
+            if ( ThemeService.CurrentTheme != theme )
```
tesar-tech, line 27: What about putting the check (`( ThemeService.CurrentTheme != theme )`) inside the `SetTheme`?
This will also "protect" from the needless re-renders when user with light theme clicks again the light theme..

Documentation/Blazorise.Docs/wwwroot/blazorise.docs.min.css
tesar-tech, line 1: my favorite topic - generated file should not be in vcs.

tesar-tech (commented):
For the dark/light theme switcher, we should also implement a third value, "use system default," which should be the default option.
 See [this](https://jonhilton.net/blazor-tailwind-dark-mode-local-storage/) for implementation (scroll down to `Respect their OS preferences`).

Documentation/Blazorise.Docs/Components/Commercial/PlanSwitch.razor.cs
stsrki: It just applies the primary color to the selected button. That's it.

Documentation/Blazorise.Docs/Components/Commercial/PlanSwitch.razor.cs
stsrki: White looks better on the primary purple color. That's why it is white in selected mode.

Documentation/Blazorise.Docs/Components/Footer.razor
stsrki: I have removed code from App.razor. It didn't work

Documentation/Blazorise.Docs/Styles/_base.scss
```diff
@@ -9,8 +12,28 @@
     -moz-osx-font-smoothing: grayscale;
 }
 
+body {
```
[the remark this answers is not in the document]
stsrki: This file must be present at all times. SCSS is a compiler with the help of VS extension during the development stage. So we need to keep the final CSS files.

Documentation/Blazorise.Docs/Routes.razor.cs
stsrki: That's the limitation at this time. We could use cookies but the GDPR consent complicates it. SO I would just leave it as it is and go with the flicker.

Documentation/Blazorise.Docs/wwwroot/blazorise.docs.min.css
stsrki: As mentioned before. CSS files must be included.

Documentation/Blazorise.Docs/Components/DocsAttributes.razor.cs
stsrki: It doesn't work all the times if I leave it at parent level. That's why I have the event handler at child components.

stsrki:
I have fixed a bunch of code according to your comments. And introduced the new System option to detect dark mode.

It's not the most beautiful code but it works. Please retest.

Documentation/Blazorise.Docs/App.razor
```diff
@@ -117,3 +117,6 @@
     </script>
 </body>
 </html>
+@code {
+    RenderFragment customIcon =@<Image Source="/img/logos/blazorise-small.svg" Text="Small Blazorise logo" Style="width:32px; height: 32px" />;
+}
```
David-Moreira, line 122: Seems like it's added and not used can you check?

Documentation/Blazorise.Docs/Components/Commercial/PlanSwitch.razor.cs
David-Moreira: I still don't understand this at all personally... But as long as it's working as expected, I'm ok with it, then.

Documentation/Blazorise.Docs/Services/ThemeService.cs
```diff
@@ -0,0 +1,54 @@
+﻿using System;
+
+namespace Blazorise.Docs.Services;
+
+public class ThemeService
```
David-Moreira, line 5: Make sure that the accessors properly encapsulate logic, just for good practices sake, if you care about it, otherwise it's fine.
i.e **SystemIsDarkMode** can be publicly set directly.

Source/Blazorise/Interfaces/Modules/IJSUtilitiesModule.cs
```diff
@@ -158,4 +173,10 @@ public interface IJSUtilitiesModule : IBaseJSModule
     /// <param name="args">Optional parameters.</param>
     /// <returns>A task that represents the asynchronous operation.</returns>
     ValueTask Log( string message, params string[] args );
+
+    /// <summary>
+    /// Checks if the current theme is in dark mode.
+    /// </summary>
+    /// <returns>A task that represents the asynchronous operation. The task result contains true if the theme is in dark mode, otherwise false.</returns>
+    ValueTask<bool> IsDarkMode();
```
David-Moreira, line 181: Since this is an utilities module, we could potentially make it clear that It's related to the Theme, i.e **IsThemeDarkMode**

David-Moreira: Actually, maybe even `IsSystemDarkMode` just to be clear, that this is never user set. but whatever it is figured out from the browser/system.

David-Moreira (approved):

Documentation/Blazorise.Docs/Components/DocsAttributes.razor.cs
tesar-tech: I see.  In that case I would suggest to investigate the reason why it doesn't work, because it doesn't appear as standard behavior and may backfire badly in the future... 
Also doing the event subscription and un-subscription in every component where you want to use a `ThemeVariable` looks quite painful to me...

My guess is that something is wrong with the non-async version of the ThemeChanged (` async void OnThemeChanged`).. Maybe making it to return Task<string> would solve the issue (really, just a guess).

Issue #5355, opened 2024-03-11
[Bug]: Autocomplete list has transparent slider background


mtbayley:
### Blazorise Version

1.4.2

### What Blazorise provider are you running on?

Bootstrap5

### Link to minimal reproduction, or a simple code snippet

https://blazorise.com/docs/extensions/autocomplete

![image](https://github.com/Megabit/Blazorise/assets/34929316/8471440e-866f-463f-9c29-94bcdc2357a1)

### Steps to reproduce

View demo

### What is expected?

Slider should not be transparent

### What is actually happening?

Slider is transparent

PR #5795, opened 2024-10-22
Web: fix transparent background on scrollbar in autocomplete

tesar-tech:
## Description
transparent bg on scrollbar fix.

image

Targeting master, because I wanted to make sure it will work with the recently implemented theme selector.

Also created a css variable for `--scrollbar-thumb-background` as it was already on 3 places...

I am not exactly sure if I placed the variable and modified css in the correct places, but this feels the most logical. 

Closes #insert_issue_number
https://github.com/Megabit/Blazorise/issues/5355

## How Has This Been Tested?
 
chromium,ff

## Types of changes

## Checklist:

- [ ] The PR is submitted to the correct branch (`master` for dev, `rel-1.x` for maintenance).


stsrki (commented):
I believe you misunderstood the problem. While the problem can be seen on our website, it might be that the original problem can be found for all our providers.

Remember. We have bootstrap 4 & 5,fluent, and others. So we need to fix it for each of them. If necessary.

First you need to be able to reproduce it in our demo. Test on bs5 first as that one is the same as we use on the web. Then go through each provider and check them. And fix css for each of them.

tesar-tech:
That's what I initially though, but the "issue" lies in the _base.scss of documentation project https://github.com/Megabit/Blazorise/blob/d5466989d186b6e5a847e97fde657e99b30520c4/Documentation/Blazorise.Docs/Styles/_base.scss#L62, if you suppress it in dev tools it's gone.

here it goes on simple demo which does not specify the transparency, as you can see the bg color of the slider happily follows the default slider style while not being transparent at all.

https://github.com/user-attachments/assets/0c57e396-c9dc-4a95-979b-55d7eefb90a2

correct me if I am wrong, but I really see no issue in the components, but rather on the blazorise website.

stsrki:
In that case I might be wrong :)

Documentation/Blazorise.Docs/Styles/_base.scss
```diff
@@ -10,6 +10,7 @@
     text-transform: none;
     -webkit-font-smoothing: antialiased;
     -moz-osx-font-smoothing: grayscale;
+    --scrollbar-thumb-background: #bfbfbf;
```
stsrki, line 13: We tend to add a prefix to the var names. Use `--b-scrollbar-thumb-background`.

stsrki (approved):
LGTM

PR #5801, opened 2024-10-23
fix on disabled button pointer

### LINKED ISSUE #5361: [Bug]: mouse pointer on disabled button in bar is wrong
### Blazorise Version

1.3.4 , 1.4.2

Bootstrap5

```tsx
<Bar>
  <BarStart>
    <BarItem>              
      <Button Color="Color.Primary" Disabled>test 1, not ok</Button>
      <Button Color="Color.Primary">test 2, ok</Button>               
    </BarItem>
  </BarStart>
  <BarItem>
    <Button Color="Color.Primary" Disabled>test 3 , ok</Button>     
  </BarItem>
</Bar>
```

### Steps to reproduce

Create a Bar with a BarStart then add a button that is disabled. When you place the mouse over the disabled button it will keep showing the 'hand' pointer instead of the arrow as would be expected on a disabled button.
When the button is used outside the Bar, BarStart it works as expected

### What is expected?

the mouse button should be the default pointer on the disabled button 

the hand pointer is showing as if the button is active

Chrome, Microsoft Edge

### Any additional comments?

_No response_

tesar-tech:
## Description

Fixes the issue where a disabled button had the `pointer` cursor.

The issue is more complicated than it seems, mainly due to how the cursor is set for buttons when they are not disabled. When a button is disabled, it inherits the `pointer` cursor from its parent, so you cannot have something like this:

```html
<div class="nav-item cursor-pointer">
  <button disabled/>
  <button />
</div>
```

and expect to have different cursors on the buttons.

The provided solution addresses this issue by moving the `cursor-pointer` to the element of `BarItem`, allowing for multiple buttons. 

It also solves the following case:

```html
<div class="nav-item">
  <div class="cursor-pointer">
    <button disabled/>
  </div>
  <button class="cursor-pointer" />
</div>
```

In this case, the `cursor-pointer` is applied to the `div` that isn't disabled, but it contains a disabled child button. This reintroduces the problem where the parent’s cursor affects the disabled child.

The solution fixes this by checking for an element with the `.disabled` class within the entire nested structure. If such an element is found, it avoids applying the `cursor-pointer` class to the direct descendant of the `nav-item`.

This clarifies why the fix is a bit more complex.

We could have resolved this by applying `cursor-default` to the disabled button directly, but that would override the cursor expected from a parent element, which may not be desirable.

So, this is a bit of a breaking change—setting the pointer behavior on the elements inside the `nav-item` instead of the `nav-item` itself.

Let me know what you think. 
It is also needed to do it for links, as they have similar problem.

Closes #5361

## How Has This Been Tested?
chromium, ff
## Types of changes

## Checklist:


Source/Blazorise.Bootstrap/Styles/components/_bar.scss
```diff
@@ -1,9 +1,9 @@
-﻿.navbar-nav .nav-item:hover *:not(:has(.disabled)):not(.disabled) {
+﻿.navbar-nav .nav-link:hover {
```
tesar-tech, line 1: With this change it makes the buttons to have the correct cursor, but that's because non of the elements in BarItem has the cursor pointer (except buttons, because they have them "naturally").

```html
<Bar>
  <BarStart>
    <BarItem>              
      <Button Color="Color.Primary" Disabled>test 1, not ok previously, ok no</Button>
      <Button Color="Color.Primary">test 2, ok</Button>               
    </BarItem>

      <BarItem>              
      <Div> 
          I had cursor-pointer, but has normal cursor now.
      </Div>
    </BarItem>
  </BarStart>
  <BarItem>
    <Button Color="Color.Primary" Disabled>test 3 , ok</Button>     
  </BarItem>
</Bar>
```
So we either:
- discard the functionality of having the BarItem to always has cursor pointer (on divs, buttons, everything) => disabled buttons will always work, but it's quite a breaking change. (this is the solution with your last SCSS structure commit)
- Place the cursor pointer only when there is no disabled button =>  buttons work as expected unless wild structure is introduced. Anything else works as before. Bit of breaking change too, because the `cursor:pointer;` style is not directly on BarItem (.nav-item), but on the descendants..

stsrki: `BarItem` should NOT have the pointer cursor. The `BarLink` should have it. So why not? Because, the BarItem is a container for links, buttons, dropdowns, and whatever else the user can place in it. If it had the cursor pointer, it would be wrong from a UX perspective as it would indicate the wrong action to the user.

tesar-tech: But previously (before this pr) the BarItem had the cursor pointer. 

removing this
```css
.navbar-nav .nav-item:hover {
    cursor: pointer;
}
```

will:
- solve the OP's issue
- make it right from ux perspective (as you described)
- introduce a breaking change

stsrki: > * introduce a breaking change

Yea, we should do it. I might have accidentally introduced that bug in 2020. As far as I can see from Git history.

tesar-tech: Ok. I agree. 
As a breaking change -> should I edit the release note, or what's the procedure?

I will also check the link, as it seems there is some unnecessary css.

stsrki: > As a breaking change -> should I edit the release note, or what's the procedure?

Yes. But close this PR and make it for 1.7 instead.

Source/Blazorise.Bootstrap/Styles/components/_bar.scss
```diff
@@ -1,12 +1,4 @@
-﻿.navbar-nav .nav-link:hover {
-    cursor: pointer;
-
-    *:not(:has(.disabled)):not(.disabled) {
-        cursor: pointer;
-    }
-}
-
-.nav .nav-link:hover {
+﻿.nav .nav-link:hover {
```
tesar-tech, line 1: This looks useless to me. 

- I didn't discover a structure where .nav and the .nav-link are direct descendants. Because the .nav-link should be place inside the BarItem (.nav-item). right?

- Also I didn't found the usage of the `.nav` (but it's bit hard to find) class as our navbars are using `navbar-nav`..
- Also link have the cursors by being a `<a>`...

stsrki: I think that rule was meant for the Tabs. Which have `nav` class. But I'm not 100% sure.

image

PR #5806, opened 2024-10-28
Update to .NET9

stsrki:
All projects are updated to the latest .NET9 RC2. Once the final .NET9 is released I will remove the RC version tag. So, we're waiting until then with the merge.


David-Moreira:
@stsrki should we remove support for any .NET version or keeping all?

David-Moreira (approved):

stsrki:
> @stsrki should we remove support for any .NET version or keeping all?

The removal will be done in 2.0. Specifically NET 6 & 7.

PR #5807, opened 2024-10-28
DatePicker: add EnabledDates parameter

### LINKED ISSUE #5754: DatePicker: add EnabledDates
We already have `DisabledDates`, so having also `EnabledDates` would be good for customization.

Reference: https://flatpickr.js.org/examples/#disabling-all-dates-except-select-few

tesar-tech:
## Description
Enabled dates in DatePicker.

Few consideration:
- Empty array means "no dates are enabled". For changing from "some enabled dates" to "enable all dates" we need to set the EnabledDates param to null (there are workarounds).. Which will eventually run this code 

```js
//datepicker.js
 picker.set("_enable", undefined);
```
which sets the internal property to undefined. Which is kind of hackie solution (others didn't work), but I didn't noticed any side effects.. Other option is to make all the dates enabled, but this will mess up the dotnet-js synchronization. Note that this code will only run when you set the EnabledDates to null and they were previously set to non-null array. 

- In release notes I fixed heading for my previous PR. Doesn't belong here, but is just a small update. 

Closes #5754

## Types of changes

## Checklist:

### DIFF
--- a/Documentation/Blazorise.Docs/Pages/Docs/Components/Dates/Examples/DatePickerEnabledDatesExample.razor
+++ b/Documentation/Blazorise.Docs/Pages/Docs/Components/Dates/Examples/DatePickerEnabledDatesExample.razor
@@ -1,0 +1,12 @@
+﻿@namespace Blazorise.Docs.Docs.Examples
+
+<DatePicker TValue="DateTime?" EnabledDates="@enabledDates" />
+
+
+@code {
+
+    DateTime?[]? enabledDates = new DateTime?[]{
+        DateTime.Now.AddDays(-1),
+        DateTime.Now.AddDays(2),
+    };
+}
--- a/Documentation/Blazorise.Docs/Pages/Docs/Components/Dates/Examples/DatePickerEnabledDatesExample.razor
+++ b/Documentation/Blazorise.Docs/Pages/Docs/Components/Dates/Examples/DatePickerEnabledDatesExample.razor
@@ -5,7 +5,7 @@
 
 @code {
 
-    DateTime?[]? enabledDates = new DateTime?[]{
+    DateTime?[] enabledDates = new DateTime?[]{
         DateTime.Now.AddDays(-1),
         DateTime.Now.AddDays(2),
     };
--- a/Documentation/Blazorise.Docs/Pages/Docs/Components/Dates/Examples/DatePickerEnabledDatesExample.razor
+++ b/Documentation/Blazorise.Docs/Pages/Docs/Components/Dates/Examples/DatePickerEnabledDatesExample.razor
@@ -2,9 +2,7 @@
 
 <DatePicker TValue="DateTime?" EnabledDates="@enabledDates" />
 
-
 @code {
-
     DateTime?[] enabledDates = new DateTime?[]{
         DateTime.Now.AddDays(-1),
         DateTime.Now.AddDays(2),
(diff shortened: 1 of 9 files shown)


Source/Blazorise/Components/DatePicker/DatePicker.razor.cs
```diff
@@ -632,6 +637,12 @@ protected override IReadOnlyList<TValue> InternalValue
     /// </summary>
     [Parameter] public IEnumerable<TValue> DisabledDates { get; set; }
 
+    #nullable enable
```
stsrki, line 640: Remove nullable.

Source/Blazorise/Components/DatePicker/DatePicker.razor.cs
```diff
@@ -70,6 +71,7 @@ public override async Task SetParametersAsync( ParameterView parameters )
                     ExecuteAfterRender( async () => await JSModule.UpdateValue( ElementRef, ElementId, formatedDateString ) );
                 }
             }
+            Console.WriteLine( $"Params called enbaled date changesd {enabledDatesChanged}" );
```
stsrki, line 74: Remove console writeline

Source/Blazorise/wwwroot/datePicker.js
```diff
@@ -301,6 +307,14 @@ export function updateOptions(element, elementId, options) {
             picker.set("disable", disableDatesOptions.concat(disableDaysOptions));
         }
 
+        if (options.enabledDates.changed) {
+            if (options.enabledDates.value === null) {
```
stsrki, line 311: Check for both null or undefined?

Source/Blazorise/Components/DatePicker/DatePicker.razor.cs
tesar-tech: ups.

Source/Blazorise/wwwroot/datePicker.js
tesar-tech: I don't think there is a pathway for it, but you never know with js.. so I changed it

David-Moreira: We should already have these kind of checks in the code base, no?
Might be worth just having a general utility for **isNullOrUndefined** to avoid having multiple ways to check across the js code base.

Documentation/Blazorise.Docs/Pages/Docs/Components/Dates/Examples/DatePickerEnabledDatesExample.razor
```diff
@@ -0,0 +1,12 @@
+﻿@namespace Blazorise.Docs.Docs.Examples
+
+<DatePicker TValue="DateTime?" EnabledDates="@enabledDates" />
+
+
+@code {
+
+    DateTime?[]? enabledDates = new DateTime?[]{
```
stsrki, line 8: Make it as `DateTime?[]`.

Source/Blazorise/Components/DatePicker/DatePicker.razor.cs
```diff
@@ -632,6 +636,10 @@ protected override IReadOnlyList<TValue> InternalValue
     /// </summary>
     [Parameter] public IEnumerable<TValue> DisabledDates { get; set; }
 
+    /// <summary>
+    /// List of enabled dates - user can pick only these.
```
stsrki, line 640: The comment might need some changes. Maybe "List of enabled dates that the user should be able to pick.".

stsrki (commented):
Also don't forget to mention new API on the datepicker page under the attributes section.

stsrki (approved):

PR #5813, opened 2024-10-30
DataGrid: Column Filter | Add Between filter mode

David-Moreira:
@stsrki can you take a look at the POC?

Initially I was trying to just use SearchValues as an object that could either be an array or a regular object, but it was getting too complicated with casting and guessing, and trying to be overly dynamic.
So I changed to just set the two values explicitly we need for the range, with SetSearchValue1 and SetSearchValue2 and corresponding getters.

1. What are your thoughts on this approach?
2. Do you think regular **DataGridFilter** should support Between also? So below every column, there would be two text fields instead of one?


Source/Extensions/Blazorise.DataGrid/Contexts/FilterContext.cs
```diff
@@ -40,5 +61,11 @@ public void TriggerFilterChange( object value )
     /// </summary>
     public object SearchValue { get; set; }
 
+    /// <summary>
+    /// Gets or sets the filter value(s).
+    /// <para>Represents the value(s) for filter methods that are range based.</para>
+    /// </summary>
+    public object[] SearchValues { get; set; }
```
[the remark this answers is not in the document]
David-Moreira: I just told you in the pr description. I started with that and didn't like all the casting. If you are really keen on it, then ok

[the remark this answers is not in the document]
stsrki: I've just now understood your description. You mentioned `SearchValues as an object`, so I was confused by a plural form.

Anyways. Yes, try to go with the existing `SearchValue as an object` and cast where needed.

David-Moreira:
@stsrki 
Please take a look at this version. 
I personally dislike the obscurity of it. But it might be "easier" for our api internally I guess.
Let me know your thoughts.

stsrki (commented):
The code seems good to me.

But I have problems with selecting dates. If I click on a date it doesn't get applied to the filter. Just stays empty-

image

David-Moreira:
> The code seems good to me.
> 
> But I have problems with selecting dates. If I click on a date it doesn't get applied to the filter. Just stays empty-
> 
> image

Seems like a regression perhaps? Or we didn't notice it when we implemented the DatePicker in the filter? It should be fixed in a separate issue as it's not caused by this PR.

stsrki:
Just update release notes and add if needed add a new example in the docs. 

I will open new ticket for the date bug.

stsrki (approved):
LGTM. Do you plan to add anything else or is it ready for merging?

Documentation/Blazorise.Docs/Models/Snippets.generated.cs
```diff
@@ -8105,7 +8118,44 @@ protected override async Task OnInitializedAsync()
 
 @code {
     private DataGrid<Employee> dataGrid;
-    private List<Employee> employeeList = new() { new() { FirstName = ""David"", LastName = ""Moreira"", Gender = ""M"" }, new() { FirstName = ""MLaden"", LastName = ""Macanovic"", Gender = ""M"" }, new() { FirstName = ""John"", LastName = ""Doe"", Gender = ""M"" }, new() { FirstName = ""Ana"", LastName = ""Chamberlain"", Gender = ""F"" }, new() { FirstName = ""Jessica"", LastName = ""Winston"", Gender = ""F"" } };
+    private List<Employee> employeeList = new()
```
David-Moreira, line 8121: You've removed the Children assignment I had added:
image

Also why change these names, and not the ones from the other example?

stsrki: I missed that, sorry.

For the new names. Just for the change. And it feels strange to have my name still there 😅

David-Moreira: The example above also has your name...
I'd go through the docs and change everywhere if that's a concern.

stsrki: > I'd go through the docs and change everywhere if that's a concern.

That's the next step.

David-Moreira:
> LGTM. Do you plan to add anything else or is it ready for merging?

Nope.

PR #5815, opened 2024-10-31
DataGrid: Group header column width fix

### LINKED ISSUE #5804: [Bug]: ShowHeaderGroupCaptions messes with DataGrid Layout
### Blazorise Version

1.6.2

Bootstrap5

https://github.com/Alexander-JM-0224/BlazoriseDataGridTestApp.git

### Steps to reproduce

Open the app, switch between the 2 pages in the navbar and see the difference.

### What is expected?

My expectation would be that when i have a nice working DataGrid where the styling is okay and the columns are the correct width. Adding the ShowHeaderGroupCaptions should not mess with the layout of the DataGrid

Well, when i add the ShowHeaderGroupCaptions is smushes soms columns. Does not respect the Width attributes we added to the columns.

First the DataGrid without ShowHeaderGroupCaptions.
Image

Then the DataGrid with ShowHeaderGroupCaptions.
Image

Now how it should be.
Image

What i need to do to make the DataGrid correct is in the inspector i need to turn 2 Width properties off, then it is showing correct spacing.
Image

Chrome, Firefox

### Any additional comments?

This is a small difference, but it is not something we want to have happen, and i sincerely doubt this should be the expected behaviour of the functionality.

Please, if i am missing something, or there is a simple workaround please let me know.

tesar-tech:
## Description

Closes #5804

The issue arises because the header group column inherits its width from the first header column with the specified `HeaderGroupCaption.` In my opinion, setting a width on the header group isn’t necessary, at least from what I can foresee.

The simplest solution is to set `Width` to `null`. Simply removing the width from the Razor file (line 42 in DataGrid) won’t be effective, as the width is also applied through `Style` property.

### DIFF
--- a/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
+++ b/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
@@ -2866,6 +2866,7 @@ public partial class DataGrid<TItem> : BaseDataGridComponent
             for ( int i = 0; i < orderedDisplayColumns.Count; i++ )
             {
                 var displayColumn = orderedDisplayColumns[i];
+                displayColumn.SetWidthToNull();
                 var colSpan = 1;
 
                 if ( !string.IsNullOrWhiteSpace( displayColumn.HeaderGroupCaption ) && orderedDisplayColumns.Count > i + 1 )
--- a/Source/Extensions/Blazorise.DataGrid/DataGridColumn.cs
+++ b/Source/Extensions/Blazorise.DataGrid/DataGridColumn.cs
@@ -383,6 +383,8 @@ public partial class DataGridColumn<TItem> : BaseDataGridColumn<TItem>
             : ParentDataGrid.FilterMethod == DataGridFilterMethod.NotEquals ? DataGridColumnFilterMethod.NotEquals
             : DataGridColumnFilterMethod.Contains;
     }
+    
+    internal void SetWidthToNull()=> Width = null;//to supress BL0005
 
     #endregion
 


Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -2866,6 +2866,7 @@ internal IEnumerable<DataGridColumn<TItem>> DisplayableColumns
             for ( int i = 0; i < orderedDisplayColumns.Count; i++ )
             {
                 var displayColumn = orderedDisplayColumns[i];
+                displayColumn.SetWidthToNull();
```
stsrki, line 2869: I don't think we should modify the Width here. What will happen when the user-defined the width on their side?

I think we should just change the Bootstrap CSS when the grouping is enabled on a table.

Here is an example from my comment in the ticket, as a reference:

```css
.table-fixed-header table.b-table
{
  width: unset:
}
```

stsrki: @David-Moreira agree?

tesar-tech: "What will happen when the user-defined the width on their side?" -> that's exactly what the bug is about. When user defines width it inherits after the first column that was defined with `HeaderGroupCaption`. Now it tries to squeeze 3 (or what ever count) columns into the size of 1...
With width disabled on header group, users can still set the width by defining it by the width of the individual items (which is exactly what is done in the bug report).

The css fix messes up the layout on bigger screens for example (isn't stretching itself to 100%)

David-Moreira: > "What will happen when the user-defined the width on their side?" -> that's exactly what the bug is about. When user defines width it inherits after the first column that was defined with HeaderGroupCaption. Now it tries to squeeze 3 (or what ever count) columns into the size of 1...

I think @stsrki meant the user defining the width in a CSS file meaning that your solution would only fix by overriding any Width explicitly defined as a Parameter and not as a css defined rule targeting that column, is this not the case?

So if @stsrki solution with css also fixes the issue, it has the advantage to also work over user's defined css. I'd personally take that approach if it works with no other issues.

David-Moreira: Additionally, I'd make it clear somewhere that this feature does not work properly with explicitly defined widths and make sure that the user knows what will happen.
Somethign along this lines:
- By enabling HeaderGroupCaption any explicitly defined width will be ignored, etc....

tesar-tech: Yes, even with my solution user can still get there (un)intended width as style or by class. There is no reason to do it by style - because user has dedicated Width property for that. If set by class, we can mitigate the effect by setting the width `unset` instead of null. 

The css solution has side effects: 

No header:

image

Header with issue:

image

When removing the 100% width (css fix)
image

My fix:
image

The css fix remove the 100% width, which is not intended (I guess) and will act differently on bigger screens. Maybe not important - you tell me.

My fix also doesn't keep the layout the same. But that is due to the fact that the stretching is now governed by the top header columns (header groups) - there are 4 of them and the stretching is happening proportionally to them, not to the lower header columns (as without the header group). Keeping the intended size of columns is only possible when the whole thing is not stretching (basically the css fix).

(My code still need some fixes, because it works with the column instance..)

stsrki: > There is no reason to do it by style - because user has dedicated Width property for that.

This is the problem because a lot of users actually use Style directly, so for them, this fix might break their UIs.

David-Moreira: > There is no reason to do it by style - because user has dedicated Width property for that.

I agree, In my onion, **never ever assume what the user should do.** There might be a myriad of reasons why he would set the width through regular styling... and it's perfectly reasonable to do so by styling.
So unless we deem something as unsuported by the framework/product we're providing, then if it's available the user might use it and we might want to handle it.

tesar-tech: Yeah, right.. 

So we have two solutions, both with breaking change.

What about letting users to do their own styling on the header group cell?

something like:

```html
<DataGridColumn Classes="what ever now user is responsible" IsHeaderGroup HeaderGroupCaption="Personal Info">
    
</DataGridColumn>
```

And instead of taking the styling from the first `DataGridColumn` with `HeaderGroupCaption="Personal Info"`, it would just take the one with  `IsHeaderGroup`. This might actually simplify few things along the line. 
Not a quick fix, but imo the cleanest solution...

stsrki: The user can already assign the class via the styling and cell styling callbacks on a column. And there is also a `Class` on a DataGrid.

I think it is best if we postpone this until we figure the best way of handling it. The problem is more in how the CSS is done on the bootstrap side. And user has a workaround by adding a bit of custom CSS if neeeded.

David-Moreira: So, just to be clear, this PR is postponed then? Should it be left hanging? OR should we update the issue with our findings and reject it for the time being?

stsrki: Yes, it is postponed. I will update the issue.

PR #5817, opened 2024-11-04
Update NuGet packages

stsrki:
This PR updates third-party packages that we use across the codebase. It also solves some vulnerability issues that VS was complaining about, specifically for the `Microsoft.AspNetCore.Http` package.


David-Moreira:
Updating bunit is also fun. Perhaps it's better to focus on bunit on a separate task, and use this task just to focus on vulnerabilities in the actual product?

stsrki:
> Updating bunit is always fun. Perhaps it's better to focus on bunit on a separate task, and use this task just to focus on vulnerabilities in the actual product?

I wanted to make sure that everything works before emerging the PR. But yeah, I can do it in another task.

David-Moreira:
@stsrki I mean, do not update any tests dependencies (bunit specially... ), and focus only on vulnerable dependencies of the actual product. 
We can handle test dependencies and test updating in a separate PR.

PR #5821, opened 2024-11-05
Table: add the capability to recalculate the fixed position for dynamically added cells

David-Moreira:


David-Moreira:
@stsrki How does this solution look to you? I kinda was on the fence whether it would be more appropriate to just do this with JS though.

Source/Blazorise/Components/Table/TableRow.razor.cs
```diff
@@ -84,19 +103,159 @@ protected Task OnMouseOverHandler( MouseEventArgs eventArgs )
...
+    }
+
+    internal void RemoveTableRowCell( TableRowCell tableRowCell )
+    {
+        tableRowCells?.Remove( tableRowCell );
+    }
+
+    private void RecalculateHashCode()
+    {
+        elementHashCode = tableHeaderCells.GetListHash() + tableRowCells.GetListHash() + tableRowHeaders.GetListHash();
```
stsrki, line 162: In theory the `elementHashCode` might have duplicate hash. I think in that case, Blazor might cause an exception on the `@key` duplicate value.

Source/Blazorise/Components/Table/TableRowHeader.razor.cs
```diff
@@ -42,14 +43,29 @@ protected override void BuildClasses( ClassBuilder builder )
...
         {
-            builder.Append( $"left:{fixedPositionStartOffset:G29}px" );
+            if ( ParentTable.FixedColumnsSync )
+            {
+                var startOffset = ParentTableRow.GetTableRowHeaderFixedPositionStartOffset( this );
+                builder.Append( $"left:{startOffset:G29}px" );
+            }
+            else if ( fixedPositionStartOffset.HasValue )
+            {
+                builder.Append( $"left:{fixedPositionStartOffset:G29}px" );
```
stsrki, line 55: We need to consider having a "number to string" utility extension method that converts number to proper string value.

stsrki (commented):
The code seems fine, although it might get even more complicated with time. I'm tempted to go the JS route but not sure how that would impact performance if we need to have many JS interop calls. Or only one would be needed...

Source/Blazorise/Components/Table/TableRow.razor.cs
David-Moreira: Any suggestion?

Source/Blazorise/Components/Table/TableRowHeader.razor.cs
David-Moreira: ANy ideas for naming? **FormatAs**?

Source/Blazorise/Components/Table/TableRow.razor.cs
stsrki: `Guid.NewGuid()`? 😅

Source/Blazorise/Components/Table/TableRowHeader.razor.cs
stsrki: - ToPreciseString
- ToG29String

David-Moreira: What's a "Precise" string :D

Source/Blazorise/Components/Table/TableRow.razor.cs
David-Moreira: hmm, or maybe append the elementId in that case?

Source/Blazorise/Components/Table/TableRowHeader.razor.cs
stsrki: The G29 "compacts" the number without rounding it. So I think precise makes sense.

Source/Blazorise/Components/Table/TableRow.razor.cs
stsrki: Yeah, that might work.

Source/Blazorise/Components/Table/TableRowHeader.razor.cs
David-Moreira: Ahh. Yea I like that descriptive name better. ToG29String would make no difference at all, it would still be obscure.

stsrki:
We'll stick with the C# code, just FYI.

stsrki (approved):

PR #5827, opened 2024-11-07
Refactor: add strongly typed serialization objects

### LINKED ISSUE #5701: Make a serialization object for JS interops
> > > Make a serialization object with the first letter in lowercase. This ensures that we get appropriately named JSON objects in case users have different JSON serialization settings. Unlikely but you never know.
> > 
> > 
> > You touch on an important point here. We should not rely on anything external or conventions. We should precisely define how our own stuff gets serialized so it can't fail due to external conditions. Is this not possible?
> 
> Well, we could introduce internal serialization class objects that make use of `JsonPropertyNameAttribute` to strongly type the name of the field. That way, it would be 100% right.

Yes. If we care about it. That's a way to go I guess, not doing guess work and hoping nothing external happens in my opinion.

_Originally posted by @David-Moreira in https://github.com/Megabit/Blazorise/issues/5691#issuecomment-2294976305_

tesar-tech:
## Description

Draft for strongly-typed options, currently focused on the button and color picker components. I need your feedback to continue further. 

Closes #5701

### Key Points
- **Type Changes**: Updated from `object` to `TheComponentJsOptions` in `JsModule` and its interface, as well as when calling methods in the module (typically within `TheComponent.razor.cs`).
- **Using `record` Types**: Given that we're targeting .NET 6, the `required` keyword isn’t available. Using `record` types is a suitable alternative, as it allows for clean initialization without needing `SomeProperty = SomeProperty` declarations. This makes both the definition and initialization more concise. It can be of course done the "old way", but I don't see any advantage there.  
- **File Organization**: The `JsOptions` files are organized within a `JsOptions` folder. I followed the codebase conventions, placing everything within the `Blazorise` namespace.

### Questions
- **Interfaces for `JsModules`**: I’m unclear on the purpose of having interfaces for `JsModules`, especially since each one seems to have only one implementation without further usage. It feels like an extra step when working with these modules. (Is there a particular rationale here?)

### Future Improvements
- **Unified `JsOption` Structure in TypeScript**: It would be beneficial to have a consistent `JsOption` DTO structure on the js side. This would involve introducing ts, which could greatly enhance readability and error-checking. With source generators, we could automate much of this without requiring manual duplication of `JsOption` on both sides. This transition could be done gradually, component by component, for a smooth adoption.


Source/Blazorise/JsOptions/ColorPickerJsOptions.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System.Collections.Generic;
+
+namespace Blazorise;
+
+public record ColorPickerJsOptions(
```
stsrki, line 5: I would prefer to not use primary constructors. IMO they don't provide much pros, and now we need to be careful about the order of providing arguments when initializing the object. 

So, use regular properties and assign them as before. `PropName = SomeName1`.

Source/Blazorise/JsOptions/ColorPickerJsOptions.cs
```diff
@@ -0,0 +1,33 @@
...
+    bool ShowHueSlider,
+    bool ShowInputField,
+    bool Disabled,
+    bool ReadOnly,
+    IReadOnlyDictionary<string, string> Localization,
+    string ColorPreviewElementSelector,
+    string ColorValueElementSelector
+    );
+
+public record OptionChange(bool Changed, bool Value);
```
stsrki, line 22: Maybe rename this to 

- `JsOptionChange`
- `JsOptionUpdate`
- something else?

Source/Blazorise/JsOptions/ColorPickerJsOptions.cs
```diff
@@ -0,0 +1,33 @@
...
+public record PaletteChange(bool Changed, string[] Value);
```
stsrki, line 23: Maybe `OptionChange` can be a generic object so that we don't need to have new this `PaletteChange` or similar? Just `OptionChange<string>`, `OptionChange<string[]>`, etc.

Source/Blazorise/JsOptions/ColorPickerJsOptions.cs
David-Moreira: **Aside Note:** I think you meant to say that you prefer not to use constructors at all **for these enhancements** in particular. They do provide **VERY VERY valuable pros**. But you prefer regular assignment here for readability.

stsrki: Yes. At least not for large option objects. The smaller OptionChange is OK.

David-Moreira: Ok, just stating it clearly, I know we are in similar positions when it comes to coding practices, but someone else reading your comment about primary constructors might take it the wrong way.

Source/Blazorise/JsOptions/ColorPickerJsOptions.cs
tesar-tech: make sense.

Source/Blazorise/JsOptions/ColorPickerJsOptions.cs
tesar-tech: Yes!

Source/Blazorise/JsOptions/ColorPickerJsOptions.cs
tesar-tech: I agree with the argument of readability and order changes. The named arguments, as @David-Moreira pointed out, solve this nicely. The advantage over the property assignment is the compiler check. If you add another property to the `ColorPickerJsOptions`, it will scream at you to initialize it too...

By the way, these aren’t primary constructors but rather positional records. The key difference is in how the value is represented: in records, it's a property, while in primary constructors, it's simply a private field (which is ignored by json serializer).

stsrki: One could argue that the named arguments would help for readability, but then we basically have the same thing in the end, just a little bit different syntax

`Propertry1: Value1` vs `Argument1 = Value1`.

David-Moreira: I don't understand your point. I actually prefer this way if readability was the concern.

Source/Blazorise/JsOptions/ButtonJsOptions.cs
stsrki, line 1: I would move new js options unded the `./Models/JSOptions` instead of having then on the root level.

or

Under the `./Modules/JSOptions` to keep the close to where they are actually going to be used.

---

Also. `ButtonJSOptions` might be better than `ButtonJsOptions` for naming. Blazor also follow the same guidelines for short abbreviations. See `IJSObjectReference`, `IJSRuntime`, `IJSInProcessRuntime`, etc.

Source/Blazorise/JsOptions/ColorPickerJsOptions.cs
David-Moreira: I personally don't care about using the primary constructor or not, it could be a regular constructor, BUT I prefer the constructor over the loose property assignment, you know... 
1. enforcement of arguments
2. encapsulation / proper access modifiers
3. opportunity to do special centralized init logic on ctor

David-Moreira: but at the end of the day I prefer consistency, so just choose one way and stick with it in my opinion.

tesar-tech:
I have a question about internal vs public data types in options. @stsrki

Currently, `VideoJSOptions` is internal, which prevents it from being used in a public module method:

```csharp
    public virtual async ValueTask Initialize(DotNetObjectReference<Video> dotNetObjectReference, ElementReference elementRef, string elementId, VideoInitializeJSOptions options)
```

One solution could be to make the `JSModule` methods `internal` as well. Is there any reason not to, or are people actively calling these methods? 

Alternatively, we could make all `*JSOptions` classes public, but placing them in the main Blazorise namespace might not be ideal. Would moving them to a dedicated namespace be a better option?

Second question is about the interfaces for the JS modules, as I already asked in the PR description—do we really need them? For example, the `JSVideoModule` doesn’t have an interface and seems to work just fine.

stsrki:
> I have a question about internal vs public data types in options. @stsrki
> 
> Currently, `VideoJSOptions` is internal, which prevents it from being used in a public module method:
> 
> ```cs
>     public virtual async ValueTask Initialize(DotNetObjectReference<Video> dotNetObjectReference, ElementReference elementRef, string elementId, VideoInitializeJSOptions options)
> ```
> 
> One solution could be to make the `JSModule` methods `internal` as well. Is there any reason not to, or are people actively calling these methods?
> 
> Alternatively, we could make all `*JSOptions` classes public, but placing them in the main Blazorise namespace might not be ideal. Would moving them to a dedicated namespace be a better option?
> 
> Second question is about the interfaces for the JS modules, as I already asked in the PR description—do we really need them? For example, the `JSVideoModule` doesn’t have an interface and seems to work just fine.

It might be that, in this case, VideoJSOptions was made public by accident. Not sure. But generally, we expose it as public in case we, or our users, need to do unit testing on the JS modules. Which, in turn, requires that JS options also be made public.

Source/Extensions/Blazorise.QRCode/QRCodeJSOptions.cs
```diff
@@ -0,0 +1,15 @@
+﻿namespace Blazorise.QRCode;
+
+public class QRCodeInitializeAndUpdateJSOptions
```
stsrki, line 3: Just name it `QRCodeJSOptions`. If there is a need for an update option, then we will introduce them in the future as a new type.

Source/Extensions/Blazorise.SignaturePad/SignaturePadJSOptions.cs
```diff
@@ -0,0 +1,37 @@
+﻿using Blazorise.Modules.JSOptions;
+
+namespace Blazorise.SignaturePad;
+
+public class SignaturePadInitializeJSOptions
```
stsrki, line 5: Remove the `*Initialize*` from the name. All initialized options should be named `ComponentJSOptions`.

Source/Extensions/Blazorise.Video/VideoJSOptions.cs
```diff
@@ -2,82 +2,84 @@
...
-    [JsonPropertyName( "controlsList" )]
     public string[] ControlsList { get; set; }
 
-    [JsonPropertyName( "settingsList" )]
     public VideoSettingsType[] SettingsList { get; set; }
 
-    [JsonPropertyName( "automaticallyHideControls" )]
     public bool AutomaticallyHideControls { get; set; }
 
-    [JsonPropertyName( "autoplay" )]
```
stsrki, line 22: Be careful about the casing. `autoplay` vs `autoPlay`. Make sure everything matches what is used in JS.

stsrki (commented):
Overall looks good: But there are still few things:

- all public classes and properties must have XML comments written. If you don't already a GH copilot is good with it. And you can also use chatgpt to generate them,
- Please use VS instead of Rider if you can. I see a lot of formating broken. Or if Rider can use .editorconfig then configure it.

stsrki:
> * Please use VS instead of Rider if you can. I see a lot of formating broken. Or if Rider can use .editorconfig then configure it.

PS. https://www.jetbrains.com/help/rider/Using_EditorConfig.html#roslyn

stsrki (approved):

PR #5828, opened 2024-11-08
Update bug_report.yaml

### LINKED ISSUE #5822: GitHub template for bug report
This would simplify the process of creating bug report repositories for users significantly.

I’m suggesting a repository under `Blazorise/BugReportProject`, which would host a basic Blazor project pre-configured with Blazorise. This repository would function as a [GitHub template](https://docs.github.com/en/repositories/creating-and-managing-repositories/creating-a-template-repository). A GitHub template adds a green "Use template" button at the top, allowing users to create a copy similar to a fork, but without linking to the base repository and with the ability to choose a custom name. It’s quite convenient.

Instructions could be included in the repository's README, and we could reference it in the GitHub issue template (https://docs.github.com/en/repositories/creating-and-managing-repositories/creating-a-template-repository).

Additionally, we could set up a GitHub Action to keep the `BugReportProject` updated with the latest version, minimizing maintenance effort.

tesar-tech:
## Description

Closes #5822

- Multiple bug report repositories were created, one for each provider, with a central hub available [here](https://github.com/Blazorise/BugReportTemplate). The automation is managed by a [GitHub Action](https://github.com/Blazorise/BugReportTemplate/blob/main/.github/workflows/build-template.yml).
  
- For each provider, the action uses `Blazorise.Templates` to generate fresh, up-to-date code.
- It pushes the generated code to the corresponding Bug Report repository (e.g., [BugReportBootstrap5](https://github.com/Blazorise/BugReportBootstrap5)). (so the provider repos are "orphaned")
- Additionally, it copies the README, license, and `.gitignore` files to each repository.
- The README provides basic instructions on how to use the template.
- The action runs daily to:
  - Ensure unique names for each repository (with MMdd suffix), helping us avoid duplicate .NET project names.
  - Keep all repositories up-to-date by using the latest `-f` flag and installing the newest `Blazorise.Templates` version.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yaml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yaml
@@ -36,22 +36,29 @@ body:
         - Bulma
         - Material
         - Tailwind
-  - type: textarea
-    id: reproduction-link
-    attributes:
-      label: Link to minimal reproduction or a simple code snippet
-      description: |
-        The easiest way to provide a reproduction is to provide a link to a public GitHub repository or a tool like [Telerik Repl](https://blazorrepl.telerik.com/).
+- type: markdown
+  id: reproduction-link
+  attributes:
+    label: Link to minimal reproduction or a simple code snippet
+    description: |
+      Provide a public GitHub repository (or code snippet). For easy setup, you can use the Blazorise Bug Report Template for each provider: 
+      [Bootstrap4](https://github.com/Blazorise/BugReportBootstrap4),
+      [Bootstrap5](https://github.com/Blazorise/BugReportBootstrap5), 
+      [Tailwind](https://github.com/Blazorise/BugReportTailwind),
+      [Material](https://github.com/Blazorise/BugReportMaterial),
+      [AntDesign](https://github.com/Blazorise/BugReportAntDesign),
+      [Bulma](https://github.com/Blazorise/BugReportBulma), or
+      [FluentUI2](https://github.com/Blazorise/BugReportFluentUI2).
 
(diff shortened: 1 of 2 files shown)


David-Moreira (approved):
A good addition might be to generate these repositories with their GH action that automatically runs the dotnet build pipeline. This will also have the additional benefit that we make sure that the generated templates build successfully.

Other then the suggestion, LGTM.

tesar-tech:
I am not sure I understand the suggestion. 

- the repositories are generated - not in their own action, but in the `BugReportTemplate` action. What would be the benefit to do it in their own gh actions?
- I can also add the `dotnet build` into the `BugReportTemplate` action. But I don't find it necessary - I somehow trust the `Blazorise.Template` will produce "buildable" repo. The right place to test this is on the `Blazorise.Template` side (but I understand the reality and this is really one line of code to add).

stsrki:
Might be good to also update label action and mention that they can use new project template for reproducible.

https://github.com/Megabit/Blazorise/blob/master/.github/label-actions.yml

stsrki (approved):

David-Moreira:
> I am not sure I understand the suggestion.
> 
> * the repositories are generated - not in their own action, but in the `BugReportTemplate` action. What would be the benefit to do it in their own gh actions?
> * I can also add the `dotnet build` into the `BugReportTemplate` action. But I don't find it necessary - I somehow trust the `Blazorise.Template` will produce "buildable" repo. The right place to test this is on the `Blazorise.Template` side (but I understand the reality and this is really one line of code to add).

No I meant the second option.
It was just a double check, no harm I guess. But you're right about the right place to making sure **Blazorise.Template** is working correctly is on its own repository.

PR #5834, opened 2024-11-11
Dropzone: Reordered property to expose current order

### LINKED ISSUE #5830: Expose internal order of DropZone items when AllowReorder is enabled
** Problem **
Droppable supports reordering of items within a zone. However, the internal order is kept only in internal indicies Dictionary.

** Solution **
Please make the internal order accessible from outside the component. With this change, reordering of the items in zone makes sense for application s using the reordering.

tesar-tech:
[Closes #5830](https://github.com/Megabit/Blazorise/issues/5830)

https://github.com/user-attachments/assets/a91132c5-6438-446f-a4a1-eeb71b3f94cf

This function triggers every time the order changes:

- when an item is dragged to a new position
- when an item is added from a different zone
- when an item is removed

While removing and adding items isn’t technically reordering, I believe these actions should also trigger the function.
 
Currently, it also fires when a user drags an item and drops it back in the same position. While this doesn’t alter the collection, the function still invokes. We could compare the old and new dictionary to prevent this, but is it worth the additional complexity?

### DIFF
--- a/Source/Blazorise/Components/Droppable/DropZone.razor.cs
+++ b/Source/Blazorise/Components/Droppable/DropZone.razor.cs
@@ -175,8 +175,13 @@ public partial class DropZone<TItem> : BaseComponent, IAsyncDisposable
                 {
                     indices[item.Key] = index++;
                 }
+                if ( AllowReorder )
+                {
+                    Reordered?.Invoke( indices );
+                }
             }
         }
+
 
         recalculateItems = true;
         shouldRerender = true;
@@ -300,8 +305,9 @@ public partial class DropZone<TItem> : BaseComponent, IAsyncDisposable
                     }
 
                     indices.Add( context, newIndex );
-                }
+                } 
             }
+            
         }
         else
         {
@@ -504,6 +510,11 @@ public partial class DropZone<TItem> : BaseComponent, IAsyncDisposable
     /// If true, the reordering of the items will be enabled.
     /// </summary>
     [Parameter] public bool AllowReorder { get; set; }
+    
+    /// <summary>
+    /// The callback that is raised when the order of items changed. Only if <see cref="AllowReorder"/> is enabled.
+    /// </summary>
+    [Parameter] public Func<Dictionary<TItem, int>, Task> Reordered { get; set; }
 
     /// <summary>
     /// If true, will only act as a dropable zone and not render any items.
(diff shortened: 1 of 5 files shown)


mkontula:
@tesar-tech From the my perspective adding and removing items for example from zone "all options" to "selected and ordered options" is *a* reordering since you can drop the new item in any position in "selected and ordered options". Also when you remove an item from "selected and ordered options", you would expect the "selected and ordered options" internal and exposed order to be coherent and without gaps.

Source/Blazorise/Components/Droppable/DropZone.razor.cs
```diff
@@ -504,6 +510,11 @@ protected override bool ShouldRender()
     /// If true, the reordering of the items will be enabled.
     /// </summary>
     [Parameter] public bool AllowReorder { get; set; }
+    
+    /// <summary>
+    /// The callback that is raised when the order of items changed. Only if <see cref="AllowReorder"/> is enabled.
+    /// </summary>
+    [Parameter] public Func<Dictionary<TItem, int>, Task> Reordered { get; set; }
```
stsrki, line 517: Now that I think of it might be better to have a "context" object instead of `Dictionary<TItem, int>`.

```cs
class DropDownOrderContext<TItem>
{
  public TItem Item {get;}
  public int Order {get;} // or Index
}
```

This way the API is more flexible for any change in the future.

tesar-tech: And then passing as a List?

What about 

```csharp
public class DropzoneOrderContext<TItem>
{
   
    public Dictionary<TItem, int> Indeces{ get; set; }

}
```

or even:

```csharp
public class DropzoneOrderContext<TItem>
{

    public List<DropzoneOrderedItem> OrderedItems { get; set; }
   //something else when needed.

}

public class DropzoneOrderedItem<TItem>
{
 public TItem Item {get;}
  public int Order {get;} 
}
```

stsrki: I like the second example.

stsrki: PS. Just match the naming accordingly to component name. `DropZoneOrderContext`, and `DropZoneOrderedItem`.

stsrki:
Got some more ideas and added dropzone names to the DropZoneOrder as I think it will be useful to the users. Please review.

PR #5838, opened 2024-11-12
DataGrid : Workaround for DatePicker being set to null no matter what when filter is set through Parameter FilterSearchValue

David-Moreira:
Apparently the bug was introduced with my code. The introduction of the parameter FilterSearchValue on _DataGridCellFilter. Such a seemingly simple change that produces a very weird behaviour where DatePicker keeps reseting the value back to null.

Only way I could figure out how to fix it, was to just go back to a similar version to the old one, and add the code that I needed the Filter.Between to work.... Not pretty, but it works...


Source/Extensions/Blazorise.DataGrid/Enums/DataGridCellFilterType.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace Blazorise.DataGrid;
+
+/// <summary>
+/// Defines the DataGrid cell filter type.
+/// </summary>
+public enum DataGridCellFilterType
+{
+    Single,
+    Range1,
```
stsrki, line 9: Range 1 and 2 are quite generic. Maybe `RangeFrom` and `RangeTo` might be better?

David-Moreira: Seems like a good idea. Only thing is that if we potentially had multiple values in the future, like 1,2,3,4... but I don't think so. So I'll go with your suggestion.

stsrki (approved):

PR #5841, opened 2024-11-13
DataGridCellDatePicker: parse valid date format

### LINKED ISSUE #5837: [Bug]: DataGrid : DatePicker Column isn't compatible with DisplayFormat
@stsrki 

I Remember now why we had a separate DateDisplayFormat.
The DisplayFormat isn't really compatible with the DatePicker format style. It seems to use a **string.format**

DisplayFormat="{0:dd.MM.yyyy}"
**UI**
Image
**DatePicker**
Image

DisplayFormat="dd.MM.yyyy"

**UI**
Image

**DatePicker**
Image

What do you think, should we revert back?

stsrki:
Closes #5837

I have chosen to not break any users code for now, and just parse valid date format if it is provided as before.

In 2.0 we can plan to migrate to a new format without the usage of `"{0:*"`.


Source/Extensions/Blazorise.DataGrid/_DataGridCellDatePicker.razor.cs
```diff
@@ -13,18 +14,34 @@ namespace Blazorise.DataGrid;
...
     protected override void OnInitialized()
     {
         valueType = Column.GetValueType( default );
         elementId = IdGenerator.Generate;
+
+        // this is a woraround for https://github.com/Megabit/Blazorise/issues/5837
+        // in 2.0 we need to remove formating as {0:dd.MM.yyyy}, and on support dd.MM.yyyy
+        if ( Column.DisplayFormat is not null && Column.DisplayFormat.StartsWith( "{0:" ) && Column.DisplayFormat.EndsWith( "}" ) )
+        {
+            dateDisplayFormat = Column.DisplayFormat.Substring( 3, Column.DisplayFormat.Length - 4 );
```
David-Moreira, line 42: `dateDisplayFormat = Column.DisplayFormat.Substring( 3, Column.DisplayFormat.Length - 4 );`

Shouldn't it be **Column.DisplayFormat.Length - 1**? Or Am I missing something?

Source/Extensions/Blazorise.DataGrid/_DataGridCellDatePicker.razor.cs
```diff
@@ -13,18 +14,34 @@ namespace Blazorise.DataGrid;
...
+
     /// <inheritdoc/>
     protected override void OnInitialized()
     {
         valueType = Column.GetValueType( default );
         elementId = IdGenerator.Generate;
+
+        // this is a woraround for https://github.com/Megabit/Blazorise/issues/5837
+        // in 2.0 we need to remove formating as {0:dd.MM.yyyy}, and on support dd.MM.yyyy
+        if ( Column.DisplayFormat is not null && Column.DisplayFormat.StartsWith( "{0:" ) && Column.DisplayFormat.EndsWith( "}" ) )
```
David-Moreira, line 40: I really like `!String.IsNullOrEmpty`, thoughts?

David-Moreira (changes requested):

Source/Extensions/Blazorise.DataGrid/_DataGridCellDatePicker.razor.cs
stsrki: No, Substring starts from index, and then takes the length.

Source/Extensions/Blazorise.DataGrid/_DataGridCellDatePicker.razor.cs
stsrki: I don't think it is necessary here.

Source/Extensions/Blazorise.DataGrid/_DataGridCellDatePicker.razor.cs
stsrki: Yeah range works like that. I always forget about the range :)

PR #5843, opened 2024-11-13
DataGrid: Comparer implementation for custom sorting

### LINKED ISSUE #5839: Custom Sorting for custom compare method
In my project there is a requirement where data is a combination of alphanumeric.
I want to define custom sort function that if it contains numbers at the end then it will need to sort like numeric sorting not string sorting
and if there are no numbers then it will behave like string sorting.

Just like WPF grid Compare(object x, object y) method.

Is there any way possible for this?

tesar-tech:
## Description

Closes #5839

Implements comparer. Usage in example page.

TODOs :
- Add comparer to datagrid api page
- update release notes


stsrki:
This should work. But we need to mention in the docs that it only makes sense for in-memory data. If the user fetch only partial data from the API then this can only sort on that amount of data.

stsrki:
> TODOs :
> 
> * Add comparer to datagrid api page
> * update release notes

Please finish this so I we can merge it.

stsrki (approved):

PR #5846, opened 2024-11-15
Docs: ApiDocsGenerator for API documentation from XML comments

tesar-tech:
## Description

- Introduced a new source generator project, `ApiDocsGenerator`, to automatically extract public properties with the `[Parameter]` attribute from Blazorise components.
- The generator produces a source file (`ComponentsApiDocsSource`) containing a dictionary to retrieve all parameters for a given component.
- **Usage Example:** In `ButtonPage.razor`:

```html
<ComponentApiDocs ComponentType="typeof(Button)" />
```

This single line will generate the API documentation at the bottom of the page, populated with XML comments.

## TODOs

1. **Verify XML Comments:**  
   Ensure the XML comments accurately represent what we want to display publicly. While this is generally true, there are discrepancies—for example, the comments for the `Button` component differ a bit (parameter `Active`)

2. **Default Values Handling:**  
   - The current implementation only displays default values for basic types but not for Blazorise enums or complex types.  
   - Defaults like `[Parameter] public int SomeParam { get; set; } = 5;` are not respected yet.  
   - These cases could likely be addressed within the source generator but how "far"?

3. **Type Differences:**  
   - For instance, `Button.Clicked` is now `EventCallback<MouseEventArgs>` but was previously just `EventCallback`.  
   - Determine the standard format we want to display for such types.

4. **Adding Extensions:**  
   - The same logic can be extended to other projects, but it will need to be configured for each `.csproj` file individually.

5. **Handling Special Cases:**  
   - For example, `DataGrid` includes `DataGridColumn`. Currently, column properties appear in the `DataGrid` API page.  
   - Decide whether this behavior is correct or if such properties should be separated.

6. **Clean up the actual source generator**
  - Currently all that is in one file, no comments, etc.. Will be done, once feature requirements are established.


stsrki:
> 1. **Verify XML Comments:**
>    Ensure the XML comments accurately represent what we want to display publicly. While this is generally true, there are discrepancies—for example, the comments for the `Button` component differ a bit (parameter `Active`)

This is something that we will need to handle from the code side. We need to rewrite comments so that they are properly formated and dramatically correct. Most comments are quite old and are not as good as they could be.

> 2. **Default Values Handling:**
>    
>    * The current implementation only displays default values for basic types but not for Blazorise enums or complex types.
>    * Defaults like `[Parameter] public int SomeParam { get; set; } = 5;` are not respected yet.
>    * These cases could likely be addressed within the source generator but how "far"?

I see. Some defaults are only set in the code further. We have two options. Define default on the property, or mention it in the comment. 

> 3. **Type Differences:**
>    
>    * For instance, `Button.Clicked` is now `EventCallback<MouseEventArgs>` but was previously just `EventCallback`.
>    * Determine the standard format we want to display for such types.

The real type should be correct, so `EventCallback<MouseEventArgs>`.

> 4. **Adding Extensions:**
>    
>    * The same logic can be extended to other projects, but it will need to be configured for each `.csproj` file individually.

Makes sense: Question! Does this new generator need to be referenced as part of the nuget once we publish blazorise?

> 5. **Handling Special Cases:**
>    
>    * For example, `DataGrid` includes `DataGridColumn`. Currently, column properties appear in the `DataGrid` API page.
>    * Decide whether this behavior is correct or if such properties should be separated.

I think whichever `"type of (ComponentName)"` is defined, we need to list it. So maybe `ComponentType` can be an array?

> 6. **Clean up the actual source generator**
> 
> * Currently all that is in one file, no comments, etc.. Will be done, once feature requirements are established.

tesar-tech:
> > 2. **Default Values Handling:**
> >    
> >    * The current implementation only displays default values for basic types but not for Blazorise enums or complex types.
> >    * Defaults like `[Parameter] public int SomeParam { get; set; } = 5;` are not respected yet.
> >    * These cases could likely be addressed within the source generator but how "far"?
> 
> I see. Some defaults are only set in the code further. We have two options. Define default on the property, or mention it in the comment.

Option 1 sounds better - one source of truth. If not doable by property we can agree on some format for handling it from comment. Like: `//__default__ stringOfSomeDefaultValue `, which will be processed and put into Default column (and removed from comment)

> 
> > 4. **Adding Extensions:**
> >    
> >    * The same logic can be extended to other projects, but it will need to be configured for each `.csproj` file individually.
> 
> Makes sense: Question! Does this new generator need to be referenced as part of the nuget once we publish blazorise?

Should be handled by `PrivateAssets="all"` in the csproj -> doesn't get to the nuget package. 

> > 5. **Handling Special Cases:**
> >    
> >    * For example, `DataGrid` includes `DataGridColumn`. Currently, column properties appear in the `DataGrid` API page.
> >    * Decide whether this behavior is correct or if such properties should be separated.
> 
> I think whichever `"type of (ComponentName)"` is defined, we need to list it. So maybe `ComponentType` can be an array?

This gonna be more complicated, Extensions are wild. We don't want just the `[Parameter]`s to be part of the docs. We also need JSOptions in some cases (Charts), etc... Maybe we'll need some customizations for individual Extensions. I would do the Components first and then look into the Extensions.

tesar-tech:
I focused on default values. From what I have checked, they should all work. 

## Default values

```csharp
[Parameter] public string SomeValue { get; set; } = "some string value";
```

It also handles cases where the default value is not a constant.

```csharp
[Parameter] public TimeSpan Interval { get; set; } = TimeSpan.FromSeconds(10);
```

Additionally, it works when the default value is set by a backing field.

```csharp
private string _backingField = "backed value";
[Parameter] public string BackedValue 
{
    get => _backingField;
    set => _backingField = value;
}
```

Moreover, it preserves the exact expression of default values, not just the computed result.

```csharp
[Parameter] public int ComplexValue { get; set; } = 20 * 200; // Output: "20 * 200", not "4000"
```

## Enum values for Blazorise enums..

I added the listing of blazorise enums:

image

the text `Possible values: Button, Submit, Reset, Link` is not part of the comment..

I didn't persuade further, because I figured there is this `Enumeration` record, which is kind of an enum, but it's bit harder to obtain the values (not impossible) from. 


## Docs pages can handle multiple components..

Component and its related components:

```html
<ComponentApiDocs ComponentTypes="[typeof(Accordion), typeof(AccordionToggle), typeof(AccordionItem), typeof(AccordionHeader), typeof(AccordionBody)]" />

```

Will focus on methods now.

stsrki:
I have noticed we will have some edge cases with inherited parameters. For example TextEdit is inherited from `BaseTextInput`, and then from `BaseInputComponent`. We need to show all of those parameters. But, I would not go all the way up to `BaseComponent`.

For `BaseComponent` I think it is good to have a separate API page in the docs and then reference it with a link. Every Blazorise component is inherited from it, and it would be too much to list all of its parameters for each component.

stsrki:
> > I have noticed we will have some edge cases with inherited parameters. For example TextEdit is inherited from `BaseTextInput`, and then from `BaseInputComponent`. We need to show all of those parameters. But, I would not go all the way up to `BaseComponent`.
> 
> What about just bringing it within the list:
> 
> ```
> <ComponentApiDocs ComponentTypes="[typeof(TextEdit), typeof(BaseTextInput)]" />
> ```
> 
> It will be separated from each other... `TextEditPage` has this now. I have added the functionality for generic types (like `BaseTextInput`).
> 
> Or we I can do something like:
> 
> ```
> <ComponentApiDocs JoinedComponentTypes=[ [typeof(TextEdit), typeof(BaseTextInput)] ] />
> ```
> 
> > For `BaseComponent` I think it is good to have a separate API page in the docs and then reference it with a link. Every Blazorise component is inherited from it, and it would be too much to list all of its parameters for each component.
> 
> I agree. We can create the page manually and add the link to the `ComponentApiDocs`
> 
> I just need to add base class to the sg.


Having them separate is not good because user should not care how it is structured internally. They only care that TextEdi has the parameters needed. And TextEdit can only be used as an external API.

The second option with `JoinedComponentTypes` might be better. But I don't like it mainly because we then need to take care of all hierarchy. Better to do it with the SC now that we are committed to it. There are going to be some rules, yes, but in the end it is a lot easier for later for us.

tesar-tech:
> Better to do it with the SC now that we are committed to it. There are going to be some rules, yes, but in the end it is a lot easier for later for us.

Agree.  I can define these "joins" in the SG. "Every component that inherits from `BaseTextInput` lists also its params".

Are there any other such "joins" or "exception" to what we already have?

stsrki:
> > Better to do it with the SC now that we are committed to it. There are going to be some rules, yes, but in the end it is a lot easier for later for us.
> 
> Agree. I can define these "joins" in the SG. "Every component that inherits from `BaseTextInput` lists also its params".
> 
> Are there any other such "joins" or "exception" to what we already have?

I would join everything that goes up to BaseComponent. Everything in BaseComponent should be excluded and optionally documented somewhere else.

stsrki:
> Also you mention some black-listed methods, what are those?

Any public method that is only meant to be used internally. For example BeginAnimation on Modal. Or, SetValue on NumericPicker. As mentioned before, we could remove it by adding the "Should only be used internally" comment. Which lead us to the second problem.

> So question is "How to deal with the inheritance of methods and their comments"?

I can only think that we need to read XML comments file to get this information.

stsrki:
BTW, my original idea was to generate API docs source code as a `*Api.razor` component in Blazorise.Docs project. Not to generate it as a ComponentsApiDocsSource.

By generating it directly into the razor files, we can immediately see what is being generated. And we don't need to run the application to see the changes. Making it easier to review what needs to be omitted by extra rules.

tesar-tech:
> > Also you mention some black-listed methods, what are those?
> 
> Any public method that is only meant to be used internally. For example BeginAnimation on Modal. Or, SetValue on NumericPicker. As mentioned before, we could remove it by adding the "Should only be used internally" comment. Which lead us to the second problem.
> 
> > So question is "How to deal with the inheritance of methods and their comments"?
> 
> I can only think that we need to read XML comments file to get this information.

What about `[NotInApiDocs]` parameter? It's more straightforward than looking for a string.

What about the "duplicates" like `OnParameterSet` ? You want them both in there? Or just the one from the top-parent (and its comment?)

tesar-tech:
> BTW, my original idea was to generate API docs source code as a `*Api.razor` component in Blazorise.Docs project. Not to generate it as a ComponentsApiDocsSource.
> 
> By generating it directly into the razor files, we can immediately see what is being generated. And we don't need to run the application to see the changes. Making it easier to review what needs to be omitted by extra rules.

You can see the output in IDE under Source Generators (In project that references the SG,, like `Blazorise`)... And you don't even need to build the source project for that.
I have setup a file watcher for the generator project and whenever I make a change it runs the `dotnet build` in root of the project and I see the generated file changes in matter of 2 seconds.

tesar-tech:
Now it works with `<inheritdoc` xml comment. Arriving at the `<inheritdoc` comment will scan through the interfaces of containing type and will find the one with the value (and its comment). 


"Should only be used internally" is also implemented. With the comment. Using an attribute would bring need for more complex solution. The current one is fairly simple.

Source/Extensions/Blazorise.Snackbar/Blazorise.Snackbar.csproj
```diff
@@ -13,6 +13,8 @@
 
   <ItemGroup>
     <ProjectReference Include="..\..\Blazorise\Blazorise.csproj" />
+    <ProjectReference Include="..\..\SourceGenerators\Blazorise.ApiDocsGenerator\Blazorise.ApiDocsGenerator.csproj" OutputItemType="Analyzer" ReferenceOutputAssembly="false" PrivateAssets="all" />
```
stsrki, line 16: I feel like adding ApiDocsGenerator to every project will bite us in the ***. I remember we had problems(and still have) with `Blazorise.Generator.Features.csproj`. Even though it is marked as `PrivateAssets="all,"`, we still had to publish it as a separate Nuget package because it wouldn't work without it.

stsrki (commented):
The API docs looks great. It even shows some params that we previously forgot to list.

Source/Extensions/Blazorise.Snackbar/Blazorise.Snackbar.csproj
tesar-tech: I hit exactly this issue yesterday when trying to figure out the `[ShouldBeOnlyUseInternally]` attribute. Creating such an attribute in a project referenced with `PrivateAssets="all"` and using it in the "parent" project causes problems. While `PrivateAssets="all"` ensures that the `SourceGeneratorProject.dll` isn't included in the NuGet package, it doesn't prevent the code from `SourceGeneratorProject` from being used in the source code. This ultimately results in the NuGet package not functioning as expected.

In this case, the `ApiDocsGenerator` project (like `ApiDocsGeneratorDtos`) isn't used in the projects from which we are trying to obtain documentation. For instance, there is no `using ApiDocsGenerator;` directive in the `Blazorise` or `Blazorise.AnExtension` projects. This contrasts with `Blazorise.Generator.Features`, which behaves differently. Therefore, I think it’s safe to proceed in this scenario.

By the way, I can think of a solution where using the attribute wouldn't require packing the source generator project either.

stsrki: Can we use Fody to remove any traces of the code that is needed for the generator?

tesar-tech:
I am gradually migrating components to the new API documentation.

- **Support for `<c>` tag in XML comments**:  
  Added support for the `<c>` tag in XML comments, which will now generate `<code>` in HTML. This replaces the use of `<Code>` for such cases.

- **Card component props discrepancy**:  
  Card components seem to inherit many props from the `BaseComponent`, as mentioned in the API docs. However, several items are missing. It’s unclear if this is intentional or a mistake. @stsrki, please double-check this.

- **`Check` component and `Checked` parameter documentation issue**:  
  The docs currently state:
  ```
  Data type of <Code>Checked</Code> value. Supported types are <Code>bool</Code> and <Code>bool?</Code>.
  ```
  However, this information is not part of the XML comments, since the `Checked` parameter comes from `BaseCheckComponent`. 

  A clean solution might be to add details about the type parameter directly in the XML comment of the base class. For example:
  ```csharp
  /// <summary>
  /// Checkboxes allow the user to select one or more items from a set.
  /// </summary>
  /// <typeparam name="TValue">Checked value type. Supported types are bool or bool?</typeparam>
  public partial class Check<TValue> : BaseCheckComponent<TValue>
  ```

  **Options to address this**:
  - Expand documentation sections to include **Type Parameters**, Parameters and Methods (rather than just Parameters and Methods).
  - Keep the current structure and document this information elsewhere.

- **Support for `<remarks>`**:  
  Added support for the `<remarks>` XML tag. For example, the `Checked.Name` property has the following XML comment:
  ```csharp
  /// <summary>
  /// Defines the name attribute of a checkbox.
  /// </summary>
  /// <remarks>
  /// The name attribute is used to identify form data after it has been submitted to the server, or to reference form data using JavaScript on the client side.
  /// </remarks>
  [Parameter] public string Name { get; set; }
  ```
  This generates a summary-detail combo in the parameter description, as shown here:  

  https://github.com/user-attachments/assets/53e23193-949f-40f3-ac4c-73654f75e5eb

  While the UI could be improved, the approach allows hiding less critical information initially.

- **`Divider.DividerType` default value update**:  
  Previously, `Solid` was the default value for `DividerType`. It now defaults to `null`, which reflects the technical reality since the actual default is determined by the `Theme`. `Solid` serves as a fallback value.  
  It’s unclear to me how problematic this change might be.

stsrki:
> I am gradually migrating components to the new API documentation.
> 
> * **Support for `<c>` tag in XML comments**:
>   Added support for the `<c>` tag in XML comments, which will now generate `<code>` in HTML. This replaces the use of `<Code>` for such cases.

All good. Just use our own components when generating the razor. So `<Code>`, not `<code>` in this case.

> * **Card component props discrepancy**:
>   Card components seem to inherit many props from the `BaseComponent`, as mentioned in the API docs. However, several items are missing. It’s unclear if this is intentional or a mistake. @stsrki, please double-check this.

This is fine. We will handle BaseComponent API separately anyways.

> * **`Check` component and `Checked` parameter documentation issue**:
>   The docs currently state:
>   ```
>   Data type of <Code>Checked</Code> value. Supported types are <Code>bool</Code> and <Code>bool?</Code>.
>   ```
>   
>   
>       
>         
>       
>   
>         
>       
>   
>       
>     
>   However, this information is not part of the XML comments, since the `Checked` parameter comes from `BaseCheckComponent`.
>   A clean solution might be to add details about the type parameter directly in the XML comment of the base class. For example:
>   ```cs
>   /// <summary>
>   /// Checkboxes allow the user to select one or more items from a set.
>   /// </summary>
>   /// <typeparam name="TValue">Checked value type. Supported types are bool or bool?</typeparam>
>   public partial class Check<TValue> : BaseCheckComponent<TValue>
>   ```
>   
>   
>       
>         
>       
>   
>         
>       
>   
>       
>     
>   **Options to address this**:
>   
>   * Expand documentation sections to include **Type Parameters**, Parameters and Methods (rather than just Parameters and Methods).
>   * Keep the current structure and document this information elsewhere.

From now on, the XML comments will be the only source of truth. So whenever we notice that something is lacking we need to make it better. One of the tasks for the future is to improve existing comments since they are not used in documentation. But focus on finishing the generator first.

> * **Support for `<remarks>`**:
>   Added support for the `<remarks>` XML tag. For example, the `Checked.Name` property has the following XML comment:
>   ```cs
>   /// <summary>
>   /// Defines the name attribute of a checkbox.
>   /// </summary>
>   /// <remarks>
>   /// The name attribute is used to identify form data after it has been submitted to the server, or to reference form data using JavaScript on the client side.
>   /// </remarks>
>   [Parameter] public string Name { get; set; }
>   ```
>   
>   
>       
>         
>       
>   
>         
>       
>   
>       
>     
>   This generates a summary-detail combo in the parameter description, as shown here:
>   
>     
>       
>       
>   
>       543c3890-9ad6-48a8-9992-31d24c213122.mp4
>       
>     
>   
>     
>   
>     
>   
>   
>   While the UI could be improved, the approach allows hiding less critical information initially.

I really like these remarks. It looks nice. 

> * **`Divider.DividerType` default value update**:
>   Previously, `Solid` was the default value for `DividerType`. It now defaults to `null`, which reflects the technical reality since the actual default is determined by the `Theme`. `Solid` serves as a fallback value.
>   It’s unclear to me how problematic this change might be.

As mentioned before. Whatever is on the code, or in the XML comments is the true.

tesar-tech:
> From now on, the XML comments will be the only source of truth. So whenever we notice that something is lacking we need to make it better. One of the tasks for the future is to improve existing comments since they are not used in documentation. But focus on finishing the generator first.

The problem here is that the current state of the generator doesn't work with the xml comments of  classes (components)(just with properties and methods). Question is: Should I make also these xml comments available (including the type parameters)?

stsrki:
> > From now on, the XML comments will be the only source of truth. So whenever we notice that something is lacking we need to make it better. One of the tasks for the future is to improve existing comments since they are not used in documentation. But focus on finishing the generator first.
> 
> The problem here is that the current state of the generator doesn't work with the xml comments of classes (components)(just with properties and methods). Question is: Should I make also these xml comments available (including the type parameters)?

No, we don't need the comments of the classes.

stsrki:
image

On TextEdit the `strong` tag is not working. 

PS. Use Blazorises Strong when you fix it.

tesar-tech:
> On TextEdit the `strong` tag is not working.


Fixed. This `<strong>` (and `<code>`) is done as part of the string, which comes to life with `MarkupString`...


> PS. Use Blazorises Strong when you fix it.

...thus using the Blazorise components isn't quite possible in this case. Well - you can use `@((MarkupString)"<Strong>Something</Strong>")`, but it's just converted to lowercase, ignoring the presence of the component. It would be somehow possible with dynamic components or some sort of string parsing or breaking it into multiple pieces from the very beginning, but I would say it's not worth the effort.

stsrki:
Can you make it so that when the `ChildContent` is defined on the "Base" or inherited class, the name of the commented class is then taken from the used component? In this case `Field` component

image

stsrki:
Recently, on the Video component, we have separated events and callbacks from the properties in the docs. See https://blazorise.com/docs/extensions/video

We can also think of making the same for all auto-generated component APIs. Anything that is `EventCallBack,` `Func,` `Action,` or another delegate type needs to have a separate `Events` section.

tesar-tech:
> Can you make it so that when the `ChildContent` is defined on the "Base" or inherited class, the name of the commented class is then taken from the used component? In this case `Field` component

Done! A rare moment when something turns out to be easier than expected.

> We can also think of making the same for all auto-generated component APIs. Anything that is `EventCallBack,` `Func,` `Action,` or another delegate type needs to have a separate `Events` section.

Done. Needed some refactoring, but works great.

stsrki (commented):
I think feature wise we're done. Just finish on replacing existing APIs in the docs for final review.

tesar-tech:
## Done:
- All the Components (with some exceptions, see below)
- Support for links (we should use `<seealso` )
- Support for `<para>`
- Basic check/edit on missing/wrong texts.

## Notes:
- From my "investigation" this automatic api docs adds about 40 % of content
- Typography had the common properties under "All typography components", now it's divided into individual components.
- `Theming` is a folder inside `Pages/Docs/Components`, which looks bit off.

## Questions/todos:
- `Repeater`, `Validations`, `Validation` doesn't inherit from `BaseComponent`, thus they don't have the generated api docs (for now). We can whitelist them or work work with some attribute or empty interface (`IBaseComponent` ? - that would be used in `BaseComponent` and these 3 places. Clean approach imo - and it would solve the "Include BaseComponent" task (see below))
- Finishing converting api for extensions
- Include BaseComponent -> we want the base component page and link to that page from every(?) component page.
- remove debug stuff, cleanup, add comments..

Source/Extensions/Blazorise.Snackbar/Blazorise.Snackbar.csproj
stsrki: What about Fody? Since we're merging into 1.7 we should make sure that we don't have to release additional NuGet package. We need to strip everything not needed for Blazorise to work in production.

Source/Blazorise/Components/Breadcrumb/BreadcrumbItem.razor.cs
```diff
@@ -97,7 +97,7 @@ internal void NotifyRelativeUriChanged( string relativeUri )
     #region Properties
 
     /// <summary>
-    /// Gets or sets the item active state.
+    /// If set to true, renders <c>span</c> instead of <c>a</c> element.
```
stsrki, line 100: This comment is incorrect. In the active state, it only means the item or a link will have some active class and can look unclickable.

tesar-tech: I copied that from the current docs.

tesar-tech:
> I'm not sure if I understand. We already explicitly define `typeof` of each component that goes into generator. Why we need to inherit for some general type then?

For component to be processed by generator it has to inherit from `BaseComponent`. The `<ComponentApiDocs ComponentTypes="[typeof(Repeater<>)]" />` only picks up what generator generated (if present).

stsrki:
Can we show just the field name here?

image

The enum type name is redundant since we already have it on the `Type` column.

stsrki:
Something is not right with the Snackbar generator

image

stsrki:
ILSpy shows that we still use these projects

image

Try to make use of Fody to remove them.

stsrki:
- For `Snackbar.Constants.DefaultAnimationDuration` 5000 is fine.
- for things like `long.MaxValue`, `int.MaxValue` we can have a dictionary of constants and names that we can use. For example if we read `9223372036854775807` we use string `long.MaxValue`. For `2147483647` we write `int.MaxValue`, etc. They are universal constants so it's not a problem.

stsrki:
I have any idea for a simplification of generators project structure. 

- Remove `Blazorise.ApiDocsDtos.csproj` and move all its shared classes into `Blazorise.Generator.Features.csproj`.
- Move generators from `Blazorise.ApiDocsGenerator.csproj` into `Blazorise.Generator.csproj`

That way we will not introduce any new project/package reference. What do you think?

tesar-tech:
> I have an idea for a simplification of generators project structure.
> 
> * Remove `Blazorise.ApiDocsDtos.csproj` and move all its shared classes into `Blazorise.Generator.Features.csproj`.
> * Move generators from `Blazorise.ApiDocsGenerator.csproj` into `Blazorise.Generator.csproj`
> 
> That way we will not introduce any new project/package reference. What do you think?

Yes, sounds good. That should work since these two "couples" serve similar purpose..

I will figure that out once I solve the presence inside nuget package..

stsrki:
> I will figure that out once I solve the presence inside nuget package.

We can figure that out later in another PR if you think it will take you too long. The existing project is already part of NuGet so it is not going to be a problem.

tesar-tech:
In the last commit there is my proposal how to not include the api docs project into Blazorise. I am not sure if there is any other `pack` command or other way to build packages, that would also need to include the `IsPack` parameter.
This is the simplest and most straightforward solution. Unfortunately I didn't find a way how to detect the project is being build for nuget packages, thus we need to use custom param...

Exploring this I realized key differences between our SG projects and why it might not be a good idea (if even feasible) to merge them into one...

## `Blazorise.SourceGenerators` and `SourceGenerators.Features`
- used to "enhance" the Blazorise library, the generated output should be part of the library - always.
- uses `Features` to signal the sg using attributes. These attributes aren't necessary to have during runtime, but they are necessary during build (every time).
- Issue: Attributes are part of the Blazorise package, thus it has to reference the `Features` project.
  - Solution: Use Fody to strip the usage and reference to `Features` project

(it might be obvious to you, but I wanted to make it clear comparation with:)

## `Blazorise.ApiDocsGenerator` and `Blazorise.ApiDocsDtos`
- Used to gather information from Blazorise library. The generated output should be part of the lib only when used by `Blazorise.Docs`, should not be part of the nuget package.
- Uses `ApiDocsDtos` in generated output, that's why it's referenced in nuget package.  
- Issue: `ApiDocsDtos` should not be part of nuget (but should be present with `Blazorise.Docs` )
  - Solution: Simply not include the `PackageReference` while being packed.


These are 2 different problems and 2 different solutions, that's why I think these sg projects should be separated... Unless I miss something obvious...

Also I made the Fody work - at least the initial setup. I agree it should be in separate pr

stsrki:
I'm aware that `Blazorise.ApiDocsDtos` and `SourceGenerators.Features` have a few differences. But, they are still both used as a shared project for the generator. So I would still merge them.

The next step would be to strip all attributes and dtos from Blazorise that were used by the source generators shared project(s). I'm still not sure if not-packing them would work. As in your last commit. I would need to first publish the NuGet package and use it to see if there are any errors when using it.

If we can use Fody, or something that is already present with the build tools then it would be great. We need to be able strip it on each build I think. And use ILSpy as a reference that is striping properly.

tesar-tech:
The "not-packing" works for `Blazorise.ApiDocsDtos` (together with not packing the generator).I  tested that on published package. It wont work on `Features`..

tesar-tech:
- SG projects are merged.
- Fody is integrated. It is able to remove reference and code usage from Blazorise nuget

Everything is documented either by comments or by README file.

NuGet/dotnet-pack.bat
```diff
@@ -31,7 +31,7 @@ dotnet pack ../Source/Extensions/Blazorise.QRCode/Blazorise.QRCode.csproj -o . -
 dotnet pack ../Source/Extensions/Blazorise.RichTextEdit/Blazorise.RichTextEdit.csproj -o . -c Release -p:IncludeSymbols=true -p:SymbolPackageFormat=snupkg
 dotnet pack ../Source/Extensions/Blazorise.Sidebar/Blazorise.Sidebar.csproj -o . -c Release -p:IncludeSymbols=true -p:SymbolPackageFormat=snupkg
 dotnet pack ../Source/Extensions/Blazorise.SignaturePad/Blazorise.SignaturePad.csproj -o . -c Release -p:IncludeSymbols=true -p:SymbolPackageFormat=snupkg
-dotnet pack ../Source/Extensions/Blazorise.Snackbar/Blazorise.Snackbar.csproj -o . -c Release -p:IncludeSymbols=true -p:SymbolPackageFormat=snupkg
+dotnet pack ../Source/Extensions/Blazorise.Snackbar/Blazorise.Snackbar.csproj -o . -c Release -p:IncludeSymbols=true -p:SymbolPackageFormat=snupkg -p:IsPack=true
```
stsrki, line 34: Do we still need IsPack if Fosy is now used?

Source/SourceGenerators/Blazorise.Generator.Features/Blazorise.Generator.Features.csproj
```diff
@@ -16,6 +16,8 @@
     <Copyright>Copyright 2018-2024 Megabit</Copyright>
 
     <TargetFramework>netstandard2.0</TargetFramework>
+
+    <LangVersion>latest</LangVersion>
```
stsrki, line 20: I noticed that you added `<LangVersion>latest</LangVersion>` in a few places. Do we need it?

stsrki (commented):
The Blazorise.Generator.Features still seems to be present as a reference.


image

image

stsrki:
You mentioned in readme

> ### Purpose of `bw` and `bwf`
> 
> - **`Blazorise.Weavers` (bw)**:
>     - Exists solely to create a NuGet package (includes `Fody.Packaging`).
>     - It acts as a wrapper for `bwf`.
> 
> - **`Blazorise.Weavers.Fody` (bwf)**:
>     - The actual implementation of the weaving logic.
>     - This is the "real hero" performing the weaving tasks.

Does this mean we now need to publish a new NuGet package?

NuGet/dotnet-pack.bat
tesar-tech: This question has 2 layers.
- Yes - we need to distinguish between packing and simple build. I updated docs in last commit for explanation.
- I wish there was something like `IsPack` automatically added, so we wouldn't  need our custom `IsPack`, but so far I haven't found it.

tesar-tech:
> You mentioned in readme
> 
> > ### Purpose of `bw` and `bwf`
> > 
> > * **`Blazorise.Weavers` (bw)**:
> >   
> >   * Exists solely to create a NuGet package (includes `Fody.Packaging`).
> >   * It acts as a wrapper for `bwf`.
> > * **`Blazorise.Weavers.Fody` (bwf)**:
> >   
> >   * The actual implementation of the weaving logic.
> >   * This is the "real hero" performing the weaving tasks.
> 
> Does this mean we now need to publish a new NuGet package?

No. It works with the nuget under the hood.. Consuming the weaver as deployed nuget would simplify some stuff, but isn't needed.

Source/SourceGenerators/Blazorise.Generator.Features/Blazorise.Generator.Features.csproj
stsrki: I agree it is convenient. But remember, we support net6 - net9, and for older frameworks, newer C# is not available.

NuGet/dotnet-pack.bat
stsrki: Why do we need to know simple build from publish? We need to remove all SG references at all times.

Source/SourceGenerators/Blazorise.Generator.Features/Blazorise.Generator.Features.csproj
tesar-tech: For some features it's not, but for most of them it is (for example the collections)...  The support is assured by `<TargetFraeworks>` not by version of c#... 
Well - it would not compile, right? So having the c# versions set to `latest` together with target frameworks should be safe... I can use file-scoped namespace and collection expression inside `Weavers.Fody` and it happily compiles for `netstandard2.0`...

stsrki:
Would this work? 

1. We add `[assembly: InternalsVisibleTo( "Blazorise.Docs" )]` into Blazorise.csproj.
2. Generate API docs in Blazorise.Docs, not in Blazorise
3. We don't need Weavers for API docs DTOs

tesar-tech:
> Would this work?
> 
> 1. We add `[assembly: InternalsVisibleTo( "Blazorise.Docs" )]` into Blazorise.csproj.
> 2. Generate API docs in Blazorise.Docs, not in Blazorise
> 3. We don't need Weavers for API docs DTOs

Unfortunately it doesn't.
What we need to access is Blazorise syntax tree. That is only available when project directly references the sg. Otherwise it gets the context already built (and shaved off the comments and syntax).

We really need the syntax tree for 2 things: xml comments and "paths" to default values. xml comments can be solved by exporting and reading the `Blazorise.xml`. Not currently implemented, but seems doable.
For default values it's more complicated. For simple constants we can still obtain it without syntax tree. For more complicated scenarios (for example when the value is "hidden" in a field and property points out to the field in getter) we need the syntax tree. 

 => The SG (at least the api docs part) needs to be referenced in Blazorise and not in Blazorise.Docs just for getting some of the default values. 

We can remove the weavering of ApiDocs by separating the SGs with known tradeoff of having another sg project.

stsrki:
If only tradeoffs were the default values, I could live without them. We can solve the problem of "hidden" value later if needed.

What I don't like about the current approach is that we force Blazorise and every extension project to generate the docs information. We are basically going in the reverse. Because of that, we are forced to do the wavering and all the extra work to clean leftovers.

So, I would like Docs to do their own thing and leave Blazorise and all the extensions intact. Don't do any SG at Docs level. Leave only the previous SG for attributes, and then leave the weaver for those attributes.

stsrki:
edit* fix typos

Documentation/Blazorise.Docs/Components/ComponentApiDocs.razor
```diff
@@ -0,0 +1,48 @@
+﻿@using Blazorise.Generator.Features.ApiDocsDtos
+<DocsPageSubtitle>
+    API
+</DocsPageSubtitle>
+
+@{
+    IReadOnlyList<ApiDocsForComponent> apiDocsForComponents = ComponentTypes.Select(x => ComponentsApiDocsSource.Instance.Components.GetValueOrDefault(x)).Where(x => x is not null).ToList();
+    var hasComponentTypes = ComponentTypes.Any();
+    var multipleComponentTypes = ComponentTypes.Count > 1;
+    var hasMethods = apiDocsForComponents.Any(x => x.Methods.Any());
```
David-Moreira, line 10: Could all these calculations be done on the OnInit method just once?
It would be a little bit more consistent and tidy? We would also be sure that this won't be recalculated needlessly?
Would you agree?

Documentation/Blazorise.Docs/Components/ComponentApiDocs.razor
```diff
@@ -0,0 +1,48 @@
...
+                              MultipleComponentTypes="multipleComponentTypes"
+                              PropertyType="ComponentApiDocsProperty.PropertyTypes.Parameters"/>
+
+
+    <ComponentApiDocsProperty ApiDocsForComponents="apiDocsForComponents"
+                              MultipleComponentTypes="multipleComponentTypes"
+                              PropertyType="ComponentApiDocsProperty.PropertyTypes.Events"/>
+}
+
+@if (hasMethods)
```
David-Moreira, line 25: I'd just do an enumerable with the methods straight away calculated **oninit** and check whether it's not empty and proceed with that single collection, instead of having hasMethod + enumerable.

i.e
@if(!apiDocsForComponentsMethods.IsNullOrEmpty())
...
foreach (var in apiDocsForComponentsMethods...)

Documentation/Blazorise.Docs/Components/ComponentApiDocsProperty.razor
```diff
@@ -0,0 +1,60 @@
...
+
+                @if (property.IsBlazoriseEnum)
+                {
+                    var enumNames = Enum.GetNames(property.Type);
+
+                    <Paragraph Margin="Margin.Is0.FromBottom">
+                        <Text>Possible values:</Text>
+                        @for (int i = 0; i < enumNames.Length; i++)
+                        {
+                            int i1 = i;
```
David-Moreira, line 33: I'd be more descriptive, localI, localIdx, etc... almost never numbers help with describing in my personal opinion

Documentation/Blazorise.Docs/Components/ComponentApiDocsProperty.razor
```diff
@@ -0,0 +1,60 @@
...
+            : "Parameters" )
+    </Heading>
+}
+
+@foreach(var apiDocsForComponent in filteredApiDocsForComponents)
+{
+    <DocsAttributes ShowDefaults="@(PropertyType != PropertyTypes.Events)"
+        Name="@(PropertyType == PropertyTypes.Events ? "Event": "Parameter")"
+        Title="@(MultipleComponentTypes ? apiDocsForComponent.TypeName : null)">
+        @foreach(var property in (PropertyType == PropertyTypes.Events? apiDocsForComponent.Events:apiDocsForComponent.Parameters).OrderBy(x => x.Name))
```
David-Moreira, line 17: Can we get some formatting in here? This almost looked like a null coalesce but it isn't right, it's a ternary operator, hard to read without proper spacing, maybe even replace with a getter propery if it's worth it.

Documentation/Blazorise.Docs/Pages/Docs/Components/Accordions/AccordionPage.razor
```diff
@@ -94,52 +94,4 @@
...
-    </DocsAttributesItem>
-</DocsAttributes>
-
-<DocsAttributes Title="AccordionBody">
-    <DocsAttributesItem Name="ChildContent" Type="RenderFragment" Default="null">
-        Specifies the content to be rendered inside the accordion body.
-    </DocsAttributesItem>
-</DocsAttributes>
\ No newline at end of file
+<ComponentApiDocs ComponentTypes="[typeof(Accordion), typeof(AccordionToggle), typeof(AccordionItem), typeof(AccordionHeader), typeof(AccordionBody)]" />
```
David-Moreira, line 97: Could generics work somehow? That could be cool. Value1, Value2, Value3
i.e 
`<ComponentApiDocs Value1"Accordion" Value2="AccordionToggle" Value3="AccordionItem" Value4="AccordionHeader" Value5="AccordionBody" />`
Not sure if worth it though? Thoughts?
@stsrki @tesar-tech ?

Source/SourceGenerators/Blazorise.Generator.Features/README.md
```diff
@@ -0,0 +1,12 @@
...
+
+Attributes like `GenerateIgnoreEqualityAttribute` are used to signal the source generator to produce additional code. These attributes have no functional purpose in the final compiled code, so they are removed post-compilation using Fody.
+
+If you add a new attribute here, make sure to account for its removal in `Blazorise.Weavers.Fody`.
+
+> It might be worthwhile to place "ready-to-be-removed" attributes in a dedicated namespace, making it easier to remove the entire namespace rather than individual attributes.
+
+## ApiDocsDtos
+
+- Is used by generated code (`ComponentApiDocsGenerator.cs`) and by `Blazorise.Docs` to present the Blazorise API.
```
David-Moreira, line 11: If I understand correctly, if **Fody** works correctly no one can access these objects by using the namespace. i.e `Blazorise.Generator.Features.ApiDocsDtos.ApiDocsForComponent` as these will not exist in the final product?

David-Moreira (changes requested):
Idea is really good, we've really wanted this in the past, but didn't really get around to it, well due to how long it does take to implement and get right. But I guess it will be great in the long run! So, awesome work!

---

Unfortunately I don't see any of the generated docs, I must be doing something wrong, but I didn't try to troubleshoot, I just built and run the docs project and API is empty, example : 
image

---

It mostly looks good, but I share same concerns as @stsrki , where polluting the main product source code with utilities for secondary stuff(docs) just doesn't feel right.

I'd personally wouldn't use a Source Generator for this, as it always require to be a dependency of the projects themselves. I'd just use the Source Generator just for the actual product (Blazorise) needs, which currently is just the **EqualitySourceGenerator**

---

I'd look at something like the `Blazorise.Docs.Generator`, which is more like a stand alone piece that is able to find stuff by paths and still inspect and extract proper information. 

After all .cs files do follow consistent structures. And there might be ways to load the information from the built .dll themselves into memory and extract everything that we need?

Documentation/Blazorise.Docs/Pages/Docs/Components/Accordions/AccordionPage.razor
stsrki: It doesn't make sense to me. We would need to have as many strongly-typed type params as there could be used at some point. Better to leave it as it is, with `typeof()`.

tesar-tech:
I share the point of view that the current solution, while functional, is too clunky and appears brittle...

I have probably found a definitive solution that has it all - no reference in Blazorise or extensions project, re-uses all the current code from ApiDocs generator (no need for new way of acquiring xml comments ) and has the functionality for default values. 

It's a cli tool, that touches the source code from files, thus have all the necessary context... The prototype works. I this one could solve all the issues.

stsrki:
> I share the point of view that the current solution, while functional, is too clunky and appears brittle...
> 
> I have probably found a definitive solution that has it all - no reference in Blazorise or extensions project, re-uses all the current code from ApiDocs generator (no need for new way of acquiring xml comments ) and has the functionality for default values.
> 
> It's a cli tool, that touches the source code from files, thus have all the necessary context... The prototype works. I this one could solve all the issues.

Will it be part of `Blazorise.Docs.Compiler`, or `Blazorise.Docs.Generator`? Both are CLI tools and both are used for source generation.

Source/SourceGenerators/Blazorise.Generator.Features/README.md
tesar-tech: yep.

tesar-tech:
Yes, it's now part of the `Blazorise.Docs.Compiler`. It works and serves its purpose well—the unnecessary references are removed. 

There are some differences between the source generator approach and this one. Primarily, the source generator context has everything already "packed," while in this approach, we need to handle references manually, identify the right namespaces to scan, etc.

The solution works for every extension out of the box. While there were some issues with `Dictionary` and default values, these have all been resolved. 

Currently, the solution mirrors the logic used by the source generator by generating `.cs` files. This was a necessity for the source generator, where generating files other than `.cs` is not feasible or recommended. However, in this context, it feels suboptimal. The `*.ApiDocs.cs` files (e.g., `Blazorise.ApiDocs.cs`, `Blazorise.Animate.ApiDocs.cs`, etc.) are used solely for data storage, and a different file type would be more appropriate for that purpose. 

The current approach also adds a significant number of lines of code that aren't strictly needed. My proposal is to convert the `ApiDocs` to JSON (or XML) and read it at runtime. This change would simplify the structure, as `ApiDocsDtos` are currently maintained in two places, which would no longer be necessary. Additionally, the code responsible for obtaining the `ApiDocs` (`ComponentsApiDocsSource`) would be simplified into a straightforward JSON deserialization, replacing the current implementation that relies on generics.

stsrki:
I've tested it, and the generator works. Once thing I have noticed is that the first navigation to a page where new API docs are listed can hang a bit. Around half a second or faster, but it is noticeable. Do you know why it would happen?

Also, what about the extensions? Can we also move them to the new system? I have tried `<ComponentApiDocs ComponentTypes="[typeof(Animate)]" />` But nothing shows.

tesar-tech:
> I've tested it, and the generator works. Once thing I have noticed is that the first navigation to a page where new API docs are listed can hang a bit. Around half a second or faster, but it is noticeable. Do you know why it would happen?

That's because of the `ComponentsApiDocsSource` singleton class that does the loading (using reflection) when initialized. Ideally that should be done on app start. But we don't really need the reflection as long as we implement the "json datasource" (that should also be done on app start). There is also some init time needed for every component, as it lists the inheritance chain and copies the properties out of it.. This also can be done ahead of time (but it's rather small operation).

> Also, what about the extensions? Can we also move them to the new system? I have tried `<ComponentApiDocs ComponentTypes="[typeof(Animate)]" />` But nothing shows.

The reason for Animate being empty is the lack of inheritance from `BaseComponent`. [Already discussed here](https://github.com/Megabit/Blazorise/pull/5846#issuecomment-2505510664). I still think the IBaseComponent would be the best solution. Otherwise it will pickup all the classes or we'll have to white list the components. (the IBaseCompnent would be added only to the components that doesn't inherit from `BaseComponent`)... With solving this, we can move to the new system.

stsrki:
> That's because of the `ComponentsApiDocsSource` singleton class that does the loading (using reflection) when initialized. Ideally that should be done on app start. But we don't really need the reflection as long as we implement the "json datasource" (that should also be done on app start). There is also some init time needed for every component, as it lists the inheritance chain and copies the properties out of it.. This also can be done ahead of time (but it's rather small operation).
> 
> >

Ok, I have optimized it by loading just the active assembly, since all `IComponentsApiDocsSource` are placed in Blazorise.Docs.

> The reason for Animate being empty is the lack of inheritance from BaseComponent. https://github.com/Megabit/Blazorise/pull/5846#issuecomment-2505510664. I still think the IBaseComponent would be the best solution. Otherwise it will pickup all the classes or we'll have to white list the components. (the IBaseCompnent would be added only to the components that doesn't inherit from BaseComponent)... With solving this, we can move to the new system.

I still don't understand why this would be a problem. We can just iterate up to the needed base class. If `BaseComponent` is the problem we can go up to the Blazor native `ComponentBase`. And in case we find Blazorise `BaseComponent` we stop there.

stsrki:
I think the current implementation is good enough. If everyone agrees, I'll merge it.

@tesar-tech @David-Moreira

tesar-tech:
> Ok, I have optimized it by loading just the active assembly, since all `IComponentsApiDocsSource` are placed in Blazorise.Docs.

Good catch. Another leftover from "true SG".

> I still don't understand why this would be a problem. We can just iterate up to the needed base class. If `BaseComponent` is the problem we can go up to the Blazor native `ComponentBase`. And in case we find Blazorise `BaseComponent` we stop there.

Yeah, I did that. The only problem I see now that we have multiple conditions to be identified as a Component for docs, but all of them are in one place.


I added additional fixes for cases that emerged from new components being added. Before merging, I would check all the extensions first with the "new API" to ensure they are addressed while the code knowledge is still fresh.

I'm not entirely comfortable with the fact that we are adding thousands of lines of code to the codebase as a data source, otherwise I would consider it completed.

Documentation/Blazorise.Docs/Components/ComponentApiDocsProperty.razor
```diff
@@ -0,0 +1,59 @@
...
+
+                @if (property.IsBlazoriseEnum)
+                {
+                    var enumNames = Enum.GetNames(property.Type);
+
+                    <Paragraph Margin="Margin.Is0.FromBottom">
+                        <Text>Possible values:</Text>
+                        @for ( int enumValue = 0; enumValue < enumNames.Length; enumValue++ )
+                        {
+                            int i1 = enumValue;
```
David-Moreira, line 33: Actually **var i** was fine. as most programmers are acquainted with that for loops.

I meant that the **i1** wasn't clear for someone that would not understand that blazor needs to somehow capture the value "inline". So I meant to clarify the i1 with localIdx, localI, or in this case localEnumValue, or something along those lines, just so it doesn't make it as obscure, and an unsuspecting programmer tries to remove the i1, thinking it's redudant to have it.

David-Moreira:
> I think the current implementation is good enough. If everyone agrees, I'll merge it.
> 
> @tesar-tech @David-Moreira

Seems good, way better to have it this way. There's alot of code, I wouldn't bother with optimizing or code standards for this specifically, what matters is that it works. I'd just bother if we need to optimize for performance reasons.

David-Moreira:
> > I think the current implementation is good enough. If everyone agrees, I'll merge it.
> > @tesar-tech @David-Moreira
> 
> Seems good, way better to have it this way. There's alot of code, I wouldn't bother with optimizing or code standards for this specifically, what matters is that it works. I'd just bother if we need to optimize for performance reasons.

As long as it's running correctly on the docs frontend, I'm fine with it.

tesar-tech:
- The generator is now able to inherit comments from `Microsoft.AspNetCore.Components`. A few such cases are present in the codebase.
- However, it was unable to inherit XML comments from `System.Runtime.IDisposable.Dispose`. This issue (pobably) arises because `System.Runtime` is an implicit reference (but I didn't dig too deep). As a result, `<inheritdoc>` cannot resolve the comment for `Dispose`. 
  - Possible solution: Avoid using `<inheritdoc>` for such cases, as `Dispose` is typically overridden to add functionality beyond simple disposal.

---

## Extensions

I added the new ApiDocs to all the extensions and add new "features" to the generator along the way...

### Major Problem
Some components include (JS)options in their documentation, but the ApiDocs generator is not designed to scan these. It only processes components. Affected extensions include: Captcha, Chart, Cropper, (possibly more)

### Minor Problems
- Captcha: The `State` property does not appear in ApiDocs because it is not marked as a `[Parameter]`. However, it is documented in the original source.

- RouterTabsPageAttribute: This is not a component, so it does not have a record in ApiDocs. This is similar to the "options" problem.

- Validation: The `Validate` method uses `<inheritdoc>`, but there is no valid source to inherit from. This is not the only instance of invalid `<inheritdoc>` usage.

- Chart: Many XML comments are missing. 

- Icon: The documentation page for the `Icon` is in the Extensions, but the `Icon` itself is defined in Blazorise.

Documentation/Blazorise.Docs.Compiler/ApiDocsGenerator/README.md
```diff
@@ -0,0 +1,98 @@
...
+- It doesn't pick up the parameters from Razor files.
+- For inheritance to be "read," it needs to be explicitly specified inside the `.cs` file:
+  - This is especially true for `ComponentBase`, which is inherited by default when creating a `.razor` file.
+    - You need to explicitly declare `SomeComponent : ComponentBase`.
+
+This limitation can be overcome by scanning through the generated files (`SomeComponent.razor.cs`).
+
+## System.Runtime
+
+Retrieving XML comments from the `System.Runtime` assembly can be problematic, particularly for the `IDisposable` interface.
```
stsrki, line 22: Have you tried this library to read XML comments? Maybe they have already solved this kind of problems?

https://github.com/loxsmoke/DocXml

stsrki:
> Major Problem
> Some components include (JS)options in their documentation, but the ApiDocs generator is not designed to scan these. It only processes components. Affected extensions include: Captcha, Chart, Cropper, (possibly more)

We define every API docs explicitly by `typeof(ComponentName)`. So by that we should be able to read whatever we want. We don't need to scan entire assembly. The only difference from regular class(JS options), and component is the inheritance. Components go to BaseComponent/ComponentBase, and classes are read as they are, in full.

---

> ### Minor Problems
> * Captcha: The `State` property does not appear in ApiDocs because it is not marked as a `[Parameter]`. However, it is documented in the original source.

Possibly and overlook from our side.

> * RouterTabsPageAttribute: This is not a component, so it does not have a record in ApiDocs. This is similar to the "options" problem.

This should be same as my previous comment.

> * Validation: The `Validate` method uses `<inheritdoc>`, but there is no valid source to inherit from. This is not the only instance of invalid `<inheritdoc>` usage.

We should fix this kind of problem. But in another PR. Once we merger this PR. We are going to improve comments. You can just make a list in an issue so that we know what to do.

> * Chart: Many XML comments are missing.

Same, we will fix it later.

> * Icon: The documentation page for the `Icon` is in the Extensions, but the `Icon` itself is defined in Blazorise.

Because Icon cannot be used without an extension, just define `typeof(Icon)` from Blazorise core.

stsrki:
For options classes, ie `ReCaptchaOptions`, we can leave them as they are now in the docs. And only focus on component SG.

stsrki:
> However, it was unable to inherit XML comments from `System.Runtime.IDisposable.Dispose`. This issue (pobably) arises because `System.Runtime` is an implicit reference (but I didn't dig too deep). As a result, `<inheritdoc>` cannot resolve the comment for `Dispose`.
> * Possible solution: Avoid using `<inheritdoc>` for such cases, as `Dispose` is typically overridden to add functionality beyond simple disposal.

Make a list of methods to ignore methods, and add Dispose onto it. The user should not be aware of it.

stsrki:
@tesar-tech you can also ignore all internal components. Every component that starts with `_` is considered internal. This is our naming convention.

image

stsrki:
image

For enums like this, we could omit `Blazorise.Video.` namespace and write it as `VideoControlsType.PlayLarge, VideoControlsType.Play, etc.`

stsrki:
For `Icon` possible values, this seems like an overkill. Maybe for Icon we don't show possible values? Or if we find any enum that has more than some N possible values?

image

stsrki:
image

This stopped working.

tesar-tech:
> This stopped working.

Just the generic type were not handled correctly. Fixed.

> For `Icon` possible values, this seems like an overkill. Maybe for Icon we don't show possible values? Or if we find any enum that has more than some N possible values?

What about displaying just first 5 values and then button `Show all` similar to how `Remarks` is handled?



> @tesar-tech you can also ignore all internal components. Every component that starts with `_` is considered internal. This is our naming convention.

I moved the code little bit ahead, so it doesn't get processed at all. Also all the conditions are now on one place within one switch expression.  

```csharp
  ( bool continueProcessing, bool skipParamAndComponentCheck ) = type switch
        {
            _ when type.TypeKind != TypeKind.Class || type.DeclaredAccessibility != Accessibility.Public  => (false, false),
            _ when type.Name.StartsWith( '_' ) => ( false, false ),
            _ when type.Name.EndsWith( "Options" ) => ( true, true ),
            _ when type.Name.EndsWith( "RouterTabsPageAttribute" ) => ( true, true ),
            _ when !type.AllInterfaces.Any( i => i.Name == "IComponent" ) => ( false, false ),
            _ => ( true, false )
        };
```

Which also solves the `RouterTabsPageAttribute` issue and is one line of code to also add the `Options`. So it's there and you can use them. Not sure if you want to still omit it, so only Cropper has that as demonstration. 

```
<ComponentApiDocs ComponentTypes="[typeof(Cropper),typeof(CropperViewer), typeof(CropperCropOptions)]" />
```

It would be maybe worth it to add them under `Options` heading (on the same level as Methods, Params, Events)...?


> For enums like this, we could omit `Blazorise.Video.` namespace and write it as `VideoControlsType.PlayLarge, VideoControlsType.Play, etc.`

Little bit hacky, but fixed.

Documentation/Blazorise.Docs.Compiler/ApiDocsGenerator/README.md
tesar-tech: The issue I am facing with `System.Runtime` is that it doesn't appear to the analyzer that `IDispose` is from `System.Runtime`. Somehow, it's blind to this assembly, so it "doesn't know which XML to load".

stsrki:
> What about displaying just first 5 values and then button `Show all` similar to how `Remarks` is handled?
> 
> >

It might work but don't worry about it now.

We can also make an exception and add a URL to the Icons page `docs/extensions/icons-available`.

> > For enums like this, we could omit `Blazorise.Video.` namespace and write it as `VideoControlsType.PlayLarge, VideoControlsType.Play, etc.`
> 
> Little bit hacky, but fixed.

Nice :)

---

We are about done for now. I want to move from this to other tasks. We can always improve documentation later. If anything goes wrong, we will fix it.

Just to have a note for later. Instead of generating `Blazorise.ApiDocs.cs`, and other files. We can generate `ComponentNameApi.razor` with real Parameters, Events, and Methods section. What are the benefits? Well, for a start, it would be a lot faster in real time because we will not need to iterate for `type of (ComponentName)` and find each parameter, etc. Everything would already be precompiled. We will do that for 2.0.

stsrki (approved):

PR #5848, opened 2024-11-18
DataGrid: Support Multiple items on DataGridSelectColumn

David-Moreira:


Source/Extensions/Blazorise.DataGrid/BaseDataGridColumn.cs
```diff
@@ -18,6 +18,11 @@ public class BaseDataGridColumn<TItem> : BaseDataGridComponent
     /// <returns>Formatted display value.</returns>
     public string FormatDisplayValue( object value )
     {
+        if (value is string[] values )
```
David-Moreira, line 21: **Reminder to self :** Can we perhaps try to just find whether it's a collection and enumerate over it while doing .ToString to support more types?

Source/Extensions/Blazorise.DataGrid/_DataGridCellSelectEdit.razor.cs
```diff
@@ -73,6 +74,11 @@ private void OnSelectedValueChanged( object value )
         }
     }
 
+    private async Task OnSelectedValuesChanged( IReadOnlyList<object> values )
+    {
+        await CellValueChanged.InvokeAsync( values?.Select( x => x.ToString() )?.ToArray() );
```
David-Moreira, line 79: **Reminder to self :** Can we somehow support more types other then string[]? int[], so on...

David-Moreira:
@stsrki Can you take a look at the current solution. I believe this should be fine. What do you think?

Source/Extensions/Blazorise.DataGrid/BaseDataGridColumn.cs
```diff
@@ -18,6 +19,24 @@ public class BaseDataGridColumn<TItem> : BaseDataGridComponent
     /// <returns>Formatted display value.</returns>
     public string FormatDisplayValue( object value )
```
stsrki, line 20: Should we make `FormatDisplayValue` virtual and override the logic in DataGridSelectColumn?

Source/Extensions/Blazorise.DataGrid/BaseDataGridColumn.cs
```diff
@@ -18,6 +19,24 @@ public class BaseDataGridColumn<TItem> : BaseDataGridComponent
...
+            foreach ( var item in values )
+            {
+                if ( firstIteration )
+                {
+                    collectionDisplayValue = item?.ToString();
+                    firstIteration = false;
+                    continue;
+                }
+
+                collectionDisplayValue += $",{item}";
```
stsrki, line 35: I would use StringBuilder for concatenation.

PR #5872, opened 2024-11-26
DataGrid: add ExpressionCompiler utility

David-Moreira:


Documentation/Blazorise.Docs/Pages/Docs/Extensions/DataGrid/Examples/DataGridLargeDataExpressionCompilerExample.razor
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    private int totalEmployees;
+
+    private async Task OnReadData(DataGridReadDataEventArgs<Employee> e)
+    {
+
+        if (!e.CancellationToken.IsCancellationRequested)
+        {
+            var query = employeeListSource.AsQueryable().ApplyDataGridSort(e.Columns).ApplyDataGridSearch(e.Columns);
```
stsrki, line 42: Why do we have `ApplyDataGridSort`, `ApplyDataGridSearch`, and `ApplyDataGridPaging` as extensions? Can't we already have a prebuilt query expression for the ReadData callback argument?

Like `e.Request`?

David-Moreira: A single method already exists, and you could use it: 
**ApplyDataGridFilters(e.columns);** but the thing is, that you need to calculate the total items. So typically you only want to build part of the query so you can **.Count()**, before you go ahead and apply **Paging**, that's why it's separated into separate methods.

image


As for the argument you suggest, we could indeed make an additional overload for the **e** => **DataGridReadDataEventArgs**, which would be a little more seamless, but I'd probably keep the **DataGridColumnsInfo** one still, what do you think?

David-Moreira: > As for the argument you suggest, we could indeed make an additional overload for the e => DataGridReadDataEventArgs, which would be a little more seamless, but I'd probably keep the DataGridColumnsInfo one still, what do you think?

And same thing reasoning applies, where you shouldn't really apply a **paging query before counting**.

David-Moreira: Also to me **IEnumerable<DataGridColumnInfo>** seems easier to handle if you have to use it on the backend, typically sending it over to an api and de-serializing it, or constructing it manually just to use the feature.

stsrki: > As for the argument you suggest, we could indeed make an additional overload for the **e** => **DataGridReadDataEventArgs**, which would be a little more seamless, but I'd probably keep the **DataGridColumnsInfo** one still, what do you think?

Yes, add it. There are always lazy devs who will appreciate it. They also have extension methods for extra customization. Also, keep the DataGridColumnsInfo .

David-Moreira: > > As for the argument you suggest, we could indeed make an additional overload for the **e** => **DataGridReadDataEventArgs**, which would be a little more seamless, but I'd probably keep the **DataGridColumnsInfo** one still, what do you think?
> 
> Yes, add it. There are always lazy devs who will appreciate it. They also have extension methods for extra customization. Also, keep the DataGridColumnsInfo .

I mean....
`var query = employeeListSource.AsQueryable().ApplyDataGridSort(e.Columns).ApplyDataGridSearch(e.Columns);`
vs
`var query = employeeListSource.AsQueryable().ApplyDataGridSort(e).ApplyDataGridSearch(e);`

Totally appreciated!! 🤔 hehe

David-Moreira: Actually, there's a good benefit to your suggestion, this overload here can encapsulate a bit of logic regarding Virtualize vs Paging:
image

stsrki (approved):

PR #5873, opened 2024-11-30
DevOps: Removes unnecessary dotnet setup

tesar-tech:
## Description

Just a quick fix from seeing warnings in gha log.   

dotnet is there on ubuntu by default


stsrki (commented):
I don't think we are safe to remove older frameworks. We need to keep them to make sure that the build is passing on all of them.

tesar-tech:
No, no.. It doesn't remove them, because they are part of the `ubuntu-latest`. This PR just removes the the unnecessary setup, that does nothing actually.

image

Issue #5657, opened 2024-07-30
[2.0] Obsolete APIs removal and changelog


stsrki:
In this ticket, we define all of the APIs that are marked as obsolete and that will eventually be removed in 2.0.

### Link

- [ ] Removed Match.Custom enum value

PR #5878, opened 2024-12-04
Next 2.0 extension apis

stsrki:
Continuation of #5657

Working on changing the API for extension components. The new API will drop `SelectedValue` and `SelectedValues` and replace them with a single parameter, `Value,` which handles all the use cases.

- [x] DropdownList
- [ ] SelectList
- [ ] TransferList
- [ ] ListView

The change of API also required that the TValue of the component cannot be defined as a single type when it works in **Multiple** modes, eg.

- **Old:** `<DropdownList TValue="string"`
- **New:** `<DropdownList TValue="<IReadOnlyList<string>"`


stsrki:
Close in favor of #6259

Issue #5871, opened 2024-11-25
[2.0] Migration tool


stsrki:
There will be a lot of breaking changes in 2.0. And, I am thinking of how to make it easy for our users to migrate. One idea is to create a CLI tool that the users can install and the run on their project to automate the migration. We will need to parse razor code and replace any parameter on component that could be renamed.

Example

**Rule 1 - two way binding**

```razor
<TextEdit @bind-Text="@value" />
```

```razor
<TextEdit @bind-Value="@value" />
```

---

**Rule 2 - one way binding**

```razor
<TextEdit Text="@value" />
```

```razor
<TextEdit Value="@value" />
```

---

**Rule 3 - manual two way binding**

```razor
<TextEdit Text="@value" TextChanged="@((v)=>value=v)" />
```

```razor
<TextEdit Value="@value" ValueChanged="@((v)=>value=v)" />
```

---

**Rule 4 - no closing tag**

```razor
<TextEdit Text="@value">
  Some content
</TextEdit/>
```

```razor
<TextEdit Value="@value">
  Some content
</TextEdit/>
```

---

**Rule 5 - expressions**

```razor
<TextEdit Text="@value" TextChanged="@((v)=>value=v)" TextExpression="@(()=>value)" />
```

```razor
<TextEdit Value="@value" ValueChanged="@((v)=>value=v)" ValueExpression="@(()=>value)" />
```

**Rule 6 - Collapse in Accordion replaced by Accordion "inner" components**

```razor
<Collapse>
    <CollapseHeader Padding="Padding.Is1.OnY">
        <Heading Size="HeadingSize.Is5" TextAlignment="TextAlignment.Start">
            <AccordionToggle>Some text</AccordionToggle>
        </Heading>
    </CollapseHeader>
    <CollapseBody TextAlignment="TextAlignment.Start">
        Some body
    </CollapseBody>
</Collapse>
```

```razor
<AccordionItem>
    <AccordionHeader Padding="Padding.Is1.OnY">
        <Heading Size="HeadingSize.Is5" TextAlignment="TextAlignment.Start">
            <AccordionToggle>Some text</AccordionToggle>
        </Heading>
    </AccordionHeader>
    <AccordionBody TextAlignment="TextAlignment.Start">
        Some body
    </AccordionBody>
</AccordionItem>
```

tesar-tech:
I spent some time thinking about this, and with more features implemented for 2.0.

We want to create a tool that will actually help developers with migration.

[...]

PR #5890, opened 2024-12-11
Accordion: strip all usage of Collapse

### LINKED ISSUE #5248: [2.0] Strip all logic that uses Collapse within the Accordion

tesar-tech:
## Description

removes the "link" between `Collapse` and `Accordion`

Closes #5248


stsrki (commented):
Codewise, looks good to me. I still need to run it.

Please, update 2.0 release notes and mention and migration steps needed for the upgrade.

Source/Blazorise.Bootstrap5/Components/Button.cs
```diff
@@ -22,8 +22,8 @@ protected override void BuildClasses( ClassBuilder builder )
     {
         if ( ParentCollapseHeader?.ParentCollapse != null )
         {
-            if ( ParentCollapseHeader.ParentCollapse.InsideAccordion )
-                builder.Append( "accordion-button" );
+            // if ( ParentCollapseHeader.ParentCollapse.InsideAccordion )
```
stsrki, line 25: I think this

```
if ( ParentCollapseHeader.ParentCollapse.InsideAccordion )
    builder.Append( "accordion-button" );
```

should go away for both collapse and accordion.

tesar-tech: I don't understand fully the connection between collapse (collapseHeader, etc) and button. 

But this seems the full removal of "collapse stuff" from button should be done? Because it seems there is no other function for that..?

I am really not sure, please check the newest commit.

stsrki: The Button was used as a toggler for collapse. Since we have introduced AccordionToggle, the Button is not needed anymore.

stsrki (commented):
LGTM. Just update the release and migration notes.

And, update https://github.com/Megabit/Blazorise/issues/5871 if needed so that you don't forget.

stsrki (approved):

PR #5893, opened 2024-12-12
[next-2.0] Changes RightAligned to EndAligned on Dropdown and DropdownList

### LINKED ISSUE #3835: Rename Dropdown RightAligned to EndAligned
Since all of Left/Right are renamed to Start/End I think it makes sense to do the same for the Dropdown RightAligned parameter.

tesar-tech:
## Description

Closes #3835

Questions/todos:

- The comments need to be updated as well. Regarding that: Should we consider unifying the source of `EndAligned` under one interface and then using `///<inheritdoc/>`? This approach would help avoid repeating the same XML comment across four places: `Dropdown.razor.cs`, `DropdownList.razor.cs`, `DropdownJSOptions`, and `DropdownState`. It’s not just about the comments; `EndAligned` has the same meaning in all these contexts, so centralizing it makes sense.

- release notes need a record, but I will do it on the end.

- There’s still one instance of `RightAligned` left - in `BarDropdown`. However, this case is genuinely "right-aligned" because, even with `dir="rtl"`, it retains its alignment. In contrast, `Dropdown` properly respects the start-end directionality. Perhaps it’s worth reconsidering the functionality of `BarDropdown` to align it with the start-end behavior (in a different task)?

### DIFF
--- a/Source/Blazorise/Components/Dropdown/Dropdown.razor.cs
+++ b/Source/Blazorise/Components/Dropdown/Dropdown.razor.cs
@@ -72,7 +72,7 @@ public partial class Dropdown : BaseComponent, IAsyncDisposable
                 options: new()
                 {
                     Direction = GetDropdownDirection().ToString( "g" ),
-                    RightAligned = RightAligned,
+                    EndAligned = EndAligned,
                     DropdownToggleClassNames = ClassProvider.DropdownToggleSelector( IsDropdownSubmenu ),
                     DropdownMenuClassNames = ClassProvider.DropdownMenuSelector(),
                     DropdownShowClassName = ClassProvider.DropdownObserverShow(),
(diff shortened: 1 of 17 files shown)


stsrki:
> ## Description
> Closes #3835
> 
> Questions/todos:
> 
> * The comments need to be updated as well. Regarding that: Should we consider unifying the source of `EndAligned` under one interface and then using `///<inheritdoc/>`? This approach would help avoid repeating the same XML comment across four places: `Dropdown.razor.cs`, `DropdownList.razor.cs`, `DropdownJSOptions`, and `DropdownState`. It’s not just about the comments; `EndAligned` has the same meaning in all these contexts, so centralizing it makes sense.

All these components, options, and states are different types of objects, so I would not like to introduce one common interface for all of them. And if we would to add interface for each specific type then we would be back to the same problem, but with more complexity. Keep it simple. Just add the same comment to all places where it is needed.

> * There’s still one instance of `RightAligned` left - in `BarDropdown`. However, this case is genuinely "right-aligned" because, even with `dir="rtl"`, it retains its alignment. In contrast, `Dropdown` properly respects the start-end directionality. Perhaps it’s worth reconsidering the functionality of `BarDropdown` to align it with the start-end behavior (in a different task)?

Makes sense. I will open a new ticket for BarDropdown.

Source/Blazorise/Components/Dropdown/Dropdown.razor.cs
```diff
@@ -428,12 +428,12 @@ public bool Visible
     /// If true, a dropdown menu will be right aligned.
```
stsrki, line 428: Just update comments to mention the end, instead of right side.

stsrki (commented):
Please, always update release notes, and/or migration notes before requesting the final review. 

Other than that, the PR looks good, but I cannot merge without them.

PR #5896, opened 2024-12-13
DataGrid: Mixes CurrentPage and PageChanged into single two-way bindable param

### LINKED ISSUE #3610: DataGrid Reading Twice When non-default PageSize changes
**Describe the bug**
DataGrid "ReadData" event is called twice when it's PageSize property is set to a non-default value; in this case 10;

The datagrid currently only has a callback for the ReadData event for server-side sorting, filtering and pagination.  DataGrid data is only populated from the ReadData callback.

This does not happen with the component page first loads - only after the value from the page size drop down is changed.

This also does not happen if PageSize is not set in markup.

StateHasChanged or InvokeAsync(StateHasChanged) are never explicitly invoked.

**To Reproduce**
1. Navigate to a page with a Blazorise DataGrid component, PageSize set to 10 - ReadData fires once as expected.
2. Change the value in the page size dropdown from 10 to 5.
3. Callback for ReadData fire twice.  Two database calls instead of one.

**Expected behavior**
ReadData callback fires one time if PageSize dropdown value is changed.

**Notes**
This a Blazor Server App, not Web Assembly.

Target Framework: net6.0

Visual Studio Enterprise 2022, version 17.0.4

All current .NET SDKs

Several custom source generators such as thisassembly installed.

Nuget Packages:
- blazorise.bootstrap5 0.9.5.6
- blazorise.charts 0.9.5.6
- blazorise.components 0.9.5.6
- blazorise.datagrid 0.9.5.6
- blazorise.fluentvalidation 0.9.5.3
- blazorise.icons.fontawesome 0.9.5.6
- blazorise.richtextedit 0.9.5.6
- blazorise.sidebar 0.9.5.6
- blazorise.spinkit 0.9.5.6

Render mode for the app is set to "Server" and not "ServerPrerendered" in _Host.cshtml because of too many issues of callbacks firing twice instead of once on initial page load.

**Example Code**
Markup Code
```
<DataGrid TItem="Item" 
                  Data="Data" 
                  PageSize="10" /* Issue only occurs if this is explicitly set */
                  ReadData="ReadData" 
                  ShowPager
                  ShowPageSizes 
                  Responsive
                  TotalItems="Data?.TotalItemCount">
</DataGrid>
```
Code Behind
```
private IPagedList<Item>? Data;

// called once on page load
// called twice if page size changes.
private async Task ReadData(DataGridReadDataEventArgs<Item> e)
{
    Data = await Service.GetPagedListAsync(e.Page, e.PageSize, e.CancellationToken).ConfigureAwait(false);
}
```

BTW, this is one of the best packages that I have ever used.  Please keep up the good work.

tesar-tech:
## Description

- Mixes CurrentPage and PageChanged into single two-way bindable param
- Fixes pagesize "bug" in datagrid test page;
- Removes unnecessary DataGridPageChangedEventArgs.cs
- updates release notes

Closes #3610


stsrki (commented):
What about renaming the `CurrentPage` in 

- `PaginationContext`
- `CurrentPage`

Does that also makes sense?

tesar-tech:
> What about renaming the `CurrentPage` in
> 
> * `PaginationContext`
> * `CurrentPage`
> 
> Does that also makes sense?

In `PaginationContext` and `DataGridState`  there the current page is really a state, information. But in Datagrid it's more like a thing you can operate with (through binding). You "go to page 20" not "go to current page 20". In pagination it's more like "changing the current page".... Just slight nuance, no strong opinion. 


Also renaming to `Page` in `PaginationContext` would require to rename all the event handlers, etc.

stsrki:
> > What about renaming the `CurrentPage` in
> > 
> > * `PaginationContext`
> > * `CurrentPage`
> > 
> > Does that also makes sense?
> 
> In `PaginationContext` and `DataGridState` there the current page is really a state, information. But in Datagrid it's more like a thing you can operate with (through binding). You "go to page 20" not "go to current page 20". In pagination it's more like "changing the current page".... Just slight nuance, no strong opinion.
> 
> Also renaming to `Page` in `PaginationContext` would require to rename all the event handlers, etc.

From a user perspective, I think it makes sense to name them all the same. That way, we will be consistent.

@David-Moreira, what do you think?

David-Moreira:
I prefer the same naming always to avoid miss interpretations from different developers.

I also have no strong feelings about `CurrentPage` vs `Page`, I guess both are understandable, but generally you don't name parameters with "Current" prefix, like `CurrentData`, `CurrentPageSize`, `CurrentFilter`, etc, etc.. You just assume they are what it's set to or will be due to being set declaratively. `Current` is redundant in my opinion.
 
I'd not use current even to represent "current state". I believe that's already implied when accessing an api that represents state.

I think it only makes sense if you have two conflicts parameters to make it explicitly which one is which, i.e `oldItem` vs `currentItem`. And I'd probably still personally use `OldItem` and `Item` instead of the `currentItem` example.

stsrki (approved):

PR #5897, opened 2024-12-13
Datagrid - prevents submitting form on Enter 

### LINKED ISSUE #5790: [Bug]: Enter on filter field submits form
### Blazorise Version

latest

None

When you use the [Enter] key in a search field of a DataGrid, the whole webpage is submitted and refreshed.
This only happens hen only 1 seachfield is present in the grid. When more textual searchfield are present(even hidden) this unwanted behavior is not present.

### Steps to reproduce

See example: https://github.com/tacoWanneeJama/BlazorAppGridFilter/settings

### What is expected?

The webpage should off course be submitted from a Blazor page.

Page refresh

Chrome

### Any additional comments?

As workaround I now add an invisible text-field.

tesar-tech:
closes #5790

`SubmitFormOnEnter` is respected even in the case of single input field is present in datagrid (case when filter is on and no other fields are present).

### DIFF
--- a/Source/Extensions/Blazorise.DataGrid/DataGrid.razor
+++ b/Source/Extensions/Blazorise.DataGrid/DataGrid.razor
@@ -12,7 +12,8 @@
     }
     <Form>
 
-        @if ( IsCellEdit && ( CommandColumn is null || !CommandColumn.SaveCommandAllowed ) )
+        @if ( (IsCellEdit && (CommandColumn is null || !CommandColumn.SaveCommandAllowed))
+            || (Filterable && !SubmitFormOnEnter) )
         {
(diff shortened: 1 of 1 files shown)


Source/Extensions/Blazorise.DataGrid/DataGrid.razor
```diff
@@ -12,7 +12,8 @@
     }
     <Form>
 
-        @if ( IsCellEdit && ( CommandColumn is null || !CommandColumn.SaveCommandAllowed ) )
+        @if ( (IsCellEdit && (CommandColumn is null || !CommandColumn.SaveCommandAllowed))
```
stsrki, line 15: Can we move this entire if statement under the getter property and name it according to what it does? For future clarity.

stsrki (approved):

Issue #5216, opened 2024-01-10
[Bug]: I want to drag the column width, but when the dragged column becomes wider, other columns will become narrower, and no scroll bar will appear in the horizontal direction.


He-Wu:
### Blazorise Version

1.2.0

### What Blazorise provider are you running on?

Bootstrap5

### Link to minimal reproduction, or a simple code snippet

https://v12x.blazorise.com/docs/extensions/datagrid/features/resizing

### Steps to reproduce

![image](https://github.com/Megabit/Blazorise/assets/56440130/d67a97d3-f294-4b40-b3af-0061f1c3f091)
I want to drag the column width, but when the dragged column becomes wider, other columns will become narrower, and no scroll bar will appear in the horizontal direction.

### What is expected?

Scroll bars appear horizontally and other column widths remain unchanged

### What is actually happening?

I want to drag the column width, but when the dragged column becomes wider, other columns will become narrower, and no scroll bar will appear in the horizontal direction.

### What browsers are you seeing the problem on?

Chrome

David-Moreira:
Hello @He-Wu Have you tried setting the `DataGrid` to `Response`?

He-Wu:
> Hello @He-Wu Have you tried setting the `DataGrid` to `Response`?

Hi,yes I've tried setting Response to true and false with the same result

David-Moreira:
Correction, I meant `Responsive` 
Ok, we need to take a look whether we can make it so you are more freely able to resize the `DataGrid` columns when it's set to `Responsive`, since the `DataGrid` overflows horizontally.

He-Wu:
> Correction, I meant `Responsive` Ok, we need to take a look whether we can make it so you are more freely able to resize the `DataGrid` columns when it's set to `Responsive`, since the `DataGrid` overflows horizontally.

Ok,thanks

David-Moreira:
> Did you ever have a look at this? We're running into the same problem right now. Are there maybe any possible workarounds that we could take?

Unfortunately not.
The resizing feature is handled by js. You'd probably need to rewrite with your own version and override our own to make it work like you want, But I don't think it's an easy feat.

stsrki:
This might be fixes with [#5908](https://github.com/Megabit/Blazorise/pull/5908)

Issue #5904, opened 2024-12-17
[Bug]: Resizable: Resizable not working using bootstrap5


Vishakha27200:
### Blazorise Version

1.7.1

### What Blazorise provider are you running on?

Bootstrap5

### Link to minimal reproduction or a simple code snippet

 ```
  <DataGrid TItem="TestResultSummary" Navigable Resizable ResizeMode="TableResizeMode.Columns"
              Data="@TRvm.TestResults.ToList()" @ref="TRvm.bgrid" RowClicked="async args => await SingleClickStaticPresenterRefresh(args)"
    @bind-SelectedRows=TRvm.SelectedTestResults RowDoubleClicked="async args => await StaticPresenterOpen(args)"
              SelectionMode="Blazorise.DataGrid.DataGridSelectionMode.Multiple" Bordered 
              RowContextMenu="@OnRowContextMenu" RowContextMenuPreventDefault="true"
              Class="custom-grid-class" Sortable SortMode="DataGridSortMode.Single" Filterable FilterMode="DataGridFilterMode.Menu" Responsive FixedHeader Virtualize>
    <DataGridMultiSelectColumn Width="200px"></DataGridMultiSelectColumn>
    <DataGridCommandColumn />
    <DataGridColumn DisplayOrder="1" Caption="#" Filterable="false" Sortable="false" TextAlignment="TextAlignment.Center" Editable="true" Width="150px">
        <DisplayTemplate>
            @(((TRvm.bgrid?.DisplayData.ToList().IndexOf(TRvm.bgrid?.DisplayData.Where(x => x.Id == context.Id).FirstOrDefault())) + 1))
           </DisplayTemplate>
       </DataGridColumn>
       <DataGridColumn DisplayOrder="2" Field="Label" Caption="@Resources.Label" TextAlignment="TextAlignment.Start" SortComparer="new AEM.TDPU.DataModels.Comparer.TestResultSummaryLabelComparer()" Editable="true" Width="150px">
           <DisplayTemplate>
               @context.Label
           </DisplayTemplate>
       </DataGridColumn>

	</DataGrid> 

```
### Steps to reproduce & bug description

In my web project, the resizable feature is not working despite adding all the required attributes in the code. I have included all necessary Bootstrap 5 links and installed the Bootstrap 5 package. Additionally, I have completed all the necessary configurations, but the functionality still isn't working as expected. Please suggest improvements to resolve this issue.

### What is expected?

[...]

PR #5908, opened 2024-12-18
Table: fix resizing of columns in responsive mode

### LINKED ISSUE #5907: Table resizing doesn't work in responsive mode
Reported on 

https://blazorise.com/support/issues/238/help-understanding-datagrid-column-sizing

stsrki:
Closes #5907, #5904, #5216

While trying to come up with a solution to resizing in responsive mode, I noticed that we just need to define `minWidth` and `maxWidth` to make it work. After that I added a check to disable resizing if the user had explicitly defined min/max width.

**Pros**: it works as expected now

**Cons**: for fixed columns, the user should explicitly define HeaderCellStyle

**Test code:**

```
<DataGrid TItem="Employee"
          Data="@employeeList"
          @bind-SelectedRow="@selectedEmployee"
          Responsive
          Resizable
          ResizeMode="TableResizeMode.Columns">
    <DataGridCommandColumn />
    <DataGridColumn Field="@nameof(Employee.Id)" Caption="#" Sortable="false" Width="400px" CellStyle="@((item)=>"min-width:400px;max-width:400px;")" HeaderCellStyle="min-width:400px;max-width:400px" />
    <DataGridColumn Field="@nameof(Employee.FirstName)" Caption="First Name" Width="2400px" CellStyle="@((item)=>"min-width:2400px;max-width:2400px;")" HeaderCellStyle="min-width:2400px;max-width:2400px" />
    <DataGridColumn Field="@nameof(Employee.LastName)" Caption="Last Name" />
    <DataGridColumn Field="@nameof(Employee.Email)" Caption="Email" />
    <DataGridColumn Field="@nameof(Employee.Salary)" Caption="Salary" />
</DataGrid>

@code {
    [Inject]
    public EmployeeData EmployeeData { get; set; }
    private List<Employee> employeeList;
    private Employee selectedEmployee;
    private TableResizeMode resizeMode = TableResizeMode.Header;

    protected override async Task OnInitializedAsync()
    {
        employeeList = await EmployeeData.GetDataAsync();
        await base.OnInitializedAsync();
    }
}
```

### DIFF
--- a/Source/Blazorise/wwwroot/table.js
+++ b/Source/Blazorise/wwwroot/table.js
@@ -193,7 +193,10 @@ export function initializeResizable(element, elementId, mode) {
                 resizer.style.height = `${calculateTableActualHeight()}px`;
 
                 // Update the width of column
-                col.style.width = `${w + dx}px`;
+                const widthStyle = `${w + dx}px`;
+                col.style.width = widthStyle;
+                col.style.minWidth = widthStyle;
+                col.style.maxWidth = widthStyle;
             };
 
             // When user releases the mouse, remove the existing event listeners
--- a/Source/Blazorise/wwwroot/table.js
+++ b/Source/Blazorise/wwwroot/table.js
@@ -85,7 +85,7 @@ export function fixedHeaderScrollTableToRow(element, elementId, row) {
 export function initializeResizable(element, elementId, mode) {
     const resizerClass = "b-table-resizer";
     const resizingClass = "b-table-resizing";
-    const resizerHeaderMode = 0;
+    const RESIZER_HEADER_MODE = 0;
     let cols = null;
 
     element = getRequiredElement(element, elementId);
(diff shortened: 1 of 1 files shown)


tesar-tech:
This pr also fixes https://github.com/Megabit/Blazorise/issues/5216  - you can get any column as wide as you want. 

But it also introduced another bug:

- go to resizable  page https://blazorise.com/docs/extensions/datagrid/features/resizing (on blazorise.com the bug is not present, only in this branch)
- Test you can resize - you can. 
- Select row by clicking it.
- Now you cannot resize 

https://github.com/user-attachments/assets/ae230f4d-1054-487f-84f6-8957ac385231

stsrki:
Good catch with the bug.

> We should probably keep a flag that would track the set minWidth origin...

This was my original idea, but I thought it would be too complex, so I didn't bother. I guess I have to now :)

stsrki:
@tesar-tech should be fixed now. Please re-test.

tesar-tech (approved):
All good now.

PR #5912, opened 2024-12-23
Steps: change NavigationAllowed to async

### LINKED ISSUE #4688: Async/await support for NavigationAllowed in Steps component
Hello!
I have the following task related to Steps. I must make it possible to go to the next step only if the form is valid. This works fine while i dont use async validators.
I would really like to have support for asynchronous methods. Like example below.

```
  async Task<bool> NavigationAllowed(StepNavigationContext context)
  {
    bool result = true;
    if (context.CurrentStepIndex == 1 && !await _validation.ValidateAsync())
      result = false;
    return result;
  }
```

tesar-tech:
## Description

Closes #4688

Description inside the [proposal](https://github.com/Megabit/Blazorise/issues/4688#issuecomment-2550807470)


Demos/Blazorise.Demo/Pages/Tests/StepsPage.razor
```diff
@@ -77,7 +77,7 @@
                 <CardText>To control the navigation based on the user input you can use <Code>NavigationAllowed</Code> function.</CardText>
             </CardBody>
             <CardBody>
-                <Steps @ref="stepsRef" @bind-SelectedStep="selectedStep" NavigationAllowed="NavigationAllowed">
+                <Steps @ref="stepsRef" InitialStep="@initialStep" NavigationAllowed="NavigationAllowed">
```
stsrki, line 80: We, and users, would still want to have `@bind-SelectedStep="selectedStep"` to know which step is currently active. So I would return this part.

Source/Blazorise/Components/Steps/Steps.razor.cs
```diff
@@ -75,52 +77,79 @@ internal void NotifyStepPanelRemoved( string name )
...
+            _ = await TrySelectStep( InitialStep );
+    }
+
     /// <summary>
     /// Sets the active step by the name.
     /// </summary>
     /// <param name="stepName"></param>
     /// <returns>A task that represents the asynchronous operation.</returns>
-    public Task SelectStep( string stepName )
+    public async Task<bool> TrySelectStep( string stepName )
```
stsrki, line 92: Revert to previous names without the Try prefix.

stsrki: PS. Same for other methods.

Source/Blazorise/Components/Steps/Steps.razor.cs
```diff
@@ -163,37 +192,14 @@ internal int IndexOfStep( string name )
...
-            // raise the changed notification
-            SelectedStepChanged.InvokeAsync( state.SelectedStep );
-
-            DirtyClasses();
-        }
-    }
+    /// <summary>
+    /// Initial step to go on the component initialization
+    /// </summary>
+    [Parameter] public string InitialStep { get; set; } = string.Empty;
```
stsrki, line 202: I don't think having an `InitialStep` makes much sense. We can always internally store the SelectedPage during the component initialization lifecycle and use it for later where needed.

Demos/Blazorise.Demo/Pages/Tests/StepsPage.razor
tesar-tech: As described in the proposal having two-way binding is not a solution that will lead to predictable behavior.    
Users expect that setting the param will result in its change, but that's not the case if the change is blocked by `NavigationAllowed`. Also with async method it's not even feasible. 

To know which step is currently activem one can either use the get-only `SelectedStep` property or subscribe to `SelectedStepChanged`.

Source/Blazorise/Components/Steps/Steps.razor.cs
tesar-tech: Since the `SelectedStep` is getter-only, this is the only way how to setup the initial value on component itself. 
But this is almost the same discussion as the one above - about having the `SelectedStep` two-way bindable or not.

Demos/Blazorise.Demo/Pages/Tests/StepsPage.razor
stsrki: `SelectedStepChanged` changed is only raised after the `NavigationAllowed` is finished with the work. And `NavigationAllowed` should only be called within "step" methods. Two-way binding should not be blocked by `NavigationAllowed` as that will not work.

tesar-tech: So you want the two-way binding to bypass the `NavigationAllowed` check? That will create inconsistent behavior for different methods of moving steps... That will also change the "old" functionality, where the `NavigationAllowed` is called in two-way binding...

stsrki: My idea was to bypass the `NavigationAllowed` during two-way binding, yes. But then I went through the code and I think it can be still done by making `public string SelectedStep` as "dumb" parameter. And move all its logic from the setter into `SetParametersAsync`. The `SetParametersAsync` is asynchronous and in there we can check if the step is changed, and we can also await for any `NavigationAllowed` result. What do you think?

tesar-tech: Yes, this will "work," but it needs to reset the `SelectedStep` when navigation is blocked. While this isn’t technically challenging, it may lead to unexpected behavior:

- When you set `SelectedStep` (e.g., from 2 to 3) on a page, it will trigger the `NavigationAllowed` check. During this time, the UI will briefly display step 3 on the page (not in the `Steps` component), even if step 3 is blocked. This depends on how quickly the async action completes.
- Additionally, it will fire the `SelectedStepChanged` event even if the `SelectedStep` change was ultimately blocked.

This also creates a distinct behavior between these two actions:
- `SelectedStep="3"` 
- `ref.SelectStep("3")`

Is this a significant issue?

stsrki: > Is this a significant issue?

> When you set SelectedStep (e.g., from 2 to 3) on a page, it will trigger the NavigationAllowed check.

I don't think it is much of an issue. One is a declarative approach, and one is imperative. So, in declarative mode, we are forced to return to a previous state of `NavigationAllowed` returns false. The problem might be the UI flicker, but I'm not sure if it will happen at all if we prevent it during `SetParametersAsync`. Have you tried it?

> Additionally, it will fire the SelectedStepChanged event even if the SelectedStep change was ultimately blocked.

This cannot be avoided. Think of how the numeric input works. If you try to set a value below Min or above the Max, it has to retrigger ValueChanged to reset it to a valid range. This is essentially the same. We are resetting to a valid step.

---

In any case. Proceed with the `SetParametersAsync,`, and we'll see how it goes.

tesar-tech: the UI flicker only appears outside of the component. Not in the component itself (steps will not jump here and there)

```
This will flicker: @selectedStep

<Steps @bind-SelectedStep="selectedStep"...
```

This is because of the `SelectedStep` really being a dumb parameter which has its setter solved in the `OnParametersSet` method and only when the changes are approved by `NavigationAllowed`..

Source/Blazorise/Components/Steps/Steps.razor.cs
tesar-tech: No longer an issue with SelectedStep being two-way bindiable.

tesar-tech:
There is one more change that could be included in this PR:

The `StepNavigationContext` used as a parameter in `NavigationAllowed` currently has `NextStepName` and `NextStepIndex`. However, in reality, it should be something like `RequestedStepName`, since it’s used also when jumping between non-sequential steps (e.g., from step 1 to 4 or 4 to 1). 

The current `NextStep` naming aligns more with its literal "next" meaning in the `Steps` context..

Is it worth renaming?

stsrki (approved):

PR #5915, opened 2024-12-26
DropdownList: Filter text input in the menu

### LINKED ISSUE #5666: Dropdownlist Checkbox Selection Filter
Hi, Dropdownlist Checkbox Selection is great. Is it possible to add filter for the options inside dropdownlist?

tesar-tech:
## Description

Closes #5666

Adds a filter to the `DropdownList`, bringing its functionality closer to what `Autocomplete` offers. 

The current solution has limited customization options for both the filter functionality and its appearance. It might be worth considering merging some features with `Autocomplete`—for example, allowing customization of the filter type.

I’ll update the documentation and release notes once the core functionality is agreed upon.


Source/Extensions/Blazorise.Components/DropdownList.razor.cs
```diff
@@ -168,6 +212,11 @@ protected bool IsSelected( TValue value )
     /// </summary>
     [Parameter] public EventCallback<TValue> SelectedValueChanged { get; set; }
 
+    /// <summary>
+    /// Enebles filter text input on the top of the items list.
+    /// </summary>
+    [Parameter] public bool FilterEnabled { get; set; }
```
stsrki, line 218: We don't usually have `FeatureEnabled`, with `*Enabled` sufix naming style. So I would propose:

- `Filterable` (similar to what we have on DataGrid)
- `ShowFilter` (similar `Show*` prefix to what we have on other components)
- `Searchable`, or `ShowSearch` as an alternative for filter.

Source/Extensions/Blazorise.Components/DropdownList.razor.cs
```diff
@@ -103,6 +105,34 @@ private bool GetItemDisabled( TItem item )
...
+        if ( query == null )
+        {
+            filteredData = Enumerable.Empty<TItem>();
+            return;
+        }
+
+        if ( TextField == null )
+            return;
+
+        filteredData = Data.Where( x => TextField.Invoke( x ).Contains( FilterText, StringComparison.OrdinalIgnoreCase ) );
```
stsrki, line 126: Do we need to store filtered data? Or can we just return it as 

`return Data.Where( x => TextField.Invoke( x ).Contains( FilterText, StringComparison.OrdinalIgnoreCase ) );`

Source/Extensions/Blazorise.Components/DropdownList.razor.cs
tesar-tech: Filterable it is.

Source/Extensions/Blazorise.Components/DropdownList.razor.cs
tesar-tech: Yeah, in this the `filteredData` field isn't needed. That implementation can go without it.

I took it form the `Autocomplete`, where it is an `IList` and the data aren't enumerated on every get. But made it `IEnumberable`, because the Data are also `IEnumerable`.

Well - it is a question now. Should the `DropdownList.FilteredData` also be an `IList` instead of `IEnumerable`?

- IEnumerable will do the filtering on every enumeration. Possibly on every re-render. This could have significant perf implications.
- If converted to List, the filtering will be done only on text change, but it needs to fit into memory.... This is also a concern for current implementation of `Autocomplete`...

stsrki: Yeah, performance can be a problem. Dropdown, from a UX perspective is meant to be used only for small menus. With this new search features we are essentially breaking that "behavior" since we are allowing for very large menus. I guess that is fine if our users would expect that. But now we have a dilemma. Should we optimize or leave it as it is.

I think we can leave it as is, and optimize with caching later if it becomes a problem.

tesar-tech: In that case, the current state is final. `filteredData` is "needed" to avoid repeating all the checks unnecessarily, even when the filter isn’t dirty.

Another option is to keep everything (`Data` and `FilteredData`) as List -> that would signal "small" dataset and avoid unnecessary filtering on every `get`...

stsrki:
Add release notes to 2.0

Issue #5847, opened 2024-11-17
[Bug]: Change options for Cropper


drma-tech:
### Blazorise Version

all

### What Blazorise provider are you running on?

None

### Link to minimal reproduction or a simple code snippet

I can't set any configuration. If I try just one, the page gives an error when running. If I try more than one, I get this error.

![Image](https://github.com/user-attachments/assets/a98397a4-7cc7-48d8-a09c-4f82ef48de36)

### Steps to reproduce & bug description

any attempt to define options.

### What is expected?

can set one or more options.

### What is actually happening?

![Image](https://github.com/user-attachments/assets/9d3bc7b5-e7d0-4f28-ba52-5d05aa172cbc)
I don't know if this is the reason (as I don't know the behavior of the record), but the options structure is a record and the properties are set to init.

### Any additional comments?

ImageOptions, SelectionOptions, GridOptions

all have the same problem

stsrki:
It looks like there is something wrong with Visual Studio IntelliSense. Can you try restarting it?

drma-tech:
It's not Visual Studio. I can't even compile. And if I use just one attribute, I get this error:

![Image](https://github.com/user-attachments/assets/705be07f-07a1-4ed8-9640-1c69f138e6cd)

stsrki:
Replace this 
![Image](https://github.com/user-attachments/assets/41cb6b02-6f19-4118-b76c-d38501d8e6a4)

with `,`

drma-tech:
Taking advantage of the cropper subject, when I'm starting and still don't have an image, it looks like this:

![Image](https://github.com/user-attachments/assets/92be7922-767f-4131-bda6-4bcc20acf202)

drma-tech:
@stsrki SetParametersAsync  was resolved also?

stsrki:
The same fix is handled for load and update of image so it should work for SetParametersAsync also.

drma-tech:
They are two different things, even though I put both of them here. But I will test them as soon as they come out.

drma-tech:
![Image](https://github.com/user-attachments/assets/69cfdfba-824b-4a81-bc6f-8eed39069ec5)

![Image](https://github.com/user-attachments/assets/3590bda8-c051-4ab2-8122-cdc3bebd24f1)

I tested it today (better late than never). It still doesn't work.

drma-tech:
also (and yet), when i use the option, i got errrors

```
[...]

PR #5916, opened 2024-12-27
Cropper: behavior on invalid Source

tesar-tech:
## Description

Mentioned in #5847

This PR aims to improve the behavior on invalid `Source` in the image cropper. The behavior mentioned in #5847 isn't actually that bad—nothing crashes, and the "transparency grid" is a good indication that something went wrong.

Introduced in this PR:

- On an invalid `Source`, the cropper is disabled (and the previous state is re-enabled when the `Source` is fixed).
- `Source` now has the `EditorRequired` attribute as it is the only way to signal to users that it's required. I don't see any reason not to use it, but I also don't see it used anywhere else. So..?
- The `ImageLoadingFailed` event was added to inform users when something goes wrong with image loading. Ensuring the `Source` is correct remains the user's responsibility, but this helps handle such cases.

(invalid source includes `null`, `""` or "image/that/does/not/exist.jpg")

### DIFF
--- a/Source/Extensions/Blazorise.Cropper/Cropper.razor.cs
+++ b/Source/Extensions/Blazorise.Cropper/Cropper.razor.cs
@@ -260,12 +260,6 @@ public partial class Cropper : BaseComponent, IAsyncDisposable
         if ( ImageReady is not null )
             await ImageReady.Invoke();
     }
-    
-    internal async Task NotifyImageLoadingFailed( string errorMessage )
-    {
-        if ( ImageLoadingFailed is not null )
-            await ImageLoadingFailed.Invoke(errorMessage);
-    }
 
     #endregion
 
@@ -284,7 +278,7 @@ public partial class Cropper : BaseComponent, IAsyncDisposable
     /// <summary>
     /// The original image source.
     /// </summary>
-    [Parameter, EditorRequired] public string Source { get; set; }
+    [Parameter] public string Source { get; set; }
 
     /// <summary>
     /// The alt text of the image.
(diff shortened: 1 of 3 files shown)


Source/Extensions/Blazorise.Cropper/Cropper.razor.cs
```diff
@@ -324,6 +330,12 @@ internal async Task NotifyImageReady()
     /// This event fires when the image is ready / loaded.
     /// </summary>
     [Parameter] public Func<Task> ImageReady { get; set; }
+    
+    /// <summary>
+    /// This event fires when the image cannot be loaded. Usually because of 404 or <see cref="Source"/> being null.
+    /// Error message inside the string parameter.
+    /// </summary>
```
stsrki, line 337: We have a strict policy of not adding, or changing any APIs in the support channel. So I would move this event, `ImageLoadingFailed `,  to the next major milestone.

Source/Extensions/Blazorise.Cropper/wwwroot/blazorise.cropper.js
```diff
@@ -298,6 +294,26 @@ export function resetSelection(element, elementId) {
     }
 }
 
+let disabledBeforeImageLoadFailed;
+let wasDisabled;
```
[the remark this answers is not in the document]
tesar-tech: makes sense.

Source/Extensions/Blazorise.Cropper/Cropper.razor.cs
tesar-tech: Ok. I relaying on the addition not being a breaking change.

It's removed, I will put it back once this pr is synced with `master`.

Source/Extensions/Blazorise.Cropper/wwwroot/blazorise.cropper.js
```diff
@@ -298,6 +294,22 @@ export function resetSelection(element, elementId) {
     }
 }
 
+function cropperImageReady(cropperImage, cropperCanvas, instance) {
+    cropperImage.$ready((image) => {
+        if (instance.wasDisabled) {
```
stsrki, line 299: The `wasDisabled` seems strange in naming. Was, and Disabled means something in the past. I would name it either simply `disabled` or something more descriptive of what it does. Eg. `disabledOnLoad`, or `disabledOnError`.

stsrki (approved):

PR #5921, opened 2025-01-04
Modal: Hides scrollbar on modal.Show()

### LINKED ISSUE #5911: [Bug]: HTML body scrollbar should be hidden when opening modal
### Blazorise Version

1.6.2-1.7.1

Bootstrap5

Image

```csharp
@page "/"
@using Blazorise

<Button Color="Color.Primary" Clicked="@ShowModal">Show Modal</Button>

<div style="min-height: 1500px">
    This is some placeholder content
</div>

<Modal @ref="modalRef">
    <ModalContent Centered>
        <ModalHeader>
            <ModalTitle>Employee edit</ModalTitle>
            <CloseButton />
        </ModalHeader>
        <ModalBody>
            <Field>
                <FieldLabel>Name</FieldLabel>
                <TextEdit Placeholder="Enter name..." />
            </Field>
            <Field>
                <FieldLabel>Surname</FieldLabel>
                <TextEdit Placeholder="Enter surname..." />
            </Field>
            <Field>
                <FieldLabel>Name</FieldLabel>
                <TextEdit Placeholder="Enter name..." />
            </Field>
  
    
        </ModalBody>
        <ModalFooter>
            <Button Color="Color.Secondary" Clicked="@HideModal">Close</Button>
            <Button Color="Color.Primary" Clicked="@HideModal">Save Changes</Button>
        </ModalFooter>
    </ModalContent>
</Modal>
@code{
    // reference to the modal component
    private Modal modalRef;

    private Task ShowModal()
    {
        return modalRef.Show();
    }

    private Task HideModal()
    {
        return modalRef.Hide();
    }
}
```

According to the document example of Bootstrap, it will hide the scroll bar when the modal is opened.

https://getbootstrap.com/docs/5.0/components/modal/

Image

Image

### What is expected?

Hide the HTML body scroll bar 

It doesn't hide the HTML body  scroll bar 

Chrome

### Any additional comments?

_No response_

tesar-tech:
## Description  

Closes #5911  

When a modal is opened, the scrollbar on the body is removed. The emphasis here is on "on the body," as many sites, including [blazorise.com](https://blazorise.com), the demo sites, and numerous other web pages, don’t use the body as the main scrolling element. 

Currently, the solution is implemented for Bootstrap 5, but I need clarification on a few points before proceeding with other providers.  

## Code Questions  

The scrollbar-hiding functionality is based on Bootstrap's [scrollbar.js](https://github.com/twbs/bootstrap/blob/main/js/src/util/scrollbar.js). It seems that Blazorise already has parts of this functionality implemented in `modal.js` for the Bootstrap 5, Bootstrap, and Tailwind providers.  

Most of the code in [modal.js](https://github.com/Megabit/Blazorise/blob/458799c524ca5f1f01370cf8e69a55d379355864/Source/Blazorise.Bootstrap5/wwwroot/modal.js) appears to handle the correction of UI elements with `fixed` and `sticky` positioning when the scrollbar is removed. This correction is likely necessary to compensate for the blank space left by the scrollbar removal.

What confuses me most is the apparent absence of actual scrollbar removal in the existing code. Without this step, there’s nothing to compensate for, making much of the code seem redundant.

Here are my key questions:  
- If this compensation is necessary, why doesn’t the corresponding code appear in other providers?  
- Am I missing something about how this is meant to function?  

## Proposal  

1. Move the scrollbar handling code into a shared `scrollbar.js` and use it for scrollbar removal and width compensation across all providers. This would require some refactoring, as the handling of `fixed` and `sticky` classes is provider-specific.  
2. Consider merging all the `modal.js` files into a single shared file to simplify maintenance.  

### A Note on Sticky and Fixed Compensation  

I’m not convinced that the compensation for sticky and fixed elements is needed. In my testing, removing the scrollbar didn’t appear to affect these elements significantly because the body padding already compensates for the scrollbar width.  

Even if there are edge cases where this is an issue, the worst-case scenario seems to be a minor (approximately 20px) horizontal shift in elements that will be inactive and covered by the modal anyway. Removing this part of the code would significantly simplify the `modal.js` script.

### DIFF
--- a/Source/Blazorise.Bootstrap5/wwwroot/modal.js
+++ b/Source/Blazorise.Bootstrap5/wwwroot/modal.js
@@ -6,6 +6,13 @@ export function open(element, scrollToTop) {
     var modals = Number(document.body.getAttribute("data-modals") || "0");
 
     if (modals === 0) {
+        // Save the original overflow value
+        const originalOverflow = document.body.style.overflow || '';
+        document.body.setAttribute('data-original-overflow', originalOverflow);
+
+        // Hide the scrollbar
+        document.body.style.overflow = 'hidden';
+
         addClassToBody("modal-open");
     }
 
@@ -32,6 +39,9 @@ export function close(element) {
     }
 
     if (modals === 0) {
+        // Restore the original overflow value
+        document.body.style.overflow = document.body.getAttribute('data-original-overflow') || '';
+        document.body.removeAttribute('data-original-overflow');
         removeClassFromBody("modal-open");
     }
 


stsrki:
I'm trying the example from the OP, and I don't think this fix works. The scrollbar still shows as usual. Nothing changes. While on the BS5 docs the scrollbar is hidden completely. https://getbootstrap.com/docs/5.3/components/modal/#modal-components

@tesar-tech can you try and investigate to see what we are still missing?

tesar-tech:
> And other providers are already working without this adjustment.

What do you mean? They are not hiding the scrollbar in any way. The adjustment inside `bootstrap/modal.js` seemed unnecessary (based on my observations) because there was nothing to adjust for (no scrollbar hiding). That's what I'm asking about—what was the adjustment for?

Additionally, this adjustment has been copied into the `tailwind/modal.js` with bootstrap classes. To me, that seems like unnecessary code.

stsrki (approved):

PR #5924, opened 2025-01-07
Docs: check for null on namespace on API docs generator

tesar-tech:


tesar-tech:
I added the null check, but it doesn't solve the real issue. As I stated already - cannot be reproduced anywhere. 

It's weird because - based on the previous screenshots - it does fail only on one namespace - but it's not specific in any way, yet it cannot somehow find it.

The only difference between our setup I see is the ide... can you try to run it with `dotnet run` ?

PR #5925, opened 2025-01-08
Table: don't resize columns with FixedPosition

stsrki:
Closes https://blazorise.com/support/issues/238/help-understanding-datagrid-column-sizing

In my previous fix I have only disabled resizing of columns that have `min-width` and `max-width`. With this fix, I also disabled resizer for columns that have FixedPosition defined.

### DIFF
--- a/Source/Blazorise/Components/Table/TableHeaderCell.razor.cs
+++ b/Source/Blazorise/Components/Table/TableHeaderCell.razor.cs
@@ -143,11 +143,12 @@ public partial class TableHeaderCell : BaseDraggableComponent, IDisposable
     /// <summary>
     /// Gets the fixed position data attribute.
     /// </summary>
-    protected string FixedPositionDataAttribute => FixedPosition == TableColumnFixedPosition.Start
-        ? "start"
-        : FixedPosition == TableColumnFixedPosition.End
-            ? "end"
-            : null;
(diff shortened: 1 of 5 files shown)


Source/Blazorise/Components/Table/TableHeaderCell.razor.cs
```diff
@@ -140,6 +140,15 @@ protected override ValueTask DisposeAsync( bool disposing )
 
     #region Properties
 
+    /// <summary>
+    /// Gets the fixed position data attribute.
+    /// </summary>
+    protected string FixedPositionDataAttribute => FixedPosition == TableColumnFixedPosition.Start
```
tesar-tech, line 146: Why the `start` vs `end` matters? the js seems to care just about the thing not being null.

stsrki: This is for the future sake. And in case we, or users would like to know that information. So it is better to have full info since we already need this attribute.

tesar-tech: ok, ok.

then, why not switch expression instead the double conditional operator??

```
   protected string FixedPositionDataAttribute => FixedPosition switch
    {
        TableColumnFixedPosition.Start => "start",
        TableColumnFixedPosition.End => "end",
        _ => null
    };
```

PR #5933, opened 2025-01-10
ToastProvider: Styling customizations

### LINKED ISSUE #5922: Styling of toasts send to ToastProvider
I am trying to do customize the styling of the toast messages created with the `ToastProvider` and `IToastService`, but there are currently no options to style them at all.

The minimum would be optional css classes that could be passed along when using the `IToastService` methods.

### Additional context
I think it would be enough to have the additional class(es) on the toast itself, not the header and body.
The css rules would allow for styling those as well.

tesar-tech:
## Description

Closes #5922 

- adds `ToastStyle`, `ToastClass` and `IconName` to

### DIFF
--- a/Source/Blazorise/Components/ToastProvider/ToastProvider.razor
+++ b/Source/Blazorise/Components/ToastProvider/ToastProvider.razor
@@ -19,7 +19,7 @@
                 <ToastHeader>
                     @if ( ShowIntentIcon )
                     {
-                        <Icon Name="@GetIconName( toastInstance.Intent )" TextColor="@GetIconColor( toastInstance.Intent )" Margin="Blazorise.Margin.Is2.FromEnd" />
+                        <Icon Name="@(toastInstance.IconName ?? GetIconName( toastInstance.Intent ))" TextColor="@GetIconColor( toastInstance.Intent )" Margin="Blazorise.Margin.Is2.FromEnd" />
                     }
                     <Strong>@toastInstance.Title</Strong>
                     @if ( ShowCloseButton )
--- a/Source/Blazorise/Components/ToastProvider/ToastProvider.razor
+++ b/Source/Blazorise/Components/ToastProvider/ToastProvider.razor
@@ -17,9 +17,9 @@
                    Style="@toastInstance.Style"
                    Class="@toastInstance.Class">
                 <ToastHeader>
-                    @if ( ShowIntentIcon )
+                    @if ( toastInstance.ShowIcon ?? ShowIntentIcon )
                     {
-                        <Icon Name="@(GetIconName( toastInstance ))" TextColor="@GetIconColor( toastInstance.Intent )" Margin="Blazorise.Margin.Is2.FromEnd" />
+                        <Icon Name="@(GetIconName( toastInstance ))" IconStyle="toastInstance.IconStyle" IconSize="toastInstance.IconSize" TextColor="@GetIconColor( toastInstance.Intent )" Margin="Blazorise.Margin.Is2.FromEnd" />
                     }
                     <Strong>@toastInstance.Title</Strong>
                     @if ( ShowCloseButton )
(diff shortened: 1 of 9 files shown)


Source/Blazorise/Models/ToastInstanceOptions.cs
```diff
@@ -50,6 +50,21 @@ public class ToastInstanceOptions
     /// Delay in milliseconds before hiding the toast.
     /// </summary>
     public double? AutohideDelay { get; set; }
+    
+    /// <summary>
+    /// Custom CSS class name to apply to the Toast
+    /// </summary>
+    public string ToastClass { get; set; }
```
stsrki, line 57: ToastInstance already represents a Toast component. So I would remove the Toast prefix from the name as it is redundant.

Source/Blazorise/Models/ToastInstanceOptions.cs
```diff
@@ -50,6 +50,21 @@ public class ToastInstanceOptions
...
+    
+    /// <summary>
+    /// Custom inline styles to apply to the Toast
+    /// </summary>
+    public string ToastStyle { get; set; }
+    
+    /// <summary>
+    /// Icon to display inside the ToastHeader.
+    /// </summary>
+    public IconName? IconName { get; set; }
```
stsrki, line 67: I think just having an IconName is not enough. If the user would like to hide an icon completely they cannot do it. If left to null, it would just fallback to our default icon.

I would add a bool ShowIcon, which would be `true` by default. And having `IconStyle`, `IconSize` seems logical to have also.

Source/Blazorise/Components/ToastProvider/ToastProvider.razor
```diff
@@ -13,11 +13,13 @@
...
                    Autohide="@toastInstance.Autohide"
-                   AutohideDelay="@toastInstance.AutohideDelay">
+                   AutohideDelay="@toastInstance.AutohideDelay"
+                   Style="@toastInstance.ToastStyle"
+                   Class="@toastInstance.ToastClass">
                 <ToastHeader>
                     @if ( ShowIntentIcon )
                     {
-                        <Icon Name="@GetIconName( toastInstance.Intent )" TextColor="@GetIconColor( toastInstance.Intent )" Margin="Blazorise.Margin.Is2.FromEnd" />
+                        <Icon Name="@(toastInstance.IconName ?? GetIconName( toastInstance.Intent ))" TextColor="@GetIconColor( toastInstance.Intent )" Margin="Blazorise.Margin.Is2.FromEnd" />
```
stsrki, line 22: Pass the `toastInstance` to the `GetIconName` and have it decide what to report based on the instance state.

Source/Blazorise/Models/ToastInstanceOptions.cs
tesar-tech: That's exactly the problem I mentioned in the original issue. You can say "not enough" also about the `ToastHeader` and `ToastBody`, `CloseButoon` or any other parameters.
What's gonna be enough? Where is the line between "just a simple API for toast notification" and a full `Toast` experience? Why does it have to be simple API anyway?

Also the icon can be styled (and hidden) by classes or styles inside the toaster.

stsrki: > That's exactly the problem I mentioned in the original issue. You can say "not enough" also about the `ToastHeader` and `ToastBody`, `CloseButoon` or any other parameters.
> What's gonna be enough? Where is the line between "just a simple API for toast notification" and a full `Toast` experience? Why does it have to be simple API anyway?

It's never enough. We're always going to add more features as they come requested. The difference between building toasts from scratch and by using the service is just that. The users don't need to build it from scratch. That the value we give. And they always have on option from doing it from scratch.

> Also the icon can be styled (and hidden) by classes or styles inside the toaster.

They can. But then they need to know the actual classnames. Which we always try to abstract as much we can. We want them to work with Blazorise. And also what happens when they changes CSS provider? Then the classnames need to be changed manually. With Blazorise handling that it all will just work.

stsrki (approved):

PR #5940, opened 2025-01-19
Docs: DetailRow samples update

### LINKED ISSUE #4745: Detail Row Trigger : Better docs with common cases
Hello everybody,

I'm trying to use the datagrid in my application and want to have a Button that toggles a detailrow. The button image should change from ExpandMore to ExpandLess when the detailrow is displayed and back to ExpandMore when the detailrow is not displayed. Is there any possibility how to achieve this?

Some code:

```
<DataGrid @ref="dataGrid" TItem="ProcessFlow" Data="processFlows" Responsive Virtualize Editable="false" Filterable Resizable ResizeMode="TableResizeMode.Columns" SelectionMode="DataGridSelectionMode.Multiple" @bind-SelectedRows="selectedProcessFlows" DetailRowStartsVisible="false" ShowPager ShowPageSizes RowSelectable="RowSelectableHandler">
            <DataGridColumns>
                <DataGridMultiSelectColumn />
[...]
<DataGridColumn Caption="@_localizer["Details"]">
                    <DisplayTemplate>
                        <Button Color="Color.Secondary" Size="Size.Small" Clicked="() => OnDetailsButtonClicked(context)"><Icon Name="IconName.ExpandMore" /></Button>
                    </DisplayTemplate>
                </DataGridColumn>
            </DataGridColumns>
            <DetailRowTemplate>
                @{
                    //Select details only when nescearry
                    var dbContext = _dbContextFactory.CreateDbContext();
                    var processFlowWithDetails = dbContext.GetProcessFlowWithReferencesById(context.Id);
                    <Processdetails ProcessFlow="@processFlowWithDetails" />
                }
            </DetailRowTemplate>
        </DataGrid>

async Task OnDetailsButtonClicked(ProcessFlow? processFlow)
    {
        if ((dataGrid != null) && (processFlow != null))
        {
            await dataGrid.ToggleDetailRow(processFlow);
        }
    }
```

_Originally posted by @svenbaus in https://github.com/Megabit/Blazorise/discussions/4744_

tesar-tech:
## Description

Closes #4745

Adds sample code that demonstrates two functionalities:

- How to distinguish between an opened and a closed detail row.
- How to indicate when a detail row can be expanded.

While the implementation works, it may not fully address the original discussion, `Datagrid check if detailrow is displayed`.

Currently, there is no straightforward way to achieve this. The current solution relies on tracking "which rows have been clicked," which is indirectly tied to the detail row display. This approach is fragile and can easily break, especially when manual triggers for detail rows are added.

A better solution could involve exposing `DataGridRowInfo.HasDetailRow` by making `DataGrid.GetRowInfo` public and optimizing its performance.

???

### DIFF
--- a/.github/workflows/blazorise-ci-basic.yml
+++ b/.github/workflows/blazorise-ci-basic.yml
@@ -5,7 +5,7 @@ on: [push]
 jobs:
   build:
 
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
 
     steps:
     - uses: actions/checkout@v3
--- a/Documentation/Blazorise.Docs/Models/Snippets.generated.cs
+++ b/Documentation/Blazorise.Docs/Models/Snippets.generated.cs
@@ -7701,15 +7701,26 @@ Install-Package Blazorise.Chart.Zoom";
         public const string DataGridDetailRowTemplateExample = @"<DataGrid TItem=""Employee""
           Data=""@employeeList""
           @bind-SelectedRow=""@selectedEmployee""
-          DetailRowTrigger=""@((e) => e.Item.Salaries?.Count > 0 && e.Item.Id == selectedEmployee?.Id)""
+          RowClicked=""RowClicked""
+          DetailRowTrigger=""@(e => DisplayDetailRow(e.Item) && e.Item.Id == selectedEmployee?.Id)""
           Responsive>
     <DataGridColumns>
         <DataGridCommandColumn />
+        <DataGridColumn TItem=Employee >
+            <DisplayTemplate>
+                @if ( DisplayDetailRow( context ) )
+                {
+                    <Button>
+                        <Icon Name=""@(rowsWithDetail.Contains( context.Id ) ? IconName.ExpandLess : IconName.ExpandMore)""/>
+                    </Button>
+                }
+            </DisplayTemplate>
+        </DataGridColumn>
         <DataGridColumn Field=""@nameof(Employee.FirstName)"" Caption=""First Name"" />
     </DataGridColumns>
     <DetailRowTemplate>
         @{
-            var salaries = ( context as Employee ).Salaries;
+            var salaries = context.Salaries;
 
             <DataGrid TItem=""Salary""
                       Data=""salaries""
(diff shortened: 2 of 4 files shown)


tesar-tech:
The test errors are weird and have nothing to do with the actual code change.

I think these are caused by the update of `ubuntu-latest`, because [this last successful run](https://github.com/Megabit/Blazorise/actions/runs/12836005926/job/35796714425) is on 22.04
image

while now `ubuntu-latest` is on 24.04

image

stsrki:
> The test errors are weird and have nothing to do with the actual code change.
> 
> I think these are caused by the update of `ubuntu-latest`, because [this last successful run](https://github.com/Megabit/Blazorise/actions/runs/12836005926/job/35796714425) is on 22.04 image
> 
> while now `ubuntu-latest` is on 24.04
> 
> image

Can we specify the exact ubuntu version to make them run?

stsrki:
> A better solution could involve exposing `DataGridRowInfo.HasDetailRow` by making `DataGrid.GetRowInfo` public and optimizing its performance.

It could be exposed, yes. But there are some API that are kinda internal and that user shouldn't need to know and not be able to use. Like `TableRow`, `SetRowDetail`, `ToggleDetailRow`, `SetTableRow`. They are part of the DataGrid lifecycle and can affect how DataGrid works if used incorrectly.

tesar-tech:
> Can we specify the exact ubuntu version to make them run?

Yes, that will work.. Also the older dotnets (6,7) are not part of the ubuntu 24.04, it is still installable and working, but we (I) removed the installation step for these, I wasn't aware it is being removed with new version. (we can easily revert that change, it's on master branch).

22.04 will be removed (probably) on April 2027.

options:

- merge this (and propagate to other branches), wait until april 27 when the issue will be fixed, documented or non-existent
- try to fix it now, revert installation for older dotnets.

tesar-tech:
> It could be exposed, yes. But there are some API that are kinda internal and that user shouldn't need to know and not be able to use. Like `TableRow`, `SetRowDetail`, `ToggleDetailRow`, `SetTableRow`. They are part of the DataGrid lifecycle and can affect how DataGrid works if used incorrectly.

Yes, I understand it is private for a reason. Question is: should I try to came up with a solution for exposing the `HasDetailRow` state? I guess that would go to `master` branch so the answers doesn't block merging of this pr.

stsrki:
> Yes, I understand it is private for a reason. Question is: should I try to came up with a solution for exposing the `HasDetailRow` state? I guess that would go to `master` branch so the answers doesn't block merging of this pr.

I guess it might be worth it. It's good to know that information. Now that we are pursuing this. I think HasDetailRow is a weird name. So we can try having something more descriptive like `DetailRowExpanded`.

I will open new issue.

stsrki (approved):

PR #5949, opened 2025-01-22
DataGrid: Expose public info about DetailRow being expanded

### LINKED ISSUE #5947: DataGrid: add API that tells us if detail row is expanded or present
> Yes, I understand it is private for a reason. Question is: should I try to came up with a solution for exposing the `HasDetailRow` state? I guess that would go to `master` branch so the answers doesn't block merging of this pr.

I guess it might be worth it. It's good to know that information. Now that we are pursuing this. I think HasDetailRow is a weird name. So we can try having something more descriptive like `DetailRowExpanded`.

I will open new issue.

_Originally posted by @stsrki in https://github.com/Megabit/Blazorise/issues/5940#issuecomment-2606579959_

tesar-tech:
## Description

Closes #5947

This update exposes `DataGridRowInfo.DetailRowExpanded`, requiring several adjustments:

- **Renamed the original `DataGridRowInfo`**:  
  Changed to `_DataGridRowInfo` to reflect its private nature while retaining `DataGridRowInfo` as the publicly exposed name. Alternatives considered include `RowInfo`, `DataGridRowInfoPublic`, etc.

- **Introduced a new public `DataGridRowInfo`**:  
  This class now directly exposes `DetailRowExpanded`. And can be used in a future to expose more a different info.

- **Renamed `HasDetailRow` to `DetailRowExpanded`**:  
  As mentioned in the issue description. An alternative like `DetailRowVisible` is possible, given the existing `DetailRowStartsVisible`, but `Expanded` feels better. It might also be worth renaming `DetailRowStartsVisible` to `DetailRowStartsExpanded` for consistency..

- **Renamed `GetRowInfo` to `GetRowInfoPrivate`**:  
  This change allows `GetRowInfo` to be used as the public method.

Merged 1.7, as it required a sample and a functional gh action for testing.


stsrki:
I think something is not right with the PR. Too many files is changed. Have you used the wrong target branch?

tesar-tech:
> I think something is not right with the PR. Too many files is changed. Have you used the wrong target branch?

That's because I merged the 1.7 to this branch, because I needed the code example.

If you merge 1.7 to master the extra code won't be part of this pr anymore.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -414,7 +414,7 @@ public bool RemoveColumn( DataGridColumn<TItem> column )
     /// Links the child row with this datagrid.
     /// </summary>
     /// <param name="row">Row to remove.</param>
-    public bool RemoveRow( DataGridRowInfo<TItem> row )
+    public bool RemoveRow( _DataGridRowInfo<TItem> row )
```
stsrki, line 417: I don't think it makes sense to mix public API and private APIs. In this case we have public methods that are working with our "private" models. 

I would leave it as previously, as `DataGridRowInfo`. And then make some fields as internal. If possible.

tesar-tech: This change comes form the rename of `DataGridRowInfo` to `_DataGridRowInfo`.
There is no change in functionality or accessibility or class usage

stsrki: > There is no change in functionality or accessibility or class usage

True, but it is still a breaking change. And now we have a private named class publicly exposed. 

Revert to the original name, as we don't want this kind if naming for our API.

tesar-tech: Ok, I reverted back and renamed created the `DataGridRowPublicInfo` ...
I still don't quite understand the conventions here and the public/private/internal logic behind this.

Why are some members marked `public`, while used only for internal stuff? Why are some of those named with `_` prefix and some are not?

stsrki: It all boils down to simplicity. I never liked too much separation of concern just for the sake of it. And your version with the new `DataGridRowPublicInfo` was going into that direction. I just find it confusing, and introduced overhead when reading the code. Making the `DataGridRowInfo` is simple enough and then hiding internal `TableRow` enough.

Please don't find this an attack onto you in any way. Both of us are still learning from each other.

stsrki:
> * As mentioned in the issue description. An alternative like `DetailRowVisible` is possible, given the existing `DetailRowStartsVisible`, but `Expanded` feels better.

🤔`DetailRowVisible` is not a bad name.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
tesar-tech: I agree this is the simpler solution.

I see you changed the `TableRow` to internal and that's exactly the place where I lack the necessary knowledge. The knowledge that cannot be extracted out of just reading the code and probably out of anywhere.

I guess it's simpler to mark everything as public, but it really makes it hard-to-impossible to see what really is supposed to be public and what can be changed without any breaking changes.

Do you have any guidelines for that? Would you consider to refactor these parts for the sake of better orientation in the code?

stsrki (approved):

PR #5952, opened 2025-01-24
MemoEdit: include border size when autosizing

stsrki:
Fixes problem with auto size showing the scrollbar

> Another question: We have the issue that a memoedit with Autosize=”true” shows a scrollbar. Its height is a few pixels too small. This is especially a problem because the scrollbar hides the green check-mark of a validation in the top right corner. I saw this has been an issue in an earlier version of Blazorise and has been fixed. Is it possible this issue has been re-introduced in 1.7?

Test code:

```razor
<Div>
    <MemoEdit Text="@memoText"
              Size="Size.Small"
              TextWeight="TextWeight.Light"
              ReplaceTab="true"
              AutoSize="true"
              Overflow="Overflow.Auto"
              ReadOnly="false">
    </MemoEdit>
</Div>

<Button Clicked="SetText">Set Text</Button>
@code {
    private string memoText;

    private void SetText()
    {
        var emp = new Employee { Id = 1, FirstName = "Alexander Maximilian", LastName = "Hohenberg Von Strassburg", Email = "alexander.maximilian@example.com", City = "Berlin", Zip = "10115", DateOfBirth = new DateTime(1985, 5, 15), Childrens = 2, Salary = 50000, IsActive = true };

        var options = new JsonSerializerOptions();

        options.WriteIndented = true;

        memoText = System.Text.Json.JsonSerializer.Serialize(emp, options);
    }
}
```

### DIFF
--- a/Source/Blazorise/wwwroot/memoEdit.js
+++ b/Source/Blazorise/wwwroot/memoEdit.js
@@ -99,9 +99,10 @@ export function updateOptions(element, elementId, options) {
 
 function onInputChanged(e) {
     if (e && e.target) {
-        e.target.style.height = 'auto';
-        e.target.style.height = this.scrollHeight + 'px';
-        e.target.style.overflowY = 'hidden';
+        const textarea = e.target;
+        textarea.style.height = 'auto';
+        textarea.style.overflowY = 'hidden';
+        textarea.style.height = `${textarea.scrollHeight + 2}px`; // The +2 accounts for browser rendering inaccuracies with scrollHeight
     }
 }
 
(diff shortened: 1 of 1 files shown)


Source/Blazorise/wwwroot/memoEdit.js
```diff
@@ -99,9 +99,10 @@ export function updateOptions(element, elementId, options) {
 
 function onInputChanged(e) {
     if (e && e.target) {
-        e.target.style.height = 'auto';
-        e.target.style.height = this.scrollHeight + 'px';
-        e.target.style.overflowY = 'hidden';
+        const textarea = e.target;
+        textarea.style.height = 'auto';
+        textarea.style.overflowY = 'hidden';
+        textarea.style.height = `${textarea.scrollHeight + 2}px`; // The +2 accounts for browser rendering inaccuracies with scrollHeight
```
tesar-tech, line 105: It's not an inaccuracy. It's caused by border, which is 1px from top and bottom.

```
 Style="border: 5px solid;"
```
will introduce the "bug" again.

image

tesar-tech (changes requested):

PR #5954, opened 2025-01-27
DateEdit: properly parse the seconds part

### LINKED ISSUE #5953: DateEdit doesn't work with seconds
Reported by the user 👇

---

I'm experiencing a problem with the DateEdit component, I want to enter a date ánd time, I bind the DateValue to a DateTime? variable which I give a value in code containing a number seconds the seconds, somehow the seconds always display as 00.

```
<DateEdit TValue="DateTime?" InputMode="DateInputMode.DateTime" @bind-Date="@DateValue" />

@code {    
    private DateTime? DateValue { get; set; } = DateTime.Today.Add(new TimeSpan(18, 03, 24));
}
```

I can't seem to paste part of screengrab but in the browser I see:

26-01-2025 18:03:00  

While I would expect 

26-01-2025 18:03:24

I created this in a test-project using v1.7.3 of Blazorise.

stsrki:
Closes #5953

Fixes the date parser by adding the seconds part.

```
<Div>
    <Paragraph>
        <DateEdit TValue="DateTime?" InputMode="DateInputMode.DateTime" @bind-Date="@DateValue" />
    </Paragraph>
    <Paragraph>
        @DateValue
    </Paragraph>
</Div>

<Div>
    <Paragraph>
        <DateEdit TValue="DateTime?" InputMode="DateInputMode.Date" @bind-Date="@Date" />
    </Paragraph>
    <Paragraph>
        @Date
    </Paragraph>
</Div>
@code {
    private DateTime? DateValue { get; set; } = DateTime.Today.Add(new TimeSpan(18, 03, 24));
    private DateTime? Date { get; set; } = DateTime.Today.Add(new TimeSpan(18, 03, 24));
}
```

### DIFF
--- a/Source/Blazorise/Utilities/Formatters/Parsers.cs
+++ b/Source/Blazorise/Utilities/Formatters/Parsers.cs
@@ -18,7 +18,7 @@ public static class Parsers
     /// <summary>
     /// Internal date-time format. Compatible with HTML date inputs.
     /// </summary>
-    public const string InternalDateTimeFormat = "yyyy-MM-ddTHH:mm";
+    public const string InternalDateTimeFormat = "yyyy-MM-ddTHH:mm:ss";
 
     /// <summary>
     /// Internal month format. Compatible with HTML date inputs.
@@ -64,8 +64,10 @@ public static class Parsers
     public static readonly string[] SupportedParseDateTimeFormats = new string[]
     {
         InternalDateTimeFormat,
+        $"{InternalDateFormat}:ss",
         ExternalDateTimeFormat,
         "yyyy-MM-ddTHH:mm",
+        "yyyy-MM-ddTHH:mm:ss",
         CultureInfo.InvariantCulture.DateTimeFormat.LongDatePattern,
         CultureInfo.InvariantCulture.DateTimeFormat.ShortDatePattern,
         "o", // a string representing UTC
--- a/Source/Blazorise/Utilities/Formatters/Parsers.cs
+++ b/Source/Blazorise/Utilities/Formatters/Parsers.cs
@@ -66,7 +66,6 @@ public static class Parsers
         InternalDateTimeFormat,
         ExternalDateTimeFormat,
         "yyyy-MM-ddTHH:mm",
-        "yyyy-MM-ddTHH:mm:ss",
         CultureInfo.InvariantCulture.DateTimeFormat.LongDatePattern,
         CultureInfo.InvariantCulture.DateTimeFormat.ShortDatePattern,
         "o", // a string representing UTC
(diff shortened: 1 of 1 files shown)


Source/Blazorise/Utilities/Formatters/Parsers.cs
```diff
@@ -64,8 +64,10 @@ public static class Parsers
     public static readonly string[] SupportedParseDateTimeFormats = new string[]
     {
         InternalDateTimeFormat,
+        $"{InternalDateFormat}:ss",
```
tesar-tech, line 67: What is this one good for? this will result in `yyyy-MM-dd:ss`. Is there a case when this will be happening?

Source/Blazorise/Utilities/Formatters/Parsers.cs
```diff
@@ -64,8 +64,10 @@ public static class Parsers
...
         ExternalDateTimeFormat,
         "yyyy-MM-ddTHH:mm",
+        "yyyy-MM-ddTHH:mm:ss",
```
tesar-tech, line 70: Isn't that a duplicate with line 66? `InternalDateTimeFormat,`

tesar-tech (commented):
+ 
`"yyyy-MM-ddTHH:mm",` inside the  `string[] SupportedParseDateFormats`...
Should't that also include seconds?

Also - why the `InputMode="DateInputMode.Date"` supports `DateTime`s with with time, shouln't that remove the time part?

stsrki:
> `"yyyy-MM-ddTHH:mm",` inside the `string[] SupportedParseDateFormats`... Should't that also include seconds?

Not sure if it's needed but I have added it just in case.

> Also - why the `InputMode="DateInputMode.Date"` supports `DateTime`s with with time, shouln't that remove the time part?

Because DateTime is the most used date type for dates and times. It's just a matter which part of it we bind.

tesar-tech:
> Because DateTime is the most used date type for dates and times. It's just a matter which part of it we bind.

Probably for a different ticket... I understand the usage of DateTime, but my concern is that it strips the time part from it when you change the date. Should't it keep the time as was the initial value?

Now it works like:
- Start the dateEdit with  time 18:03:24 and date 2025-01-29 
- submit the form 
- 2025-01-29  18:03:24 is sent. But for user it seems like only the date is taken care of.

Second example:
- (the same) Start the dateEdit with  time 18:03:24 and date 2025-01-29 
- change date to something else and back to 2025-01-29 
- submit the form
- 2025-01-29  00:00:00 is sent. Different result, but for user it appears the same..

stsrki:
> Probably for a different ticket... I understand the usage of DateTime, but my concern is that it strips the time part from it when you change the date. Should't it keep the time as was the initial value?

We had a similar report in one of the tickets. Usually, this is not a problem since the input will bind to a date field. But in a situation when the user binds **two** inputs, one for the date part and one for the time part, all to the **same date value**, then it would make sense to do this. But, considering how rare case it is and how much more complex it would be to make the component lifecycle while keeping track of the time part to synchronize it with the date part, it just doesn't make sense to do it.

PR #5963, opened 2025-02-04
Docs: Introduces category, add Theme to api docs

### LINKED ISSUE #4932: `DataGridCheckColumn` does not use `Theme.ColorOptions.Primary`
**Describe the bug**
According to @stsrki  `DataGridCheckColumn` should use `Theme.ColorOptions.Primary` as base of its color.
This does not work.

**To Reproduce**
see reproduction in https://github.com/oaldrian/BlazoriseReproductions/tree/datagrid-checkcolumn-color 
Index.razor and App.razor

**Expected behavior**
`DataGridCheckColumn` should use `Theme.ColorOptions.Primary` as base of its color.

**Screenshots**
Screenshot 2023-08-16 191000

**Additional context**
Tested on 1.3

tesar-tech:
## Description

Closes #4932 , but is more than that.

- Introduces Category and Subcategory, that allow us to put any class into the ApiDocs.
- Adds all the classes from Theme folder to ApiDocs. 
  - `/docs/theming` has all the subfolder listed as links/pages.
  - Every subfolder (subcategory) has all the classes in containing folder. Palettes, Colors, Options... After I saw the API I realized we probably don't want everything to be listed there. (so - which ones?)


Documentation/Blazorise.Docs/Pages/Docs/Components/Theming/ThemingApiColors.razor
stsrki, line 1: Add Page suffix to the page razor file. Same for other theming API pages.

Documentation/Blazorise.Docs/Pages/Docs/Components/Theming/ThemingApiColors.razor
```diff
@@ -0,0 +1,8 @@
+﻿@page "/docs/theming/api/colors"
+
+<Seo Canonical="/docs/theming/api/colors" Title="Blazorise Colors API" Description="Blazorise Colors API" />
```
stsrki, line 3: For better SEO, we need to make sure that every page has the following:

1. H1 heading
2. Breadcrumbs
3. Lead text that briefly describes the content of the page
4. Longer paragraph with more context
5. SEO description must be between 100 and 320 characters

stsrki (commented):
It seems a lot of options don't have any comments

image

It might be good to pass them through chatgpt to generate it?

stsrki:
> ## Description
> Closes #4932 , but is more than that.
> 
> * Introduces Category and Subcategory, that allow us to put any class into the ApiDocs.
> * Adds all the classes from Theme folder to ApiDocs.
>   
>   * `/docs/theming` has all the subfolder listed as links/pages.
>   * Every subfolder (subcategory) has all the classes in containing folder. Palettes, Colors, Options... After I saw the API I realized we probably don't want everything to be listed there. (so - which ones?)

Can you explain more what this solves? Not sure if I understand it.

Documentation/Blazorise.Docs/Pages/Docs/Components/Theming/ThemingApiColors.razor
tesar-tech: Yes, will do.

Do you want the pages as they are now?

For example the `ThemeSwatch` and the whole `Palettes` folder is something I don't understand (and I am not if we need the apidocs for that). May I have a brief explanation?

Also this:

image

Is something I don't know what to do with, because the xml comments warning is suppressed there. (same for all the colors)... This seems like another case of "public private members"...

tesar-tech:
> Can you explain more what this solves? Not sure if I understand it.

The whole api docs was constructed for components only. This adds the possibility to make api docs from all the classes in the codebase. This pr also simplifies the usage, leveraging the categories:

```
<ComponentApiDocs Category="Theme"  Subcategory="Options" />
```

vs

```
<ComponentApiDocs ComponentTypes="list manually all the classes that are in themes/options folder (which would get outdated once new options is added)"  />
```

Categories also paved the way for more complicated scenarios, where the category isn't just copy of the folder structure.





> It might be good to pass them through chatgpt to generate it?

yep.

Documentation/Blazorise.Docs/Pages/Docs/Components/Theming/ThemingApiColors.razor
stsrki: > For example the ThemeSwatch and the whole Palettes folder is something I don't understand (and I am not if we need the apidocs for that). May I have a brief explanation?

Yes, the pages are OK as separated like you make them.

> Is something I don't know what to do with, because the xml comments warning is suppressed there. (same for all the colors)... This seems like another case of "public private members"...

I didn't bother commenting them years ago. But with power of GPT it is really easy now. So yes, do it.

tesar-tech:
All done. Now just the colors are bit weird.

It would be probably helpful to see them as colors, maybe with background of that specific color...

Also the sort is alphabetical, which places the `_50` between 400 and 500...

image

Worth fixing?

PR #5973, opened 2025-02-14
Autocomplete and Dropdown: improves checkbox selection

### LINKED ISSUE #5950: Autocomplete: close menu when clicked outside
Support ticket https://blazorise.com/support/issues/250

tesar-tech:
## Description

Closes #5950 – Fixed an issue where clicking outside the autocomplete did not close the menu.  

While solving this issue, I ended up doing a broader refactoring. Unnecessary code naturally fell away as I worked through the root cause. This is still a work in progress, as some of the changes are breaking, and I'd like to get your opinion before finalizing them.  

### Changes & Refactoring  

- **OnBlur Handling**  
  The previous implementation did not prevent the `onblur` event from firing on the autocomplete textbox (it did for the checkbox but not for the text). This is now resolved by re-arranging event handling and placing the text outside the checkbox in `DropdownItem`.  
  - Fixing this at the `DropdownItem` level allows us to remove all code related to handling the blur scenario, primarily the logic around `clickFromCheck` and `focusFromCheck`. This part of the refactoring is not yet complete.  

- **Merging Click & Checkbox Events**  
  `DropdownItem` previously had two separate events: click and checkbox check. With these changes, I don't think we need them both.  
  - Both events essentially signal a state change (i.e., a selection occurred).  
  - Click and check are used interchangeably, adding unnecessary complexity without real benefits.  
  - I merged them into a single `SelectedChanged` event with a `bool`, since selection state change is what truly matters. The checkbox merely reflects the selected state at the `DropdownItem` level.  
  - We can still differentiate between check and click events under the new `onblur` handling approach.  

### Another notes

- The `DropdownList` is probably broken right now, but I think it’s an easy fix.  
- We can remove the obsolete `Multiple` property from Autocomplete.  
- This refactoring might also allow us to eliminate some provider-specific `DropdownItem.razor` files.  

Tested on bootstrap5.


stsrki:
Tested on BS5 and it works nicely.

> OnBlur Handling

This seems to be working much better. Nice catch. Although please test on all providers and make sure spacing of the DropdownItem and Checkbox is not broken. Compare it to live demos for reference.

> Merging Click & Checkbox Events

We still need to have two separate events. Clicked event is a must-have for the dropdownitem. So it needs to stay. Rememeber, showing checkboxes is an extra feature of the dropdownitem. When we have regular dropdown we dont check items. we click on them to start an action. So, revert to include Clicked event.

I would leave the olf CheckedChanged naming, as it conveys better the usage of the dropdown item.

---

Also, make sure to test on all providers. There was always a reason why we did something i a specific way in the past- When removing old code, we need to make sure we don't break any edge case.

tesar-tech:
> We still need to have two separate events. Clicked event is a must-have for the dropdownitem. So it needs to stay. Rememeber, showing checkboxes is an extra feature of the dropdownitem. When we have regular dropdown we dont check items. we click on them to start an action. So, revert to include Clicked event.

The `Clicked` isn't removed, it is just part of the `SelectedChanged`, because even with simple dropdown you selecting an item.
The clicking will behave the same, just will fire different event. Checkbox is just a status indicator. Is there any use-case where clicking the checkbox will cause a different action than clicking the text next to it? 

Also the `Value` that was part of the `Clicked` event seems unnecessary. You cannot do anything with that inside the `DropdownItem` and the `Clicked` event is transferring the `Value` to the place where it's defined (component usage)... Am I missing here something?

tesar-tech:
I still don't think it is necessary to have both events. The `CheckedChanged` event is also triggered by `Clicked`. Merging them into one (with whatever name) would provide the same functionality as we have now, just with `EventArgs` containing `Item` and `Checked`. This would also allow us to eliminate code like:  

```
CheckedChanged="() => isCheckbox ? OnDropdownItemSelected(value) : Task.CompletedTask"
Clicked="() => isCheckbox ? Task.CompletedTask : OnDropdownItemSelected(value)"
```  

...anyway, the current state of this branch reflects what was there before.  

I figured out the issue with "content inside `Check` causing the dropdown to close." The alternative approach would have made styling more complicated (because it was outside of the `Check` container), so now it just includes an extra `div`.  

This solution works for all providers. I also refactored the AntDesign and Tailwind implementations, as these two have custom `DropdownItem` components. Now, they share as much code as possible without significantly affecting the base component. I'm not sure if this is the desired approach, but to me, it makes sense because I struggled to understand why AntDesign and Tailwind had their own separate components. Now, the shared logic is extracted, and the rest is documented.

Source/Blazorise/Components/Dropdown/DropdownItem.razor
```diff
@@ -1,16 +1,30 @@
...
-        <Div Display="Blazorise.Display.InlineBlock" @onclick:stopPropagation @onmouseup:stopPropagation>
-            <Check TValue="bool" Value="@(@checked)" ValueChanged="@CheckedChangedHandler" ValueExpression="@checkedExpression" Disabled="@Disabled" @onclick:stopPropagation>
+
+<a @ref="@ElementRef" id="@ElementId" class="@ClassNames" style="@StyleNames" @onmousedown:preventDefault @onmousedown="@EventCallback.Empty" @onclick="@ClickHandler" @attributes="@Attributes">
+    @Core
+</a>
+
+@code {
+    private protected RenderFragment Core =>
+        @<text>
```
stsrki, line 10: I understand that `<text>` is an SVG tag and that it will be removed by Blazor from the final HTML. I like the reusability of it, but I still have problems with it. 

What happens if, one day, the Blazor team decides to render it as a valid HTML tag? We will break down the components. That's why I prefer to almost always write it explicitely.

tesar-tech: The `<text>` is used in SVG, yes. But it is also a razor directive https://learn.microsoft.com/en-us/aspnet/core/mvc/views/razor?view=aspnetcore-9.0#explicit-delimited-transition

> What happens if, one day, the Blazor team decides to render it as a valid HTML tag?

This seems really unrealistic. A lot of breaking changes for no real benefit.

You can have the text tag if you want:

image
 
resulting in:

image

stsrki: Ah, I didn't see that in the Blazor docs. So they have a special rule for it. In that case, I agree that it is highly unrealistic it will be changed in the future.

PR #5975, opened 2025-02-19
Markdown: ImageUploadChanged with multiple files fix

### LINKED ISSUE #5961: [Bug]: Markdown ImageUploadChanged throws exception and has "odd" behavior
### Blazorise Version

1.7.3

Bootstrap5

```razor
<Markdown ImageUploadChanged="ImageUploadChanged"></Markdown>

@code {
    private async Task ImageUploadChanged(FileChangedEventArgs events)
    {
        try
        {
            var count = events.Files.Length; // Prints always one 

            foreach (var file in events.Files)
            {
                using var memoryStream = new MemoryStream();
                await file.WriteToStreamAsync(memoryStream);
            }
        }
        catch (Exception e)
        {
            Console.WriteLine(e);
            throw;
        }
    }
}
```

Using the `Markdown` component in combination with uploading **multiple** files/images at once leads to a `NullReferenceException`. 

The "odd" thing here is that `events.Files.Length` will always be one, independent of how many items you drag and drop onto the markdown field. That said, for each file dropped, the `ImageUploadChanged` will be invoked.

In a given scenario where you drop at least two or more files, `await file.WriteToStreamAsync(memoryStream);` will throw with the given exception.

### What is expected?

Aside from the exception, that shouldn't throw. I think it would be more obvious for folks if it either follows more closely what `InputFileChangeEventArgs` is doing (aka `GetMultipleFiles`) or populate the array with all the items, rather than invoking the callback once per file.

1. `NullReferenceException`
2. The `FileChangedEventArgs.Files` array only consists of one item

_No response_

### Any additional comments?

Here the stack trace of the second invocation:
```no-class
   at Blazorise.RemoteFileEntryStreamReader.OpenReadStreamAsync(CancellationToken cancellationToken)
   at Blazorise.RemoteFileEntryStreamReader.CopyFileDataIntoBuffer(Memory`1 destination, CancellationToken cancellationToken)
   at Blazorise.RemoteFileEntryStreamReader.WriteToStreamAsync(Stream stream, CancellationToken cancellationToken)
   at Blazorise.RemoteFileEntryStreamReader.WriteToStreamAsync(Stream stream, CancellationToken cancellationToken)
   at Blazorise.FileEntry.WriteToStreamAsync(Stream stream, CancellationToken cancellationToken)
   at Bug.Components.Pages.Home.ImageUploadChanged(FileChangedEventArgs events) 
```

tesar-tech:
## Description
Closes #5961

This effectively fixes the issue. The previous implementation overwrote the `_blazorFilesById` dictionary, which resulted in [this error](https://github.com/Megabit/Blazorise/blob/bd5b05a8e27a51861b7b6476e42a6930598ff1fe/Source/Blazorise/wwwroot/io.js#L46)  
when calling `await file.WriteToStreamAsync(memoryStream);`

The fix works, but there are plenty of areas for improvement.

## Files One by One, Not Grouped
- As @linkdotnet pointed out in the issue, files are processed one by one, even though `FileChangedEventArgs` supports multiple files at once. This behavior originates from the underlying Easy Markdown Editor and its `imageUploadFunction`, which handles only one file at a time. The relevant implementation is here:  
  [EasyMDE source](https://github.com/Ionaru/easy-markdown-editor/blob/fe680f09f66357243db0b2b4564cdf0dc6d20a1b/src/js/easymde.js#L1956).  

  On the Blazorise side, event handling follows a similar pattern to `FileEdit`, which maintains the same event arguments (`Files` instead of `File`). That said, I think we should improve this, at least by documenting it. Additionally, I believe I can "group" the individual calls from js into a single call to .NET.

## Unreachable Files
The current fixed implementation follows one flow—it keeps files in memory "forever". This is by design because Blazorise cannot effectively predict when a user will call `WriteToStream`.  

This issue could be mitigated by deleting the `blob` from the dictionary (where the files persist "forever") once it has been written to a stream. However, doing so would prevent the user from writing to the stream again or later:

```
private async Task ImageUploadChanged(FileChangedEventArgs events)
{ 
    fileForLater = events.Files.First();
}

IFileEntry? fileForLater;

async Task SomeMethodThatIsCalledLater()
{
    if (fileForLater is null) return; 
    // C# file reference is not null, but the actual blob state in JS is unknown 
    using var memoryStream = new MemoryStream();
    await fileForLater.WriteToStreamAsync(memoryStream);
}
```
The implementation problem lies in the fact that the C# reference is not synchronized with its content on the js side, and there is currently no way to verify this from C# code.

(also) for this reason, the documentation includes a `try-catch` block around this logic. However, I think it would be better to streamline the solution with something like `TryToWriteToStreamAsync`...

Is it worth improving that @stsrki ?

### DIFF
--- a/Documentation/Blazorise.Docs/Pages/Docs/Extensions/Markdown/MarkdownPage.razor
+++ b/Documentation/Blazorise.Docs/Pages/Docs/Extensions/Markdown/MarkdownPage.razor
@@ -58,7 +58,9 @@
 
 <DocsPageSection>
     <DocsPageSectionHeader Title="Upload image">
-        Uploading image has a similar API to our FileEdit component and it is fairly simple to do.
+        Uploading an image has a similar API to our <Code>FileEdit</Code> component and is fairly simple to use.
+        <Strong>Note that</Strong> the events related to file uploads fire separately for each file instead of grouping them.
+        This behavior comes from the underlying Easy Markdown Editor, which processes files individually.
     </DocsPageSectionHeader>
     <DocsPageSectionContent Outlined FullWidth>
         <MarkdownUploadImageExample />
--- a/Source/Extensions/Blazorise.Markdown/wwwroot/markdown.js
+++ b/Source/Extensions/Blazorise.Markdown/wwwroot/markdown.js
@@ -80,7 +80,9 @@ export function initialize(dotNetObjectRef, element, elementId, options) {
             imageUploadNotifier.onError = onError;
 
             // Reduce to purely serializable data, plus build an index by ID
-            element._blazorFilesById = {};
+            if (element._blazorFilesById == null) {
+                element._blazorFilesById = {};
+            }
 
             var fileEntry = {
                 id: ++nextFileId,


stsrki:
The fix seems to be working fine. That's good.

Regarding the suggested changes. If you think you could improve it and make it without introducing any new APIs while still making it sync between C# and JS reference, then you're free to go. I will open new ticket for that.

Anyways, before merging, please update the docs and mention the current behavior and the reasoning behind it so that people are aware of it.

stsrki (approved):

Issue #5857, opened 2024-11-19
1.7 Support Change Log


stsrki:
## 1.7.8

- #6145

## 1.7.7

- #6069
- #6079
- #3162
- #6101
- #5955
- #6099
- #5909
- #5935
- #5852
- #6117
- #6120
- #6123
- #6124
- #6127
- #6121

## 1.7.6

- #5998
- #4317
- #6006
- #6004
- #6000
- #6022
- #5612
- #5976
- #6025
- #6029
- #6038
- #6047
- #6040
- #6018
- #5994
- #6068
- #6056
- #6077

### Contributions

- #6008
- #6014
- #6032

## 1.7.5

- #4932
- #5970
- #5961
- #5981
- #5881
- #5985
- #5919
- #5977
- #5979
- #5978

## 1.7.4

- #4826
- #5928
- #4745
- #5952
- #5953

## 1.7.3

- #5899
- #5925
- #5930
- #5911
- #5712
- #5928

## 1.7.2

- #5889
- #5770
- #5790
- #5888
- #5905
- #5907
- #5904
- #5216
- #5847

## 1.7.1

- #5850
- #5875
- #5876
- #5858
- #5607

PR #5980, opened 2025-02-21
Markdown: merge uploaded files callback

### LINKED ISSUE #5977: Markdown: merge uploaded files callback
The fix seems to be working fine. That's good.

Regarding the suggested changes. If you think you could improve it and make it without introducing any new APIs while still making it sync between C# and JS reference, then you're free to go. I will open new ticket for that.

Anyways, before merging, please update the docs and mention the current behavior and the reasoning behind it so that people are aware of it.

_Originally posted by @stsrki in https://github.com/Megabit/Blazorise/issues/5975#issuecomment-2668133661_

tesar-tech:
## Description

Closes #5977 

- When user uploads multiple files at once they got all in one batch to the `ImageUploadChanged`. This is done by waiting for the `imageUploadFunction` in js for 100ms. 
- Also fixes issues with deleted reference in js, while keeping the reference in c#. This was also an issue with `FileEdit`. Implementation leverages the finalizer of `FileEntry` object.  (the sub-issue 2)
- Also fixes the sub-issue (1)

### DIFF
--- a/Source/Blazorise/Utilities/IO/FileEntry.cs
+++ b/Source/Blazorise/Utilities/IO/FileEntry.cs
@@ -57,6 +57,11 @@ public class FileEntry : IFileEntry
 
         return Task.CompletedTask;
     }
+    
+    ~FileEntry()
+    {
+        Owner.RemoveFileEntry( Id );
+    }
 
     #endregion
 
--- a/Source/Blazorise/wwwroot/fileEdit.js
+++ b/Source/Blazorise/wwwroot/fileEdit.js
@@ -75,7 +75,10 @@ export function open(element, elementId) {
 
 // Reduce to purely serializable data, plus build an index by ID
 function mapElementFilesToFileEntries(element) {
-    element._blazorFilesById = {};
+
+    if (element._blazorFilesById == null) {
+        element._blazorFilesById = {};
+    }
 
     let fileList = Array.prototype.map.call(element.files, function (file) {
         file.id = file.id ?? ++nextFileId;
--- a/Source/Extensions/Blazorise.Markdown/Markdown.razor.cs
+++ b/Source/Extensions/Blazorise.Markdown/Markdown.razor.cs
@@ -272,20 +272,26 @@ public partial class Markdown : BaseComponent,
     /// <summary>
     /// Notifies the component that file input value has changed. This method is intended for internal framework use only and should not be called directly by user code.
     /// </summary>
-    /// <param name="file">Changed file.</param>
+    /// <param name="fileEntries">Array of changed file</param>
     /// <returns>A task that represents the asynchronous operation.</returns>
     [JSInvokable]
-    public async Task NotifyImageUpload( FileEntry file )
+    public async Task NotifyImageUpload( FileEntry[] fileEntries )
     {
-        file.FileUploadEndedCallback = new();
+        foreach ( var file in fileEntries )
+        {
+            file.FileUploadEndedCallback = new();
 
-        // So that method invocations on the file can be dispatched back here
-        file.Owner = (IFileEntryOwner)(object)this;
-
+            // So that method invocations on the file can be dispatched back here
+            file.Owner = this;
+        }
         if ( ImageUploadChanged is not null )
-            await ImageUploadChanged.Invoke( new( file ) );
+            await ImageUploadChanged.Invoke( new( fileEntries ) );
 
-        file.FileUploadEndedCallback.SetResult();
+        foreach ( var file in fileEntries )
+        {            
+            file.FileUploadEndedCallback.SetResult();
+        }
+
         await InvokeAsync( StateHasChanged );
     }
 
--- a/Documentation/Blazorise.Docs/Pages/Docs/Extensions/Markdown/MarkdownPage.razor
+++ b/Documentation/Blazorise.Docs/Pages/Docs/Extensions/Markdown/MarkdownPage.razor
@@ -59,8 +59,6 @@
 <DocsPageSection>
     <DocsPageSectionHeader Title="Upload image">
         Uploading an image has a similar API to our <Code>FileEdit</Code> component and is fairly simple to use.
-        <Strong>Note that</Strong> the events related to file uploads fire separately for each file instead of grouping them.
-        This behavior comes from the underlying Easy Markdown Editor, which processes files individually.
     </DocsPageSectionHeader>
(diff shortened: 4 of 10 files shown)


Source/Blazorise/wwwroot/fileEdit.js
```diff
@@ -75,7 +75,10 @@ export function open(element, elementId) {
 
 // Reduce to purely serializable data, plus build an index by ID
 function mapElementFilesToFileEntries(element) {
-    element._blazorFilesById = {};
+
+    if (element._blazorFilesById == null) {
```
stsrki, line 79: I think it is best to check it as `if (!element._blazorFilesById)` instead of specifically for `null`.

Source/Blazorise/Utilities/IO/FileEntry.cs
```diff
@@ -57,6 +57,11 @@ public Task Cancel()
 
         return Task.CompletedTask;
     }
+    
+    ~FileEntry()
+    {
+        Owner.RemoveFileEntry( Id );
```
stsrki, line 63: Is this thread safe to call JS during the finalizer? What happens if JS interop is disposed of before finishing the operation?

Source/Blazorise/wwwroot/fileEdit.js
tesar-tech: yeah.. This is so alien to me, but makes "jsense"...

Source/Blazorise/Utilities/IO/FileEntry.cs
tesar-tech: So I iterated to a solution with try catch wrapper... I was trying something with weak reference, but it didn't work as expected. 
As long as the Owner is alive it should handle the js module.. So if the js interop will be disposed before finishing the operation it will just go through the catch block now.. The problem here is unpredictability of the moment when finalizer is called. It can be easily finalized as the last object when the Owner or js interop is long gone. In that case again - catch block. I don't think we have a different option here and it's not that bad imo.

The initial problem is the split nature of the `FileEntry` object, where half lives in js, and half lives in .net...
I can think of a solution where that's not the case, but it would require some api changes and changes of the Markdown event handling and arguments. 

`WriteToStreamIfAvailableAsync` instead of `WriteToStreamAsync` returning bool and with guaranteed access to the blob just within the `OnImageUploadChanged`. + other refactoring like forcing users to specify the `UploadUrl`...Because it creates a lots of workarounds around the normal flow of the uploads..

I guess it's not needed now. But if you were ever thinking about some improvements on the Markdown component internals - now I have it fresh in my memory.

Source/Blazorise/Utilities/IO/FileEntry.cs
```diff
@@ -57,6 +58,21 @@ public Task Cancel()
 
         return Task.CompletedTask;
     }
+    
+    ~FileEntry()
+    {
+        try
+        {
+            Owner.RemoveFileEntry( Id );//when the owner is alive this will work
```
stsrki, line 66: Still, the problem might be that we're calling asynchronous code from the finalizer, which is synchronous. This might lead to deadlock at some point. Can we not use finalizer at all for this? Can we JS dispose of fileentry when it is removed from a list somewhere else?

tesar-tech: I don't like using a finalizer either. If there were another non-API-breaking solution, I would use it, but I don't see one. Do you have any ideas?  

We can't reliably determine when the C# `FileEntry` object is no longer needed, so the finalizer is the only place to dispose of the JS resources as well.  

I don't think the current solution would cause deadlocks because the async code in the finalizer runs in a fire-and-forget fashion. It doesn't wait for anything, so there’s no point where it could get locked.

I would still prefer to break the api with  `WriteToStreamIfAvailableAsync` or `TryWriteToStreamAsync` and removing the blob from js after the `OnImageUploadChanged` has been invoked.

stsrki:
@tesar-tech, the name of your branch indicates 1.7, but you're targeting `master`. Mistake or intentional?

Source/Extensions/Blazorise.Markdown/Markdown.razor.cs
```diff
@@ -272,20 +272,34 @@ public Task NotifyCustomButtonClicked( string name, object value )
...
+        {
+            await ImageUploadChanged.Invoke( new( fileEntries ) );
+        }
+
+        foreach ( var file in fileEntries )
+        {
+            file.FileUploadEndedCallback.SetResult();
+        }
+
+        foreach ( var file in fileEntries )
```
stsrki, line 298: This seems to be working fine on my side. Not sure if there will be any sideefects now with your solution so please check.

Also, the finalizer was never reached for me in debug mode. Maybe we don't need it anymore?

tesar-tech: It is working, until user code hits the https://github.com/Megabit/Blazorise/issues/5977 sub-issue 2. If we can live with that than the finalizer isn't needed...

stsrki: We never meant to support that anyway. The file should be handled asap during the changed event.

tesar-tech: Yes, that's why I suggested the rename and the api change to `TryWriteToStreamAsync`, so users don't shoot their own leg with our "shoulds".

stsrki: Nah :)

OK, please remove finalizers from final versions. And add release notes before I merge it.

tesar-tech: Ok.

You mean release notes here https://github.com/Megabit/Blazorise/issues/5857 or creating a release note 175 .razor?

stsrki (approved):

Source/Extensions/Blazorise.Markdown/Markdown.razor.cs
stsrki: > It is working, until user code hits the #5977 sub-issue 2. If we can live with that than the finalizer isn't needed...

I should have done more tests. It seems we have now broken FilePicker. And FIlePicker works by saving references for later upload. So, I think we will need to go back to the finalizer or find an alternative for 

```
foreach ( var file in files )
{
    await file.Owner.RemoveFileEntry( file.Id );
}
```

PR #5995, opened 2025-03-02
PdfViewer: Introduce PrintRequested EventCallback

11bthornton:
## Description

Hello, thank you for taking the time to read this. Thought about raising this as an issue for discussion first but as implementation was simple, thought I'd just get straight to the point.

Currently, the PdfViewer component provides a callback as a parameter for the following events:

- `Loaded`
- `PageNumberChanged`
- `ScaleChanged`

However, it does not provide a callback that can be invoked when the document is requested to be printed.

This adds a `PrintRequested` Eventcallback parameter to `PdfViewer`.

### Context
We have an application that uses `PdfViewer` in combination with `PdfViewerToolbar`. As a result, the user uses the toolbar's print button rather than a custom button that invokes `Print` directly on a reference to the PdfViewer component. Since we're not doing the latter, we can't invoke some custom functionality at time of print request.

If there's a better way of achieving the desired effect, please let me know.

#### Discussion point:

This solution invokes the callback before the browser print window opens - so in my case, I am aware that the user may or may not actually go on to print / export the document. The other option could be a `Printed` EventCallback, but this would be a bit more complex to implement.

Example use:

```razor
@using Blazorise.PdfViewer

@page "/"

<PageTitle>Home</PageTitle>

<PdfViewerContainer>
    <PdfViewerToolbar />
    <PdfViewer PrintRequested="OnPrintRequested" Source="@($"data:application/pdf;base64,{base64String}")" />
</PdfViewerContainer>

@code {
    string base64String = "Omitted for brevity";

    private void OnPrintRequested()
    {
        Console.WriteLine("Print requested");
    }
}
```

## How Has This Been Tested?
Manually - correct me if I'm wrong but other events in this component do not have tests either. Happy to look into doing so if desired.

## Types of changes

## Checklist:

- [ ] I've added relevant tests. - **see section above**

### DIFF
--- a/Documentation/Blazorise.Docs/Pages/News/2025-03-10-release-notes-180.razor
+++ b/Documentation/Blazorise.Docs/Pages/News/2025-03-10-release-notes-180.razor
@@ -78,4 +78,12 @@
     The <Code>Cropper</Code> component has been enhanced with the <Code>ImageLoadingFailed</Code> event, which triggers when an image fails to load. This event provides a way to handle errors and display a custom message or image in place of the failed image.
 </Paragraph>
 
+<Heading Size="HeadingSize.Is3">
+    PdfViewer Component Enhancements
+</Heading>
+
+<Paragraph>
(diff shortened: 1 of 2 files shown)


stsrki (commented):
@11bthornton Thank you for the contribution. All is looking good to me. Can I just ask you to accept the CLA before I merge it? Thanks

11bthornton:

stsrki (approved):

PR #5999, opened 2025-03-04
DataGrid: Add New method overload

### LINKED ISSUE #5996: DataGrid: Functionality to clone a given row without changing the source Data
https://blazorise.com/support/issues/265#comment_2325

tesar-tech:
## Description

Closes #5996

Introduces the `AddItemToBatchChanges` method, enabling programmatic insertion of items into the batch edit collection when batch editing is enabled.  

It addresses the original request:  

> **"DataGrid: Functionality to clone a given row without changing the source data."**  

Rather than duplicating rows directly, this approach adds items to the batch collection, allowing them to be saved or discarded like any other batch edit change.  

Users can insert a new instance of `TItem` into the batch edit collection as shown:  

```csharp
<Button Clicked="OnAddNewRowClicked">Add new row</Button>

private async Task OnAddNewRowClicked(MouseEventArgs obj)
{
    var newEmployee = new Employee
    {
        FirstName = "John", 
        LastName = "Doe", 
        Email = "john.doe@example.com",
        Salary = 50000, 
        Id = 99999
    };

    await dataGridRef.AddItemToBatchChanges(newEmployee);
}
```

If users need to duplicate a row, they must create a new TItem, ensuring control over cloning behavior.

### DIFF
--- a/Documentation/Blazorise.Docs/Pages/News/2025-03-10-release-notes-180.razor
+++ b/Documentation/Blazorise.Docs/Pages/News/2025-03-10-release-notes-180.razor
@@ -86,4 +86,14 @@
     From our community member <Anchor To="https://github.com/11bthornton" Title="Link to 11bthornton">11bthornton</Anchor>, we have received a contribution that enhances the <Code>PdfViewer</Code> component. The component now supports the <Code>PrintRequested</Code> event callback, which triggers when the user requests to print the PDF document. It is a small but significant improvement that enhances the user experience.
 </Paragraph>
 
+<Heading Size="HeadingSize.Is3">
+    DataGrid: Programmatic Addition of Items to Batch Edit Collection
+</Heading>
+
+<Paragraph>
+    Introduced the <Code>AddItemToBatchItems</Code> method in DataGrid, enabling the programmatic addition of new items to the batch edit collection when batch editing is enabled.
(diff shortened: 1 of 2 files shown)


stsrki (commented):
I'm not sure about the new API. Seems redundant to me and it targets specifically the batch edit mode, which moves away from our current "way".

Can we try expanding on the existing `public Task New()`? We can add new overload `public Task New( TItem item )` and then based on the `BatchEdit` state we do the rest internally.

tesar-tech:
Yes, this approach works and also enables adding new items with predefined values when batch editing is disabled.

However, I'm not entirely satisfied with merging two distinct features into a single method (adding to batchItems and creating UI for adding new item). If we want to avoid letting the batch-related logic spill over into the main DataGrid class, we could encapsulate somehow (for example `dataGridRef.BatchStuff.AddBatchItem`).

stsrki:
> Yes, this approach works and also enables adding new items with predefined values when batch editing is disabled.
> 
> However, I'm not entirely satisfied with merging two distinct features into a single method (adding to batchItems and creating UI for adding new item). If we want to avoid letting the batch-related logic spill over into the main DataGrid class, we could encapsulate somehow (for example `dataGridRef.BatchStuff.AddBatchItem`).

We already do that in other methods. Like `Delete` for example. This way we only extend the same to the `New` method.

stsrki (approved):

PR #6014, opened 2025-03-12
DataGrid: Fix CellsEditableOn behavior for edit modes other than 'Cell'

### LINKED ISSUE #5976: Regression: CellsEditableOnNewCommand and CellsEditableOnEditCommand Ignored in DataGridEditMode.Popup
https://blazorise.com/support/issues/261

paulosdfigueiredo:
## Description

This PR fixes the behavior of the `CellValueIsEditable` property in the **DataGrid** component.

- **For edit modes other than `DataGridEditMode.Cell`**: it preserves the previous behavior, ensuring no unexpected changes.  
- **For `DataGridEditMode.Cell`**: it maintains the current functionality, allowing cell-level edits as intended.  

Closes #5976.

---

## How Has This Been Tested?

Tested using the demo pages with the following properties:  
- `CellsEditableOnNewCommand="false"`  
- `CellsEditableOnEditCommand="false"`  

Validated across different `DataGridEditMode` configurations to ensure correct behavior.

---

## Types of Changes

**Bug fix** (non-breaking change that resolves the issue)

### DIFF
--- a/Source/Extensions/Blazorise.DataGrid/DataGridColumn.cs
+++ b/Source/Extensions/Blazorise.DataGrid/DataGridColumn.cs
@@ -646,29 +646,13 @@ public partial class DataGridColumn<TItem> : BaseDataGridColumn<TItem>
     /// <summary>
     /// Returns true if the cell value is editable.
     /// </summary>
-    public bool CellValueIsEditable
-        => Editable && IsCellEditablePerCommand &&
-        (
-            ParentDataGrid.EditMode != DataGridEditMode.Cell 
-            || 
-            ( 
-                ParentDataGrid.EditMode == DataGridEditMode.Cell
-                &&
-                ( 
-                    ( ParentDataGrid.EditState == DataGridEditState.Edit && CellEditing ) 
-                    ||
-                    //We don't have data, let's keep a regular editable row for New.
-                    ParentDataGrid.EditState == DataGridEditState.New 
-                ) 
-            )
-        );
-
-    protected bool IsCellEditablePerCommand
-        => (
-                        ( CellsEditableOnNewCommand && ParentDataGrid.EditState == DataGridEditState.New )
-                        ||
-                        ( CellsEditableOnEditCommand && ParentDataGrid.EditState == DataGridEditState.Edit )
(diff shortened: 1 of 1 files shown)


paulosdfigueiredo:

tesar-tech:
Hey, thank you for this PR.

Just a question:
The "issue-point" in the old code was this part:

image

where the `ParentDataGrid.EditMode` being `Cell` would bypass the `IsCellEditablePerCommand` set by `CellsEditableOnNewCommand="false"`

is that right or is there some other decision path you changed?

tesar-tech (approved):
I merged the `CellValueIsEditable` and `IsCellEditablePerCommand` properties into one switch expression and now it appears significantly more readable.

This version emit the same results as your provided code and from my testing it does the right thing - 

allows the `CellValueIsEditable` only when `CellsEditableOn<Edit/New>Command` matches the edit state of the parent grid. With some extra conditions for edit state.

I also run some testing code and all the mismatches between old and current version are related exactly to that.

paulosdfigueiredo:
I believe the issue was that `ParentDataGrid.EditMode` **NOT** being `Cell` the whole expression short-circuited incorrectly and would bypass the `IsCellEditablePerCommand`.

This was my refactoring process:

## 1. Factor out common logic

- I noticed that `Editable` is checked once at the top, but `IsCellEditablePerCommand` is being repeated in both branches.
- Since `IsCellEditablePerCommand` applies in both cases, I pulled it out and combined it with `Editable`.
- This is also short-circuits earlier — if Editable or IsCellEditablePerCommand is false, it skips all the inner checks

---

## 2. Simplify the inner conditions

Next, I tackled the inner logic by separating the two key cases:

- **If `EditMode` is NOT "Cell"** → the cell is always editable.  (keep the behaviour before Rapid Editing existed)
- **If `EditMode` is "Cell"** → further checks are needed: (keep the behavior for Rapid Editing)
  - The cell is in **Edit** mode and is actively being edited, **or**
  - The cell is part of a **new row**.

paulosdfigueiredo:
Your commit LGTM

PR #6017, opened 2025-03-13
IFileEntryOwner: FileEntry to IFileEntry

### LINKED ISSUE #5991: Change IFileEntryOwner to work with interface types
Change 

```cs
Task WriteToStreamAsync( FileEntry fileEntry, Stream stream, CancellationToken cancellationToken = default );
```

into

```cs
Task WriteToStreamAsync( IFileEntry fileEntry, Stream stream, CancellationToken cancellationToken = default );
```

tesar-tech:
## Description

Closes #5991

Now the IFileEntryOwner uses the interface IFileEntry.

I went through the code, but I don't see a reason for keeping the interface. Only the one implementation exists and it worked just fine with it. 

What's the reason to keep the interface - is there any upcoming feature that would utilize it?

The failing test isn't related to this change. It was already in here: https://github.com/Megabit/Blazorise/pull/5993/checks . Any idea?

### DIFF
--- a/Source/Blazorise/Components/FileEdit/FileEdit.razor.cs
+++ b/Source/Blazorise/Components/FileEdit/FileEdit.razor.cs
@@ -382,7 +382,7 @@ public partial class FileEdit : BaseInputComponent<IFileEntry[]>, IFileEdit,
 
     /// <summary>
     /// Gets or sets the max chunk size when uploading the file.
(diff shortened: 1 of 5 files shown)


stsrki (approved):

stsrki:
> What's the reason to keep the interface - is there any upcoming feature that would utilize it?

Just to have a general type that could be shared for different implementations. We could change it but I don't find it worth it at this stage.

PR #6028, opened 2025-03-19
Chart: Plugin optimization

tesar-tech:
## Description

- Puts common logic and properties into BaseChartPlugin.
- Solves pain-points of async eventing between chart and plugins.
- So far the optimization is done of ChartZoom plugin only...
-  ...because of that the `BaseChartOfT` supports both - the old way and new way - and look relatively dirty.
- This optimization will reduce the necessary code in every plugin significantly, can solve the disposing on one place in initializing on one place. Potential bugs will be solved for all of the plugins at one place.
- The current state is perfectly applicable for `DataLables` and  `Annotations`, for `Streaming` and `Trendline`  it will need few adjustments.

---
There are some questions left:
- Why `Trendline` doesn't have the `DisposeAsync` implementation ?
- Why `Streaming` doesn't have the `OnParameterSet` implementation?

### DIFF
--- a/Source/Extensions/Blazorise.Charts/BaseChartOfT.cs
+++ b/Source/Extensions/Blazorise.Charts/BaseChartOfT.cs
@@ -1,8 +1,10 @@
 ﻿#region Using directives
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Threading.Tasks;
 using Blazorise.Extensions;
+using Blazorise.Modules;
 using Blazorise.Utilities;
 using Microsoft.AspNetCore.Components;
 using Microsoft.JSInterop;
@@ -22,6 +24,8 @@ public class BaseChart<TItem> : BaseComponent, IAsyncDisposable
     /// Occures after the chart has initialized successfully.
     /// </summary>
     public event EventHandler Initialized;
+    
+    public  List<BaseChartPlugin> ChartPlugins = [];
 
     /// <summary>
     /// List of registered plugins for this chart.
@@ -70,8 +74,11 @@ public class BaseChart<TItem> : BaseComponent, IAsyncDisposable
     /// <summary>
     /// Notifies the chart that it is being properly initialized.
     /// </summary>
-    protected void NotifyInitialized()
+    protected async Task NotifyInitialized()
     {
+        var tasks = ChartPlugins.Select(handler => handler.OnParentChartInitialized());
+        await Task.WhenAll(tasks); 
+        
         Initialized?.Invoke( this, EventArgs.Empty );
     }
 
@@ -86,6 +93,11 @@ public class BaseChart<TItem> : BaseComponent, IAsyncDisposable
             pluginNames.Add( pluginName );
         }
     }
+    
+    public void NotifyBasePluginInitialized( BaseChartPlugin plugin )
+    {
+       ChartPlugins.Add( plugin );
+    }
 
     /// <summary>
     /// Notifies the chart that it should remove the plugin.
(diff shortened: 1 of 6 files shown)


Source/Extensions/Blazorise.Charts/BaseChartOfT.cs
```diff
@@ -86,6 +93,11 @@ public void NotifyPluginInitialized( string pluginName )
             pluginNames.Add( pluginName );
         }
     }
+    
+    public void NotifyBasePluginInitialized( BaseChartPlugin plugin )
```
stsrki, line 97: Do we still need `NotifyBasePluginInitialized`? Can we merge it with `NotifyPluginInitialized`?

Source/Extensions/Blazorise.Charts/BaseChartPlugin.cs
```diff
@@ -0,0 +1,102 @@
+using System;
+using System.Threading.Tasks;
+using Blazorise.Extensions;
+using Blazorise.Modules;
+using Microsoft.AspNetCore.Components;
+using Microsoft.JSInterop;
+
+namespace Blazorise.Charts;
+
+public abstract class BaseChartPlugin : BaseComponent, IAsyncDisposable
```
stsrki, line 10: Should we name this simple as ChartPlugin? Drop the Base?

Source/Extensions/Blazorise.Charts/BaseChartPlugin.cs
```diff
@@ -0,0 +1,102 @@
...
+
+public abstract class BaseChartPlugin<TItem, TJSModule> : BaseChartPlugin
+where TJSModule : IBaseJSModule, IAsyncDisposable
+{
+
+    protected abstract TJSModule GetNewJsModule();
+
+    protected abstract Task AddPlugin();
+
+    protected abstract bool ProceedWithExecution( ParameterView parameterView );
```
stsrki, line 24: ProceedWithExecution name seems strange. Execution of what? Might need better name for future clarity.

stsrki (commented):
Codewise seems OK to me. There are quite a lot of changes. Not to mention new APIs. This is my major concern. because we could potentially break users code since I know personally a lot of them inherit our components and modify them. So this will break their code on a patch release.

So I think we should move this to 1.8 at least.

You can proceed with updating other plugins.

stsrki:
> There are some questions left:
> 
> * Why `Trendline` doesn't have the `DisposeAsync` implementation ?
> * Why `Streaming` doesn't have the `OnParameterSet` implementation?

To tell you the truth, I don't remember. Probably they didn't need this.

Source/Extensions/Blazorise.Charts/BaseChartOfT.cs
tesar-tech: When proceeding with the update, there will be just the `NotifyPluginInitialized( BaseChartPlugin plugin )`, this is just to keep the both "systems" running.

Source/Extensions/Blazorise.Charts/BaseChartPlugin.cs
tesar-tech: Yes. That's right.

Source/Extensions/Blazorise.Charts/BaseChartPlugin.cs
tesar-tech: It is weird, yeah, I will do something about it.

tesar-tech:
closing this in favor of https://github.com/Megabit/Blazorise/pull/6030 which targets master branch.

PR #6030, opened 2025-03-19
Chart: Plugin optimization

tesar-tech:
## Description

(originally in https://github.com/Megabit/Blazorise/pull/6028)

- Puts common logic and properties into ChartPlugin.
- Solves pain-points of async eventing between chart and plugins.
- Changes how plugins are connected to the BaseChartOfT
- Adds few comments about the connection

This optimization will reduce the necessary code in every plugin significantly, can solve the disposing on one place in initializing on one place. Potential bugs will be solved for all of the plugins at one place.

- `Trendline` plugin is different - I added comments inside the `ChartTrendline.razor.cs`

---
To answer the questions from previous pr:

>  Why Trendline doesn't have the `DisposeAsync` implementation ?

 I think it is just missing the same disposal logic as the other plugins, it's about the `await JSModule.SafeDisposeAsync();` that is missing.

> Why Streaming doesn't have the `OnParameterSet` implementation?

Part of the "refresh" strategy is done in `Refresh` method, but the parameters `Vertical` and `Options` wont't get there. They are used only in plugin initialization. Thus I think, the `OnParameterSet` should have its implementation. But maybe I am missing something.


stsrki (commented):
Looking good. I made some changes on the method names and structure of the code, but overall the principle is the same.

If there is any need update the docs. And also add release notes.

stsrki (approved):

PR #6032, opened 2025-03-19
IFileEntryOwner: Prevent ConstructorContainsNullParameterNames when AOT Compiled

### LINKED ISSUE #6029: [Bug]: FileEdit Throws System.NotSupportedException: ConstructorContainsNullParameterNames when AOT compiled
### Blazorise Version

1.7.5

Bootstrap5

https://github.com/mtbayley/BlazoriseFileEdit

1. Create new Blazor Web App from default template
2. Add FileEdit to page
3. Set the following on the Webassembly Project
```xml
<RunAOTCompilation>true</RunAOTCompilation>
<WasmEnableSIMD>false</WasmEnableSIMD>
```
4.Publish self contained
5. Select a file using file picker

### What is expected?

File is selected

Exception is thrown in browser
```
Uncaught (in promise) Error: Error: System.NotSupportedException: ConstructorContainsNullParameterNames, System.Threading.Tasks.Task`1[Microsoft.JSInterop.IJSObjectReference]
   at System.Text.Json.Serialization.Metadata.DefaultJsonTypeInfoResolver.GetTypeInfo(Type , JsonSerializerOptions )
   at System.Text.Json.JsonSerializerOptions.CachingContext.CreateCacheEntry(Type type, CachingContext context)
--- End of stack trace from previous location ---
   at System.Text.Json.Serialization.Metadata.DefaultJsonTypeInfoResolver.GetTypeInfo(Type , JsonSerializerOptions )
   at System.Text.Json.Serialization.Metadata.JsonTypeInfo.ConfigureProperties()
   at System.Text.Json.Serialization.Metadata.JsonTypeInfo.<EnsureConfigured>g__ConfigureSynchronized|172_0()
   at System.Text.Json.Serialization.Metadata.JsonTypeInfo.ConfigureProperties()
   at System.Text.Json.Serialization.Metadata.JsonTypeInfo.<EnsureConfigured>g__ConfigureSynchronized|172_0()
   at System.Text.Json.Serialization.Metadata.JsonTypeInfo.ConfigureProperties()
   at System.Text.Json.Serialization.Metadata.JsonTypeInfo.<EnsureConfigured>g__ConfigureSynchronized|172_0()
   at System.Text.Json.Serialization.Metadata.JsonTypeInfo.<EnsureConfigured>g__ConfigureSynchronized|172_0()
   at Microsoft.JSInterop.Infrastructure.DotNetDispatcher.ParseArguments(JSRuntime , String , String , Type[] )
   at Microsoft.JSInterop.Infrastructure.DotNetDispatcher.InvokeSynchronously(JSRuntime , DotNetInvocationInfo& , IDotNetObjectReference , String )
   at Microsoft.JSInterop.Infrastructure.DotNetDispatcher.BeginInvokeDotNet(JSRuntime , DotNetInvocationInfo , String )
    at http://localhost/_content/Blazorise/fileEdit.js?v=1.7.5.0:20:19
```

Microsoft Edge

### Any additional comments?

Related:
https://github.com/dotnet/aspnetcore/issues/58454

> The problem seems to be that System.Text.Json (S.T.J) does not know how to deserialize the `Lazy<Task<IJSObjectReference>> LazyFileTask`. Before you publish, S.T.J able to inspect `Lazy<T>`, determine it's not deserializable, and skip over the property. But after publishing and [trimming](https://learn.microsoft.com/en-us/aspnet/core/blazor/javascript-interoperability/?view=aspnetcore-8.0#object-serialization), S.T.J does not have enough metadata about `Lazy<T>` to determine if it's deserializable or not and errors out.
> 
> You can add `[JsonIgnore]` to `LazyFileTask` to tell S.T.J to always skip over it without checking if it's deserializable and get rid of the JS interop exception. 

 _Originally posted by @halter73 in [#58454](https://github.com/dotnet/aspnetcore/issues/58454#issuecomment-2418132388)_

I believe this bug was caused by the recent change in `IFileEntryOwner`
https://github.com/Megabit/Blazorise/pull/5980/files#diff-e50ac6ff04962affa021ea951cf80799f918c4b2312585922a1d00cd58be0024

Can we add [JsonIgnore] to these new properties? (I will test)

Reverting to 1.7.4 solves the problem for now

mtbayley:
## Description
Closes #6029

## How Has This Been Tested?

Tested with Demo App

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [X] The PR is submitted to the correct branch (`master` for dev, `rel-1.x` for maintenance).
- [X] My code follows the code style of this project.
- [X] I've added relevant tests.

### DIFF
--- a/Source/Blazorise/Utilities/IO/IFileEntryOwner.cs
+++ b/Source/Blazorise/Utilities/IO/IFileEntryOwner.cs
@@ -1,5 +1,6 @@
 ﻿#region Using directives
 using System.IO;
+using System.Text.Json.Serialization;
 using System.Threading;
 using System.Threading.Tasks;
 using Blazorise.Modules;
@@ -44,10 +45,12 @@ public interface IFileEntryOwner
     /// <summary>
     /// Element reference.
     /// </summary>
+    [JsonIgnore]
     ElementReference ElementRef { get; set; }
 
     /// <summary>
     /// JS file module.
-    /// </summary>    
+    /// </summary>
+    [JsonIgnore]
     IJSFileModule JSFileModule { get; set; }
 }
--- a/Source/Blazorise/Utilities/IO/IFileEntryOwner.cs
+++ b/Source/Blazorise/Utilities/IO/IFileEntryOwner.cs
@@ -27,4 +27,12 @@ public interface IFileEntryOwner
     /// <param name="cancellationToken">A cancellation token to signal the cancellation of streaming file data.</param>
     /// <returns>Returns the stream for the uploaded file entry.</returns>
     Stream OpenReadStream( FileEntry fileEntry, CancellationToken cancellationToken = default );
+
+    /// <summary>
+    /// Removes the file entry from js dictionary.
+    /// </summary>
+    /// <param name="fileEntry">Currently processed file entry.</param>
+    /// <param name="cancellationToken">A cancellation token to signal the cancellation of streaming file data.</param>
+    /// <returns>A task that represents the asynchronous operation.</returns>
+    Task RemoveFileEntry( IFileEntry fileEntry, CancellationToken cancellationToken = default );
 }
--- a/Source/Blazorise/Utilities/IO/IFileEntryOwner.cs
+++ b/Source/Blazorise/Utilities/IO/IFileEntryOwner.cs
@@ -45,7 +45,6 @@ public interface IFileEntryOwner
     /// <summary>
     /// Element reference.
     /// </summary>
-    [JsonIgnore]
     ElementReference ElementRef { get; set; }
 
     /// <summary>
--- a/Source/Blazorise/Utilities/IO/IFileEntryOwner.cs
+++ b/Source/Blazorise/Utilities/IO/IFileEntryOwner.cs
@@ -1,10 +1,7 @@
 ﻿#region Using directives
 using System.IO;
-using System.Text.Json.Serialization;
 using System.Threading;
 using System.Threading.Tasks;
-using Blazorise.Modules;
-using Microsoft.AspNetCore.Components;
 #endregion
 
 namespace Blazorise;
@@ -30,26 +27,4 @@ public interface IFileEntryOwner
     /// <param name="cancellationToken">A cancellation token to signal the cancellation of streaming file data.</param>
     /// <returns>Returns the stream for the uploaded file entry.</returns>
     Stream OpenReadStream( FileEntry fileEntry, CancellationToken cancellationToken = default );
-
-    /// <summary>
-    /// Removes the file entry from js dictionary.
-    /// </summary>
-    /// <param name="fileEntry">Currently processed file entry.</param>
-    /// <param name="cancellationToken">A cancellation token to signal the cancellation of streaming file data.</param>
-    /// <returns>A task that represents the asynchronous operation.</returns>
-    async Task RemoveFileEntry( IFileEntry fileEntry, CancellationToken cancellationToken = default )
-    {
-        await JSFileModule.RemoveFileEntry( ElementRef, fileEntry.Id );
-    }
-
-    /// <summary>
-    /// Element reference.
-    /// </summary>
-    ElementReference ElementRef { get; set; }
-
-    /// <summary>
-    /// JS file module.
-    /// </summary>
-    [JsonIgnore]
-    IJSFileModule JSFileModule { get; set; }
 }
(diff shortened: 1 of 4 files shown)


Source/Blazorise/Utilities/IO/IFileEntryOwner.cs
```diff
@@ -44,10 +45,12 @@ public interface IFileEntryOwner
     /// <summary>
     /// Element reference.
     /// </summary>
+    [JsonIgnore]
```
stsrki, line 48: I believe ElementRef should be able to be serialized. ElementReference is a known Blazor related type.

stsrki: Thank you!

tesar-tech:
To summarize how the bug works (for me and for future reference):

It starts here in js:

https://github.com/Megabit/Blazorise/blob/cd7958690b655cf6dfeaaef0c0201e9006db6407/Source/Blazorise/wwwroot/fileEdit.js#L19

which calls a dotnet method, implemented here:

https://github.com/Megabit/Blazorise/blob/cd7958690b655cf6dfeaaef0c0201e9006db6407/Source/Blazorise/Adapters/FileEditAdapter.cs#L30

`FileEntry` contains the property `IFileEntryOwner Owner`, which has this property:

https://github.com/Megabit/Blazorise/blob/cd7958690b655cf6dfeaaef0c0201e9006db6407/Source/Blazorise/Utilities/IO/IFileEntryOwner.cs#L52

This fails to be marked as "unable to be deserialized" due to AOT trimming.

Is that correct?

---

The bug fix works, yes. We can merge it. But I don't like the solution:

- There is `[JsonIgnore]` appearing out of nowhere within an interface that isn’t supposed to touch JSON at all.  
  - Someone will probably (rightfully) question why it’s there.
- How do we know no other properties are causing the same issue? (There’s probably a way, but you really don’t want to have to keep that in mind.)

---

I suggest creating `FileEntryDto` to clearly define what should and shouldn’t be transferred between js and dotnet.

Similar issues exist on the js side, where some "hacks" are used to avoid serialization:  
https://github.com/Megabit/Blazorise/blob/cd7958690b655cf6dfeaaef0c0201e9006db6407/Source/Blazorise/wwwroot/fileEdit.js#L107

So we might want to solve it on both sides.

Not part of this PR—we can move this to another ticket.

---

@mtbayley Can you add a simple comment explaining why the attribute is there?

btw thanks for this and for the referenced issue, I had no idea such issues might happen.

tesar-tech (changes requested):
just the comment

mtbayley:
Yes, you understand the issue correctly. I have run into a similar issue with my AuthenticationStateProvider in my work.

`[JsonIgnore]` tells `System.Text.Json` to not bother serializing/deserializing during the `NotifyChange` invocation.

If you have a better solution, I'm up for testing that as well. I agree, the error only occurs during runtime for a released AOT compiled app. It would be hard to catch during development or testing.

tesar-tech:
@stsrki what's the strategy?

- merge the fix, create solution for 1.8 (or another pr)
- merge the fix and leave it
- create the solution right here.

stsrki:
I think we don't even need it. Why can't we remove `JSFileModule` from `IFileEntryOwner`? And just call `await JSFileModule.RemoveFileEntry( ElementRef, fileEntry.Id );` directly from FileEdit and Markdown?

Or, change `RemoveFileEntry` as

```
async Task RemoveFileEntry( IJSFileModule module, IFileEntry fileEntry, CancellationToken cancellationToken = default )
{
    module.RemoveFileEntry( ElementRef, fileEntry.Id );
}
```

tesar-tech:
That will just eliminate the purpose of the interface and duplicate the code. The shared functionality is there, because it's the same for both of the classes. 

And it wont solve the "might happen for other types that should not be in dto" issue.

stsrki:
The thing is, we shouldn't have introduced a new API in the 1.7 release. We broke our own rules, basically. At the time, I had wrongly assumed it was not a big change, so it wouldn't hurt. So I would rather to duplicate code then introducing JSON attribute to the interface. 

We can optimize it in 1.8.

tesar-tech:
- It doesn't change the API - the `JSFileModule` is ignored either way. The attribute just confirm that on AOT.
- I suspect the `IFileEntryOwner` is more like the "private public" blazorise interface than really an interface for users to utilize.
- I really don't think someone uses the `IFileEntryOwner`.
- Even if someone does use the `IFileEntryOwner`, I don't really think someone is eager to serialize the `IJSFileModule`. Maybe it is even not possible.

The `[JsonIngore]` attribute won't affect anyone in a negative way.

> We can optimize it in 1.8.

sounds like you don't really care. If that's the case, I have no objections against the latest changes.

stsrki (approved):

PR #6034, opened 2025-03-22
DataGrid: optimize utility allocations

### LINKED ISSUE #5959: DataGrid: optimize utility allocations
We need to reuse constant utility values by creating read-only, or const members in code. The main problem with too many utilities is that for heavy components like DataGrid, it will allocate a new instance of each "fluent" method on each render. Usually, this is not a problem for simple use cases.

For example for:

- `Display="Display.None"` we can have `Display="Constants.DisplayNone"`
- `Margin="@Blazorise.Margin.Is1.FromEnd"` we can have `Margin="Constants.Margin1FromEnd"`

Where:

```cs
static class Constants
{
    public static readonly IFluentDisplay DisplayNone = Display.None;

    public static readonly IFluentSpacing Margin1FromEnd = Margin.Is1.FromEnd;

    // etc.
}
```

tesar-tech:
## Description

Closes #5959

### DIFF
--- a/Source/Extensions/Blazorise.DataGrid/Constants.cs
+++ b/Source/Extensions/Blazorise.DataGrid/Constants.cs
@@ -1,0 +1,32 @@
+namespace Blazorise.DataGrid;
+
+static class Constants
+{
+    // Display
+    internal static readonly IFluentDisplay DisplayInlineBlock = Display.InlineBlock;
+    internal static readonly IFluentDisplay DisplayNone = Display.None;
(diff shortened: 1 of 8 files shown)


stsrki (commented):
LGTM. I think it makes sense to move it to 1.8 since it is beneficial and we will release 1.8 a lot sooner.

stsrki:
PS. I will review some tickets from 2.0 and see if we can do them for 1.8.

tesar-tech:
closing for branch change 

https://github.com/Megabit/Blazorise/pull/6036

PR #6043, opened 2025-03-27
PdfViewer: Add localization and download button

### LINKED ISSUE #5942: PdfViewer: add localization, and download button
https://blazorise.com/support/issues/246

tesar-tech:
## Description

Closes #5942

Localization: 
- I am not sure about the translations. It's from chatGPT. I can say the `cs` and `sk` is ok.

Download Button:
- It can be also done with a direct link, but that would break the normal eventing of pdfViewer. Current implementation is done the same way as printing was implemented. Also it easily allows to extend the download functionality by edits in pdf (if we decide this functionality is important)


stsrki (approved):
Looking good. Good job!

PR #6050, opened 2025-04-02
DataGrid: Fixes Toggleable for DetailRow

### LINKED ISSUE #6049: [Bug]: DataGrid: It's impossible to not to toggle `DetailRow` by clicking when DetailRow is open
### Blazorise Version

1.7.5 

None

```
@page "/employee-grid5"
@using Blazorise.DataGrid

<DataGrid TItem="Employee"
            @ref="dataGridRef"
          Data="@employeeList"
          DetailRowTrigger="@(e => {e.Toggleable = false;  return false; })"
          >
    <DataGridColumns>
        
        <DataGridColumn Field="@nameof(Employee.FirstName)" Caption="First Name" />
        <DataGridColumn Field="@nameof(Employee.LastName)" Caption="Last Name" />
    </DataGridColumns>
    <DetailRowTemplate>
      <Div>These are details...</Div>
    </DetailRowTemplate>
</DataGrid>

<Button Clicked="() => dataGridRef.ToggleDetailRow(employeeList.First(),true)"  Color="Color.Info">Toggle First Detail Row</Button>


@code {
    private List<Employee> employeeList;
    private Employee selectedEmployee;
    DataGrid<Employee> dataGridRef;
    
    protected override async Task OnInitializedAsync()
    {
        employeeList = await EmployeeData.GetDataAsync();
    }

    // Mock data provider
    public static class EmployeeData
    {
        public static Task<List<Employee>> GetDataAsync()
        {
            var data = Enumerable.Range(1, 25)
                                 .Select(i => new Employee
                                              {
                                              Id = i,
                                              FirstName = $"FirstName{i}",
                                              LastName = $"LastName{i}",
                                              Salaries = Enumerable.Range(1, 30)
                                                                   .Select(month => new Salary
                                                                                    {
                                                                                    Date = new DateTime(2023, month % 12 + 1, 1),
                                                                                    Total = Random.Shared.Next(4000, 10000)
                                                                                    })
                                                                   .ToList()
                                              })
                                 .ToList();


            return Task.FromResult(data);
        }
    }

    // Employee class
    public class Employee
    {
        public int Id { get; set; }
        public string LastName { get; set; }
        public List<Salary> Salaries { get; set; } = new();
    }

    // Salary class
    public class Salary
    {
        public DateTime Date { get; set; }
        public decimal Total { get; set; }
    }
}

```

- run the code
- Click the button:
- 

Image

- The first DetailRow is open now.
- Click the first row - The detail row is closed now, but it should not be

### What is expected?

the `DetailRowTrigger="@(e => {e.Toggleable = false;  return false; })"`
should prevent DetailRow closing

The detail row is closing but it should not

Other

### Any additional comments?

_No response_

tesar-tech:
## Description

Closes #6049

- Checks Toggleable first
- Simplifies `SetRowDetail`
- Clarify comments

### DIFF
--- a/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
+++ b/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
@@ -1741,18 +1741,19 @@ public partial class DataGrid<TItem> : BaseDataGridComponent
             else if ( DetailRowTrigger is not null )
             {
                 var detailRowTriggerContext = new DetailRowTriggerEventArgs<TItem>( rowInfo.Item );
-                var detailRowTriggerResult = DetailRowTrigger( detailRowTriggerContext );
+                var shouldOpenWhenClosed = DetailRowTrigger( detailRowTriggerContext );
 
                 if ( !skipDetailRowTriggerType && detailRowTriggerType != detailRowTriggerContext.DetailRowTriggerType )
                     return;
-
-                rowInfo.SetRowDetail( detailRowTriggerResult, detailRowTriggerContext.Toggleable );
+                
+                if (detailRowTriggerContext.Toggleable)
+                    rowInfo.SetRowDetail(!rowInfo.HasDetailRow && shouldOpenWhenClosed);
 
                 if ( rowInfo.HasDetailRow && detailRowTriggerContext.Single )
                 {
-                    foreach ( var row in Rows.Where( x => !x.IsEqual( rowInfo ) ) )
+                    foreach ( var row in Rows.Where( x => !x.IsEqual( rowInfo ) ) )//close other detail rows
                     {
-                        row.SetRowDetail( false, false );
+                        row.SetRowDetail( false);
                     }
                 }
             }
@@ -3608,6 +3609,8 @@ public partial class DataGrid<TItem> : BaseDataGridComponent
 
     /// <summary>
     /// A trigger function used to handle the visibility of detail row.
+    /// Return true if you want the detail row to be visible when row is clicked and detail row is not visible.
+    /// Doesn't affect the closing behaviro (will be always close on click)
     /// </summary>
     [Parameter] public Func<DetailRowTriggerEventArgs<TItem>, bool> DetailRowTrigger { get; set; }
 
--- a/Source/Extensions/Blazorise.DataGrid/Models/DataGridRowInfo.cs
+++ b/Source/Extensions/Blazorise.DataGrid/Models/DataGridRowInfo.cs
@@ -46,10 +46,9 @@ public class DataGridRowInfo<TItem>
     /// <summary>
     /// Sets the detail row for the current row.
     /// </summary>
-    /// <param name="hasDetailRow">Indicates whether the detail row is present.</param>
-    /// <param name="toggleable">Indicates whether the detail row can be toggled.</param>
-    public void SetRowDetail( bool hasDetailRow, bool toggleable )
-        => this.hasDetailRow = ( toggleable && !this.hasDetailRow & hasDetailRow ) || ( !toggleable && hasDetailRow );
+    /// <param name="hasDetailRowNew">Open or close the detail row</param>
+    public void SetRowDetail( bool hasDetailRowNew )
+        => this.hasDetailRow = hasDetailRowNew;
 
     /// <summary>
     /// Toggles the visibility of the detail row.


tesar-tech:
The test doesn't pass, because it goes against the idea of `Toggleable`.

https://github.com/Megabit/Blazorise/blob/f6d75dcb1c3932b697319d960cdd6e31b0340762/Tests/Blazorise.Tests/Components/DataGridDetailRowComponentTest.cs#L105

I am not sure what was the intention for `Toggleable`, but I would suspect it should prevents toggling when set to `false`, but the test implies otherwise.

Is there something behind such behavior @stsrki ?

With `Toggleable` being ignored under some conditions it's impossible to fix the bug.

stsrki:
> The test doesn't pass, because it goes against the idea of `Toggleable`.
> 
> https://github.com/Megabit/Blazorise/blob/f6d75dcb1c3932b697319d960cdd6e31b0340762/Tests/Blazorise.Tests/Components/DataGridDetailRowComponentTest.cs#L105
> 
> I am not sure what was the intention for `Toggleable`, but I would suspect it should prevents toggling when set to `false`, but the test implies otherwise.
> 
> Is there something behind such behavior @stsrki ?
> 
> With `Toggleable` being ignored under some conditions it's impossible to fix the bug.

Based on a name, yeah, it should not be able to toggle it. But I was not the original creator, so I'm not sure.

@David-Moreira any chance you could remember this?

David-Moreira:
> > The test doesn't pass, because it goes against the idea of `Toggleable`.
> > https://github.com/Megabit/Blazorise/blob/f6d75dcb1c3932b697319d960cdd6e31b0340762/Tests/Blazorise.Tests/Components/DataGridDetailRowComponentTest.cs#L105
> > 
> > I am not sure what was the intention for `Toggleable`, but I would suspect it should prevents toggling when set to `false`, but the test implies otherwise.
> > Is there something behind such behavior @stsrki ?
> > With `Toggleable` being ignored under some conditions it's impossible to fix the bug.
> 
> Based on a name, yeah, it should not be able to toggle it. But I was not the original creator, so I'm not sure.
> 
> @David-Moreira any chance you could remember this?

Oh man, having to remember things developed centuries ago!!
If I remember correctly toggle as the name implies, makes it so the **DetailRow** can have a toggle mechanism regardless of the evaluated function for RowDetail in subsequent evaluations.

I.e
Row 1 evaluates that it should show rowdetail => FUNC returns true
As a user if **toggleable** is set to true, then I can click on it and even though FUNC still returns true, it will be set to false and hide the Row Detail instead just like a toggle.
image

Practical example : 
Let's say that you show Detail Rows for every customer that has ages over 30.
The func will evaluate and display every detail row. 
- If toggleable is set to false, whenever you try to click on a row with customer age > 30 
  - It should stay with detail if im not mistaken
- If toggleable is set to true, whenever you try to click on a row with customer age > 30 
 - it will collapse the detail as per the toggle
 
- If toggleable is set to false, whenever you try you try to click on a row with customer age < 30 
  - Nothing happens 
- If toggleable is set to true, whenever you try to click on a row with customer age < 30 
  - Even though the condition is not verified I believe it will show the detail as per the toggle

So as you can see, by enabling or disabling the flag you can have slight variations of the behaviour and that's why it exists and also why it's important to have nice examples of these variations in the docs, because the more flags you have the harder it is to understand the different outcomes. Also the thing with evolving features, trying to keep old behaviours in, while allowing new behaviours...

The behaviour I described is what I remember and might not be working exactly as I described, but I believe it should.

https://github.com/Megabit/Blazorise/pull/3008 - PR that introduces new mechanism for handling **DetailRow**
https://github.com/Megabit/Blazorise/pull/3330 - PR that seems to introduce the **Toggleable** feature

David-Moreira:
**PS:** I was just looking briefly at the code of this PR, it seems that we'll might be loosing the behaviour that I explained and this will be a breaking change, but I'm not sure please check

image
image

The new code seems to only call the SetRowDetail function ONLY if **Toggeable is set to true**, where before it would always call it, so I'm not 100% this logic is equivalent to the one before.

Although I like the idea of the change to simplify **SetRowDetail**. It's of note that this is a public api and I beleive it's also a breaking change for whoever that might be using that "special" logic.

stsrki:
@David-Moreira thank you very much for taking the time to help us!!

tesar-tech:
Yes, thank you for the intervention.

> If toggleable is set to false, whenever you try to click on a row with customer age > 30

>  -  It should stay with detail if im not mistaken

That's actually true and it answers the original question: "How to prevent the closing of the `DetailRow`".

Who would have thought that the answer would be "by returning `true` in `DetailRowTrigger`".

Which means this PR is no longer needed—at least not for fixing *the* bug.

What I don't like is the absolutely overengineered solution where one state is controlled in three different ways that interact with each other in unpredictable ways (unless you are the author of the code). 

- `Toggleable` means: `Toggleable_Sometimes_When_Some_Other_Conditions_Are_Met_And_Can_Be_Overridden_By_Force_Toggle`
- `DetailRowTrigger` means: `Yeah_I_Might_Open_The_Detail_Row_But_It_Dependz_Go_Read_Docs_Good_Luck_Lol`

The only one that makes sense is the forced toggle—it does exactly what the name/argument suggests.

I understand the overengineering came from stacking up features over time.

It would require some breaking changes, but I can imagine a simple solution where variable naming makes sense and actually does what it’s supposed to do.

I feel like I know the answer, but I’ll ask anyway: @stsrki Is it something worth fixing?

tesar-tech:
Hm, that might have sounded like I was "bashing" you for knowing "what’s actually going on." It's quite the opposite — I'm glad you shared your knowledge and took the time to explain it to us.

> If you can come with a simpler solution that is clearer and complies with every feature without breaking things for users

Ha ha - No, I don't think so. It would have to be a breaking change.

stsrki:
Converting to draft until we figure out what to do here. If it's a breaking change or behavior, then we will do it in the next release. If at all.

PR #6058, opened 2025-04-04
Highlighter: List of HighlightedTexts

### LINKED ISSUE #6053: Allow Highlighter component to accept a list of texts to be highlighted
Currently the Highlighter component only allows one string to be inputted as the text to be highlighted.

Create a property called HighlightedTexts which accepts a list of strings to be highlighted.

tesar-tech:
## Description

Closes #6053


Source/Blazorise/Components/Highlighter/Highlighter.razor.cs
```diff
@@ -23,31 +26,33 @@ public partial class Highlighter : BaseComponent
...
+        var escaped = highlightedTexts
+                      .Where(s => !string.IsNullOrWhiteSpace(s))
+                      .Select(Regex.Escape).ToList();
+
+        string pattern = untilNextBoundary
+          ? string.Join("|", escaped.Select(h => h + nextBoundary))
+          : string.Join("|", escaped);
+
+        
+        var regex = new Regex($"({pattern})", caseSensitive ? RegexOptions.None : RegexOptions.IgnoreCase);
```
stsrki, line 50: It's usually good to cache regex variable instead of recreating it every time.

stsrki (commented):
The API seems good to me. Need to add docs and release notes.

Source/Blazorise/Components/Highlighter/Highlighter.razor.cs
```diff
@@ -80,4 +90,7 @@ private static IEnumerable<string> GetFragments( string text, string highlighted
     [Parameter] public bool UntilNextBoundary { get; set; }
 
     #endregion
+    
+    record Fragment(string Text, bool IsMatch);
```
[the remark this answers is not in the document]
tesar-tech: There is quite a good reason for  `Fragment`.

It belongs to the class and will be used only there. That's why it's private and inside the class file. It's not a dto that is used project wise.

That also explains the naming - it's only used in Highlighter, so no benefit from naming it Highlighter again. (we have GetFragments method too).

Or why you think it should move?

Source/Blazorise/Components/Highlighter/Highlighter.razor.cs
tesar-tech: But how? It can change every time the method is called.

I can somehow create a dictionary with `(pattern,caseSensitive)` key, but I don't think it's worth it.

stsrki:
> There is also the `protected override void OnParametersSet()` consideration. Now the method is called twice, I guess it's because of the List being a parameter and Blazor does call the OnParamSet "just to be sure" when type is not a primitive one.
> 
> Yeah - I think this should be handled better. Probably with `SetParams` override???

Maybe override `SetParametersAsync` and compare new parameter to see if they have changed?

stsrki:
It's left to do:

- Try to optimize fragment builder to recalculate only when needed
- Add example in the docs
- Add release notes

Source/Blazorise/Components/Highlighter/Highlighter.razor.cs
```diff
@@ -35,11 +36,26 @@ record Fragment( string Text, bool IsMatch );
...
-        allHighlightedTexts = new List<string>( HighlightedTexts ?? Enumerable.Empty<string>() ) { HighlightedText };
-        fragments = GetFragments( Text, allHighlightedTexts, CaseSensitive, UntilNextBoundary, NextBoundary );
+        var changed =
+        ( parameters.TryGetValue<string>( nameof( Text ), out var text ) && Text != text )
+        || ( parameters.TryGetValue<string>( nameof( HighlightedText ), out var highlightedText ) && HighlightedText != highlightedText ) 
+        || ( parameters.TryGetValue<bool>( nameof( CaseSensitive ), out var caseSensitive ) && CaseSensitive != caseSensitive ) 
+        || ( parameters.TryGetValue<bool>( nameof( UntilNextBoundary ), out var untilNext ) && UntilNextBoundary != untilNext ) 
+        || ( parameters.TryGetValue<string>( nameof( NextBoundary ), out var nextBoundary ) && NextBoundary != nextBoundary )
+        || ( parameters.TryGetValue<IEnumerable<string>>( nameof( HighlightedTexts ), out var highlightedTexts ) 
+             && !(HighlightedTexts ?? Enumerable.Empty<string>()).SequenceEqual( highlightedTexts ?? Enumerable.Empty<string>() ) );
```
stsrki, line 49: We have `AreEqual` extension method to compare collections.

Documentation/Blazorise.Docs/Pages/Docs/Components/Highlighters/HighlighterPage.razor
```diff
@@ -27,6 +27,14 @@
     <DocsPageSectionSource Code="BasicHighlighterExample" />
 </DocsPageSection>
 
+<DocsPageSection>
+    <DocsPageSectionHeader Title="Multiple Highlights">
+    </DocsPageSectionHeader>
```
stsrki, line 32: Add a description of the example.

Source/Blazorise/Components/Highlighter/Highlighter.razor.cs
```diff
@@ -64,6 +93,11 @@ private static IEnumerable<string> GetFragments( string text, string highlighted
     /// </summary>
     [Parameter] public string HighlightedText { get; set; }
 
+    /// <summary>
+    /// Array of search terms to be highlighted.
+    /// </summary>
+    [Parameter] public List<string> HighlightedTexts { get; set; }
```
stsrki, line 99: I think it would be best to have it as an array

`public string[] HighlightedTexts`

That way we will force users to reasign HighlightedTexts parameter and we will not have problems with mutating of list and comparing reference types as we have in other component.

stsrki: They would be less likely to update an array as `myArray[1] = "something else";` I agree with you. The problem still occurs. 

I think the best way to say it in the documentation is that mutation is generally not supported for reference types and that users need to recreate a new list if they wish to change the parameter.

stsrki (approved):

PR #6067, opened 2025-04-10
DataGrid: Export: Architecture, csv exporter, external binary data exporter

### LINKED ISSUE #6042: DataGrid export to Excel
https://blazorise.com/support/issues/273#comment_2360

We need to see how complex this might be. One thing I am certain of is that we don't want it as part of the DataGrid project. So, the only other way is to create a separate Blazorise project.

We can make it so that we can convert any `Data` to Excel, XML, CSV, or whatever we can think of.

Now, the only problem is how we connect this "project" with DataGrid. Because users can interact with DataGrid export through the context menu or through the Export button. Or anything else.

I'm open to proposals.

tesar-tech:
## Description

Closes #6042

This pull request introduces a new **exporter architecture** for the DataGrid component.

As part of this architecture, I've also included two example supported formats
- `Csv` – a **text-based** format that is bundled in the `Blazorise.DataGrid` package.
- `Bson` – a **binary** format that is simple to work with and demonstrates external extensibility. It requires an additional dependency and is implemented in a separate project (`Blazorise.DataGrid.Exporters`). (This is where Excel exporters should go too).

While the new architecture is not small, it enables a powerful level of extensibility. Creating a CSV exporter, for example, could be done with much less code — but the purpose of this system is to ensure **external exporters can be plugged into DataGrid easily**.

With this architecture, no changes are needed in the `Blazorise.DataGrid` core codebase to support new exporters. A developer only needs to implement a single abstract class, focusing purely on the core export logic and options. Thanks to these design choices, the boilerplate for creating a new exporter (e.g. a potential JSON exporter) is reduced to a minimum (see the `BsonToFileExporter`).

While developing the architecture, I also re-evaluated what "export" should mean in the DataGrid context. It doesn't have to mean only exporting to a file — it can also mean exporting to a string (TextExporter), the clipboard, or any other target. Currently, we support 3 targets:
- **File**
- **Clipboard**
- **Text** (in-memory string)

This concept of "targets" opens up future extensibility such as:
- API-based export
- Printer
- Email or sharing
- Chat/Telegram integration

(Many of these could work with the simple `TextTarget` or minor variations of it, so I din't go much deeper)

---

## Nomenclature

- **Target**: The destination where exported DataGrid content ends up (File, Clipboard, Text).
- **Format**: The format in which data is encoded (CSV, BSON, XML, etc.).
- **Exporter**: A combination of format and target, e.g. `CsvToFileExporter`. This is why the `Csv` directory contains multiple files — it supports multiple targets.

---

## Getting data from DataGrid

This area still needs improvement, as my knowledge of the full DataGrid capabilities is limited. Currently, only columns with a defined `Field` are exported, and the value is either formatted (for text) or raw (for binary).

One potential future enhancement:  
A property like `BaseDataGridColumn.Func<object, object> ExportValue` could be used to customize how each column is exported — giving the developer full control per column.

The data passed to exporters is structured as a `List<List<object>>` or `List<List<string>>`, depending on whether the export is binary or text-based.

---

## Usage

As simple as:

```csharp
// Simple export to CSV with default options
dataGridRef.Export(new CsvToFileExporter());

// Copy to clipboard without headers
dataGridRef.Export(new CsvToClipboardExporter(new() { ExportHeader = false }));

// Export to CSV file with a custom file name and only the first 3 rows
dataGridRef.Export(
    new CsvToFileExporter(new() { FileNameNoExtension = "custom-csv-file" }),
    new DataGridExportOptions { NumberOfRows = 3 }
);

// Bson export (defined in an external project)
dataGridRef.Export(new BsonToFileExporter());
```

As shown above, only one public method in `DataGrid.razor.cs` is needed to support all export formats and targets.

---

- This PR does not include UI changes (e.g. export button). That could come in a future PR.
- `Blazorise.DataGrid.Exporters` is an external test project designed to validate the extensibility of the exporter system. It serves as a foundation for future projects where we'll support additional exporters that require external dependencies (e.g., Excel). It can be merged in this test form - it does nothing wrong.

### DIFF
--- a/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
+++ b/Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
@@ -3,7 +3,9 @@ using System;
 using System.Collections.Generic;
 using System.Collections.Specialized;
 using System.Dynamic;
+using System.IO;
 using System.Linq;
+using System.Reflection.Metadata;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
@@ -1802,6 +1804,63 @@ public partial class DataGrid<TItem> : BaseDataGridComponent
     /// <returns>A task that represents the asynchronous operation.</returns>
     public ValueTask ScrollToRow( int row )
         => tableRef.ScrollToRow( row );
+    
+    public async Task<TExportResult> Export<TExportResult, TCellValue>( IDataGridExporter< TExportResult, TCellValue> exporter, DataGridExportOptions options = null )
+    where TExportResult: IExportResult, new()
+    {
+        if ( exporter is IFileExportTarget fileExporter )
+        {
+            Console.WriteLine( $"I am here in fileexpor" );
+            fileExporter.JsDataGridModule = JSModule;
+        }
+        else if ( exporter is IClipboardExportTarget clipboardExporter )
+        {
+            clipboardExporter.JSUtilitiesModule = JSUtilitiesModule;
+        }
+        
+        var data = ExportData<TCellValue>(options);
+
+        TExportResult exportResult = await exporter.Export( data.exportedData, data.columnNames );
+        return exportResult;    
+    }
+    
+    private (List<List<TCellValue>> exportedData, List<string> columnNames) ExportData<TCellValue>( DataGridExportOptions options )
+    {
+        options ??= new();
+
+        // Filter columns (exclude Command, MultiSelect, and DisplayTemplate columns)
+        var columnsToExport = Columns
+                              .Where( column => column.ColumnType != DataGridColumnType.Command && column.ColumnType != DataGridColumnType.MultiSelect && column.Field != null && column.DisplayTemplate == null )
+                              .ToList();
+
+        var exportedData = new List<List<TCellValue>>();
+
+        var columnNames = columnsToExport.Select( c => c.Caption ).ToList();
+
+        var filteredDataToTake = options.NumberOfRows == -1? FilteredData : FilteredData.Take( options.NumberOfRows );
+        
+        bool isCellValueString= typeof( TCellValue) == typeof( string);
+        
+        foreach ( var item in filteredDataToTake )
+        {
+            var rowValues = new List<TCellValue>();
+
+            foreach ( var column in columnsToExport )
+            {
+                var cellValue = column.GetValue( item );
+                object formattedValue = isCellValueString
+                                        ? column.FormatDisplayValue( cellValue ) ?? ""
+                                        : cellValue;
+
+                rowValues.Add((TCellValue)formattedValue );
+            }
+
+            exportedData.Add( rowValues );
+        }
+
+        return (exportedData, columnNames);
+    }
+
 
     #endregion
 
--- a/Blazorise.sln
+++ b/Blazorise.sln
@@ -161,6 +161,8 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Blazorise.Weavers", "Source
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Blazorise.Weavers.Fody", "Source\SourceGenerators\Blazorise.Weavers.Fody\Blazorise.Weavers.Fody.csproj", "{FFC4A285-1A16-4DD4-8B8C-141521E405B0}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Blazorise.DataGrid.Exporters", "Source\Extensions\Blazorise.DataGrid.Exporters\Blazorise.DataGrid.Exporters.csproj", "{01A482C0-8DD8-4A9D-95FF-5CC2F71DB41B}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
@@ -423,6 +425,10 @@ Global
 		{FFC4A285-1A16-4DD4-8B8C-141521E405B0}.Debug|Any CPU.Build.0 = Debug|Any CPU
 		{FFC4A285-1A16-4DD4-8B8C-141521E405B0}.Release|Any CPU.ActiveCfg = Release|Any CPU
 		{FFC4A285-1A16-4DD4-8B8C-141521E405B0}.Release|Any CPU.Build.0 = Release|Any CPU
+		{01A482C0-8DD8-4A9D-95FF-5CC2F71DB41B}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{01A482C0-8DD8-4A9D-95FF-5CC2F71DB41B}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{01A482C0-8DD8-4A9D-95FF-5CC2F71DB41B}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{01A482C0-8DD8-4A9D-95FF-5CC2F71DB41B}.Release|Any CPU.Build.0 = Release|Any CPU
 	EndGlobalSection
 	GlobalSection(SolutionProperties) = preSolution
 		HideSolutionNode = FALSE
@@ -495,6 +501,7 @@ Global
 		{EAB7EC89-900A-4280-B24A-152B9DD2B503} = {9731051E-0AA7-411E-A76A-987854F034DA}
 		{BF5FFB8C-45AD-4875-BB01-2DA388890419} = {0538DB67-B4F3-4D00-B969-D3874A52E405}
 		{FFC4A285-1A16-4DD4-8B8C-141521E405B0} = {0538DB67-B4F3-4D00-B969-D3874A52E405}
+		{01A482C0-8DD8-4A9D-95FF-5CC2F71DB41B} = {9731051E-0AA7-411E-A76A-987854F034DA}
 	EndGlobalSection
 	GlobalSection(ExtensibilityGlobals) = postSolution
 		SolutionGuid = {205B3EA4-470F-45DA-911E-346AF7D0A9A5}
--- a/Source/Blazorise/Interfaces/Modules/IJSUtilitiesModule.cs
+++ b/Source/Blazorise/Interfaces/Modules/IJSUtilitiesModule.cs
@@ -165,6 +165,14 @@ public interface IJSUtilitiesModule : IBaseJSModule
     /// <param name="elementId">ID of the rendered element.</param>
     /// <returns>A task that represents the asynchronous operation.</returns>
     ValueTask CopyToClipboard( ElementReference elementRef, string elementId );
+    
+    /// <summary>
+    /// Copies the specified string content to the clipboard.
+    /// </summary>
+    /// <param name="stringToCopy">The string content to copy to the clipboard.</param>
+    /// <returns>A task that represents the asynchronous operation.</returns>
+    ValueTask CopyStringToClipboard(string stringToCopy);
+
 
     /// <summary>
     /// Writes a log message to the browser console.
--- a/Source/Blazorise/Modules/JSUtilitiesModule.cs
+++ b/Source/Blazorise/Modules/JSUtilitiesModule.cs
@@ -107,6 +107,10 @@ public class JSUtilitiesModule : BaseJSModule, IJSUtilitiesModule
     /// <inheritdoc/>
     public ValueTask CopyToClipboard( ElementReference elementRef, string elementId )
         => InvokeSafeVoidAsync( "copyToClipboard", elementRef, elementId );
+    
+    /// <inheritdoc/>
+    public ValueTask CopyStringToClipboard(string stringToCopy)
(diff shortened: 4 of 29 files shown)


stsrki:
There are many things to review here. But overall I like the approach that will help us to extend it, and with minimal changes to DataGrid.

I will properly review it later but for now I would like to focus more on high level architecture.

1. I would keep only basic interfaces and utilities in the `Blazorise.DataGrid.Exporters`, and move each of specific implementation into a separate sub-package. eg, `Blazorise.DataGrid.Exporters.Bson`, `Blazorise.DataGrid.Exporters.Excel`, etc. This is because these dependencies can be quite large to download and we should keep the separate so that users can choose what they want without downloading everything at runtime.
2. There are many new utilities and helpers in the DataGrid package. So I think we should move them to the core Blazorise package. Why? Because we might need them for other components. Like Markdown, Scheduler, etc. They can also have export at some point.
3. I like the idea for ExportValue. We should add it but in a separate PR. I want to first merge this in a working state.

tesar-tech:
Hmm. Ok, so now `Sources` are added alongside `Targets` and `Formats`. An `Exporter` is defined by these three components.

- The `DataGridExporter` no longer exists. You can now export any type of data to any kind of result. The `DataGrid.Export` method still requires `TabularData`, but it can also handle other formats—image, text, vector, or any other data structure.
- This makes it a universal data transfer mechanism:
  - Between Blazorise components (e.g., DataGrid to Chart)
  - To the outside world (e.g., DataGrid to Excel)
  - From the outside world to Blazorise components (e.g., CSV to Table)

---

Also, there's no place anymore for `Blazorise.DataGrid.Exporters.Bson`, because if it's intended for use with other components, it should be in `Blazorise.Exporters.Bson`. The same applies to `Csv`—which now has its own project. We can evaluate whether referencing it directly from `DataGrid` makes sense.

---

The common export logic has been moved to the `Blazorise` project. It might even be worth giving it its own project—`Blazorise.Export`.

Source/Blazorise/Export/ExportResult.cs
```diff
@@ -0,0 +1,20 @@
+namespace Blazorise.Export;
+
+/// <summary>
+/// Represents a generic result of a DataGrid export operation, indicating whether the export was successful.
+/// </summary>
+public interface IExportResult
+{
+    bool IsSuccess { get; init; }
```
stsrki, line 8: We usually don't use prefixes like `Is` for public APIs. There might be exception but we tend to not go with it. 

**Little background:** 

`Is` is usually used for internal getters when we need to have some logic behind the public parameter.

Lets say we have 


`[Parameter] public bool Disabled`

then somewhere else, we would have

`protected bool IsDisabled => ParentDropdown?.Disabled ?? Disabled;`

Source/Blazorise/Export/IExporter.cs
```diff
@@ -0,0 +1,42 @@
+using System.Threading.Tasks;
+
+namespace Blazorise.Export;
+
+/// <summary>
+/// Represents a DataGrid exporter that handles string-based (textual) cell values, such as CSV or plain text formats.
+/// </summary>
+public interface IExporterText<TExportResult, in TSourceData> : IExporter<TExportResult, TSourceData>
+where TExportResult : IExportResult
```
stsrki, line 9: Better naming would be `ITextExporter` and `IBinaryExporter`.

Source/Blazorise/Export/Sources/TabularSourceData.cs
```diff
@@ -0,0 +1,9 @@
+using System.Collections.Generic;
+
+namespace Blazorise.Export;
+
+public class TabularSourceData<TDataType>: IExportableData<TDataType>
+{
+    public List<List<TDataType>> Data { get; set; } = new();
+    public List<string> ColumnNames { get; set; } = new();
```
stsrki, line 8: Not sure if `ColumnNames` is a good name for generic exporters? Maybe `KeyNames`, or `FieldNames`?

stsrki:
> The common export logic has been moved to the `Blazorise` project. It might even be worth giving it its own project—`Blazorise.Export`.

Yeah, I think this might be a better approach, as Blazorise core might become too bloated.

So have:

- `Blazorise.Exporters` project with all the shared logic
- `Blazorise.Exporters.Bson`
- `Blazorise.Exporters.Csv`

Source/Blazorise/Export/ExportResult.cs
tesar-tech: This is heavily under my skin and I always forget...

Source/Blazorise/Export/Sources/TabularSourceData.cs
tesar-tech: I am also not sure. ColumnNames refer to the table as a structure which has columns and it has some names for them. Csv calls them "Headers".  KeyNames and FieldNames sounds too generic. Which is good if we wanna refer to the output structure (which can be anything).

So - do we wanna refer to the original structure or the output structure?

stsrki: Columns make sense when you speak in terms of tabular data, ot csv/excel files. But since we want to make it as generic as possible, then I think Field is a good approach. We already use Field on DataGrid, and soon on Scheduler. And it also represents the real data that we wish to read or process.

tesar-tech: But here we **are** talking about `TabularSourceData` class, it doesn't get more generic than that.

stsrki: 😅 ignore me.

stsrki (commented):
There are quite some inhertiance going around. I'm still trying to unwrap it in my head...

Source/Extensions/Blazorise.Exporters/Targets/File/IFileExportOptions.cs
```diff
@@ -0,0 +1,27 @@
...
+
+/// <summary>
+/// Defines configuration options for exporting content to a file, including file name, extension, and MIME type.
+/// </summary>
+public interface IFileExportOptions
+{
+    /// <summary>
+    /// Constructs the full file name by combining the file name without extension and the file extension.
+    /// </summary>
+    public string FileName => $"{FileNameNoExtension}.{FileExtension}";
```
stsrki, line 11: Does it really matters that we need too have separate props for `FileName`, `FileNameNoExtension`, `FileExtension`? I think just one, `FileName` is enough. If the user provides the filename without extension, it's OK. If the user user provides extension it's still good. Either way I don't think it matters.

We can always have a default extension applied internally if the user doesn't specify it.

Source/Extensions/Blazorise.DataGrid/Configuration/DataGridExportOptions.cs
```diff
@@ -0,0 +1,12 @@
...
+
+/// <summary>
+/// Specifies the number of rows to export from a data grid.
+/// </summary>
+public class DataGridExportOptions
+{
+    /// <summary>
+    /// -1 means all rows
+    /// </summary>
+    public int NumberOfRows { get; init; } = -1;
```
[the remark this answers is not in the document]
tesar-tech: That's a question. My idea was to keep it consistent with the "rich markup properties". Basically specify these in markup using column parameters - for example the `BaseDataGridColumn.Func<object, object> ExportValue` or `ExportHeader` , `SupressExport` or similar.

And keep the `DataGridExportOptions` only for the export itself (that cannot be specified per-column).
The `ExportValue` is an ultimate customization that would have to be there anyway. And keeping the "column export customization" inside `DataGridExportOptions` will duplicate the functionality in potentially confusing way.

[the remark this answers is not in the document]
stsrki: I don't think adding them in markup makes much since we are already having an `Export()` method that works in an "imperative" way. So it is only natural that we want to expand on `DataGridExportOptions` on what to actually export.

Source/Extensions/Blazorise.Exporters/Targets/File/IFileExportOptions.cs
stsrki: Yes, that is an option. But my taking is this, we only need one `FileName` property. And prevent users from confusing them. Should they use only one, all of them, or a combination of different properties? 

Better just give them one to use.

stsrki: Much better now. Thanks.

Source/Extensions/Blazorise.Exporters/IExportResult.cs
```diff
@@ -0,0 +1,12 @@
...
+
+/// <summary>
+/// Represents a generic result of an export operation, indicating whether the export was successful.
+/// </summary>
+public interface IExportResult
+{
+    /// <summary>
+    /// Indicates whether an operation was successful.
+    /// </summary>
+    bool Success { get; init; }
```
stsrki, line 11: Should we also have `Errors` if `Success == false`?

tesar-tech: Error handling isn't quite there yet.
Yeah, somehow yes.
It has to start with the js functions - for example when the clipboard isn't available or saving file fails.

tesar-tech: Should I improve on the error handling now?

Source/Extensions/Blazorise.DataGrid/Configuration/DataGridExportOptions.cs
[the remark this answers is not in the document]
tesar-tech: Well, I still don't think it's a good idea:

- `public string[] Fields { get; init; } // if null, then all fields are exported` 
  - If you want to export just 2 fields, it's fine. But if you want to exclude just 2 fields (out of, say, 15), you need to specify all the remaining 13 here.
  - When you add a column to the datagrid, you need to remember to add it to `Fields`.
  - Using the pattern "null for everything, empty for nothing" just feels wrong.
  - This will clash with `ExportValue` if implemented.

- `public bool UseCaptions { get; init; } // if true, the Caption will be used for Csv column names`
  - There's no way to customize this "per column". Essentially, it suffers from the same issues as `Fields`.

If you're sure you'll never want "per column export customization" (preferably by using `ExportedValue`), then it's fine. But if you think this might be useful in the future, it's probably better not to duplicate the functionality by allowing configuration through `ExportOptions` as well.

Having configuration in two places might seem like a step forward for users, but the opposite is true—there will be more code to maintain and more to explain, more prone to errors on both sides.

There's also the option for "per column customization" using `ExportOptions`, but that would require using `object`, strings for property names, or something else not type-safe.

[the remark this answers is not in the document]
stsrki: > That's a question. My idea was to keep it consistent with the "rich markup properties". Basically specify these in markup using column parameters - for example the BaseDataGridColumn.Func<object, object> ExportValue or ExportHeader , SupressExport or similar.

It's not a bad idea. But it is maybe a lot easier for the users to modify export options than to adjust parameters. Not sure. Both ways make sense. 🤔

It's just that since we already have options, why not use them fully?

---

> public string[] Fields { get; init; } // if null, then all fields are exported
> 
> If you want to export just 2 fields, it's fine. But if you want to exclude just 2 fields (out of, say, 15), you need to specify all the remaining 13 here.
> When you add a column to the datagrid, you need to remember to add it to Fields.
> Using the pattern "null for everything, empty for nothing" just feels wrong.
> This will clash with ExportValue if implemented.

Nothing wrong with that approach. We can either go with all by default, or with the "negate" logic where we could have `string[] IgnoreFields`. IMO the negate logic is always harder to grasp as it brings some mental overhead. 

So, `public string[] Fields { get; init; } // if null, then all fields are exported` is fine. 

We have a similar thing for autogenerated columns. When the user just defines `TItem`, all the model fields are used. If they wish to customize, they need to manually define DataGridColumn's.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -1835,12 +1835,17 @@ private TabularSourceData<TCellValue> ExportData<TCellValue>( DataGridExportOpti
...
+                                              column.Field != null && 
+                                              column.DisplayTemplate == null &&
+                                              ( options.Fields == null || options.Fields.Contains( column.Field ) ) )
                               .ToList();
-
         var exportedData = new List<List<TCellValue>>();
-
-        var columnNames = columnsToExport.Select( c => c.Caption ).ToList();
+        
+        var columnNames = (options.UseCaptions
```
stsrki, line 1846: I think we should fallback to field name in case when caption is not defined?

tesar-tech: Is't that too much magic? 

`UseCaptionsButUseFieldIfCaptionNotSpecified`

stsrki: Not really. We only need to document `UseCaptions` by writing XML comment. And mention the behavior in the docs for example, if there is any.

PR #6074, opened 2025-04-14
DataGrid: add FilterDefaultItem* parameters to DataGridSelectColumn

### LINKED ISSUE #6041: No Unselected Datagrid Option
Hello! Is there a way to have an unselected column pop up first? Now with a datagridselectcolumn it shows the first value, but not filters by it, and then I have to click the second value, then click back to it. I tried adding Data="" and concating my data with a generic "Select" item, which shows up, I gave it ID of 0 and -1 but 1 still is the default selected item, which causes the issue where you have to select it then select another item. How do we resolve this?

tesar-tech:
## Description

Closes #6041

Adds support for custom default item in filter dropdowns of `DataGridSelectColumn` via `FilterDefaultItem*` parameters.


Source/Extensions/Blazorise.DataGrid/_DataGridCellFilter.razor
```diff
@@ -3,23 +3,19 @@
...
     {
-        <_DataGridCellNumericEdit Column="@numericColumn" CellValue="@GetSearchValue()" CellValueChanged="@FilterChanged"></_DataGridCellNumericEdit>
-
+        <_DataGridCellNumericEdit Column="@numericColumn" CellValue="@GetSearchValue()" CellValueChanged="@FilterChanged" />
     }
     else if ( Column is DataGridSelectColumn<TItem> selectColumn )
     {
-        <_DataGridCellSelectEdit Column="@selectColumn" CellValue="@GetSearchValue()" CellValueChanged="@FilterChanged"></_DataGridCellSelectEdit>
-
+        <_DataGridCellSelectEdit Column="@selectColumn" CellValue="@GetSearchValue()" CellValueChanged="@FilterChanged" FilterDisplay="true" />
```
stsrki, line 10: Is ` FilterDisplay="true"` supposed to be `FilterDisplay="@selectColumn.FilterDisplay"`?

Source/Extensions/Blazorise.DataGrid/DataGridSelectColumn.cs
```diff
@@ -72,6 +72,26 @@ public override string FormatDisplayValue( object value )
...
+
+    /// <summary>
+    /// Value of the default item used in filter dropdowns.
+    /// </summary>
+    [Parameter] public object FilterDefaultItemValue { get; set; }
+
+    /// <summary>
+    /// If true, disables the default item in filter dropdowns.
+    /// </summary>
+    [Parameter] public bool FilterDefaultItemDisabled { get; set; } = false;
```
stsrki, line 88: I think `= false;` is not necesary. Not sure why we have it for the above parameters too. 🤔

Source/Extensions/Blazorise.DataGrid/_DataGridCellFilter.razor
tesar-tech: the `FilterDisplay` on `_DataGridCellSelectEdit` is a flag that is saying "I am being inside the filter row" `selectColumn.FilterDisplay` has different meaning.

I am not sure about the naming `IsInFilterRow`?

stsrki: Maybe something that clearly says in which area it might be used. In this it seems to me all other are basically placed on the filter area.

```
if ( Column is DataGridNumericColumn<TItem> numericColumn )
{
    <_DataGridCellNumericEdit Column="@numericColumn" CellValue="@GetSearchValue()" CellValueChanged="@FilterChanged" />
}
else if ( Column is DataGridSelectColumn<TItem> selectColumn )
{
    <_DataGridCellSelectEdit Column="@selectColumn" CellValue="@GetSearchValue()" CellValueChanged="@FilterChanged" FilterDisplay="true" />
}
else if ( Column is DataGridCheckColumn<TItem> checkColumn )
{
    <_DataGridCellCheckEdit Column="@checkColumn" CellValue="@GetSearchValue()" CellValueChanged="@FilterChanged" />
}
else if ( Column is DataGridDateColumn<TItem> dateColumn )
{
    <_DataGridCellDatePicker Column="@dateColumn" CellValue="@GetSearchValue()" CellValueChanged="@FilterChanged" />
}
else
{
    <TextEdit Text="@GetSearchValueAsString()" TextChanged="@((newValue) => FilterChanged(newValue))" />
}
```

For example, currently on the Scheduler component that I am working on, I have a `SchedulerDragArea` enum that has

- Header
- MonthlyView
- WeekView
- etc

And then I know what to do based on that information.

---

So we can have something like that, eg `DataGridArea` with

- Default, for main area
- Filter
- Footer
- Aggregate
- or anything else

tesar-tech: Like this?

We can add more when needed. 

We can also make abstract class for cells with this one param.

stsrki (approved):

PR #6080, opened 2025-04-16
MessageService: add Choose method

stsrki:
While working on the Scheduler component, I needed to have a confirmation dialog with multiple choices. So this PR solves it. 

It adds:

- `Choose` method that awaits the result from the user
- new `MessageOptionsChoice` that holds the information about the buttons on the dialog
- New examples in the docs
- Release notes

The only problem is that I had to change `TaskCompletionSource<bool> Callback` into `TaskCompletionSource<object> Callback`. I'm not sure if this parameter is even used by anyone, considering that `IMessageService` is used to give all information to the component.

### DIFF
--- a/Source/Blazorise/Components/MessageAlert/MessageAlert.razor.cs
+++ b/Source/Blazorise/Components/MessageAlert/MessageAlert.razor.cs
@@ -1,5 +1,7 @@
 ﻿#region Using directives
 using System;
+using System.Collections.Generic;
+using System.Linq;
 using System.Threading.Tasks;
 using Blazorise.Localization;
 using Microsoft.AspNetCore.Components;
@@ -101,6 +103,26 @@ public partial class MessageAlert : BaseComponent, IDisposable
     }
 
     /// <summary>
+    /// Handles the event when a choice button is clicked, hiding the modal and invoking callbacks as necessary.
+    /// </summary>
+    /// <param name="choice">Represents the button that was clicked, providing its key for further processing.</param>
+    /// <returns>Returns a task that represents the asynchronous operation of handling the button click.</returns>
+    protected Task OnChoiceClicked( MessageOptionsChoice choice )
+    {
+        return InvokeAsync( async () =>
+        {
+            await ModalRef.Hide();
+
+            if ( IsChoice && Callback is not null && !Callback.Task.IsCompleted )
+            {
+                await InvokeAsync( () => Callback.SetResult( choice.Key ) );
+            }
+
+            await Confirmed.InvokeAsync();
+        } );
+    }
+
+    /// <summary>
     /// Handles the <see cref="Modal"/> closing event.
     /// </summary>
     /// <param name="eventArgs">Provides the data for the modal closing event.</param>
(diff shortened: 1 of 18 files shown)


Source/Blazorise/Components/MessageAlert/MessageAlert.razor.cs
```diff
@@ -100,6 +102,26 @@ protected Task OnCancelClicked()
...
+    /// </summary>
+    /// <param name="choice">Represents the button that was clicked, providing its key for further processing.</param>
+    /// <returns>Returns a task that represents the asynchronous operation of handling the button click.</returns>
+    protected Task OnChoiceClicked( MessageOptionsChoice choice )
+    {
+        return InvokeAsync( async () =>
+        {
+            await ModalRef.Hide();
+
+            if ( IsChoice && Callback is not null && !Callback.Task.IsCompleted )
```
tesar-tech, line 116: I wonder if there is a case there the `Callback.Task.IsCompleted` would be null?

Source/Blazorise/Components/MessageAlert/MessageAlert.razor.cs
```diff
@@ -305,7 +339,7 @@ protected virtual ModalSize Size
     /// <summary>
     /// Occurs after the user respond with an action button.
     /// </summary>
-    [Parameter] public TaskCompletionSource<bool> Callback { get; set; }
+    [Parameter] public TaskCompletionSource<object> Callback { get; set; }
```
tesar-tech, line 342: I think people are using it for confirmation. 
What about making it generic instead of the object?

Source/Blazorise/Models/MessageOptionsChoice.cs
```diff
@@ -0,0 +1,79 @@
...
+        Color = color;
+    }
+
+    /// <summary>
+    /// Constructs a new instance with specified text and color for a message option.
+    /// </summary>
+    /// <param name="key">The unique key for the result of the choice.</param>
+    /// <param name="text">The content that will be displayed as the message option.</param>
+    /// <param name="color">The color that will be used to represent the message option visually.</param>
+    public MessageOptionsChoice( object key, string text, Color color )
```
tesar-tech, line 38: optional parameters here and keeping it in one method? (that way you will also add overload where only the key and text is specified)

Source/Blazorise/Components/MessageAlert/MessageAlert.razor.cs
stsrki: I have no idea how to make this part more generic TBH.

Source/Blazorise/Models/MessageOptionsChoice.cs
stsrki: I'm not sure if I understand. Can you elaborate?

Source/Blazorise/Components/MessageAlert/MessageAlert.razor.cs
stsrki: I've checked the internals of `TaskCompletionSource`, and Task is `readonly` and initialized in the constructor. I doubt it can ever be null.

Source/Blazorise/Models/MessageOptionsChoice.cs
tesar-tech: something like

```
public MessageOptionsChoice(object key = null, string text = "", Color color = default)
    {
```

stsrki: I see. Not a fan, TBH.

tesar-tech (approved):

PR #6084, opened 2025-04-19
Layout: Fix layout issues with fixed header

### LINKED ISSUE #6069: [Bug]: Layout Functionality + Documentation
### Blazorise Version

1.7.5

Bootstrap5

https://github.com/Alexander-JM-0224/BlazoriseTestANN

Pull the project and run, in the MainLayout.razor are a couple of examples.

### What is expected?

When reading your documentation there are 3 examples of Layouts that should be possible to use, but some (and especially the one we want to use) do not work as advertised.

Even in your Demo page.

We want for a frontend overhaul for our product to be able to change the layout. I looked at the documentation and "Sider with header on top" is exactly the layout we want. BUT when using the given example code the result is not satisfying. The Header goes underneath the Sider as seen in my included project, it is Example 1. (and also visible on your demo page)

Image

So there are some problems here.

In my included project there are 3 examples of the MainLayout:
- The first is the one that gets closest to the result we want, the "Sider with header on top". But this one adds the height of the header to the rest of the page and makes it scrolling, its close to what we want but without the scrolling. (this one i made myself, but am stuck with the header height being added to the page)
- The second one is your example code, that just does not do what we want/what the picture in the documentation says it should be. (Example code for "Sider with header on top")
- The third example is a working one, that is correct with the documentation, i just put that one in there to make shure i did not mess something else up. ("With sider")

Chrome, Safari, Brave

### Any additional comments?

I'm not exactly shure what is going wrong here, seeing as the "Header with sider on top" is also broken in your demo envoirment i expect this to be a bug.
But also the lacking documentation makes it a bit hard to come up with a solution that works as intended. 

So if it indeed is a bug, we would like it fixed offcourse, if it is my own lack of understanding of how the layout functionality should work i would really appreciate a quick explanation/(working)example.

Thanks in advance,
Alexander.

tesar-tech:
## Description

Closes #6069

The issue was a missing variable in the CSS. The only provider that sets it is Tailwind. I'm not sure why that is—or why the value is different (`88px` vs `60` on the others).

Using the variable from `blazorise.css` doesn't appear to be common practice here, but it seems the most logical solution to me.

An alternative solution might be what the FluentUI provider was previously doing. But that would mean duplicating the code across all  the providers.


stsrki (commented):
I would rather copy it accross all providers instead of having it core Blazorise.

Also, don't forget to run web compiler on all main SCSS files

image

tesar-tech:
Ok, now the same CSS is used across all providers.

I have done the rebuild. Unfortunately, I cannot use the WebCompiler extensions, as they don't support platforms other than Windows. I had some luck with this WebCompiler: https://github.com/excubo-ag/WebCompiler, but the output syntax is slightly different.

I've created a script that compiles the CSS for all providers while centralizing the settings in one place. It's ignored by Git on my side, but I can share it if needed. It requires install of the webcompiler globall tool, then iterates through all the providers.

stsrki (approved):

stsrki:
> Ok, now the same CSS is used across all providers.
> 
> I have done the rebuild. Unfortunately, I cannot use the WebCompiler extensions, as they don't support platforms other than Windows. I had some luck with this WebCompiler: https://github.com/excubo-ag/WebCompiler, but the output syntax is slightly different.
> 
> I've created a script that compiles the CSS for all providers while centralizing the settings in one place. It's ignored by Git on my side, but I can share it if needed. It requires install of the webcompiler globall tool, then iterates through all the providers.

It seems they have some problem with `watch` on the newer version, which I find problematic. The original VS extension works best, and I don't find it problematic that it comes bundled with Node in the background. As long as I don't have to fight with it.

Which brings us to another problem. You mostly work with Rider, so often I have to fix formatting, as you have probably already noticed. Is there anything in Rided that can make it work to behave the same as VS?

PR #6085, opened 2025-04-19
Dropdown: add `hide` middleware to floatingUi for hiding when scrolling

### LINKED ISSUE #6079: [Bug]: Should add `data-popper-reference-hidden` when dropdown is invisible
### Blazorise Version

1.7.3

Bootstrap5

https://popper.js.org/docs/v2/modifiers/hide/

```csharp
<Card style="overflow-x: auto; height: 200px; max-width: 400px;">
    <CardHeader>
        <CardTitle>Horizontal Scroll Example</CardTitle>
    </CardHeader>
    <CardBody>
        <div>
            <Dropdown>
                <DropdownToggle Color="Color.Primary">
                    Dropdown
                </DropdownToggle>
                <DropdownMenu>
                    <DropdownItem>Action</DropdownItem>
                    <DropdownDivider />
                    <DropdownItem>Another Action</DropdownItem>
                </DropdownMenu>
            </Dropdown>
            <div style="min-width: 2000px;">
                test
            </div>
        </div>
    </CardBody>
</Card>

```

Image

### What is expected?

Add `data-popper-reference-hidden` and `data-popper-escaped` to dropdown

Like this 

Image

No `data-popper-reference-hidden` and `data-popper-escaped` added

_No response_

### Any additional comments?

_No response_

tesar-tech:
## Description

Closes #6079

based on https://floating-ui.com/docs/hide#hide

### DIFF
--- a/Source/Blazorise/wwwroot/floatingUi.js
+++ b/Source/Blazorise/wwwroot/floatingUi.js
@@ -9,10 +9,17 @@ const DIRECTION_START = 'Start'
 export function createFloatingUiAutoUpdate(targetElement, menuElement, options) {
     //https://floating-ui.com/docs/autoUpdate
     return autoUpdate(targetElement, menuElement, () => {
+        const rect = targetElement.getBoundingClientRect();
+        const padding = {
+            top: rect.height * 0.7,  
+            bottom: rect.height * 0.7,
+            left: rect.width * 0.7,  
+            right: rect.width * 0.7
+        };
         computePosition(targetElement, menuElement, { //https://floating-ui.com/docs/computePosition#anchoring
             placement: getPlacementDirection(options.direction, options.rightAligned), //https://floating-ui.com/docs/computePosition#placement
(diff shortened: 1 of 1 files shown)


stsrki:
The problem seems to be still, although it is less problematic than before. See the screenshot:

image

tesar-tech:
It was hiding when the whole element disappeared.

Which is not exactly wrong - even inside the docs for the `hide` they have it like that.

There is `padding`  options that can "speed up" the hiding. I made it dynamic based on the size of the element. I don't think it is completely correct to hide right away (when the button is just slightly hidden). `0.7` looks like a good compromise.

[Screencast From 2025-04-19 22-02-31.webm](https://github.com/user-attachments/assets/b9661bc8-7299-462b-b3f5-68f6da8e62f7)

stsrki:
You're right. It makes more sense to show the menu as long as the parent element is visible. I have reverted it to previous code.

stsrki (approved):

PR #6087, opened 2025-04-20
Localization: Add culture-specific overloads for text localization APIs

### LINKED ISSUE #6015: Extend localizatrion APIs
C/P from support email:

In an earlier discussion we talked about two new features in the area of the `ITextLocalizer` we would like to have.  I just wanted to make sure I “officially” placed a request for those and I hope they can make it into the next release 😊 

**Feature Request 1**

Could we have a public property on the ITextLocalizer that exposes the CultureInfo of the `TextLocalizerService`

```cs
public CultureInfo CurrentCulture
{
     get => textLocalizerServer.CurrentCulture.
}
```
 

**Feature Request 2**

Could we have a second GetString method on the ITextLocalizer that allows to request a string for a specific culture (potentially different from the one currently set on the TextLocalizerService)

```cs
public string GetString( CultureInfo culture, string name, params object[] arguments )
```

tesar-tech:
## Description

Closes #6015

Adds `GetString`, `GetStrings` and relevant indexer.

The issue also asks about exposing current culture though `TextLocalizerService`, but it's already there as `SelectedCulture` property.

### DIFF
--- a/Documentation/Blazorise.Docs/Pages/News/2025-03-10-release-notes-180.razor
+++ b/Documentation/Blazorise.Docs/Pages/News/2025-03-10-release-notes-180.razor
@@ -170,4 +170,12 @@
     The <Code>MessageService</Code> now supports the <Code>Choices</Code> method, allowing you to display a list of options in a message dialog. This feature is useful for scenarios where you want to present users with multiple choices and handle their selection accordingly.
 </Paragraph>
 
+<Heading Size="HeadingSize.Is3">
+    Culture-specific overloads for text localization
+</Heading>
+
+<Paragraph>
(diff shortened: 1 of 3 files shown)


stsrki (commented):
Seems good to me. I only have doubts about the "Feature Request 1" from the ticket. So I have contacted our customer for more feedback about it.

stsrki:
> Seems good to me. I only have doubts about the "Feature Request 1" from the ticket. So I have contacted our customer for more feedback about it.

@tesar-tech feedback from the user

> The background for request 1 is:
> 
> In many code-places where we need a localizer we also need to know which culture the localizer currently uses. The reason is we have a masterdata table and we have to retriev a value according to the current culture (so a different value for “de” than for “en”).
> 
> This can easily be done by having an ITextLocalizerService injected. But for simplicity we though it would be nice to only have the ITextLocalizer injected and ask it for the current culture. So you only need one service having injected instead of two (in many places).
> 
>  
> 
> So nothing you could not do with the current version, just a simplification.

tesar-tech:
>  This can easily be done by having an ITextLocalizerService injected. But for simplicity we though it would be nice to only have the ITextLocalizer injected and ask it for the current culture. So you only need one service having injected instead of two (in many places).
> So nothing you could not do with the current version, just a simplification.

Oh - I see.

stsrki (approved):

PR #6089, opened 2025-04-23
DatePicker: Adds "Today" button

### LINKED ISSUE #6061: "Today" Button in DatePicker component.
Our company uses the DateEdit and DatePicker components troughout our product.

- DateEdit because it has the "Today" button to automatically go to the day of today.
- DatePicker because it supports disabled dates.

Now the moment has come we need a combination of both, we need disabled dates, but also need the today button.

Flatpickr seems to support this functionality so i was a bit surprised the DatePicker component does not, which would make it much more complete.

DateEdit with Today button ("Vandaag" means Today in dutch and gets translated automatically).
Image

DatePicker on which we wish to have the button.
Image

Do not know if it is better for you guys to add disabled dates to the DateEdit component, or Today to the DatePicker. But either would help our product ux a fair bit.

Is this something that is on the roadmap already, or something you are willing/capable to support?

Thanks in advance,
Alexander.

tesar-tech:
## Description

Closes #6061
image

Not the cleanest solution, but since flatpicker support is lacking we don't have much other options.

One option can be to change our flatpicker fork, but I am not sure if it's work it.

The current solution does what plugins do, just in our custom js.

### DIFF
--- a/Documentation/Blazorise.Docs/Pages/News/2025-03-10-release-notes-180.razor
+++ b/Documentation/Blazorise.Docs/Pages/News/2025-03-10-release-notes-180.razor
@@ -127,6 +127,14 @@
 </Paragraph>
 
 <Heading Size="HeadingSize.Is3">
+    Show "Today" Button on DatePicker
+</Heading>
+
+<Paragraph>
+    The <Code>DatePicker</Code> component now includes a <Code>ShowTodayButton</Code> parameter. When enabled, a "Today" button appears at the bottom of the calendar, allowing users to quickly jump to the current date.
+</Paragraph>
+
+<Heading Size="HeadingSize.Is3">
     Optimized Chart Plugins
 </Heading>
 
--- a/Source/Blazorise/Components/DatePicker/DatePicker.razor.cs
+++ b/Source/Blazorise/Components/DatePicker/DatePicker.razor.cs
@@ -58,6 +58,7 @@ public partial class DatePicker<TValue> : BaseTextInput<IReadOnlyList<TValue>>, 
             var placeholderChanged = parameters.TryGetValue( nameof( Placeholder ), out string paramPlaceholder ) && Placeholder != paramPlaceholder;
             var staticPickerChanged = parameters.TryGetValue( nameof( StaticPicker ), out bool paramStaticPicker ) && StaticPicker != paramStaticPicker;
             var showWeekNumbersChanged = parameters.TryGetValue( nameof( ShowWeekNumbers ), out bool paramShowWeekNumbers ) && ShowWeekNumbers != paramShowWeekNumbers;
+            var showTodayButtonChanged = parameters.TryGetValue( nameof( ShowTodayButton ), out bool paramShowTodayButton ) && ShowTodayButton != paramShowTodayButton;
 
             if ( dateChanged || datesChanged )
             {
@@ -89,7 +90,8 @@ public partial class DatePicker<TValue> : BaseTextInput<IReadOnlyList<TValue>>, 
                  || disableMobileChanged
                  || placeholderChanged
                  || staticPickerChanged
-                 || showWeekNumbersChanged )
+                 || showWeekNumbersChanged
+                 || showTodayButtonChanged )
             {
                 ExecuteAfterRender( async () => await JSModule.UpdateOptions( ElementRef, ElementId, new DatePickerUpdateJSOptions()
                 {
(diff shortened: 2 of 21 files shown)


stsrki:
It works, so that is good. Although I would like to make it part of our flatpickr fork if possible. This is a good candidate for the first work.

I don't really like how the button looks as a link. And it doesn't look like it belongs to the framework styling.

**BS4**

image

**FluentUI**

image

---

For reference, you can see the real fluent design at https://react.fluentui.dev/?path=/docs/compat-components-datepicker--docs

image

---

So, instead of having an inline style in JS

```css
<button type="button" class="today-button"
        style="color: #007bff; background-color: transparent; border: none; padding: 4px 8px; cursor: pointer; text-decoration: underline;">
  ${localizedToday}
</button>
```

I would do this:

1. Define only a class name on the button, `flatpickr-today-btn`.
2. For every one of our CSS providers,, define CSS styling in `_date-picker.scss`

But even before that, I would first implement it on our flatpickr fork. And make it resemble their default day item styling.

stsrki:
PS. Throwing in also material design for reference

image

stsrki:
https://github.com/jcsmorais/shortcut-buttons-flatpickr

tesar-tech:
The link buttons try to mimic what `DateEdit` (input type date) provides.

---

I added the "Today" feature rather as a hotfix.

About `flatpickr`... Are you sure it’s a good idea to start working on it? I feel like poor `flatpickr` has done its duty — it’s really old(dead) now, and there’s no point in beating a dead horse.

The question is whether you truly want to commit to keeping this horse alive — it contains a decade-old npm packages, not part of our tech stack, and it carries a lot of technical baggage. It's destined to fail sooner or later. Unfortunately, there's no direct alternative that matches `flatpickr` feature for feature (we can't just switch libraries).

Spending resources on this dead horse is unwise unless you're ready to commit long-term. And I don't event see point in that.

From my understanding, date picking isn’t rocket science (I have done calendar once). There are three main parts:

- Make the calendar behave as expected — show days, months, navigation buttons, etc.
- Position it correctly under the input (we already have `floatingUI` for this) and hide it at the appropriate time.
- Support the necessary parameters and features that `flatpickr` provided.

There aren’t that many features. I suggest we rewrite it properly in Blazor:

- We won’t need any js interop — there are no required APIs missing in Blazor for date-picking capabilities.
- We can style it based on the provider.
- We can extend functionality beyond just a "Today" button, e.g., via `RenderFragment ButtonBar`, which is currently impossible without some serious Blazor hackery.
- We eliminate one (rather heavy) dependency and its JS interop.
- We keep it fully within the Blazor ecosystem.

stsrki:
It's not such an easy thing to do in Blazor. That's why we used Flatpickr in the first place. For example, we have almost no control over how to properly handle input masks. There is no way to prevent invalid value natively in Blazor. So we would then had to fall back to JSInterop for every edge case. We actually had this in the past on one of the early Blazorise components. It was a nightmare to maintain after a while. So I removed it sometime in 0.9.x version, I think 🤔

And input mask is just one feature we need. There is also a display format, which is then again different from internal value. 

I mean, we could come close to what we need but why solve something that is already solved.

tesar-tech:
> but why solve something that is already solved

Because it no longer fits our overall needs. Instead of solving the input mask issue, we now carry the burden of keeping flatpickr "alive."

---
Anyway, it's done — our fork of flatpickr now has the `showTodayButton` option. I also made sure it works when the option changes from Blazor, and localization is functional as well.

Initialization, option changes, and localization changes all call the same function in js — all the logic is centralized.

I styled it using the `_flatpickr.scss` file, but I don't consider this the best approach. Better would be to reuse the `btn btn-primary` classes. Similarly, this will likely be the case for other providers. I suggest letting styling be governed by individual class providers instead.

Source/Blazorise.Bootstrap5/wwwroot/blazorise.bootstrap5.css
stsrki, line 1: On flatpickr repo, there is `src\style\flatpickr.styl`. Maybe we should define the default button and its container styles in there?

tesar-tech: The container is the same for all the providers, so yes. 

But the button will be style differently for different providers and we'll have to do this anyway. So it would serve as a backup or you think something else?

stsrki: Better to have default styles that we can just override on our side per need. I have already done some work on this.

stsrki:
> Better would be to reuse the `btn btn-primary` classes.

This is not that bad of an idea. That's why I have exposed todayButtonElement on the picker instance. Ba doing that we could pass the CSS class name for each of our providers and assign it in Blazorise.

stsrki (approved):

PR #6092, opened 2025-05-03
Scheduler component

### LINKED ISSUE #475: Scheduler Component
It would be great to have a Scheduler Component... I know that it must be hard to implement one, but maybe it could be done...

stsrki:
Closes #475

It's a large PR, so I will try to break it down.

The Scheduler component is made as a generic component, with an API similar to DataGrid, where `TItem` is meant to define the collection of items that is edited and handled by the component. The `Data` parameter is used to populate the component.

At minimum, the `TItem` must have `Title`, `Start`, and `End` fields named as such. However, they can be overridden if needed with Scheduler `*Field` parameters.

---

Scheduler has several supported view types.

- Day
- Week
- WorkWeek
- Month

Each of them has a separate parameter that affects it independently of other views. Such as Start and End time, work hours, etc.

---

Scheduler has a built-in editor that is enabled with `Editable` parameter. And when used it always checks if there exists a Field on TItem, that can be edited. If found, it will be possible to edit in the editor. 

If `Editable` is disabled, then users have to use callbacks to store and edit their Data items.

---

We also support recurring items. They work by defining `RFC-5545`(see https://icalendar.org/Home.html) rules on a base `TItem`. 

This `Item` will then show any recurring item in a view. However, these recurring items will **not be persisted** or exist in the `Data` collection or parameter. It is up to the user to handle real occurrences independent of the UI in their system. Only exceptions to the rules, like **deleted occurrences** and **rescheduled occurrences**, will be stored as part of the Item.


tesar-tech (commented):
I scanned through the code and I have some objections.  
Some parts are just begging to be abstracted into an interface or an abstract class.  
With some use of generics, a lot of code could be reduced and many issues avoided.

I can name the concrete examples, but I'll start with the most obvious one:  
Why isn't `TItem` sheltered by some interface?

```
public interface ISchedulerItem
{
    DateTime Start { get; set; }
    DateTime End { get; set; }
    //...
}

```

---

The component works without an issue. I have found several, rather ui/ux problems or suggestions:

**Overall:**
- Adding an event by dragging would be really useful.  

  [Screencast From 2025-05-06 22-50-32.webm](https://github.com/user-attachments/assets/7bec9c61-4a01-4b2e-805a-007a2b8ed2e3)

- Hover state is the same as the selected state, which isn't great UX:  
  image

- Sunday is labeled as `7` instead of `S`.

- Hover vs. selected state in the view selector is confusing.

- Title and description should have some kind of length limiter.

- Clicking the `All day` row does nothing. It would be useful if this triggered a new event popup with "all day" pre-selected.

- It's hard to set an event in the middle of another one—it jumps to the beginning.  

  [Screencast From 2025-05-06 23-09-41.webm](https://github.com/user-attachments/assets/3ac598fb-5763-4fed-a16f-a8c5416e2770)  

  (Well, I see I can if I place the cursor in the right empty space, but it's a really small target area.)

- When moved to the middle, the event renders incorrectly:  
  image

- Similarly, in the "All day" events section, it's very hard to move a multi-day event to the right.  

  [Screencast From 2025-05-06 23-15-05.webm](https://github.com/user-attachments/assets/3efa96a7-6067-43c6-8234-8c7c80a82c06)

---

**Month View:**
- When hovering over an event, the background should change. It's unclear whether clicking will create a new event or select an existing one.

- The top part of a day cell is not clickable or visually distinct, so when a day is full, clicking precisely between events becomes difficult.

- If there are more than three events, they overflow into the next day and become unclickable.

---

**Day View:**
- Why does the day end at 16:00? Not sure if that's a feature or a bug. I would expect the day to be scrollable.

---

**Week / Work Week View:**
- Cannot be dismissed by clicking outside:  
  image

stsrki:
> * Adding an event by dragging would be really useful.

Can you explain what you mean by this?

> * Hover state is the same as the selected state, which isn't great UX:

This is how Outline buttons work. And how it appears when it is active. It comes from bootstrap. Not sure if there is way around it.



> * Sunday is labeled as `7` instead of `S`.
> *

This works on my side. Maybe it is related to cultureinfo?



> * Hover vs. selected state in the view selector is confusing.
> *

Again, the same as previous. It comes from Bootstrap.

> * Clicking the `All day` row does nothing. It would be useful if this triggered a new event popup with "all day" pre-selected.
> *

It's not supposed to do anything. It's just a label to indicate the row meaning.



> * Cannot be dismissed by clicking outside:

This is by design for MessageProvider. We don't want them clicking on the backdrop to trigger an action. We want them to do the action explicitly. One could argue that if this is good for UX, but it has been working like this for years. I don't want to break this.

tesar-tech:
> > * Adding an event by dragging would be really useful.
> 
> Can you explain what you mean by this?
> 
Now you can add an event by clicking and selecting time from TimePicker. Faster ux is to let user drag the time right away. I would say it's even the expected experience. Just like Jira has it. Or this random js component. https://javascript.daypilot.org/demo/scheduler/scalehours.html

[Screencast From 2025-05-07 17-43-19.webm](https://github.com/user-attachments/assets/5a0cfe0f-3624-477d-be1d-e73bb51a2a7c)


> This works on my side. Maybe it is related to cultureinfo?
I have it fixed now

> > * Clicking the `All day` row does nothing. It would be useful if this triggered a new event popup with "all day" pre-selected.
> 
> It's not supposed to do anything. It's just a label to indicate the row meaning.

I don't mean the text, but the row. When you click in the "sheet" it adds an event, I would expect the same from the space above the sheet.

image

 
> > * Cannot be dismissed by clicking outside:
> 
> This is by design for MessageProvider. We don't want them clicking on the backdrop to trigger an action. We want them to do the action explicitly. One could argue that if this is good for UX, but it has been working like this for years. I don't want to break this.

That's a ux driven by technical "limitation", which I understand, but user rarely do.
When you click normal (non-recurring) event you can cancel it by clicking outside.
When you click recurring one, you cannot.  There is not a good (apart from the technicality) reason to do that - users are not forced to take action to continue. There is still the cancel button, thus it doesn't make sense to not allow the canceling through backdrop click.

tesar-tech:
> This is a deliberate choice. We don't want to force users to use explicit type. And with `ISchedulerItem` we would force it into him. We want them to have freedom to choose their types, and have their field names changed if necessary. And if in the future we wish to add support for ExpandoObject for example, with this approach we would be able.

I just can’t agree with that. Maybe it doesn’t matter, since you’re the boss, but I honestly can’t grasp the logic behind “not wanting to force users.”

Right now, you’re forcing users to juggle string values instead. Developers have to keep those in mind, manually check all the properties, and set the *Field values correctly. An interface would make that explicit. If they’re dealing with a fixed type they can’t change, they could still expose the required properties via simple getters, or use a decorator or view model. C# has solid tools for these scenarios.

In the current state, if there’s a mismatch—say, no property named "End"—the app just silently starts without showing the End editor. I can’t imagine that being the intended behavior. That’s not freedom, that’s a footgun. It should stop you at build time—that’s exactly what interfaces are for: to define a contract and make sure things line up.

---
Also the `*Field`s should be strongly type - but would operate differently, but I would say more smoothly. 

```
<Scheduler EndFunc="item => item.EndDateTime"
```
That would solve the extreme case where developers cannot touch existing type, decorate them or create viewModels out of them. All strongly typed an safe.

tesar-tech:
- It appears there is some trouble with displaying recurring events: Week vs WorkWeek (also missing in day view):

[Screencast From 2025-05-07 18-30-20.webm](https://github.com/user-attachments/assets/15c9ea3e-4a57-4d9c-ab8d-94dac075142e)

- Setting `-1` in for recurring count will remove the event. (when editing)

image

- This will result in not adding an event and not notify user about impossible conditions (recurring date is older than start)

image

- Repeat with different count than `1` on `Never` will result in not adding the event at all. The repeat field should be disabled when `Never` is used (or maybe even hidden).

image

- this will ignore the end date. Keeping the endDate with the count is bit confusing. It should be either ended by number of occurrences (count) or by the end date (???)

image

- Many of the languages don't have date translations (roman, danish, ...). Probably not related to the scheduler, but it is present there.

image

- Days are also lacking translation (here for german)

image

- If you highlight some text and drag an event, it does this 🤷.

[Screencast From 2025-05-07 19-03-58.webm](https://github.com/user-attachments/assets/434db16a-9519-4a49-801d-ab040b6101ae)

- The drag&drop doesn't work on touchscreens. I guess that's because the drag&drop html api doesn't support touch. But it is possible to do it with `ontouch` and similar events..

stsrki:
> > This is a deliberate choice. We don't want to force users to use explicit type. And with `ISchedulerItem` we would force it into him. We want them to have freedom to choose their types, and have their field names changed if necessary. And if in the future we wish to add support for ExpandoObject for example, with this approach we would be able.
> 
> I just can’t agree with that. Maybe it doesn’t matter, since you’re the boss, but I honestly can’t grasp the logic behind “not wanting to force users.”
> 
> Right now, you’re forcing users to juggle string values instead. Developers have to keep those in mind, manually check all the properties, and set the *Field values correctly. An interface would make that explicit. If they’re dealing with a fixed type they can’t change, they could still expose the required properties via simple getters, or use a decorator or view model. C# has solid tools for these scenarios.
> 
> In the current state, if there’s a mismatch—say, no property named "End"—the app just silently starts without showing the End editor. I can’t imagine that being the intended behavior. That’s not freedom, that’s a footgun. It should stop you at build time—that’s exactly what interfaces are for: to define a contract and make sure things line up.
> 
> Also the `*Field`s should be strongly type - they would operate differently, but I would say more smoothly.
> 
> ```
> <Scheduler EndFunc="item => item.EndDateTime"
> ```
> 
> That would solve the extreme case where developers cannot touch existing type, decorate them or create viewModels out of them. All strongly typed an safe.

It's not so simple. Yes, the interface would make it much easier for us and for the user. But, we would lose any flexibility for the future.

The thing is. Everyone(third-party vendors) is doing it in such a way that the fields are configured as string names. And it makes perfect sense for me. Even though it is somewhat harder to maintain. This way we will always be able to support almost any type.

For example

- If users request `DateTimeOffset` types instead of `DateTime`, we could easily handle it with a simple internal mapping changes.
- If users request support for `ExpandoObject`, or `DataTable`, or anything else that makes sense in enterprise worlds, and if we agree on such a change, then we have a way to make it. 

With a constraint on the generic type, we would lose any flexibility for the future. And I wish I could do it. It would make my world much easier with the implementation.

stsrki:
I believe most of the bugs are now fixed.

As for the other things you have mentioned:

> Adding an event by dragging. 

I like it, but would rather leave this for a future version. 

> If you highlight some text and drag an event, it does this 🤷.

Not sure what to think about this. It's kinda an invalid state of the UI. I have tried some other schdulers and the also behave the same. SO I would just leave it as it is.

> * It appears there is some trouble with displaying recurring events: Week vs WorkWeek (also missing in day view):

Still trying to figure this.

tesar-tech:
Just few maybe opinionated details:

- This is something about not-ideal user experience. The hover box (the black border) is for the place where my mouse is. It would be better if the hover box was shown where the center of the hover object is. Here is an "extreme" image where almost the full event is in right, but it still shows the hoverbox on left.

image

---
- Simple repeat event (by clicking to repeat row):
 
image

results in:
image

which is probably technically correct, but I somehow didn't expect the event to repeat forever. So maybe forever option next to the End and Count? Or forcing the End or count...? Maybe the `End` itself is confusing, because there are two `End`s so maybe explaining the one more? ("Repeat ends at: ")

---

- Maybe I mentioned already, not sure. But the `1` on never should not be there:

image
then it reads: "Repeat every 1 never"

---


Also - I didn't yet tested on other provider, rather waiting for this to be feature complete - or you have this coverd?

stsrki:
> * Simple repeat event (by clicking to repeat row):
> 
> image
> 
> results in: image
> 
> which is probably technically correct, but I somehow didn't expect the event to repeat forever. So maybe forever option next to the End and Count? Or forcing the End or count...? Maybe the `End` itself is confusing, because there are two `End`s so maybe explaining the one more? ("Repeat ends at: ")

I think we can just add nother label that explains it to the user what is going to happen

image

tesar-tech:
> image

I don't have this in the `New appointment`. Am I missing something?

stsrki:
> > image
> 
> I don't have this in the `New appointment`. Am I missing something?

I don't see an image. What are you asking about?

tesar-tech:
https://github.com/Megabit/Blazorise/pull/6092#issuecomment-2882971568
img from this comment

stsrki:
> [#6092 (comment)](https://github.com/Megabit/Blazorise/pull/6092#issuecomment-2882971568) img from this comment

It is a screenshot from the Teams app. I think it is a good approach for telling users what they have selected in a human-readable way.

tesar-tech:
> > [#6092 (comment)](https://github.com/Megabit/Blazorise/pull/6092#issuecomment-2882971568) img from this comment
> 
> It is a screenshot from the Teams app. I think it is a good approach for telling users what they have selected in a human-readable way.

Oh. Yeah, agree. I was under impression you already done it.

PR #6094, opened 2025-05-05
Chart: fix Cutout option

### LINKED ISSUE #6009: Chart: add Cutout to chart options
> Cutout is still not working 3 years later, how is this possible? 

 _Originally posted by @alen-resser in [#3977](https://github.com/Megabit/Blazorise/issues/3977#issuecomment-2708892106)_

stsrki:
Closes #6009

My first idea was to just add the JSON property name, but since this field is already wrong, it shouldn't affect anyone. So I have just renamed it. And to make it more flexible, I made it an `object` so that it can accept string and number as per chartjs specs.

### DIFF
--- a/Documentation/Blazorise.Docs/Pages/News/2025-03-10-release-notes-180.razor
+++ b/Documentation/Blazorise.Docs/Pages/News/2025-03-10-release-notes-180.razor
@@ -155,6 +155,18 @@
 </Paragraph>
 
 <Heading Size="HeadingSize.Is3">
+    Renamed Chart Options
+</Heading>
+
+<Paragraph>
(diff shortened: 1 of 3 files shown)


tesar-tech:
Don't you wanna keep the `CutoutPercentage` and just add `CutoutPixels`?

I know it doesn't align with chart js, but imo better than using `object` and forcing devs to append the `%` into the string when using percentage.

The only thing left to be solved is to prioritize these two options. When both are set, which one to send to chartjs? Maybe percentage? It doesn't really matter, but should be documented.

When just one of the `CutoutPercentage` and `CutoutPixels` is set, that one will be used. For `CutoutPixels` it will append the `%`.

wdyt?

stsrki:
> Don't you wanna keep the `CutoutPercentage` and just add `CutoutPixels`?
> 
> I know it doesn't align with chart js, but imo better than using `object` and forcing devs to append the `%` into the string when using percentage.
> 
> The only thing left to be solved is to prioritize these two options. When both are set, which one to send to chartjs? Maybe percentage? It doesn't really matter, but should be documented.
> 
> When just one of the `CutoutPercentage` and `CutoutPixels` is set, that one will be used. For `CutoutPixels` it will append the `%`.
> 
> wdyt?

I wouldn't go that route, as we would have to add special logic for determining which field to apply. It would be too specific for a very narrow use case.

Best to just go with one field that is mapped 1:1 with chartjs.

tesar-tech (approved):

PR #6109, opened 2025-05-16
ChartZoom: add zoom event and helper methods

### LINKED ISSUE #6078: Synchronized charts
I have two linecharts that are on the same X scale. I'm using the chart zoom plugin to allow zooming and panning, but to be useful the charts would need to be synchronized. So if the user zooms or pans one chart, the other chart will be updated to the same zoom and pan.

Add an API to allow synchronization between multiple charts. There seems to be a plugin already for chartjs that is able to do this:

https://github.com/AbelHeinsbroek/chartjs-plugin-crosshair

Unfortunately this hasn't been updated in about two years, so I'm not sure if it's a good idea to use it or not.

### Additional context
--

stsrki:
Closes #6078

- Added new `Zoomed` event callback that is used to listen for any zoom changes, and it receives the `zoomLevel` as an argument
- Added `GetZoomLevel`, `SetZoomLevel`, and `ResetZoomLevel` methods for more control over zoom functionality.


Source/Extensions/Blazorise.Charts.Zoom/ChartZoom.razor.cs
```diff
@@ -16,23 +17,139 @@ public partial class ChartZoom<TItem> : ChartPlugin<TItem, JSChartZoomModule>
 
     /// <inheritdoc/>
     protected override JSChartZoomModule CreatePluginJsModule()
-        => new( JSRuntime, VersionProvider, BlazoriseOptions );
+    {
+        DotNetObjectRef ??= DotNetObjectReference.Create( new ChartZoomAdapter<TItem>( this ) );
```
tesar-tech, line 21: What's the logic behind the `??=` operator here?  
It suggests a scenario where `CreatePluginJsModule` might be called multiple times. But that's not the case, right?  
If that were true, it would imply `OnParentChartInitialized` is called more than once. That would be an exceptional situation, so throwing an exception here should be the correct approach. (otherwise it would fail silently)

Source/Extensions/Blazorise.Charts.Zoom/ChartZoomAdapter.cs
```diff
@@ -0,0 +1,49 @@
...
+using System.Threading.Tasks;
+using Microsoft.JSInterop;
+#endregion
+
+namespace Blazorise.Charts.Zoom;
+
+/// <summary>
+/// Middleman class for the ChartZoom plugin.
+/// </summary>
+public class ChartZoomAdapter<TItem>
```
tesar-tech, line 12: Why the adapter?

The one method can simply go to the razor.cs file. (??)

Source/Extensions/Blazorise.Charts.Zoom/ChartZoom.razor.cs
```diff
@@ -16,23 +17,139 @@ public partial class ChartZoom<TItem> : ChartPlugin<TItem, JSChartZoomModule>
...
+    /// </summary>
+    /// <remarks>The zoom level represents the scaling factor applied to the chart. A value of <see
+    /// langword="1.0"/>  indicates no zoom, while values greater than <see langword="1.0"/> indicate zooming in, and
+    /// values  less than <see langword="1.0"/> indicate zooming out.</remarks>
+    /// <returns>A <see cref="ValueTask{TResult}"/> representing the asynchronous operation.  The result is a <see
+    /// cref="double"/> value indicating the current zoom level of the chart.  The default value is <see
+    /// langword="1.0"/> if the JavaScript module is not initialized.</returns>
+    public async ValueTask<double> GetZoomLevel()
+    {
+        if ( JSModule is null )
```
tesar-tech, line 60: The same - is there a normal workflow state where the `JSModule` would be null? Or is it exceptional state?

tesar-tech (commented):
All good, just small code questions.

Source/Extensions/Blazorise.Charts.Zoom/ChartZoom.razor.cs
stsrki: Not sure what you mean. `??=` is just a shorthand for null check and then assigning the new value to a var.

Source/Extensions/Blazorise.Charts.Zoom/ChartZoomAdapter.cs
stsrki: It can, yes. But I wanted to prevent the public methods that are meant for internal use from leaking to the users. Remember, we have some public methods in other parts of the code that have a comment "internal use only". This is to prevent that.

Source/Extensions/Blazorise.Charts.Zoom/ChartZoom.razor.cs
stsrki: Wouldn't that break the pre-rendering since initialize is called twice in SSR?

tesar-tech (approved):

PR #6124, opened 2025-05-27
Video: bundle JS and CSS files

stsrki:
Based on the comment from https://github.com/vidstack/player/discussions/1560#discussioncomment-13271325

I have made changes to how we import vidstack files. Now everything is prebundled, and there is no need to manually define chunks.

This PR:

- removes all chunks
- removes vidstack and plyr files
- adds new bundled files

I have also created a new project for future reference https://github.com/stsrki/vidstack-bundle


tesar-tech:
I tested the video player and nothing went wrong.
I am glad this is done as it removes a lot from the repo.

Maybe I would consider to put the build "project" under the `build` folder in `Blazorise.Video` as it will be used with every js update - it is part of blazorise build step time to time.

stsrki:
> Maybe I would consider to put the build "project" under the `build` folder in `Blazorise.Video` as it will be used with every js update - it is part of blazorise build step time to time.

I'm not a fan of bringing the JS build process into the Blazorise codebase. Better to keep it out.

PR #6126, opened 2025-05-27
DataGrid: columns reordering

### LINKED ISSUE #5708: Datagrid columns reorder
Hi :)

Given a datagrid we want to give users the possibility to set columns in the order they want and save it as preference somewhere.

Most common solution is a drag&drop on columns but right now it is not possible with blazorise.

So 
- either giving drag&drop events on columns so we can manage ourself actions while drag&drop
- or getting the datagrid manage completely the reorder feature and have the possibility to 2way bind a DisplayOrders dictionnary to handle columns order ?

Any plan for this ?
Having to develop our own datagrid just because of this feature sounds not good 
Thank you :)

stsrki:
Closes #5708

First draft or columns reordering. Instead of doing it on the column picker I decided to first try on the table directly. The change was fairly easy, so apart from adding `ondrop` event, I only had to add some additional properites to store the state of the grid.


Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -787,6 +786,37 @@ private Task OnColumnDragEnded( DragEventArgs e )
...
+        var orderedColumns = Columns.OrderBy( c => c.DisplayOrder ).ToList();
+
+        // If any column has DisplayOrder of 0, reinitialize all
+        if ( orderedColumns.Any( c => c.DisplayOrder == 0 ) )
+        {
+            for ( int i = 0; i < orderedColumns.Count; i++ )
+                orderedColumns[i].DisplayOrder = i + 1;
+        }
+
+        orderedColumns = Columns.OrderBy( c => c.DisplayOrder ).ToList();
```
tesar-tech, line 803: I think this is not needed.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -787,6 +786,37 @@ private Task OnColumnDragEnded( DragEventArgs e )
         return Task.CompletedTask;
     }
 
+    private Task OnColumnDropped( DataGridColumn<TItem> columnBeingDropped )
```
tesar-tech, line 789: `columnBeingDropped`
Is it the one that is dragged and now it's dropping? or is it the one that is registering the drop event on itself?

I would call them `columnTarget` and `collumnDragging`. or somethin similar.

Source/Extensions/Blazorise.DataGrid/_DataGridPagination.razor
```diff
@@ -127,14 +127,14 @@
                         {
                             @foreach ( var column in ParentDataGrid.GetColumns().Where( x => x.IsRegularColumn ) )
                             {
-                                <DropdownItem ShowCheckbox Checked=@column.Displaying CheckedChanged="@(value => ColumnDisplayingChanged( column, value ))">@column.Caption</DropdownItem>
+                                <DropdownItem ShowCheckbox Checked=@column.Displaying CheckedChanged="@(value => ColumnDisplayingChanged( column, value, column.DisplayOrder ))">@column.Caption</DropdownItem>
```
tesar-tech, line 130: This display order passing has no value. It will hit the `Column.SetDisplaying` where it will met the same displayOrder value.

tesar-tech (changes requested):
It works without any bug. It's nice and simple.

It lacks few things:
- some better drag&drop ui - maybe borders, transparency, target highlighting...
- You cannot move something to be first column.

The actual algorithm of reordering columns (`OnColumnDropped`) can be done in a smoother and more efficient way imo.

stsrki:
> I agree, the drag and drop needs some better styling.
> 
> I would also like to see an easy way to reset the column order and display to default with a click of a button. Perhaps somewhere in the column chooser menu and within the API.

I think making it through the API would be the best way. Although, it should still work with the DisplayOrder parameter.

stsrki:
@tesar-tech please re-review. 

I have reverted your changes and modified the logic so that it works without rearranging columns. Now it only works by changing the `DisplayOrder`.

Source/Extensions/Blazorise.DataGrid/DataGrid.razor.cs
```diff
@@ -782,11 +793,59 @@ private Task OnColumnDragStarted( DataGridColumn<TItem> col )
...
 
+    private async Task OnColumnDropped( DataGridColumn<TItem> columnDropped )
+    {
+        if ( columnDragStarted is null || columnDropped is null )
+            return;
+
+        // Normalize DisplayOrder if needed
+        var orderedColumns = Columns.OrderBy( c => c.DisplayOrder ).ToList();
+
+        if ( orderedColumns.Any( c => c.DisplayOrder == 0 ) )
```
tesar-tech, line 810: I still don't get it.
But this can wait for now.

Demos/Blazorise.Demo/Pages/Tests/DataGrid/ColumnsPage.razor
```diff
@@ -36,11 +36,11 @@
                 <DataGrid TItem="Employee" Data="@inMemoryData" Responsive Editable ShowPager Filterable>
                     <DataGridColumns>
                         <DataGridCommandColumn TItem="Employee"></DataGridCommandColumn>
-                        <DataGridColumn Editable Field="@nameof(Employee.FirstName)" Caption="First Name"></DataGridColumn>
-                        <DataGridNumericColumn Editable Field="@nameof(Employee.Salary)" Caption="Salary"></DataGridNumericColumn>
-                        <DataGridDateColumn DisplayFormat="{0:dd.MM.yyyy}" Editable Field="@nameof(Employee.DateOfBirth)" Caption="Date Of Birth"></DataGridDateColumn>
-                        <DataGridCheckColumn Editable Field="@nameof(Employee.IsActive)" Caption="Active"></DataGridCheckColumn>
-                        <DataGridSelectColumn Caption="Gender" Field="@nameof(Employee.Gender)" Data="EmployeeData.Genders" ValueField="(x) => ((Gender)x).Code" TextField="(x) => ((Gender)x).Description" Editable />
+                        <DataGridColumn Editable Field="@nameof( Employee.FirstName )" Caption="First Name" Reorderable></DataGridColumn>
```
tesar-tech, line 39: There is one fatal flaw. Not in this example, but in the way how `DisplayOrder` is being edited.

Consider this code:

```
        <DataGridColumns>
                        <DataGridCommandColumn TItem="Employee"></DataGridCommandColumn>
                        <DataGridColumn Editable HeaderDisplay="Display.Always" DisplayOrder="1" Field="@nameof( Employee.FirstName )" Caption="First Name" Reorderable></DataGridColumn>
                        <DataGridNumericColumn Editable Field="@nameof( Employee.Salary )" Caption="Salary" Reorderable></DataGridNumericColumn>
                        <DataGridDateColumn  DisplayFormat="{0:dd.MM.yyyy}" Editable Field="@nameof( Employee.DateOfBirth )" Caption="Date Of Birth" Reorderable></DataGridDateColumn>
                        <DataGridCheckColumn Editable Field="@nameof( Employee.IsActive )" Caption="Active"></DataGridCheckColumn>
                        <DataGridSelectColumn Caption="Gender" Field="@nameof( Employee.Gender )" Data="EmployeeData.Genders" ValueField="( x ) => ( (Gender)x ).Code" TextField="( x ) => ( (Gender)x ).Description" Editable />
                    </DataGridColumns>
```
It's the same code from `ColumnsPage.razor`, just this line was altered:

<DataGridColumn Editable **HeaderDisplay="Display.Always" DisplayOrder="1"** Field="@nameof( Employee.FirstName )" Caption="First Name" Reorderable></DataGridColumn>

This forces the `FirstName` column to stay in spot `#1`. First it is on the end bcs the others have 0, when you move to  "anywhere", it jumps to 1, because others now have higher number, then you cannot move it at all.

[Screencast From 2025-05-29 23-29-17.webm](https://github.com/user-attachments/assets/b788834a-79f8-4ad2-82f7-b14c610cbfec)

This is caused by: https://learn.microsoft.com/en-us/aspnet/core/blazor/components/overwriting-parameters?view=aspnetcore-9.0

The issue isn't present in the original source code, because the column has only primitive-type params, that blazor checks for changes (and there are non). But adding the `HeaderDisplay` or any other non-primitive type paramater forces blazor to re-render the column every-time. For example when some drag and dropping events occur.

Solution might be to introduce `DisplayOrderInitial`, or keep the `DisplayOrder` and make it behave as only the initial value (and keep the real state hidden). This would also solve the requested feature "return to initial order state".

Documentation/Blazorise.Docs/NewFilesToBuild.txt
```diff
@@ -0,0 +1 @@
+D:\Projects\Megabit\Blazorise\Documentation\Blazorise.Docs\Pages\Docs\Extensions\DataGrid\Code\DataGridReorderingColumnsExampleCode.html
```
tesar-tech, line 1: does this belongs here?

Source/Extensions/Blazorise.DataGrid/DataGridColumn.cs
```diff
@@ -297,6 +297,15 @@ public async Task SetDisplaying( bool displaying )
         await ParentDataGrid.Refresh();
     }
 
+    public async Task SetDisplayOrder( int displayOrder, bool forceParentRefresh = false )
+    {
+        DisplayOrder = displayOrder;
+        await ParentDataGrid.ColumnDisplayOrderChanged.InvokeAsync( new ColumnDisplayOrderChangedEventArgs<TItem>( this, displayOrder ) );
```
tesar-tech, line 303: Wouldn't be better to make the EventCallback for the whole reorder batch?
Because with this users will need to keep track of the columns order anyway. What about just pushing all the columns with their order?? Or only the change ones?
idk. just an idea

PR #6132, opened 2025-05-28
DataGrid: simplifies OnColumnDropped method

tesar-tech:
## Description

Solves https://github.com/Megabit/Blazorise/pull/6126#issuecomment-2916550269

I just removed few lines. It does the same thing still.


stsrki (approved):
Seems to be working. Thanks.

PR #6400, opened 2026-02-14
DataGrid: Self-Reference Hierarchy mode

### LINKED ISSUE #6304: Feature: DataGrid Self-Reference Hierarchy
### Discussed in https://github.com/Megabit/Blazorise/discussions/6301

<sup>Originally posted by **Gareth064** December  5, 2025</sup>
Would it be possible to add a new feature to the DatGrid which allows you to render records in a Hierarchy (Parent/Child). Basically a tree type DataGrid.

I hate referencing other Libraries in these conversations, but this is the only other example I could find that demonstrates what I mean.

This is from Radzen and they named it "Self-Referencing Hierarchy"
image
As you can see the nested/child data is just displayed, aligned to the same columns as the main DataGrid etc.
With a little indent there on the first column to demonstrate that its a child of the row above it.</div>

stsrki:
Closes #6304

### DIFF
--- a/Source/Blazorise/Components/NumericPicker/NumericPicker.razor.cs
+++ b/Source/Blazorise/Components/NumericPicker/NumericPicker.razor.cs
@@ -231,7 +231,7 @@ public partial class NumericPicker<TValue> : BaseBufferedTextInput<TValue, Numer
 
         if ( paramValue.Changed )
         {
-            ExecuteAfterRender( async () => await JSModule.UpdateValue( ElementRef, ElementId, paramValue ) );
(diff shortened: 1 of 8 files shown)


njannink (commented):
- [x] Code reviewed
- [x] Documentation tested and reviewed

Demos tested:
- [x] Ant Design
- [x] Bootstrap
- [x] Bootstrap server
- [x] Bootstrap 5
- [x] Bulma
- [x] FluentUI2
- [x] Material
- [x] Tailwind

Issue #6159, opened 2025-06-26
[Bug]: Treeview not responding after changes to HasChildNodes


njannink:
### Blazorise Version

1.7

### What Blazorise provider are you running on?

Material

### Link to minimal reproduction or a simple code snippet

            <TreeView Nodes="@Root"
                      GetChildNodes="@(x => x is FolderModel f ? f.Items : [])"
                      HasChildNodes="@(x => x is FolderViewModel f && f.Items.Count > 0)">
                <NodeContent>@context.Name</NodeContent>
            </TreeView>
 

### Steps to reproduce & bug description

Initial the Root is empty so HasChildNodes is false, after adding a new folder to the root the tree doesn't update. The Items list is an ObservableCollection

### What is expected?

On changes of the ObservableCollection also reevaluate the HasChildNodes this can become true or false when the collection changes

### What is actually happening?

new nodes dont appear

njannink:
@stsrki this one is fixed by #6498

PR #6426, opened 2026-02-26
TreeView: Reload node support to reload only a specific part of the tree

njannink:
Add support for reloading a specific node of the tree, eg for when a child is added/removed from a specific node to force reevaluation of the expand toggle visability.

### DIFF
--- a/Source/Extensions/Blazorise.TreeView/Internal/_TreeViewNode.razor.cs
+++ b/Source/Extensions/Blazorise.TreeView/Internal/_TreeViewNode.razor.cs
@@ -5,7 +5,6 @@ using System.Collections.Specialized;
 using System.Linq;
 using System.Threading.Tasks;
 using Blazorise.Extensions;
-using Blazorise.TreeView.Extensions;
 using Blazorise.Utilities;
 using Microsoft.AspNetCore.Components;
 using Microsoft.AspNetCore.Components.Web;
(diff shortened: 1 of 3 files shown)


stsrki:
@njannink made some optimizations. Can you recheck the behavior on your side?

njannink:
Will check will also see if this fixes #6159

njannink:
@stsrki I don't really know what is the best approach to continue with this. Biggest issue is that the TreeView does support `ObservableCollection` but only after a node has been expanded so that is making the API a bit confusing. Ideally for HasChildNodes you use something like an IObservable<bool> that you subscribe, but that would introduce a dependency on System.Reactive.

We could introduce:
```
 interface ITreeNode
{
    IEnumerable<T> GetChildren();
    bool HasChildren();
    event EventHandler ChildrenChanged;
}
```

And then do some automatic binding and also trigger reevalution when the ChildrenChanged is triggered.

Thoughts are welcome

stsrki:
Let's first start with reproducible. Can you create a test demo in the demo app that shows this exact problem in action? Because I don't really see it in our current demo.

njannink:
@stsrki see the following test app

https://github.com/njannink/ObservableBlazoriseTreeView

You can see that the root of tree1 isn't auto expanded even if items are added. tree2 does auto expand, because it already had childs so events are hooked up. Only after an explicit reload after added a first child to tree1 you can see it starts working.

stsrki:
Made a different approach. Can you try it?

https://github.com/Megabit/Blazorise/pull/6498

stsrki:
Should we close this in favor of #6498

njannink:
I will close this one and create a new one on top of 6498. I think the reload a specific part of the tree is still usefull for users who don't use `ObservableCollection` and then I can also pickup the drag+drop again

PR #6455, opened 2026-03-11
RangeSlider component

### LINKED ISSUE #5584: Slider with range selection
I would like to have a slider that can handle 2 values so that the user can select a range instead of a single value. I would like to use it in a filter selection for years (begin end year). Something like this (from mdbootstrap pro). I would suggest to colorize the range between the selected points though.
image

stsrki:
Closes #5584

Introduces `RangeSlider`, a new generic Blazorise input component for selecting a numeric range with two handles and a strongly typed `RangeSliderValue<TValue>`. 

The component includes min/max/step support, optional value tooltips, accessible handle  labels, configurable handle crossing behavior (`ClampToOtherHandle `/ `AllowEqualValues`), JS-assisted drag clamping, providerstyling/renderers across supported UI providers, theme color integration, tests, and full documentation with examples.


njannink:
## Test 

### General
- Optionally show start and end values, also for regular slider?

### Provider Specific
- [ ] AntDesign
  - Not sure if the animation off the selected value has any added value seems a bit slow aswell
- [x] Bootstrap
- [ ] Bootstrap Server
  - Black box rendered when control has focus
  - Rendering issues of selected range
  - Values rendered behind bar
  image
- [x] Bootstrap5
- [x] Bulma
- [ ] FluentUI2
  - regular range slider shows end value maybe also show start / end value on new range slider
- [ ] Material
  - hitbox of grabber seems to be very small (also for regular range slider)
  - range bubbles very big visible overlapping other UI
- [x] Tailwind

stsrki:
I think you had an older CSS file on some providers. Can you clear the cache and try again?

PR #6464, opened 2026-03-13
Accessibility: FieldSet and Legend component

### LINKED ISSUE #6441: Support fieldset and legend
https://blazorise.com/support/issues/413/fieldset

stsrki:
Closes #6441

This PR adds first-class support for semantic form grouping in Blazorise with new independent `FieldSet` and `Legend` components. They render native `fieldset` and `legend` elements, have dedicated class provider methods across all supported providers, and integrate with existing accessibility options so grouped controls such as `RadioGroup` can automatically use the parent `Legend` as their accessible label through `aria-labelledby`. The PR also adds demo and documentation examples showing the recommended pattern for radio groups and related checkbox sets, while keeping `Field` and `FieldLabel` focused on single-input scenarios.


njannink:
[x] Code Reviewed

### General Test Remarks
- Was not able to test validation logic. Maybe add to ValidationsPage

### Provider Specific Tests
- [x] AntDesign
  - Alignment for vertical fields off
  image
- [x] Bootstrap
- [x] Bootstrap Server
- [x] Bootstrap5
- [x] Bulma
  Legend missing and alignments off
  image
- [x] FluentUI2
- [x] Material
  - padding / margin of elements is missing
  image
- [x] Tailwind

stsrki:
The latest refactoring moves semantic grouping support onto the existing `Field` and `Fields` APIs, instead of treating `FieldSet` and `Legend` as specialized versions of Field and FieldLabel. Field and Fields now opt into grouped semantics explicitly, and their label components switch to rendering a legend when used in group mode. At the same time, FieldSet and Legend were simplified into standalone native wrappers that render only the `fieldset` and `legend`, without inheriting the behavior, accessibility wiring, or layout assumptions of Field and FieldLabel. 

The reason for this split is to avoid forcing native `fieldset` and `legend` usage into the styling and behavioral model of a single-field container. A real fieldset can group many different kinds of controls, not just the scenarios that map cleanly to `Field`, so coupling those components too tightly would artificially limit their use. By reusing `Field` and `Fields` for the higher-level grouped API, and keeping `FieldSet` and `Legend` as dumb native elements with their own class-provider hooks for horizontal, validation, required-indicator, and screenreader styling, the API becomes clearer: the existing components handle Blazorise form composition, while the standalone wrappers remain available for direct semantic HTML when needed.

stsrki:
@njannink, please recheck the API and UI

stsrki:
PS. I'm thinking of renaming the `Group` parameter to `SemanticGroup`.

njannink:
SemanticGroup says the same to me as Group. Maybe use (Is)FieldSet to really point at the underlying html that will be rendered?

PR #6586, opened 2026-05-04
PivotGrid component

### LINKED ISSUE #2668: Create a PivotGrid component
image

image

stsrki:
Closes #2668 

PivotGrid provides a read-only pivot table for grouping records by row and column dimensions, calculating aggregate values, and rendering totals/subtotals in a table layout. It supports declarative field definitions, runtime layout changes through a field chooser, local data binding, external data loading, paging, and row virtualization.

The public API includes the main `PivotGrid<TItem>` component, row/column/aggregate/field child components, prepared result models, data request/result models, and event args for provider-based data loading and cell click handling.

It also includes:

- local and external data support through `Data`, `ReadData`, and `DataSource`
- paging and page-size support
- virtualized row rendering
- provider-side paging/virtualization contracts
- row and column totals/subtotals
- expandable row and column groups
- runtime field chooser
- filtering through the field chooser state
- custom aggregate functions through Aggregator
- templates for headers, values, and cells
- cell, row header, and column header styling hooks
- accessible expand/collapse controls
- docs, API docs coverage, and demo examples for the main usage scenarios

The demo and documentation cover basic pivoting, runtime field selection, paging, virtualization, remote data, templates/events, filters/expansion, custom aggregation, and conditional styling.


David-Moreira:
**1.** Whether each row is expanded or not, it seems like a vertical scroll is missing I can't see all data: 
**Expanded**
image
**Not Expanded**
image

In the meantime I figured out that you need to change the page, should there be a scrollable option? Typically pivot tables show everything at once, no?
I guess there's Virtualize perhaps never mind.

**2.** Setting a **DataSource** ReadData or DataSource makes it so the totals row no longer shows bold numbers

**Data:**
image
vs
**ReadData**
image


**3.** 
Paging is a little confusing? I think there's some write up improvement needed
image

**4.** 
When ReadData is set I cannot change the page sizes

**5.**
Is there a fundamental difference between **Data** and **DataSource**?

**6.**
Expandable rows does not work with **Virtualzie**

**7.**
If **Virtualize is active And DataSource is Data** and Row Totals is removed the content is not properly updated. Same thing the other way around, if RowTotals starts inactive and Virtualize is activated and then RowTotals is changed to active, the content is not properly updated, which points to a possible issue with any parameter that changes the content rendering
image

**8.**
I have no idea what the Column subtotals, column totals, column total position do, I try them and nothing happens.

**9.** 
Column Captions.
Leaf shows Berlin, Field shows City, and Full Path still shows Berlin, should it be different then Leaf? Or in this example is not applicable?

**10.**
When Expandable rows is set, the name of the row section is Cloud Total. I understand this is because the totals are shown at this level. But would it be "more correct" to show just "Cloud" which is the actual section name?
Because if you have Expandable rows set, but disable Row subtotals, then there's no section at all. Shouldn't there still be a section even if the subtotals are not active but expandable rows are set?

**11**
The SubRow Totals shows a new row with totals by each section, however the Row Totals renders new columns with totals by row, is this intended? The way each one renders at different levels does not seem consistent? Is this the intended behaviour?

**12**
I'm not sure what's intended with Page by groups does? If data source is data, everything shows, if read data only a few items show?

**13**
Is it intended for the Row Captions Field and Column captions Field to be the actual field names **Product** and **City**? Doesn't seem to make much sense, because you cannot distinguish any data at all?
image

**14**
I personally like the DataGrid question marks where additional information is provided for each option explaining what it does

David-Moreira:
**Docs**

a. Dark mode is a bit jarring : 
image

Other then that docs seem fine to me

stsrki:
> **5.**
> Is there a fundamental difference between **Data** and **DataSource**?

Yes.

Data is for the normal case: you already have the items in memory, so PivotGrid does the pivot work itself.

DataSource is for delegated loading. PivotGrid sends a PivotGridDataRequest to your data source, and the data source returns either rows/data to pivot or a prepared pivot result. That is the better fit for server-side paging, virtualization, or keeping query logic outside the component.

So, short version:

- Data: component owns the work.
- DataSource: your provider owns the read/result work.
- ReadData: same idea as DataSource, but as an event callback instead of a reusable provider object.

David-Moreira:
> So, short version:
> 
> Data: component owns the work.
> DataSource: your provider owns the read/result work.
> ReadData: same idea as DataSource, but as an event callback instead of a reusable provider object.

So is **DataSource** something new or being introduced in this PR?
Your description of **DataSource** makes me think **DataProvider** is a better naming, that's what would immediately make me think this is some kind of unit / class I can provide that handles the data concerns.
**DataSource** naming just makes me think about an **IEnumerable** or **Func** 
So my personal suggestion is **DataProvider** naming or something similar if this is a new feature.

PR #6690, opened 2026-07-17
Reporting: introduce report designer, PDF generation, DockLayout, and ContextMenu

stsrki:
## Summary

Introduces the new Blazorise Reporting extension together with the supporting PDF, DockLayout, and ContextMenu components. The branch also adds the required core infrastructure, styling, documentation, and demos.

## Reporting

Adds a declarative, band-based reporting system with headers, footers, detail and group bands, text, fields, images, lines, rectangles, panels, page breaks, tables, and subreports.

The interactive designer supports toolbox and field drag-and-drop, properties editing, rulers, grid snapping, multi-selection, moving, resizing, alignment, ordering, collision warnings, context menus, keyboard shortcuts, and undo/redo history.

Reports support object, `DataSet`, `DataTable`, `CSV`, and SQL data sources. Expressions include formulas, functions, aggregates, running totals, formatting, grouping, and special fields.

Report definitions and designer state can be serialized, saved, and loaded. The toolbar is command-based and supports custom groups, items, templates, and command visibility. Preview is available as HTML or PDF, with PDF display through `PdfViewer`.

Rendering was optimized through targeted refreshes, cached pagination, reduced cloning, and simplified data resolution. 

## DockLayout

Adds a new core docking layout with horizontal and vertical splits, resizable panes, tabbed panes, drag-and-drop docking, compass drop zones, closable panes, pinning, auto-hide rails, and flyout panes.

Layout state can be captured and restored, including pane placement, size, selected tabs, closed panes, and pinned state. Rendering and resize handling were optimized so only affected nodes refresh during interactions.

## ContextMenu

Adds a new core context menu with regular items, headers, dividers, groups, submenus, checked items, toggles, and toolbar layouts.

Menus support programmatic visibility, lifecycle events, pointer positioning, scrolling, and nested placement. Outside-click and Escape listeners are registered only while a menu is visible to avoid unnecessary document events and JS interop.

## PDF and PdfViewer

Adds the `Blazorise.Pdf` extension with declarative Razor components and fluent builders for creating documents, pages, text, images, lines, rectangles, tables, rows, and cells.

The built-in renderer supports page sizing, margins, borders, colors, image fitting, Unicode text, custom TrueType fonts, and shared Blazorise `TextAlignment` and `VerticalAlignment` enums. Public PDF types were split into matching files and standardized around PdfGenerationOptions and PdfGenerationResult.

Reporting uses the PDF generator for preview and download, with generated documents displayed directly through PdfViewer.

## Core infrastructure

Adds `BaseStyledComponent`, a lightweight base class that provides initialization, disposal, class and style builders, and `Class` and `Style` parameters without the complete `BaseComponent` utility API. `BaseComponent` now builds on this smaller base.

Document observation was extended for global pointer interactions, and a new Duplicate icon was added across the supported icon providers. Provider styles were added for Reporting, DockLayout, and ContextMenu.


David-Moreira:
**PDF Generator**
1. I know this is a demo, but could everything have better spacing, look better, etc or is that not worth it?
image

2. Specially this one?
image

3. The Generate PDF and Open button don't really seem to do anything, what are they supposed to do?

4. Is this probe correct?
It seems like perhaps it's not rendering some chars at the end, not sure?
image

5. Perhaps have the PDF have more then 1 page, so we could see paging work?

**Pdf Viewer**
Works really good from what I can tell. Just have to make sure there's enough window space to view full document. Zoom could potentially be manually editable, and have other pdf tools, like selecting text, etc... not sure if possible

**Context Menu**
1. What's up with the location. Is that useful information? Should it be 0,0 if I click on the edge of a container with right click?
Also if I click other menus, it tracks the last command but not the location.
image

2. The last command, whenever a command is clicked is Menu Closed, which in practice is true, but not helpful information as part of the command logging, no?  Menu Open and Menu Close, could be separate commands, from actual menu commands? Worth it?

3. The Paste command is never enabled, I take it, it's just an example?

4. What is Keep menu open supposed to do? I switched it on, and the menu still gets closed?
image

**Reporting**
1. Reseting the report is in itself an undoable option. Is this per design?
2. I'd like to see some visual feedback for the Save and Load buttons. 
3. The connect datasource to me is a little confusing, took me a while to understand. If I understand correctly the first option always creates a new one, but the other options are existing data sources? Could there be a better UX for this, or is it just me?
Also it seems, that from all the options, the only usable in the web UI is the csv?
4. I tried changing the data source for the CSV lines, but it seems to have no effect?
5. It would be nice to preview the data source data if possible?
6. If the Report Explorer is in the dock, the properties doesn't work anymore, I can select elements, or go to the report explorer and select it, but it never updates it anymore.
7. If I unpin Toolbox or FieldsExplorer when opening, it goes on top of the menu bar hiding it.
8. If I unpin the Properties or Solution Explorer, when opening it goes on top of the Design and Preview 
9. It's an uncommon choice to see Close in context menus.
10. Should Load warn there are unsaved changes?
11. If the Properties window is unpined, the Report Explorer also seems not to sync with the active report or sub report.
12. Since I can name the different section, should I be able to search for them in the Report Explorer? 
13. Why are there (Demo) font families?
14. Also I'm not sure about the others but isn't Helvetica typically a licensed font? Not sure how that works for web?
15. I don't think the Snap To Grid is working, I tried setting it, but when I move the element it doesn't snap to it, Am I missing something? Also shouldn't that be more like a global option instead of a per field option?
16. Sometimes you want to make tiny adjustments, I feel like a zoom feature would be interesting to explore
17. Also I worked on a similar software before to build this kind of reports, it had an option to add lines that would extend all the way vertically and horizontally where my mouse was hovering, this was nice so you could understand if the elements were aligned vertically or horizontally all the way across.
18. Formula doesn't recognize a data source field:
image

19. If I type Width 9999 or Height 9999, realistically it should just default to a standard maximum for the page as it would be a shortcut for a seasoned report builder to quickly set the width and/or height to max
20. Apparently I can freely do negative numbers for most inputs, font-size, height, x, y, etc... is this per design?
21. In the format box property, there's a Culture input, I have no idea what goes in here, it's a free type text. If it's controller values, then it should be a dropdown, otherwise a tooltip can help define what it does
22. Isn't it Visible a better name then Supress?
23. Putting elements on top of each other makes them red, almost like they are invalid, but if I understand correctly, they still are valid, maybe the visual cue should be different?
24. In the properties window, I think the type could be shown somewhere just to be extra explicit. If the report explorer is closed you don't know which type something is.
25. Pasting in a table cell only works through the context menu, ctrl + v doesn't seem to work.

The report did crash my webpage twice over the course of my tests, it just became unresponsive. I tested it during about 1h or so. I couldn't quite reproduce it.

PR #6774, opened 2026-08-29
Theme Generator: avoid string alloc when parsing

SimonCropp:


stsrki (approved):
Looking good. Thank you for the contribution.

PR #6775, opened 2026-08-30
Components: avoid allocations when rendering boolean attributes

SimonCropp:
Replace `ToString().ToLowerInvariant()` patterns with explicit `"true"`/`"false"` literals across core, provider, and extension components.


stsrki (approved):
LGTM. Thanks for the contribution.

